PR #63, opened 2024-12-09
Geocoder street address and route long

dgpaci:


cgabriel (approved):
I've just updated the case when street_number is empty when dealing with subpremise.

PR #85, opened 2025-01-20
Title: Improve Column Type Compatibility and Migration Handling

fporcari:
Description:
This PR introduces a robust mechanism to handle column type changes in the ORM, ensuring compatibility and safe migrations. Key updates include:
	1.	Compatibility Mapping:
Added a COMPATIBLE_TYPES dictionary to define which type conversions are automatically supported between ORM and SQL types.
	2.	Column Type Change Logic:
	•	For compatible type changes, the system directly applies the ALTER COLUMN operation.
	•	For incompatible type changes:
	•	If the column is empty, it is dropped and recreated with the new type.
	•	If the column is not empty, the operation raises an exception to prevent unintended data loss.
	3.	Utility Methods:
	•	Implemented struct_is_empty_column_sql and struct_is_empty_column to verify if a column contains non-NULL data before proceeding with type changes.
	4.	Extensive Testing:
	•	Verified behavior for compatible and incompatible type changes.
	•	Tested edge cases like empty columns and non-convertible data.

This update ensures safe, predictable, and efficient handling of schema migrations.


cgabriel:
Not suitable for merge in master.

PR #86, opened 2025-01-20
changing column dtype

fporcari:


cgabriel:
Reviewed LLAMA comments with @fporcari , approved for merge

Issue #110, opened 2025-02-04
[mkthresource] Wrong column size calculation from model data size


EdoFede:
The calculation of the column width made by mkthresource with -g option is not working.
The output is 35em for text columns (of any size) and 7em for other columns.

EdoFede:
https://github.com/genropy/genropy/tree/feature/mkthresource_width_calc

EdoFede:
Fixed in #113

PR #111, opened 2025-02-04
Refactor column width calculation in ThResourceMaker (-g option)

EdoFede:
Refactoring of wrong-working column width calculation in ThResourceMaker (launched with -g option).
Since the direct calculation of the width in "em" seems to be not so accurate, I've used a lookup table instead.
Not so elegant, but improves accuracy of the width calculation.

#110


EdoFede:
Error

PR #118, opened 2025-02-20
Feature/nginx proxy x forwaded for

cgabriel:
When users are using forward proxies/vpn like ZScaler, their endpoint adds a X-Forwarded-For headers, which then gets summed up with their original source IP if using $proxy_add_forwarded_for in nginx reverse proxy configuration. This could cause issues when expecting that a single IP address is searched in the request headers.

This change includes:

* Change to the nginx configuration  generation, using $remote_addr rather than $proxy_add_forwarded_for, to have only the user endpoint in the reuqest
* Search for commas in the x-forwarded-for header (for old nginx deployments), and if found, use only the last IP listed there.


cgabriel:
Closing this, also related to traefik changes, which will may be needing a more globally thought approach.

PR #257, opened 2025-10-30
Feature/improvements in bagfield

fporcari:
Fixed validation in bagfields


gnrpy/gnr/web/gnrwebpage.py
cgabriel, line 1: The docstring must be updated to document this new safeMode parameter

cgabriel (changes requested):

PR #297, opened 2025-11-14
Fix external project packages check to verify main.py existence

fporcari:
## Summary

This PR fixes an issue in the external project packages feature introduced in PR #283.

## Problem

The original implementation used `os.path.isdir()` to check if a package exists locally before falling back to external projects. However, when updating from git, deleted packages can leave empty directories behind. The `os.path.isdir()` check returns `True` for these empty directories, preventing the proper fallback to external projects.

## Solution

Changed the package existence check from:
```python
if not os.path.isdir(os.path.join(path, pkgid)):
```

To:
```python
if not os.path.exists(os.path.join(path, pkgid, 'main.py')):
```

This ensures only valid Genropy packages (with `main.py`) are recognized, allowing proper fallback to external projects when needed.

## Changes

- Modified `gnrpy/gnr/app/gnrapp.py` line 1026 to check for `main.py` existence instead of directory existence

## Testing

- Verified that the check correctly identifies valid packages
- Tested that empty directories left by git updates no longer block external project lookup

### DIFF
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -1025,6 +1025,13 @@ class GnrApp(object):
                 path = os.path.join(project_path,'packages')
                 if not os.path.isdir(os.path.join(path, pkgid)):
                     path=None
+                    external_project_path = os.path.join(project_path,'external_projects.xml')
+                    if os.path.exists(external_project_path):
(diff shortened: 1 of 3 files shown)


fporcari:
Chiudo questa PR perché creata erroneamente su master. Le PR devono essere create su develop.

PR #226, opened 2025-08-31
Set capture = 'environment' in the image input  widget to use the rear camera

Giammi77:
example:

# -*- coding: utf-8 -*-
from gnr.core.gnrdecorator import public_method

class GnrCustomWebPage(object):
    py_requires = 'th/th:TableHandler'

    def main(self,pane,**kwargs):
        gb = pane.gridbox(columns=1,gap='5px', datapath='main')
        div = gb.img(src='^.img_targa',id='div',
                    lbl='Fotografia',
                    lbl_side='top',
                    crop_height='100px',
                    crop_width='100px',
                    capture='environment',
                    # crop_border='2px dotted silver',
                    # crop_rounded=6,
                    edit=True,
                    placeholder=True,
                    takePicture=True,
                    )

PR #301, opened 2025-11-18
fix: add capture='environment' support for rear camera on mobile

dgpaci:
This PR improves support for the `capture` attribute in the img widget for camera selection on mobile devices.

## Changes from original PR #226

**Fixes:**
- Fixed `capture_environment` → `capture_environment` typo
- Removed duplicate variable declaration

**Improvements:**
- Full support for both camera modes:
- `capture='environment'` - rear camera
- `capture='user'` - front camera
- Support for reactive bindings (e.g., `capture='^.camera_mode'`)
- Correct attribute propagation in both crop and non-crop scenarios
- Read the value from the datasource when the capture dialog is opened

## Files Changed

- `gnrjs/gnr_d11/js/genro_widgets.js`:
- `startCapture` method: Reads the capture from the datasource or attribute and configures the camera accordingly
- Propagation of the capture attribute in both paths (crop (and non-crop)
- Dialog connect_show: Reads the current capture value and passes it to the video widget.

## Test

Added `test_6_camera_toggle` to `projects/gnrcore/packages/test/webpages/html/img.py`, which demonstrates use with a static rear camera.


projects/gnrcore/packages/test/webpages/html/img.py
```diff
@@ -66,17 +79,6 @@ def deleteImage(self, file_url=None, **kwargs):
         print('**DELETED IMAGE: ', file_url)
         return 'Image deleted'
 
-    def test_4_s3Storage(self,pane):
```
cgabriel, line 69: Why the removal of this test in this PR?

cgabriel (changes requested):
Please review the extra S3 test removed, which is not, to me, related to the PR topic.

projects/gnrcore/packages/test/webpages/html/img.py
dgpaci: As you point out, it's not related, but a test to store an image on S3 bucket was wrong in the first place, because it's a storage service test and not a true img widget test. That's why I removed it...
If someone wants to test s3 storage of an image, he can use test 1 with a storage service active, it's not necessary to use such test

dgpaci (commented):
PR is consistent, the test which was removed is not related but it can be deleted without particular problems since it's totally unuseful

cgabriel (approved):

PR #225, opened 2025-08-26
Feature/mail review

dgpaci:
### **User description**
Closes #208. Please run a few tests as well, everything seems fine to me

___

### **PR Type**
Enhancement

___

### **Description**
- Refactored mail service with improved SMTP connection handling

- Added context manager for SMTP sessions and connection reuse

- Enhanced email account configuration with reply-to and priority support

- Fixed parameter handling bugs in mail service methods

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Mail Service"] --> B["SMTP Session Manager"]
  B --> C["Connection Reuse"]
  A --> D["Account Parameters"]
  D --> E["Reply-to Support"]
  D --> F["Priority Handling"]
  A --> G["Bug Fixes"]
  G --> H["Parameter Mapping"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>12 files</summary><table>
<tr>
  <td><strong>mail.py</strong><dd><code>Core mail service refactoring with SMTP session management</code></dd></td>
  <td><a href="https://github.com/genropy/genropy/pull/225/files#diff-96254fea4e5bc5ee0df7af46a357f732d2c58623b7646862539e109868390bbf">+53/-33</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>mail.py</strong><dd><code>Enhanced mail template service with reply-to and priority</code></dd></td>
  <td><a href="https://github.com/genropy/genropy/pull/225/files#diff-52b23e9384df73db6ca3d683b1665ebde61ca4c72b1e699ce76258269f967243">+12/-8</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>account.py</strong><dd><code>Enhanced account model with reply-to and test email functionality</code></dd></td>
  <td><a href="https://github.com/genropy/genropy/pull/225/files#diff-5386b7d4b4595a6f0af6b2852ba58c00ce3c7ee86a6637295107476903ff729e">+50/-26</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>message.py</strong><dd><code>Added priority support and improved message sending logic</code></dd></td>
[...]

Issue #208, opened 2025-07-28
General optimization of e-mail sending workflow


dgpaci:
E-mail sending workflow can benefit from a few general optimizations:

- [ ] when a lot of e-mails is sent, connection seems to be opened and closed every time. It would probably be better to keep it open somehow..
- [ ] furthermore, if attachments are present these are opened and closed every time (probably only with S3), therefore at some point a "Too many open files error" is generated
- [ ] email service and its method are redefined at many levels, a general refactoring could make it more clear 

We already started to make a few improvements in a dedicated branch `feature > mail-review`, check to really understand what's needed and what's not

dgpaci:
There's a `send_one_email` in `btcmail` which must be refactored, an old "legacy" method is called and is unnecessary

```
def send_one_email(self,**kwargs):
        #DP This method could probably just call sendmail 
        if self.db.package('email'):
            self.mail_handler.sendmail(account_id=self.batch_parameters.get('account_id'),**kwargs)
        else:
            self._send_one_email_legacy(**kwargs)
```

dgpaci:
Bug found in sendUserTemplateMail due to this kwargs extraction in email > mailservice (line 51):

@extract_kwargs(headers=True, message=True)
    def sendmail(self, scheduler=None, account_id=None, attachments=None, headers_kwargs=None, 
                 message_kwargs=None, doCommit=None, smtp_connection=None, **kwargs):
It's not clear to me why it was done in the first place... Anyway, since message kwargs are put in message_kwargs, if this function is called without preparing correctly the dictionary, messages are basically empty.

This function is called many times in our code even in customer applications, therefore it should be managed accordingly

cgabriel:
Closing with PR #315 merged

PR #315, opened 2025-11-28
Email features: reply-to support, priority, and fixes

dgpaci:
## Summary

This PR adds minimal, production-ready email features extracted from PR #225:

- **Reply-to support** at three levels (global preference / account / message)
- **Message priority** field with values: immediate (-1), high (1), standard (2), low (3), no send (9)
- **Debug address fix** - properly redirects emails when debug_address is set
- **Weak attachments fix** - handles empty arrays correctly
- **Test UI improvements** - better SMTP test button with default user email
- **Global email preferences** - new preference pane for default settings

## What's excluded from PR #225

This PR intentionally **excludes** the SMTP connection management refactoring from PR #225, which had several critical bugs and will be redesigned separately. Specifically excluded:

- smtp_session() context manager
- Connection pooling/reuse logic
- smtp_connection parameter passing
- Modified send_messages action
- Changes to mailservice.py with @extract_kwargs bug

### Models
- email/model/account.py - Add smtp_reply_to column, update getSmtpAccountPref(), add sendEmailFromParams() test method
- email/model/message.py - Add reply_to and priority columns, fix debug_address usage, fix weak_attachments handling

### UI
- email/resources/tables/account/th_account.py - Add reply-to field, improve test button to use model method
- email/resources/tables/message/th_message.py - Add reply-to field to form

### Services
- adm/lib/services/mail.py - Add reply_to and priority support to template mail methods

### Configuration
- email/resources/preference.py - **NEW** global email preferences pane
- email/localization.xml - Add translations for new fields

## Testing

- Test SMTP settings from account UI
- Create message with reply-to
- Create message with priority
- Verify debug_address redirect works
- Test with empty weak_attachments array
- Set global reply-to preference

## Breaking changes

None - this is backward compatible.

## Related issues

Relates to #208 - Email workflow optimization (partially addresses)

## Notes

The priority field is ready but the "immediate send" logic (priority=-1) is not fully functional in this PR as it depends on the connection management refactoring that was excluded. For now, priority is informational only.

---


projects/gnrcore/packages/email/model/account.py
```diff
@@ -41,12 +45,14 @@ def getSmtpAccountPref(self,account=None,account_name=None):
         mp['smtp_host'] = account['smtp_host']
         mp['from_address'] = account['smtp_from_address']
         mp['user'] = account['smtp_username']
+        mp['reply_to'] = account['smtp_reply_to'] or self.db.application.getPreference('dflt_reply_to',pkg='email')
```
cgabriel, line 48: What if the preference is not set? Maybe using the from_address as a second backup value would make the implementation more solid?

projects/gnrcore/packages/email/model/account.py
```diff
@@ -67,4 +73,24 @@ def trigger_onUpdated(self, record_data,old_record=None):
...
-    
+
+    @public_method
+    def sendEmailFromParams(self, host=None, from_address=None, to_address=None, reply_to=None, subject=None, body=None,
+                                username=None, password=None, tls=None, ssl=None, port=None, **kwargs):
+        account_params = dict(smtp_host=host, port=port, user=username, password=password, ssl=ssl, tls=tls, **kwargs)
+        mh = MailService()
+        subject = subject or 'This is a test message'
+        body = body or f'This is a test message from {from_address} to {to_address}'
+        reply_to = reply_to or self.db.application.getPreference('dflt_reply_to',pkg='email')
```
cgabriel, line 84: Same as before, if the preference is not set the reply-to would be populated with nothing. Better add a second backup using the from address (reply_to or getpreference or from_address)

projects/gnrcore/packages/email/model/message.py
```diff
@@ -217,16 +222,21 @@ def getAttachmentNode(self,date=None,filename=None, new_mail = None, atc_counter
...
                  message_id=None,message_date=None,message_type=None,
                  html=False,doCommit=False,headers_kwargs=None,**kwargs):
-        
+
         message_date = message_date or self.db.workdate
-        extra_headers = Bag(dict(message_id=message_id,message_date=str(message_date),reply_to=reply_to))
+        # Fix debug address - if system_debug_address in kwargs, use it instead of to_address
+        to_address = kwargs.get('system_debug_address') or to_address
+        # Reply-to default from preferences
+        reply_to = reply_to or self.db.application.getPreference('dflt_reply_to',pkg='email')
```
cgabriel, line 234: Same here, what if the preference is not set?

projects/gnrcore/packages/email/resources/tables/message/th_message.py
```diff
@@ -246,6 +246,7 @@ def th_form(self, form):
         fb.field('from_address',colspan=2)
         fb.field('cc_address',colspan=2)
         fb.field('bcc_address',colspan=2)
+        fb.field('reply_to',colspan=2)
```
cgabriel, line 249: can this be validated as an email address?

projects/gnrcore/packages/email/model/account.py
```diff
@@ -19,6 +22,7 @@ def config_db(self, pkg):
         tbl.column('last_uid',name_long='!!Last UID')
         tbl.column('smtp_host',name_long='!!SMTP host')
         tbl.column('smtp_from_address',name_long='!!From address')
+        tbl.column('smtp_reply_to',name_long='!!Reply to')
```
cgabriel: any validation?

cgabriel (changes requested):

projects/gnrcore/packages/email/model/account.py
dgpaci: If it's not set actually I think it's actually more clear to use nothing at all: if nothing is used, there's no reply-to and therefore when e-mail is answered, it is answered to the sender (from address).

projects/gnrcore/packages/email/model/account.py
dgpaci: I agree, I think we could add `validate_email=True` directly as column attribute

projects/gnrcore/packages/email/model/account.py
cgabriel: Yes, my worries were about an empty reply-to: header being added to the email, which maybe have caused issues on MUAs when the user would it 'reply'. But I've checked the lower implementation of sendmail, which gets added only if the value evaluates to true, so it's fine.

cgabriel (approved):

PR #318, opened 2025-11-29
feat: add double-click to lock buttons in MultiButton for mobile (#306)

### LINKED ISSUE #306: Multibutton shift key issue on mobile plaftorm
It has been reported that the MultiButton widget, use in table handler's sections, has the issue that it request the shift key usage when it's configured for multiple value. The shiftkey upon click is not possible on mobile/touch platform, causing the inability to make a multiple selection.

The reporter provided the following diff for genro_components:

```
diff --git a/gnrjs/gnr_d11/js/genro_components.js b/gnrjs/gnr_d11/js/genro_components.js
index cfa8763dd..bd80b6698 100644
--- a/gnrjs/gnr_d11/js/genro_components.js
+++ b/gnrjs/gnr_d11/js/genro_components.js
@@ -4379,14 +4379,17 @@ dojo.declare("gnr.widgets.MultiButton", gnr.widgets.gnrwdg, {
         var containerKw = {_class:'multibutton_container'};
         var btn_action;
         if (sticky){
+            var single_section = true
             btn_action = function(_kwargs){
                 let event = _kwargs.event;
                 var sn = event.target?genro.dom.getBaseSourceNode(event.target):null;
                 if(sn){
                     var mcode = sn.getInheritedAttributes()['multibutton_code'];
                     if(mcode){
-                        if(event.shiftKey && multivalue){
-                            var prevmcode = sourceNode.getRelativeData(value);
+                        if(multivalue){
+                            var prevmcode = single_section ? '' : sourceNode.getRelativeData(value);
+                            single_section = _kwargs['single_section'] ?? false
+                            if (single_section) {prevmcode = ''}
                             if(prevmcode){
                                 prevmcode = prevmcode.split(',');
                                 var j = prevmcode.indexOf(mcode);
```

This could be a possible solution, by removing the shift key or, otherwise, we should handle it differently with different events when used on mobile/touch.

genro:
## Summary

Fixes #306 - MultiButton requires shift+click for multiple selection, impossible on mobile

### Problem
The MultiButton widget supports multiple selection but requires `shift+click`, which is not available on mobile/touch devices.

### Solution
Implemented **double-tap to "lock" buttons**:
- Double click on a button makes it "sticky" (locked)
- Locked buttons remain selected even when clicking on other buttons
- Another double click unlocks the button
- Visual indicator: **underlined text** for locked buttons

## Changes

- `gnrjs/gnr_d11/js/genro_components.js`: 
  - Added `getLockedCodes()` function to get codes of locked buttons
  - Added `btn_dblaction` handler for toggling `data-locked` attribute
  - Modified `btn_action` to include locked buttons in selection

- `gnrjs/gnr_d11/css/gnrbase.css`:
  - Added style for buttons with `data-locked="true"` (text-decoration: underline)

## Test plan

@fporcari could you please test this?

- [ ] Verify that double click on a button locks it (underlined)
- [ ] Verify that locked buttons remain selected when clicking on others
- [ ] Verify that another double click unlocks the button
- [ ] Test on mobile/touch device

### DIFF
--- a/gnrjs/gnr_d11/css/gnrbase.css
+++ b/gnrjs/gnr_d11/css/gnrbase.css
@@ -4542,6 +4542,13 @@ p {
 .multibutton_selected.multibutton:hover{
 	text-decoration:none;
 }
+
+/* Locked button style (double-click to lock/unlock for multivalue selection) */
+.multibutton[data-locked="true"] .multibutton_caption{
+    text-decoration: underline;
+    text-underline-offset: 2px;
+}
+
 .multibutton_selected.multibutton[disabled='true']{
     opacity: 1;
     cursor: not-allowed;
--- a/gnrjs/gnr_d11/js/genro_components.js
+++ b/gnrjs/gnr_d11/js/genro_components.js
@@ -4378,6 +4378,25 @@ dojo.declare("gnr.widgets.MultiButton", gnr.widgets.gnrwdg, {
         sourceNode.attr.items = items;
         var containerKw = {_class:'multibutton_container'};
         var btn_action;
+        var btn_dblaction;
+
+        // Helper function to get all locked button codes
+        var getLockedCodes = function(){
+            var lockedCodes = [];
+            var mb = gnrwdg.multibuttonSource;
+            if(mb){
+                mb.forEach(function(n){
+                    if(n.domNode && n.domNode.dataset.locked === 'true'){
+                        var code = n.attr.multibutton_code || n.attr[gnrwdg.identifier] || n.attr['code'] || n.label;
+                        if(code){
+                            lockedCodes.push(code);
+                        }
+                    }
+                });
+            }
+            return lockedCodes;
+        };
+
         if (sticky){
             btn_action = function(_kwargs){
                 let event = _kwargs.event;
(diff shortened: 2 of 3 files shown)


fporcari (changes requested):
## Code Review for PR #318

Thanks for this PR! The double-tap to lock approach is an elegant solution for mobile multi-selection. The implementation is clean and maintains backward compatibility with the existing shift+click behavior.

### Issues Found

**P1 - Critical: Selection loss when unlocking the last button**

In the `btn_dblaction` handler, when the last locked button is unlocked, the selection collapses to only the last element:

```javascript
}else if(currentValue){
    // No locked buttons, keep only the last selected
    var currentCodes = currentValue.split(',');
    if(currentCodes.length > 0){
        sourceNode.setRelativeData(value, currentCodes[currentCodes.length-1]);
    }
}
```

This could cause unexpected data loss if the user had multiple selections (e.g., via shift+click) before unlocking. I'd suggest removing this `else` branch entirely and keeping the current selection unchanged when there are no locked buttons.

**P2 - Behavior clarification needed: single-click with locked buttons**

When locked buttons exist and the user single-clicks a new button, the new button is added to the locked ones, but any previously selected (non-locked) buttons are lost. Is this the intended behavior? If so, it might be worth documenting.

### Suggestions

1. **Touch optimization**: Consider adding `touch-action: manipulation` to `.multibutton_container` to prevent zoom conflicts on mobile devices.

2. **Accessibility**: Adding `aria-pressed="true"` to locked buttons would improve screen reader support.

Overall, the approach is solid. Just the P1 issue should be a

cgabriel:
> ## Code Review for PR #318
> Thanks for this PR! The double-tap to lock approach is an elegant solution for mobile multi-selection. The implementation is clean and maintains backward compatibility with the existing shift+click behavior.


Do have any feedback to provide about the Test Plan requested by the PR opener @genro ?

dgpaci:
I'm adding a test case to this branch, I tried on a mobile device and the result is totally different from the expected behaviour (in particular, no multiple selection is possibile, exactly as it was before). Since no particular improvement has been introduced, I suggest to close this PR and try finding a solution for the issue from a more global point of view.

cgabriel:
Closing the PR as requested, leaving the branch with the new test case.

PR #358, opened 2025-12-15
fix: prevent duplicate service_name parameter in storage handler

### LINKED ISSUE #357: Storage handler: duplicate service_name parameter in getService() call
## Description

After PR #331 and #349, there's an error when calling `addAttachment` in a mail sending procedure:

```
getService() got multiple values for keyword argument 'service_name'
```

The error occurs in `gnrstoragehandler.py` at line 452.

## Root Cause

In the `LegacyStorageHandler.storage()` method (line 416-456), at line 452 the code calls:

```python
return self.site.getService(
    service_type='storage',
    service_name=storage_name,  # Passed explicitly
    **service_params              # Unpacked from dict
)
```

The problem occurs when `service_params` (which comes from `stored_params` merged with `kwargs`) **already contains a `service_name` key**. This causes `service_name` to be passed twice:
1. Once as an explicit argument: `service_name=storage_name`
2. Again in the unpacking: `**service_params`

This can happen in two scenarios:

### Scenario 1: Database parameters
If in the `sys.service` table, the `parameters` field (Bag/XML) contains a `service_name` key, it gets loaded into `stored_params` via `_loadStorageParametersFromDb()` (line 191-218) and then expanded at line 452.

### Scenario 2: Kwargs with service_name
If the `storage()` method is called with `service_name` in kwargs (e.g., `storage('mystore', service_name='xyz')`), this goes into `kwargs`, gets added to `service_params` at line 450, and then gets passed twice.

## Solution

Before unpacking `service_params`, remove any keys that are already being passed explicitly to avoid conflicts:

```python
# Remove parameters that are passed explicitly to avoid conflicts
service_params.pop('service_name', None)
service_params.pop('service_type', None)

return self.site.getService(
    service_type='storage',
    service_name=storage_name,
    **service_params
)
```

## Location

`gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py`
Line: 452

fporcari:
## Summary
- Add `trigger_onInserted` to `sys.service` model to update `storage_params` when a new storage service is added to the database
- Pop `service_name` from `service_params` before passing to `getService()` to avoid duplicate keyword argument error
- Add tests for service_name parameter handling

Fixes #357

## Test plan
- [x] Run existing storage handler tests
- [x] Add new tests for `service_name` parameter handling
- [x] Verify all 588 tests pass

### DIFF
--- a/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py
+++ b/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py
@@ -416,13 +416,15 @@ class LegacyStorageHandler(BaseStorageHandler):
     def storage(self, storage_name, **kwargs):
         """Get a storage service by name using stored parameters.
 
-        Concrete implementation that retrieves parameters from storage_params registry
-        and uses ServiceHandler to get/create the actual storage service instance.
+        Retrieves parameters from storage_params registry and uses ServiceHandler
+        to get/create the actual storage service instance.
 
         All storage services are pre-loaded at initialization and kept in sync
-        via database triggers. If storage_name is not found in storage_params,
-        falls back to legacy behavior: creates a local storage with storage_name
-        as subdirectory of site_static_dir.
+        via database triggers (onInserted, onUpdated, onDeleted).
+
+        If storage_name is not found in storage_params, falls back to legacy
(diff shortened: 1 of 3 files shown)


dgpaci (approved):
After testing related issue #357, I confirm the problem seems to be solved

cgabriel (approved):

PR #360, opened 2025-12-16
Fix potential mutation of readOnlyAttrs dictionary in addPackage

fporcari:
## Summary
- Fix dictionary mutation issue when passing `readOnlyAttrs` to `addPackage` during required_packages iteration
- Create a copy of the dictionary with `dict(readOnlyAttrs)` to prevent cross-contamination between package configurations
- Initialize `readOnlyAttrs` as empty dict instead of None to ensure `dict()` always works

## Test plan
- [x] Existing tests pass
- [x] Verify that packages with `readOnly=True` propagate the attribute correctly to required packages
- [x] Verify that modifications to `pkgattrs` in `addPackage` don't affect other iterations

### DIFF
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -1029,9 +1029,9 @@ class GnrApp(object):
             attrs['path'] = self.realPath(attrs['path'])
         apppkg = GnrPackage(pkgid, self, **attrs)
         apppkg.content = pkgcontent or Bag()
-        readOnlyAttrs = {'readOnly':True} if attrs.get('readOnly') else None
+        readOnlyAttrs = {'readOnly':True} if attrs.get('readOnly') else dict()
         for reqpkgid in apppkg.required_packages():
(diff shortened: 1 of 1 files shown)


dgpaci:
This PR fixes the issue described in #361 where auxInstance packages fail to import due to dictionary mutation.

dgpaci (approved):
## Test Results

I tested the situation described in issue #361, where the presence of an auxInstance caused a package import error:

```
Error code GNR-001: Cannot import package er_gov from /home/ubuntu/genropy/projects/gnrcore/packages/er_gov/main.py: [Errno 2] No such file or directory: '/home/ubuntu/genropy/projects/gnrcore/packages/er_gov/main.py'.
```

## Result

With the changes introduced in this PR, the error no longer occurs. The fix that creates a copy of the `readOnlyAttrs` dictionary with `dict(readOnlyAttrs)` effectively solves the mutation problem that was causing cross-contamination between package configurations.

## Conclusion

✅ The PR fixes the documented bug
✅ Existing tests pass
✅ The solution is clean and minimal

**Approved**

cgabriel (approved):

Issue #274, opened 2025-11-05
[mkthresource] `-m` option overwrites entire menu instead of adding entry


EdoFede:
## Description
When using the `-m` option with `gnr dev mkthresource` command to automatically create a menu entry for a new resource, the command overwrites the entire `menu.py` file instead of appending the new entry to the existing menu structure.

## Steps to Reproduce
1. Create an application with an existing `menu.py` containing multiple menu entries
2. Execute: `gnr dev mkthresource -g -m <instance>:<package>.<table>`
3. Check the `menu.py` file

## Expected Behavior
The command should append the new resource entry to the existing menu structure, preserving all existing menu branches and entries.

**Example of expected menu structure after command execution:**
```python
class Menu(object):
    def config(self,root,**kwargs):
        recstudio = root.branch(u'!![en]Recording Studio')
        recstudio.thpage(u'!![en]Projects', table='recstudio.project')
        recstudio.webpage(u"!![en]Calendar", filepath='/recstudio/calendar')
        recstudio.webpage(u"!![en]Calendar (Worker)", filepath='/recstudio/calendarWorker')
        recstudio.thpage(u'!![en]Customers', table='er_core.cliente')
        recstudio.thpage(u'!![en]Jobs', table='recstudio.job')
        recstudio.thpage(u'!![en]Quotes', table='recstudio.quote')
        recstudio.thpage(u'!![en]Rates', table='recstudio.rate')
        recstudio.thpage(u'!![en]Studios allocation', table='recstudio.studio_allocation')
        recstudio.thpage(u'!![en]Workers allocation', table='recstudio.worker_allocation')
        recstudio.lookupBranch('!![en]Setup tables', pkg='recstudio')
        # New entry added here or in appropriate branch
        auto = root.branch(u"auto")
        auto.thpage(u"!!Worker", table="recstudio.worker")
```

## Actual Behavior
The entire `menu.py` file is overwritten with only the new resource entry.

**Example of actual result after executing `gnr dev mkthresource -g -m edodevel:recstudio.worker`:**
```python
class Menu(object):
    def config(self,root,**kwargs):
        auto = root.branch(u"auto")
        auto.thpage(u"!!Worker", table="recstudio.worker")
```

## Impact
- All existing menu entries are lost
- Requires manual restoration of the menu structure from version control
- Makes the `-m` option unsafe to use in projects with established menu structures

## Suggested Fix
The command should:
1. Check the existence of `menu.py` file
[...]

PR #366, opened 2025-12-23
Don't overwrite menu.py when adding a new resource via 'gnr dev mkthresource'

cgabriel:
When creating a resource and using the '-m' switch to create the menu item, check if the menu.py already exists, and if so, parse it with ast to find the variable name used for the root branch, find the line number where the method ends and append at the end of it all the package menu items for the newly created resources (they can be more than one)

refs #274

### DIFF
--- a/gnrpy/gnr/dev/cli/gnrmkthresource.py
+++ b/gnrpy/gnr/dev/cli/gnrmkthresource.py
@@ -147,7 +147,7 @@ class ThResourceMaker(object):
                         raise RuntimeError("Root branch not found")
                     insert_at = finder.config_func.end_lineno
                     for item in self.packageMenus[package]['auto']:
-                        new_line = f'        {finder.branch_var}.webpage(u"{item.attr['label']}", table="{item.attr['table']}")\n'
+                        new_line = f"        {finder.branch_var}.webpage(u'{item.attr['label']}', table='{item.attr['table']}')\n"
                         lines.insert(insert_at, new_line)
                         insert_at += 1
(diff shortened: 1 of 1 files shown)


EdoFede (changes requested):
Now the entire menu is no longer overwritten as before, great!

But it seems two problems have been introduced:

1. When creating a th resource for a table (`gnr dev mkthresource -g -m edodevel:recstudio.worker` in my example from issue #274), a `webpage` type entry is inserted in the menu instead of `thpage`.
2. The line is appended without a newline (it seems that the newline character is placed after the inserted line rather than before it)

```
...
        recstudio.lookupBranch('!![en]Setup tables', pkg='recstudio')        recstudio.webpage(u'!!Worker', table='recstudio.worker')

```

cgabriel:
> Now the entire menu is no longer overwritten as before, great!
> 
> But it seems two problems have been introduced:
> 
> 1. When creating a th resource for a table (`gnr dev mkthresource -g -m edodevel:recstudio.worker` in my example from issue #274), a `webpage` type entry is inserted in the menu instead of `thpage`.
> 2. The line is appended without a newline (it seems that the newline character is placed after the inserted line rather than before it)
> 
> ```
> ...
>         recstudio.lookupBranch('!![en]Setup tables', pkg='recstudio')        recstudio.webpage(u'!!Worker', table='recstudio.worker')
> ```


nice catch thank, the test case for the newline issue was with a menu.py without an empty new line at the end.

EdoFede (approved):
Checked, works fine now!

PR #367, opened 2025-12-28
argument "aux_instance" in the menu thpage

dedalus2000:
The aux_instance argument in the menu’s thpage does not seem to work generating the error "gnr.sql.gnrsql_exceptions.GnrSqlMissingTable: !!Error code GNR-001 : Missing package providing table xxx.yyyy" at line 520 of gnrmenu.py.

If ```table``` is on the aux_instance, how can ```self._page.db.table(table)``` handle it?

This patch appears to make aux_instance work correctly.


fporcari (approved):
looks fine to me

Issue #372, opened 2026-01-07
[FEATURE] GNR_DB_NAME env var should always be used if defined.


cgabriel:
### Summary

GNR_DB_NAME env var doesn't get used if an instance defines <db name="">, and this is a problem for deployments re-using the same image.

### Problem or Need

if the GNR_DB_NAME env var is set, it should always be used for the value of the database name, regardless if it's set in the default instance config file or in the instance config itself, otherwise it won't allow reuse of application images for different deployments.

### Proposed Solution

GnrSqlApp should check if the variable exists and has a value, and if so, use such value ignoring everything else. The environment always must have precedence.

### Impact and Scope

This will allow to control the deployments of the same image for instance which defined a <db name> in their config through the environment.

PR #373, opened 2026-01-07
GNR_DB_NAME, if set, is used with precedence over configurations.

cgabriel:
refs #372


fporcari (approved):
Approved, but maybe we should extend for other variables

mbertoldi:
> Approved, but maybe we should extend for other variables

I agree, we should be able to  override all db variables

cgabriel:
I'm closing this, which is just a workaround, and will use the new feature from #383 so all variables can be used.

PR #387, opened 2026-01-08
feat(bag): add Docker-style template variable syntax with defaults

genro:
## Summary

- Add `TemplateDict` class for Docker-style variable substitution in Bag XML templates
- Support for default values using `{VAR:-default}` syntax
- Deprecate `$VAR` and `${VAR}` syntax with warning (still works but emits DeprecationWarning)

## New syntax

| Syntax | Description |
|--------|-------------|
| `{VAR}` | Variable (keeps placeholder if missing) |
| `{VAR:-default}` | Variable with default value |
| `$VAR`, `${VAR}` | **Deprecated** - converted to `{VAR}` with warning |

## Example

```xml
<!-- Old syntax (deprecated) -->
<db host="$DB_HOST" port="${DB_PORT}"/>

<!-- New syntax with defaults -->
<db host="{DB_HOST:-localhost}" port="{DB_PORT:-5432}"/>
```

```python
# If DB_HOST is provided but DB_PORT is not:
b = Bag(source, _template_kwargs=dict(DB_HOST="myserver"))
# Result: host="myserver", port="5432" (default)
```

## Backward compatibility

- Existing `{VAR}` syntax continues to work unchanged
- `$VAR` and `${VAR}` syntax still works but emits a deprecation warning

## Test plan

- [x] Test existing `{VAR}` syntax still works
- [x] Test `{VAR:-default}` syntax with and without provided values
- [x] Test `$VAR` syntax works but raises DeprecationWarning

### DIFF
--- a/gnrpy/gnr/core/gnrbag.py
+++ b/gnrpy/gnr/core/gnrbag.py
@@ -77,6 +77,48 @@ from gnr.core.gnrlang import file_types
 class AllowMissingDict(dict):
     def __missing__(self, key):
         return "{"+key+"}"
+
+class TemplateDict(dict):
+    """Dict for Docker-style variable substitution with defaults.
+
+    Supports:
+    - {VAR} - variable (keeps placeholder if missing)
+    - {VAR:-default} - variable with default value
+    - $VAR, ${VAR} - deprecated syntax (converted to {VAR} with warning)
+    """
+
+    PATTERN = re.compile(r'\{([^}:]+)(?::-([^}]*))?\}')
+    DOLLAR_PATTERN = re.compile(r'\$\{?([A-Za-z_][A-Za-z0-9_]*)\}?')
+
+    def substitute(self, source):
+        """Substitute variables in source string.
+
+        :param source: string with variable placeholders
+        :returns: string with variables substituted
+        """
+        import warnings
+
+        # Deprecation warning for $ syntax
+        if self.DOLLAR_PATTERN.search(source):
+            warnings.warn(
+                "Dollar syntax ($VAR, ${VAR}) is deprecated. "
+                "Use Docker-style {VAR} or {VAR:-default}",
+                DeprecationWarning,
+                stacklevel=4
+            )
+            source = self.DOLLAR_PATTERN.sub(r'{\1}', source)
+
+        def replacer(match):
+            var_name = match.group(1)
+            default = match.group(2)
+            if var_name in self:
+                return str(self[var_name])
+            elif default is not None:
+                return default
+            else:
+                return match.group(0)  # keep placeholder if missing
+
+        return self.PATTERN.sub(replacer, source)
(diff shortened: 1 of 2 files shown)


gnrpy/gnr/core/gnrbag.py
```diff
@@ -1992,7 +2034,7 @@ def _fromSource(self, source, fromFile, mode, _template_kwargs=None):
             # source = source.format_map(AllowMissingDict(_template_kwargs))
 
             if self._template_kwargs:
-                source = source.format_map(self._template_kwargs)
+                source = TemplateDict(self._template_kwargs).substitute(source)
```
cgabriel, line 2042: 'source' here is the source filename of the bag, with an absolute path, not its content. So the template is trying to replace keyword on the path name, not on his content.

gnrpy/tests/core/gnrbag_test.py
```diff
@@ -39,7 +39,28 @@ def test_template_kwargs(self):
         test_kwargs = dict(TESTVAR="admin", TESTVAR2="admin2")
         b = Bag(source, _template_kwargs=test_kwargs)
         assert b['name'] == "admin"
-        
+
+    def test_template_kwargs_with_default(self):
+        # Test Docker-style {VAR:-default} syntax
+        source = "<db host='{DB_HOST:-localhost}' port='{DB_PORT:-5432}'/>"
```
cgabriel, line 45: The test is working because a xml string is passed to it, while in the Bag implementation, a filename is passed.

cgabriel (changes requested):
The current changes won't work, the replacemente is not being made on the content of the file but rather on a file path, so no env variables are being used to replace the attributes values defined in the bag.

gnrpy/gnr/core/gnrbag.py
```diff
@@ -2034,6 +2034,11 @@ def _fromSource(self, source, fromFile, mode, _template_kwargs=None):
             # source = source.format_map(AllowMissingDict(_template_kwargs))
 
             if self._template_kwargs:
+                if fromFile:
+                    # Read file content before template substitution
+                    with open(source, 'r', encoding='utf-8') as f:
```
cgabriel, line 2039: The API is a little confused here, if the mode is XML and there are template_kwargs, they get substituted by opening and reading the file, then the "fromFile" is passed to gnr.core.gnrbaxml.BagFromXml, which decides to open the file based on the parameters, and it does env vars expansions as well. Why do we have 2 file open and 2 different method to handle the env variables expansion in bag?

cgabriel (changes requested):

genro:
Closing: the review correctly identified that template substitution was operating on the file path rather than on the file content. The existing env-var expansion in BagFromXml already covers the use case. This PR needs a full redesign and is no longer a priority.

Issue #379, opened 2026-01-08
[BUG] Memory Leak in Thread-Local Storage Dictionaries


fporcari:
## Problem Description

The thread-local dictionaries in `GnrWsgiSite` (`_currentPages`, `_currentRequests`, `_currentAuxInstanceNames`) grow indefinitely because keys (thread IDs) are never removed after request completion.

**Observed behavior:** Dictionary entries accumulate over time, consuming memory.
**Expected behavior:** Thread-local entries should be cleaned up after each request completes.

## Reproduction Steps

1. Start a GenroPy site with multiple worker threads
2. Make many HTTP requests over time
3. Monitor memory usage or inspect `site._currentPages` dict size
4. Observed result: Dictionary size grows continuously, entries are never removed

## If It Cannot Be Reproduced Externally

- Module/feature involved: `GnrWsgiSite` request handling
- The `cleanup()` method sets `currentPage = None` but doesn't remove the thread ID key from dictionaries
- With thread pools (e.g., gunicorn workers), thread IDs are reused but old entries persist

## Environment

develop branch, all versions

## Additional Material

**Affected file:** `gnrpy/gnr/web/gnrwsgisite.py`

**Current code (line ~1335):**
```python
def cleanup(self):
    debugger = getattr(self.currentPage, 'debugger', None)
    if debugger:
        debugger.onClosePage()
    self.currentPage = None
    self.db.closeConnection()
```

**Proposed fix:**
```python
def cleanup(self):
    debugger = getattr(self.currentPage, 'debugger', None)
    if debugger:
        debugger.onClosePage()
    self.currentPage = None
    self.db.closeConnection()
    # Cleanup thread-local storage
    thread_id = _thread.get_ident()
    self._currentPages.pop(thread_id, None)
    self._currentRequests.pop(thread_id, None)
    self._currentAuxInstanceNames.pop(thread_id, None)
```

**Impact:** Long-running servers accumulate dead entries. With 100 threads and 1M requests = potentially millions of useless memory entries.

**Estimated effort:** 30 minutes

fporcari:
can be solved defining a 
class ThreadDict:
    def __init__(self):
        self._data = {}

    def get(self):
        return self._data.get(_thread.get_ident())

    def set(self, value):
        tid = _thread.get_ident()
        if value is None:
            self._data.pop(tid, None)
        else:
            self._data[tid] = value

in gnrlang

use ThreadDict for currentPage,currentRequest and currentAuxInstanceNames
[...]

PR #396, opened 2026-01-13
defined a new ThreadedDict object with automatic keys based on thread…

cgabriel:
… ident

use the new object to store thread-related data (requests, pages, auxinstances)

added thread-related storage cleanup in cleanup function

refs #379


fporcari (approved):
Looks fine

PR #400, opened 2026-01-14
Add readOnly support for checkboxcolumn and checkBoxSet

### LINKED ISSUE #399: Add readOnly support for checkboxcolumn and checkBoxSet
## Problem

Currently, checkbox columns in grids don't support a `readOnly` attribute to prevent user interaction in specific conditions. While `disabled` exists at the row level (checked via `currNode.attr.disabled`), there's no way to make individual checkboxes read-only based on row data or other dynamic conditions.

## Use Case

When displaying mixed data in a grid (e.g., regular records and virtual/computed records), it's necessary to prevent editing of checkbox columns for specific rows while keeping them editable for others.

Example: Displaying both manual assignments and automatic assignments (derived from other data) in the same grid, where automatic assignments should be read-only.

## Proposed Solution

Add support for a `readOnly` attribute in checkbox columns that:
1. Can be a boolean value (`true`/`false`)
2. Can be a path string using framework conventions:
   - `=#ROW.fieldname` - evaluates a field from the current row attributes
   - `^#FORM.path` - evaluates a form datapath
   - Any other datasource path

The check should be performed in `mixin_onCheckedColumn_one` before processing the click, similar to the existing `disabled` check.

## Implementation

Modified `gnr_d11/js/genro_grid.js` in `mixin_onCheckedColumn_one` (around line 3937):

```javascript
if (cellkw.readOnly) {
    var isReadOnly = cellkw.readOnly;
    if (typeof isReadOnly === 'string' && isReadOnly.startsWith('=#ROW.')) {
        var fieldName = isReadOnly.substring(6);
        isReadOnly = currNode.attr[fieldName];
    } else if (typeof isReadOnly === 'string' && (isReadOnly.charAt(0) === '=' || isReadOnly.charAt(0) === '^')) {
        isReadOnly = sourceNode.currentFromDatasource(isReadOnly);
    }
    if (isReadOnly) {
        return;
    }
}
```

### Python (gnrwebstruct.py):
```python
r.checkBoxSet('tag_pkeys', name='Permissions', cells_width='8em', columns=columns,
              cells_readOnly='=#ROW._is_readonly')
```

### Result:
Checkboxes are disabled for rows where `_is_readonly` field is `true`, while remaining interactive for other rows.

## Benefits

- Consistent with framework patterns (uses `=#ROW.` notation like other attributes)
- Generic solution that works with any datasource path
- Minimal code change
- No breaking changes to existing functionality

dgpaci:
## Summary

Add support for `readOnly` attribute in checkbox columns to prevent user interaction based on dynamic conditions (row data, form state, etc).

## Problem

Currently, checkbox columns don't support a `readOnly` attribute to conditionally prevent editing. While `disabled` exists at the row level, there's no way to make individual checkboxes read-only based on row-specific data.

## Solution

Added `readOnly` check in `mixin_onCheckedColumn_one` (genro_grid.js) before processing checkbox clicks.

The `readOnly` attribute supports:
- **Boolean values**: `true`/`false`
- **Path strings** using framework conventions:
  - `=#ROW.fieldname` - evaluates field from current row attributes
  - `^#FORM.path` - evaluates form datapath
  - Any other datasource path

## Implementation Details

Modified `gnr_d11/js/genro_grid.js` around line 3937:

```javascript
if (cellkw.readOnly) {
    var isReadOnly = cellkw.readOnly;
    if (typeof isReadOnly === 'string' && isReadOnly.startsWith('=#ROW.')) {
        var fieldName = isReadOnly.substring(6);
        isReadOnly = currNode.attr[fieldName];
    } else if (typeof isReadOnly === 'string' && (isReadOnly.charAt(0) === '=' || isReadOnly.charAt(0) === '^')) {
        isReadOnly = sourceNode.currentFromDatasource(isReadOnly);
    }
    if (isReadOnly) {
        return;
    }
}
```

## Usage Example

```python
# Python - gnrwebstruct.py
r.checkBoxSet('tag_pkeys', name='Permissions', 
              cells_width='8em', 
              columns=columns,
              cells_readOnly='=#ROW._is_readonly')
```

Result: Checkboxes are disabled for rows where `_is_readonly` is `true`.

## Benefits

- Consistent with framework patterns (uses `=#ROW.` notation)
- Generic solution that works with any datasource path
- Minimal code change (12 lines)
- No breaking changes to existing functionality
- Similar to `editDisabled` for other cell types

## Testing

Tested with mixed grid containing editable and non-editable rows based on row data flag.

Closes #399

### DIFF
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -3934,6 +3934,18 @@ dojo.declare("gnr.widgets.IncludedView", gnr.widgets.VirtualStaticGrid, {
         if (currNode.attr.disabled) {
             return;
         }
+        if (cellkw.readOnly) {
+            var isReadOnly = cellkw.readOnly;
+            if (typeof isReadOnly === 'string' && isReadOnly.startsWith('=#ROW.')) {
+                var fieldName = isReadOnly.substring(6);
+                isReadOnly = currNode.attr[fieldName];
+            } else if (typeof isReadOnly === 'string' && isReadOnly.charAt(0) === '=' || isReadOnly.charAt(0) === '^') {
+                isReadOnly = sourceNode.currentFromDatasource(isReadOnly);
+            }
+            if (isReadOnly) {
+                return;
+            }
+        }
         var newval = !checked;
         if(cellkw.radioButton){
             if(cellkw.radioButton===true){


fporcari (commented):
it seems seems some commit are not related to the topic of the PR

gnrpy/gnr/app/gnrlocalization.py
```diff
@@ -31,7 +31,7 @@
 from gnr.app import logger
 
 SAFEAUTOTRANSLATE = re.compile(r"""(\%(?:\((?:.*?)\))?(?:.*?)[s|d|e|E|f|g|G|o|x|X|c|i|\%])""")
-LOCREGEXP = re.compile(r"""("{3}|'|")\!\!(?:\[(?P<lang_emb>.{2})\])?(?:{(?P<key_emb>\w*)})?(?P<text_emb>.*?)\1|\[\!\!(?:\[(?P<lang>.{2})\])?(?:{(?P<key>\w*)})?(?P<text>.*?)\]|\b_T\(("{3}|'|")(?P<text_func>.*?)\6\)""")
```
fporcari, line 34: it should belong to another PR

gnrjs/gnr_d11/js/genro_widgets.js
```diff
@@ -5340,7 +5340,7 @@ dojo.declare("gnr.widgets.uploadable", gnr.widgets.baseHtml, {
                                                     }});
         let cropPage = dlg.center._('div',{width:boundaryWidth+'px',height:(boundaryHeight+40)+'px',position:'relative',margin:'10px'})._('div',{nodeId:frameCode+'_cropper',position:'absolute',top:0,bottom:0,left:0,right:0});
         var slotbar = dlg.bottom._('slotBar',{slots:'*,5,confirmImage,5'});
-        slotbar._('button','confirmImage',{label:'Confirm',
```
fporcari, line 5343: it should belong to another PR

fporcari (changes requested):
Please clean up this PR in order to check checkboxtext readOnly feature only

dgpaci:
You're right, it was a mistake. The branch started from the wrong place and included localization commits that had nothing to do with this PR.

I cleaned up the branch by cherry-picking only the correct commit from develop. Now the PR only contains the changes for readOnly checkbox support.

Ready for review! 👍

gnrjs/gnr_d11/js/genro_grid.js
```diff
@@ -3934,6 +3934,18 @@ dojo.declare("gnr.widgets.IncludedView", gnr.widgets.VirtualStaticGrid, {
         if (currNode.attr.disabled) {
             return;
         }
+        if (cellkw.readOnly) {
+            var isReadOnly = cellkw.readOnly;
+            if (typeof isReadOnly === 'string' && isReadOnly.startsWith('=#ROW.')) {
+                var fieldName = isReadOnly.substring(6);
+                isReadOnly = currNode.attr[fieldName];
+            } else if (typeof isReadOnly === 'string' && isReadOnly.charAt(0) === '=' || isReadOnly.charAt(0) === '^') {
```
fporcari, line 3942: Exist a method "isPointerPath"

fporcari (approved):

gnrjs/gnr_d11/js/genro_grid.js
cgabriel: @fporcari so it's approved as stated or changes are needed?

dgpaci:
The proposed solution is not optimal. An alternative solution will be found to implement #399

PR #403, opened 2026-01-14
feat: add JSON/XML body parsing methods to BaseRpc component

fporcari:
## Summary

Add explicit methods for parsing JSON and XML request bodies in the `BaseRpc` component, providing developers with fine-grained control over body parsing instead of automatic parsing at the framework level.

### New Methods

- **`get_request_body_json()`**: Parse request body as JSON
  - Returns `dict`/`list` or `None` on parsing errors
  - Only processes POST/PUT/PATCH requests with JSON Content-Type
  
- **`get_request_body_xml()`**: Parse request body as XML converted to Bag
  - Returns `Bag` object or `None` on parsing errors
  - Only processes POST/PUT/PATCH requests with XML Content-Type

Both methods are available in `BaseRpc`, `NetBagRpc`, and `XmlRpc` classes.

### Usage Example

```python
class GnrCustomWebPage(object):
    py_requires='gnrcomponents/externalcall:BaseRpc'

    @public_method
    def webhook_handler(self, **kwargs):
        json_data = self.get_request_body_json()
        if json_data:
            user_id = json_data.get('user_id')
            # Process data...
        else:
            return Bag({'error': 'Invalid or missing JSON body'})
```

### Benefits

- **Explicit over implicit**: Developers explicitly call methods when needed
- **No overhead**: No automatic parsing for requests that don't need it
- **Clean API**: No pollution of kwargs with `_json_body`/`_xml_body`
- **Control**: Developer decides when and how to parse the body

### Technical Details

- Silent error handling (returns `None` on failure)
- Content-Type detection via `mimetype` header
- XML parsing uses application catalog for type conversion
- Backward compatible (additive change only)

### DIFF
--- a/resources/common/gnrcomponents/externalcall.py
+++ b/resources/common/gnrcomponents/externalcall.py
@@ -90,6 +90,54 @@ class BaseRpc(BaseComponent):
     def rpc_error(self, method, *args, **kwargs):
         return 'Not existing method %s' % method
 
+    def get_request_body_json(self):
+        """
+        Parse request body as JSON.
+
+        Returns:
+            dict/list: Parsed JSON data, or None if parsing fails or body is empty
+        """
+        if getattr(self.request, 'method', None) not in ('POST', 'PUT', 'PATCH'):
+            return None
+
+        mimetype = (self.request.mimetype or '').lower()
+        is_json = getattr(self.request, 'is_json', False)
+        if not (is_json or 'json' in mimetype):
+            return None
+
+        return self.request.get_json(silent=True)
+
+    def get_request_body_xml(self):
+        """
+        Parse request body as XML and convert to Bag.
+
+        Returns:
+            Bag: Parsed XML as Bag object, or None if parsing fails or body is empty
+        """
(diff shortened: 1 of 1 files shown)


resources/common/gnrcomponents/externalcall.py
```diff
@@ -90,6 +90,54 @@ def rpc_index(self, *args, **kwargs):
...
+        """
+        Parse request body as JSON.
+
+        Returns:
+            dict/list: Parsed JSON data, or None if parsing fails or body is empty
+        """
+        if getattr(self.request, 'method', None) not in ('POST', 'PUT', 'PATCH'):
+            return None
+
+        mimetype = (self.request.mimetype or '').lower()
```
cgabriel, line 103: Can't this check just be avoided, since werkzeug's request.get_json() already checks for the correct mimetype itself? From docs:

```
If the mimetype does not indicate JSON
        (:mimetype:`application/json`, see :attr:`is_json`), or parsing
        fails, :meth:`on_json_loading_failed` is called and
        its return value is used as the return value. By default this
        raises a 415 Unsupported Media Type resp.
```

cgabriel:
Refusing this,  overridden by #408

PR #406, opened 2026-01-16
fix empty to_address

dedalus2000:
The "mail" service is unable to send emails with only BCC recipients when the TO field is empty.
This patch fixes the issue: smtplib accepts an empty TO field, but Genropy splits it by comma without checking whether it is empty.

Tested in production.

### DIFF
--- a/gnrpy/gnr/lib/services/mail.py
+++ b/gnrpy/gnr/lib/services/mail.py
@@ -349,7 +349,12 @@ class MailService(GnrBaseService):
         from_address = account_params['from_address']
         msg = self.build_base_message(subject, body, attachments=attachments, html=html, charset=charset)
         msg['From'] = from_address
-        msg['To'] = to_address
+        if to_address:
+            # to_address could be None
+            msg['To'] = to_address
+            if ',' in to_address:
+                to_address = to_address.split(',')
+
         headers_kwargs = headers_kwargs or {}
         message_id = message_id or headers_kwargs.pop('message_id',None)
         reply_to = reply_to or headers_kwargs.pop('reply_to',None)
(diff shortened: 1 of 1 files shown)


dgpaci (approved):
The fix correctly addresses the issue of sending emails with only BCC recipients when the TO field is empty/None.

The added null check prevents the TypeError when trying to use the 'in' operator on a None value, and the rest of the code flow (especially in _sendmail at line 413) already handles None values correctly.

**Minor observation** (non-blocking): At line 352,  sets the header even when to_address is None, which might set the header to the string "None". However, since this has been tested in production, it's apparently not an issue in practice.

If you wanted to be extra thorough, you could add:
```python
if to_address:
    msg['To'] = to_address
```

But given the production testing, the current fix is sufficient.

LGTM! ✓

dgpaci (approved):
The fix correctly addresses the issue of sending emails with only BCC recipients when the TO field is empty/None.

The added null check prevents the TypeError when trying to use the 'in' operator on a None value, and the rest of the code flow (especially in _sendmail at line 413) already handles None values correctly.

**Minor observation** (non-blocking): At line 352, msg['To'] = to_address sets the header even when to_address is None, which might set the header to the string "None". However, since this has been tested in production, it's apparently not an issue in practice.

If you wanted to be extra thorough, you could add:
```python
if to_address:
    msg['To'] = to_address
```

But given the production testing, the current fix is sufficient.

LGTM!

cgabriel:
> **Minor observation** (non-blocking): At line 352, msg['To'] = to_address sets the header even when to_address is None, which might set the header to the string "None". However, since this has been tested in production, it's apparently not an issue in practice.
> 
> If you wanted to be extra thorough, you could add:
> 
> ```python
> if to_address:
>     msg['To'] = to_address
> ```

@dedalus2000 are you going to implement this?

dedalus2000:
@dgpaci  right: I use Gmail, but some other email providers might reject the message.

I have now updated the patch and tested it in production.

Note: it seems to me that sendmail_many has the same “empty to-address” issue (depending on multiple_mode). It uses handle_addresses, which could also be useful for sendmail. This is not handled in this pull request.

cgabriel (approved):

dgpaci (approved):
Correct and minimal solution. Fixes the crash when to_address is None, allowing emails to be sent only to BCC recipients. This is compatible with `_sendmail`, which accepts None/strings/lists.

PR #412, opened 2026-01-16
fix: normalize locale and language format across codebase

fporcari:
## Summary

- `language` property is now **read-only**, derived from `avatar.language` or `locale_language`
- `language` is always normalized to **lowercase** (e.g., `it`, `en`)
- `locale` is normalized to `xx_XX` format when complete (e.g., `it_IT`), or lowercase when 2-char only
- Replaced all `locale.split('-')[0]` patterns with `locale[:2]` or `self.language`
- Removed redundant `.lower()` calls where values are already normalized
- Simplified `TableScriptToHtml` to delegate `locale`/`language` to page

## Files Changed

| File | Change |
|------|--------|
| `gnrwebpage.py` | `language` read-only property, `locale` normalization |
| `gnrbaseclasses.py` | Simplified `TableScriptToHtml` |
| `gnrsqlmodel.py` | Removed redundant `.lower()` |
| `gnrdate.py`, `gnrrlab.py` | `locale[:2]` instead of `split('-')` |
| `th_tree.py`, `translate_fields.py` | Removed redundant `.lower()` |
| `utils.py` | `locale[:2]` instead of `split('-')` |
| `docserver.py`, `index.py`, `viewer.py` | Use `self.language` |

## Test plan

- [x] All 699 tests pass
- [ ] Manual testing of locale/language in web pages
- [ ] Verify localized columns work correctly

### DIFF
--- a/gnrpy/gnr/core/gnrdate.py
+++ b/gnrpy/gnr/core/gnrdate.py
@@ -62,8 +62,8 @@ def checkDateKeywords(keywords,datestr,locale):
 def monthsFromDateRange(date_start=None,date_end=None,locale=None):
     curr_date = datetime.date(date_start.year,date_start.month,1)
     result = []
-    if locale and '-' in locale:
-        locale = locale.split('-')[0]
+    if locale and len(locale) > 2:
+        locale = locale[:2]
     months = dates.get_month_names(width='wide', locale=locale)
     while curr_date<=date_end:
         m = curr_date.month
--- a/gnrpy/gnr/core/gnrrlab.py
+++ b/gnrpy/gnr/core/gnrrlab.py
@@ -46,7 +46,7 @@ class RlabResource(object):
             self.locale = locale or '{language}-{languageUPPER}'.format(language=self.language,
                                         languageUPPER=self.language.upper())
         elif self.locale:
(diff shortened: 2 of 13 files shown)


dgpaci (approved):
Tested on sandbox: changing the user language preference correctly updates field labels to the selected language. The normalization logic works as expected.

## Approved ✓

The PR successfully normalizes locale and language formats across the codebase, making the handling more consistent and predictable.

## Suggestions for future improvements

While the current implementation works correctly, there are two areas that could be improved:

1. **User Form locale field**: The `locale` field in the user form is currently a text input. It should be a `filteringSelect` with the available locale values to prevent invalid entries and improve UX.

2. **Hardcoded locale values**: The available locale values in user preferences (en_GB, en_US, en_AU, it_IT) appear to be hardcoded. These should be:
   - Defined in the `adm.language` table (which currently only stores the language code)
   - Retrieved dynamically from that table rather than being hardcoded
   
   This would make it easier to add new locales without code changes and provide a single source of truth for supported languages/locales.

These improvements can be addressed in a follow-up PR and don't block this fix.

cgabriel (approved):

PR #413, opened 2026-01-16
Fix counter initialization for packages without tables

fporcari:
## Summary
- Handle the case where a package has no tables during counter sequence initialization
- Prevents error when `pkgobj['tables']` returns None

## Changes
- Added early return check in `initializePackageSequences` method when package has no tables
- Simplified the table iteration logic

## Test plan
- [x] Verify counter initialization works for packages with tables
- [ ] Verify no error occurs for packages without tables during initialization


cgabriel (approved):
Was not able to reproduce the issue, so I think the fix is correct.

PR #419, opened 2026-01-21
Export DHZ content to XLSX as string problem

dedalus2000:
In a 1:N relationship (such as Customers:Invoices), dragging a DHZ column onto the view causes an error during Excel export: the content is a string, but it is treated as a DHZ and therefore Genropy attempts to remove the timezone.

This patch checks the content: if it is a string, it skips timezone handling.

err1
err2


cgabriel (approved):
Looking good to me

dedalus2000:
Note this detail: the patch fixes an issue in the aggregator. In fact the ORM produces a DHZ column, but the aggregator produces a string with the DHZ values separated by commas.
The Excel exporter should probably be told that this field is a string and not a DHZ.

I think this patch is a simple a fast way to solve this problem

fporcari (approved):
locks good to me

PR #422, opened 2026-01-22
Fix: Parametric query dialog loses content in nested form structures

### LINKED ISSUE #174: Parametric query in webpages with a particular structure
Parametric query (such as "?Name" in query editor, showing a dialog where it is possible to specify a value) show a weird behaviour when used in particular webpages built as follows:
- a webpage with a thFormHandler
- inside the called Form, a th with possibility to run a query

This structure is replicated in Sandbox, branch **tests/parametric_query**, where I created:
- a webpage sandbox/clienti, with a thFormHandler showing the Form for the first cliente in db and Form ParametricQueryBugForm
- a cliente Form Resource "ParametricQueryBugForm", where fatture are shown with `extendedQuery` on

If a query with Protocollo ?Number is run, a dialog appears:

Image

When I write something and push "Confirm" or simply push tab button, the content goes blank. 

Image

If something is forced into the query where path through inspector, everything's fine:
`main.form.fatt_fattura.view.query.where.c_0`

dgpaci:
## Problem

Fixes #174

When using parametric queries (with `?Parameter` syntax) in a `plainTableHandler` inside a form resource that is itself inside a `thFormHandler`, the query parameter dialog would appear correctly, but after entering a value and pressing Confirm or Tab, the content would go blank.

### Structure that causes the issue:
- Webpage with `thFormHandler`
- Form resource with a `plainTableHandler` using `extendedQuery=True`
- Parametric query (e.g., `Protocollo ?Number`)

### Observed behavior:
1. Query parameter dialog appears ✓
2. User enters value in the textbox
3. User presses Confirm or Tab
4. Content goes blank (dialog shows empty) ✗

## Root Cause Analysis

Through extensive debugging with console logging, we discovered the issue was **NOT** related to unstable datapath in nested forms as initially suspected. The datapath remained stable throughout. 

The actual problem is **widget interference through shared datapath binding**:

### The Interference Mechanism:

1. **Two widgets bind to the same datapath** (`main.form.fatt_fattura.view.query.where.c_0`):
   - Main query builder `searchboxTextbox` (with `validate_onAccept` handler calling `checkQueryLineValue`)
   - Parametric dialog `textbox` (without handler)

2. **When user types in the dialog textbox**, it writes to the shared datapath

3. **The main query builder searchboxTextbox detects the change** in its bound datapath and automatically triggers its `validate_onAccept` handler (even though the user didn't interact with it directly)

4. **The handler calls `checkQueryLineValue()`** passing the searchboxTextbox's cached value (`?Parameter`)

5. **`checkQueryLineValue()` sees the `?` prefix and sets the value to `null`**:
   ```javascript
   if (value.indexOf('?') == 0) {
       sourceNode.setRelativeData(relpath, null);  // Mark as parametric placeholder
   }
   ```
   This is the intended behavior for marking parametric placeholders in the query builder.

6. **Result**: User's input is immediately overwritten with `null`

### Why the Original TempPath Approach Was Rejected

The initial PR proposed using a temporary isolated datapath but was rejected because:
- The reviewer (correctly) wanted to understand the root cause
- "In nested form structures, this wherepath becomes unstable" was the stated reason
- The reviewer wanted to find the bug reason before applying a workaround

### What We Discovered

- The wherepath is **NOT unstable** in nested forms (confirmed via logging)
- The problem is `validate_onAccept` being triggered on the wrong widget
- The tempPath approach is **NOT a workaround** but the **correct architectural solution**

## Solution

Use a temporary isolated datapath (`gnr.temp_querypars_XXX`) for the parametric dialog:

### Implementation in `buildParsDialog`:

1. **Create temporary datapath**: Generate unique path using `genro.time36Id()`
2. **Copy current values**: Copy existing where clause data to temporary path
3. **Clean parametric markers**: Remove `?Parameter` values (these are placeholders, not initial values)
4. **Isolate editing**: User edits values in the isolated temporary context
5. **Copy back on confirm**: Copy values from temporary path back to wherepath
6. **Run query**: Execute with properly saved values

### Code Changes:

```javascript
buildParsDialog:function(parslist) {
    var sourceNode = this.sourceNode;
    var that = this;

    // Use a temporary isolated datapath to avoid interference
    var tempPath = 'gnr.temp_querypars_'+genro.time36Id();
    var dlg = genro.dlg.quickDialog('Complete query',{datapath:tempPath,...});

    // Copy and clean
    var whereData = sourceNode.getRelativeData(this.wherepath);
    genro.setData(tempPath, whereData.deepCopy());
    var tempData = genro.getData(tempPath);
    
    // Remove ?Parameter markers
    for (var i = 0; i < parslist.length; i++) {
        var relpath = parslist[i].relpath;
        var currentValue = tempData.getItem(relpath);
        if (currentValue && currentValue.indexOf('?') === 0) {
            tempData.setItem(relpath, null);
        }
    }

    var confirm = function(){
        // Copy back to wherepath
        var tempConfirmData = genro.getData(tempPath);
        sourceNode.setRelativeData(that.wherepath, tempConfirmData);
        that.runQuery()
        dlg.close_action();
    };
    
    // ... dialog setup
}
```

## Why This Solution is Correct

The tempPath provides proper architectural isolation:

✅ **Main query builder** monitors `main.form.fatt_fattura.view.query.where.c_0`
✅ **Dialog** edits in `gnr.temp_querypars_XXX.c_0` (isolated)
✅ No interference between the two contexts
✅ `validate_onAccept` only triggers on actual query builder interactions
✅ User input is preserved during dialog editing
✅ Values copied to wherepath only on Confirm

This is not a workaround but a correct separation of concerns between:
- The persistent query builder state (monitored by UI widgets)
- The temporary parameter input dialog (isolated editing context)

## Testing

Verified with test case in sandbox repository, branch `tests/parametric_query`:

1. Navigate to `/sandbox/clienti` webpage
2. The page shows a `thFormHandler` with cliente form containing a table of fatture
3. Create parametric query: `Protocollo ?Number`
4. Run query - parameter dialog appears
5. Enter value (e.g., `123`)
6. Press Confirm
7. ✅ Query executes correctly with entered value
8. ✅ No blank dialog
9. ✅ Results filtered properly

## Request for Re-Review

@fporcari This PR now includes:
- Full root cause analysis showing the issue is widget interference, not unstable paths
- Justification for why tempPath is the correct architectural solution
- Clean implementation without workarounds
- Comprehensive explanation of the mechanism

The tempPath approach provides proper isolation and is the standard pattern for dialog-based editing in Genropy. Could you please re-review?


fporcari (changes requested):
@dgpaci "In nested form structures (form inside form), this wherepath becomes unstable" 
we need to find the reason of this behavior.
this code is very old, maybe we can use prompt dialog instead of parsdialog
anyway we should find the bug reason before find a workaround

resources/common/th/th_querytool.js
```diff
@@ -818,9 +818,31 @@ dojo.declare("gnr.QueryManager", null, {
     
     buildParsDialog:function(parslist) {
         var sourceNode = this.sourceNode;
-        var dlg = genro.dlg.quickDialog('Complete query',{datapath:this.wherepath,width:'250px',autoSize:true});
         var that = this;
+
+        // Use a temporary isolated datapath to avoid interference with the main query builder
```
fporcari, line 823: have you considered to use genro.dlg.prompt or genro.dlg.askParameters instead of quickDialog?
genro.dlg.askParameters(function(_askResult){
                sourceNode.setRelativeData(wherepath,_askResult);
            },{title:_T('CompleteQUery'),fields:runQueryPars},d,this.sourceNode);

fporcari (commented):
try to use genro.dlg.prompt or genro.dlg.askParameters it would provide the same pattern withouth defining specal tempPath

resources/common/th/th_querytool.js
cgabriel: @fporcari could you please complete the PR yourself since your suggestion is a little obscure?

cgabriel:
Superseded by #505

PR #426, opened 2026-01-22
feat: Multidomain Workspace Mode - Isolated tenants under single instance

### LINKED ISSUE #346: [FEATURE] Multidomain Workspace Mode - Isolated tenants under single instance
## Overview

This proposal introduces a **multidomain** capability for Genropy applications, allowing multiple completely isolated workspaces (tenants) under a single application instance, accessible via URL paths rather than separate deployments.

## Problem Statement

Current Genropy multitenancy options:
1. **Separate instances**: One server per customer (high infrastructure cost)
2. **Classic multidb**: Shared instance with data synchronization (not truly isolated)

Neither fits SaaS requirements where each customer needs independent databases, users, and sessions without synchronization overhead.

### URL Structure
- `app.example.com/customer1/` → Customer workspace (isolated)
- `app.example.com/customer2/` → Customer workspace (isolated)
- etc.

### Master/Admin Instance
The **master/admin workspace** will be managed by a **separate application instance** connecting to the same database cluster (e.g., `instances.example.com` or `admin.example.com`). This provides:
- Clean architecture without special `_main_` handling within the multidomain app
- Clear separation of responsibilities
- Independent deployment of admin functionality
- Enhanced security through physical isolation between admin and tenants

### Key Features
- **Complete session isolation**: Separate cookie namespaces per workspace
- **Independent databases**: Each workspace has its own database
- **Startup data copy**: Initial data copied from template (no automatic sync)
- **Per-domain site register**: Isolated register instances

### Files to Modify
- `gnrwsgisite.py`: Domain/path routing for workspaces
- `gnrsql.py`: Multidomain property handling
- `gnrwebpage.py`: Session management per workspace
- `gnrsiteregister.py`: Per-domain register instances
- `multidb` package: Startup data methods

## Open Questions

1. **Site register isolation**: Best approach for per-workspace registers
2. **Service handlers**: Shared vs. per-workspace strategy
3. **Preferences isolation**: Scope of preference separation
4. **Background jobs**: Compatibility with PR #310
5. **Startup data updates**: Strategy for updating tenant data after initial copy

## Implementation Status

A working prototype exists in the `feature/multidomain-workspace` branch. This formal proposal aims to document requirements, resolve design questions, and reimplement with cleaner architecture.

fporcari:
## Summary
- Implement multidomain capability allowing multiple isolated workspaces (tenants) under a single application instance
- Add `GnrDomainProxy` and `GnrDomainHandler` classes for managing per-domain isolated resources (register, services, storage)
- Introduce `DbStoreRouter` for unified URL routing supporting both standard and multidomain modes
- Isolate authentication cookies per domain with workspace-specific paths
- Add JavaScript `buildContextUrl()` helper for proper URL generation in multidomain context
- Support per-domain application preferences with isolated cache keys

### Thread-safety and domain isolation fixes
- Add lock protection to `GnrDomainHandler.__contains__` and `__getitem__` to prevent TOCTOU race conditions
- Pass explicit domain parameter to `LegacyStorageHandler` to ensure correct domain context during lazy initialization
- Update `BaseStorageHandler` to use explicit domain in `_loadStorageParametersFromDb` and `updateStorageParams`
- Restore multidb compatibility for `use_dbstores` check (fix regression)

## Linked Issue
Closes #346

## Test plan
- [x] Test suite passes (699 tests)
- [x] Manual testing of multidomain routing
- [x] Verify cookie isolation between workspaces
- [x] Test backward compatibility in single-domain mode
- [x] Verify multidb works correctly without multidomain enabled
- [x] Verify preference isolation per domain
- [x] Verify storage handler uses correct domain context under concurrent access


cgabriel (changes requested):
I'm not able to activate a new dbstore (using sandbox-md instance), data is not copied upon initialization.

gnrpy/gnr/web/gnrdaemonprocesses.py
cgabriel, line 1: The new gnr task scheduler has not been updated to reflect this changes.

cgabriel (changes requested):
I'm trying to summarize here my comments, we can split them into specific issue if needed:

1. using the definition 'multidomain' is IMHO misleading - the common meaning for 'domain', related to web applications, is pretty precise, while the goal of this feature is to run multiple tenants under a single domain name. It would be much better to use a less-misleading term. The whole things is related to tenancy rather the domains, which the goal to serve everything under just one single domain. I know that naming things is second most difficult thing in IT, but the room for improvement here is easy

2. I was not initially able to use the feature, because there is no documentation at all, and I've discovered that there's a lot to do on the application which uses this feature in order to be able to use. I've checked the changes made to genropy/sandbox@sandbox-md in order to use the feature, and I'm pretty sure it needs documentation and/or components in the framework to help introducing the feature in applications. Maybe the multidb package, which apparently is a requirements to use the feature, could provide the needed?

4. There is no validation in name/dbstore when creating sub-tenants, so a name that correspondes to static tools or similar in the dispatcher can be used, leading to unuseable tenants.

3. Due to the fact that this approach needs load balancing to scale up computing resources, rather than a simple name-based one, I've conducted several tests to verify path based balancing, in particular, with several different hosts serving the application:

  1. one host serving common static files, 2 hosts handling [a-m]* and [n-z]* tenants
  2. 3 hosts serving everything in round-robin scheme ( used in auto-scaling)
  3. 2 hosts serviing everything except a single tenant, handle by a specific host (used in auto-scaling + static resource allocation)
  
All these tests were positive!

fporcari (commented):
Thanks for the thorough review and especially for the load balancing tests — great to know all scenarios passed.

Addressing each point:

**1. dbstore data initialization**
This is not in scope for this PR. The dbstore data initialization is a concern of the multidb package in general (the relevant PR is the StoreTable refactoring). The decision was made earlier to provide hooks and utilities in the `StoreTable` class, leaving it up to the application developer to decide *how* to initialize the data (via templatedb, copy from mainstore, or other strategies). See `multidb_setStartupData` in the StoreTable lib.

**2. Task scheduler not updated**
Acknowledged, I'll update it. Is there any documentation or specific guidance on how to adapt the new task scheduler for this? Any pointers would be appreciated.

**3. "Multidomain" naming**
Our benevolent dictator @genro imposed the name (personally I would have chosen "workspace"). Technically this is a multi-tenant with multidb where the option `multidomain=True` is set. From what I've heard from @genro, the current approach is `app.foo.com/tenantname`, but we may reserve the option to support `tenantname.app.foo.com` (subdomain-based routing) in the future — which is why "multidomain" was chosen to keep that door open.

**4. Name validation for tenants**
Agreed, this should have been addressed in the other PR but we can add validation in `trigger_onInserting` to prevent collisions with static paths and other reserved names.

**5. Documentation**
Writing documentation is necessary, but we don't keep doc sources in the genropy repository. This will be handled separately.

gnrpy/gnr/web/gnrdaemonprocesses.py
cgabriel: Resolved with the task configuration loading endpoint.

mbertoldi (approved):

cgabriel (approved):

Issue #433, opened 2026-01-28
[BUG] Task scheduler regression


cgabriel:
### Problem Description

THere is regression in the old task scheduler, which tasks are not scheduled anymore after the first one.

### Reproduction Steps

Just schedule a task with a 1 minute frequency.

### Environment

latest develop branch

cgabriel:
Merged in #434

PR #434, opened 2026-01-28
Fixed task issue while finding new tasks to be scheduled

cgabriel:
Fixed task issue while finding new tasks to be scheduled: the check between the current timestamp and the last execution timestamp was failing due to comparison between a tz-aware one and a non-aware one. Issue apparently was introduced with timezone changes, since this is still the old code from 2017.

Added debug logging in the scheduling stack to simplify handling of possible future issues.

refs #433


fporcari (approved):
looks good

PR #440, opened 2026-02-01
Fix text file handling in email attachments

Giammi77:
## Summary
Fixes incorrect file opening mode when attaching text files to emails.

## Changes
- Added MIME type check in `_attachments()` method
- Text files (`mime_family=='text'`) now open with `mode='r'`
- Binary files continue to open with `mode='rb'`

## Issue
Previously all attachments were opened in binary mode, which could cause encoding issues with text files.

### DIFF
--- a/gnrpy/gnr/lib/services/mail.py
+++ b/gnrpy/gnr/lib/services/mail.py
@@ -243,11 +243,19 @@ class MailService(GnrBaseService):
             if not mime_type:
                 mime_type = attachment_node.mimetype
             mime_family, mime_subtype = mime_type.split('/')
-            with attachment_node.local_path() as attachment_path:
-                with open(attachment_path, mode='rb') as attachment_file:
-                    email_attachment = mime_mapping[mime_family](attachment_file.read(), mime_subtype)
-                    email_attachment.add_header('content-disposition', 'attachment', filename=attachment_node.basename)
-                    msg.attach(email_attachment)
+            # check mime_family to fix file open mode
+            if mime_family=='text':
+                with attachment_node.local_path() as attachment_path:
+                    with open(attachment_path, mode='r') as attachment_file:
+                        email_attachment = mime_mapping[mime_family](attachment_file.read(), mime_subtype)
+                        email_attachment.add_header('content-disposition', 'attachment', filename=attachment_node.basename)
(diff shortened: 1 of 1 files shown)


cgabriel:
can you provide a issue to reproduce the problem, to understand better what this PR is addressing?

Giammi77:
Issue: Email attachments fail when attaching text files
Problem Description
When sending emails with text file attachments (e.g., .txt, .csv, .log files), the code was opening ALL files in binary mode ('rb'), regardless of their MIME type. This causes issues because:

Text files should be opened in text mode ('r') to properly handle character encoding
MIMEText constructor expects strings, not bytes - when you pass bytes from a binary-mode read to MIMEText(), it can cause encoding errors or unexpected behavior
Steps to Reproduce

# Example code that would fail before the fix
from gnr.lib.services.mail import MailService

mail_service = MailService(
    from_address='sender@example.com',
    smtp_host='smtp.example.com',
    user='user',
    password='pass'
)

# Try to send an email with a .txt file attachment
mail_service.sendmail(
    to_address='recipient@example.com',
    subject='Test with text attachment',
    body='Please find the attached text file',
    attachments=['path/to/file.txt']  # This would cause issues
)
Expected Behavior
Text files (MIME type text/*) should be properly encoded and attached to emails without encoding errors.

Actual Behavior (before fix)
The code opened text files in binary mode, which could cause:

Encoding errors when creating MIME attachments
Incorrect character handling in text files
Root Cause
In [mail.py:246-250](vscode-webview://1it6q1qb0kuvav05hkaempo0jccm8i1j8h85f71fmiuguetmmaaq/mail.py#L246-L250), all files were opened with mode='rb':


with open(attachment_path, mode='rb') as attachment_file:
    email_attachment = mime_mapping[mime_family](attachment_file.read(), mime_subtype)
But MIMEText expects text strings, not bytes.

Solution
The fix checks the mime_family and uses:

mode='r' for text files (mime_family=='text')
mode='rb' for binary files (images, applications, etc.)

cgabriel:
I was able to reproduce the issue, thank you. I'm closing this since the problem has been addressed in 7a1b8511a47e752ffbeabb12c048c589725d0aea with a cleaner approach which reduces code redundancy.

PR #447, opened 2026-02-05
refactor: remove dest_user_id and ViewMobile from email.message

dgpaci:
## Summary
- Remove `dest_user_id` concept (subtable and formulaColumn) from email.message model
- Remove `ViewMobile` class from message table handler - mobile views should be redefined in specific packages as needed
- Add indexed flag to `to_address` column for better query performance
- Add `message_address` table resource (th_message_address.py) for address validation/control

## Rationale
The `dest_user_id` and `ViewMobile` concepts are too generic and should be redefined by individual packages based on their specific requirements rather than being part of the core email package.


cgabriel (approved):
Watchout when adding such index to an existing huge table.

PR #448, opened 2026-02-05
feat(email): make message_address collection optional

genro:
## Problem

The `email.message_address` table stores all email addresses (to, from, cc, bcc) for every message processed by the email package. This enables useful queries like "show all messages sent to user@example.com".

However, in practice:
- **The table grows very large** - every email generates 2-10+ rows (one per address)
- **It's rarely used** - most installations never query this table
- **No way to disable it** - the collection happens unconditionally on every message

Over time, this table can contain millions of rows, impacting database performance and backup times, with no benefit for installations that don't use the address search feature.

## Solution

Make address collection **opt-in** via package preferences:

### New Preferences

| Preference | Default | Description |
|------------|---------|-------------|
| `collect_addresses` | `false` | Enable address collection in `message_address` table |
| `exclude_from_address` | `false` | When collecting, exclude sender (FROM) addresses |

The second option is only visible when the first is enabled.

### Configuration

In the admin preferences panel under "Email":

1. **Collect addresses** - Check to enable address collection
2. **Exclude FROM address** - (only shown if above is checked) Exclude sender addresses

### Behavior

| `collect_addresses` | `exclude_from_address` | Result |
|---------------------|------------------------|--------|
| `false` (default)   | -                      | No addresses stored |
| `true`              | `false`                | All addresses stored (to, from, cc, bcc) |
| `true`              | `true`                 | Only recipient addresses stored (to, cc, bcc) |

## Migration Notes

- **Existing installations**: No change needed. Default behavior is now disabled (previously always enabled)
- **Existing data**: The `message_address` table and its data are preserved; this change only affects new messages
- **To enable collection**: Go to Admin → Preferences → Email → Check "Collect addresses"

## Files Changed

- [message.py](projects/gnrcore/packages/email/model/message.py) - Check preferences before collecting addresses
- [preference.py](projects/gnrcore/packages/email/resources/preference.py) - Add preference UI fields

## Test Plan

- [ ] Verify default behavior: new messages don't populate `message_address`
- [ ] Enable `collect_addresses`: verify addresses are collected
- [ ] Enable `exclude_from_address`: verify FROM addresses are excluded
- [ ] Verify UI: second checkbox only visible when first is checked


dgpaci (approved):
Tests are positive

PR #444, opened 2026-02-04
refactor: centralize menu line badge handling

fporcari:
## Summary

Centralize and extend menu badge handling with the new `menuLineBadge` attribute.
titleCounter is still compatible but is deprecated in menu.

### New `menuLineBadge` attribute syntax

| Syntax | Description |
|--------|-------------|
| `menuLineBadge="methodName"` | Calls the specified RPC method on the table |
| `menuLineBadge="#"` | Returns total record count for the table |
| `menuLineBadge="#:fieldpath"` | Returns count where field has a "positive" value |
| `menuLineBadge="#:!fieldpath"` | Returns count where field has a "negative" value |
| `menuLineBadge=True` | (tableBranch only) Counts branch children without RPC |

**Fieldpath prefix**: can start with `$` (column) or `@` (relation), defaults to `$` if omitted.

**Filter conditions by dtype**:
- Boolean (`B`): `IS TRUE` / `IS NOT TRUE`
- Numeric (`N`, `L`, `I`, `R`): `IS NOT NULL AND !=0` / `IS NULL OR =0`
- Other (text, date, etc.): `IS NOT NULL` / `IS NULL`

### Other changes
- Add new `getMenuLineBadge` method in menuproxy for centralized badge logic
- Replace `titleCounter_count` with unified `badgeContent` attribute
- Add `menuline_badge` CSS class for badge styling
- Simplify label rendering using `innerHTML:` prefix for HTML badges
- Add deprecation warnings for legacy `condition` parameter and `titleCounter`

## Test plan
- [x] Test badge display on menu items
- [x] Verify `titleCounter` backwards compatibility
- [x] Test custom RPC handler with `menuLineBadge="methodName"`
- [x] Test `#` syntax for total record count
- [x] Test `#:columnName` syntax for filtered count
- [x] Test `#:!columnName` syntax for negated filter
- [ ] Test `True` syntax on tableBranch for children count

PR #452, opened 2026-02-06
Fix menuLineBadge check for virtual columns

### LINKED ISSUE #451: Fix menuLineBadge check for virtual columns after #444
## Problem

After the merge of #444, the check `if not column:` in `getMenuLineBadge` is always true for virtual columns because `tblobj.column()` returns an empty Bag instead of None.

An empty Bag is falsy in Python, so the condition incorrectly triggers and returns 0, breaking menulinebadge functionality for virtual columns.

## Solution

Change the check from `if not column:` to `if column is None:` to properly distinguish between:
- Column not found: returns `None`
- Virtual column: returns empty Bag (falsy but not None)

## Impact

Currently, all menu badges using virtual columns show 0 instead of the correct count.

## Related

Introduced by: #444

dgpaci:
## Summary
- Fixes menulinebadge returning 0 for virtual columns after #444
- Changes column existence check from `if not column:` to `if column is None:`
- Replaces error logging with ValueError exception for missing columns

## Problem
After merging #444, the check `if not column:` incorrectly treats empty Bags (returned for virtual columns) as non-existent columns. This causes all menulinebadge using virtual columns to always return 0.

## Solution
Use explicit `if column is None:` check to distinguish between:
- Column not found: `tblobj.column()` returns `None`
- Virtual column: `tblobj.column()` returns empty Bag (falsy but not None)

Also improves error handling by raising ValueError instead of logging and returning 0.

## Test plan
- [x] Test menulinebadge with virtual columns
- [x] Test menulinebadge with regular columns
- [x] Verify proper error when column doesn't exist

Fixes #451

### DIFF
--- a/gnrpy/gnr/web/gnrwebpage_proxy/menuproxy.py
+++ b/gnrpy/gnr/web/gnrwebpage_proxy/menuproxy.py
@@ -57,9 +57,8 @@ class GnrMenuProxy(GnrBaseProxy):
 
             # build WHERE clause based on column dtype
             column = tblobj.column(fieldpath)
-            if not column:
-                logger.error(f"Column {fieldpath} not found in table {table}")
-                return 0
+            if column is None:
+                raise ValueError(f"Column {fieldpath} not found in table {table}")
             dtype = column.attributes.get('dtype')
             if dtype == 'B':
                 # boolean: check for TRUE/NOT TRUE


cgabriel:
Merging with external review from @dgpaci @fporcari @gporcari, confirmed by @dgpaci

PR #456, opened 2026-02-06
Fix tree search highlight with innerHTML labels

fporcari:
## Summary
- Strip `innerHTML:` prefix from tree node labels before applying search highlight markup
- Prevents the raw prefix from appearing as visible text in filtered tree results
- Consistent with existing `innerHTML:` handling in `setLabelNode` (genro_patch.js)

## Test plan
- [x] Open a tree with HTML-formatted labels (using `innerHTML:` prefix)
- [x] Use the tree search/filter feature
- [x] Verify the `innerHTML:` prefix does not appear in search results
- [ ] Verify search highlighting works correctly on HTML labels


cgabriel (approved):
Have not idea how to test this, but since it has been requested as urgent, I'll leave dealing with possible errors to @fporcari

PR #457, opened 2026-02-06
feat(sql): SqlCompoundQuery — set operations on SqlQuery

genro:
## Summary

Add Python operators to `SqlQuery` for composing queries with SQL set operations, producing `SqlCompoundQuery` objects:

| Python operator | SQL operation |
|----------------|---------------|
| `q1 + q2` | `UNION` |
| `q1 \| q2` | `UNION ALL` |
| `q1 & q2` | `INTERSECT` |
| `q1 - q2` | `EXCEPT` |

### Why

Composing SQL set operations previously required writing raw SQL or building complex query strings manually. With this change, developers can express set operations naturally using Python's operator syntax, and the framework handles parameter namespacing, SQL generation, and cross-database compatibility automatically.

### Design

**Minimal subclass approach**: `SqlCompoundQuery` inherits from `SqlQuery` and overrides only `sqltext` and `sqlparams`. All other methods (`fetch()`, `cursor()`, `count()`, `test()`) work through inheritance because they rely on `self.sqltext` and `self.sqlparams`. This leverages the existing lazy compilation: accessing `sqltext` on each sub-query triggers its compilation automatically.

**Template-based composition**: Each compound stores a dict of sub-queries (keyed by mangler) and a template string. For example, `q1 + q2` produces template `{q0} UNION {q1}`, and `sqltext` formats the template with each sub-query's compiled SQL.

**Parameter namespacing via mangler** (from PR #1, commit `e3882949c`): Each sub-query gets a unique mangler prefix (`q0_`, `q1_`, ...) so parameters like `:year` become `:q0_year` and `:q1_year`, preventing name conflicts.

**Per-thread counter in `db.currentEnv`**: A monotonic counter in the thread-local environment generates unique mangler keys, ensuring no conflicts even with parenthesized expressions like `(q1 + q2) & (q3 + q4)` where two independent compound queries are composed.

### Usage

```python
# Simple set operations
q1 = db.query('hr.employee', columns='$name', where='$dept = :dept', dept='Engineering')
q2 = db.query('hr.employee', columns='$name', where='$dept = :dept', dept='Sales')

# UNION — all employees from either department (no duplicates)
result = (q1 + q2).fetch()

# UNION ALL — including duplicates
result = (q1 | q2).fetch()

# Chaining — evaluated left to right
result = (q1 + q2 + q3).fetch()

# Parenthesized grouping — controls evaluation order
result = ((q1 + q2) & (q3 + q4)).fetch()

# count() works on compound queries
total = (q1 + q2).count()

# test() returns the SQL text and parameters for inspection
sql, params = (q1 + q2).test()
```

### Cross-database compatibility

- **PostgreSQL**: Compound sub-queries are wrapped as `SELECT * FROM (...) AS alias` when parenthesized grouping is needed (PostgreSQL requires aliases on subqueries in FROM)
- **SQLite**: Same wrapping strategy (SQLite does not support bare parentheses around compound SELECT statements)

### What does NOT work on SqlCompoundQuery

Methods that access `self.compiled` internals (like `selection()`, `handlePyColumns`) are inherited but will raise errors if called. This is intentional — compound queries are meant for `fetch()`, `count()`, `cursor()`, and `test()`. Future PRs may extend support if needed.

## Test plan

- [x] `test_compound_union_sqltext` — verifies SQL text structure and mangled parameter names
- [x] `test_compound_union_fetch` — UNION returns correct titles across years
- [x] `test_compound_union_all` — UNION ALL preserves duplicates
- [x] `test_compound_intersect` — INTERSECT returns only common rows
- [x] `test_compound_except` — EXCEPT excludes matching rows
- [x] `test_compound_chain` — `q1 + q2 + q3` chains correctly
- [x] `test_compound_count` — count() on compound queries
- [x] `test_compound_parentheses` — `(q1 + q2) & (q3 + q4)` with correct grouping
- [x] All 96 tests pass on sqlite, postgres, and postgres3 backends

### DIFF
--- a/gnrpy/gnr/sql/gnrsqldata.py
+++ b/gnrpy/gnr/sql/gnrsqldata.py
@@ -1379,6 +1379,89 @@ class SqlQuery(object):
             cursor.close()
         return n
 
+    def _next_compound_key(self):
+        env = self.db.currentEnv
+        idx = env.get('_compound_counter', 0)
+        env['_compound_counter'] = idx + 1
+        return 'q%d' % idx
+
+    def _compound(self, other, operator):
+        if isinstance(other, SqlCompoundQuery):
+            self_key = self._next_compound_key()
+            self.mangler = self_key
+            queries = dict(other.queries)
+            queries[self_key] = self
+            template = '{%s} %s SELECT * FROM (%s) AS _cr' % (self_key, operator, other._template)
+        else:
+            self_key = self._next_compound_key()
+            other_key = other._next_compound_key()
+            self.mangler = self_key
+            other.mangler = other_key
+            queries = {self_key: self, other_key: other}
+            template = '{%s} %s {%s}' % (self_key, operator, other_key)
+        return SqlCompoundQuery(queries=queries, template=template,
+                                dbtable=self.dbtable, db=self.db)
+
+    def __add__(self, other):
+        return self._compound(other, 'UNION')
+
+    def __or__(self, other):
+        return self._compound(other, 'UNION ALL')
+
+    def __and__(self, other):
+        return self._compound(other, 'INTERSECT')
+
+    def __sub__(self, other):
+        return self._compound(other, 'EXCEPT')
+
+
+class SqlCompoundQuery(SqlQuery):
+
+    def __init__(self, queries, template, dbtable, db):
+        self.queries = queries
+        self._template = template
+        self.dbtable = dbtable
+        self.db = db
+        self.storename = None
+
+    def _get_sqltext(self):
+        return self._template.format(**{k: q.sqltext for k, q in self.queries.items()})
+
+    sqltext = property(_get_sqltext)
+
+    def _get_compiled(self):
+        return next(iter(self.queries.values())).compiled
+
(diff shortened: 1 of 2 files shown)


genro:
## Fix: non-destructive mangleParams

Added commit `a50eb0fb0` fixing an issue in `mangleParams`:

**Problem**: `mangleParams` used `pop()` to rename parameters with the mangler prefix, removing the original keys from the shared `sqlparams` dict. This is destructive because the dict is passed by reference and could cause regressions in contexts where original parameters are still needed.

**Fix**: `mangleParams` now adds mangled keys without removing the originals.

```python
# Before (destructive)
self.sqlparams['%s_%s' % (self.mangler, k)] = self.sqlparams.pop(k)

# After (additive)
self.sqlparams['%s_%s' % (self.mangler, k)] = v
```

E.g.: `year=2005` with mangler `q0` → dict has both `year=2005` and `q0_year=2005`.

Test updated to verify original parameters are preserved in the dict.

genro:
Superseded — compound query feature merged into subquery_to_join branch

PR #479, opened 2026-02-12
fix(sqlite): correct notnull detection in getColInfo

genro:
## Summary

- Fix bug in `gnrsqlite.py:272` where `PRAGMA table_info` returns `notnull` as integer (1/0) but the code compared it to the string `'NO'`, always yielding `False`
- This caused `gnrdbsetup` to generate spurious `ALTER COLUMN SET NOT NULL` for every NOT NULL column on SQLite, which SQLite does not support

## Root cause

```python
# BUG: always False because int != str
col['notnull'] = (col['notnull'] == 'NO')

# FIX:
col['notnull'] = bool(col['notnull'])
```

## Test plan

- [x] Run `gnrdbsetup test_invoice` on a fresh SQLite DB — should no longer fail on NOT NULL ALTER
- [x] Verify existing SQL tests still pass

Ref #478

### DIFF
--- a/gnrpy/gnr/sql/adapters/gnrsqlite.py
+++ b/gnrpy/gnr/sql/adapters/gnrsqlite.py
@@ -269,7 +269,7 @@ class SqlDbAdapter(SqlDbBaseAdapter):
                 colType = colType[:colType.find('(')]
             colType = colType.strip()
             col['dtype'] = self.typesDict[colType] if colType else 'T'
-            col['notnull'] = (col['notnull'] == 'NO')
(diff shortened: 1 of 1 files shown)


mbertoldi (approved):

cgabriel (commented):
I'm not able to reproduce the issue with the provided test plan, and no changes in behaviour using the proposed fix. I suspected there's something more subtle going on, but being unable to reproduce the problem, I can't say more about the matter. I'm approving the PR based on other reviewer's actions

PR #484, opened 2026-02-12
Fix: Skip broadcast with invalid pkey in onExternalChange

### LINKED ISSUE #483: Invalid pkey (false) in broadcast when inserting child records with subtable on parent
## Description

When inserting child records in a table with a foreign key relationship to a parent table that has subtables defined with reverse relation conditions, an invalid broadcast is generated on the parent table with `pkey: false` instead of a valid ID.

This causes a SQL error when grids attempt to reload with `store_chpkeys=[false]`:
```
psycopg2.errors.UndefinedFunction: operator does not exist: character = boolean
LINE: ... WHERE ... AND ( "t0"."id" IN (true) )
```

## Steps to Reproduce

1. Create a parent table with a subtable definition using reverse relation:
```python
tbl.subtable('mie_giornate', 
    condition='@persone_giornata.@partecipazione_persona_id.@partecipazione_id.user_id=:env_user_id')
```

2. Create a child table with foreign key to parent:
```python
tbl.column('parent_id').relation('parent.id', relation_name='children', mode='foreignkey', onDelete='cascade')
```

3. In parent's trigger_onInserted, insert child records:
```python
    self.db.table('child_table').insert(...)
```

4. Insert a parent record and observe the broadcast in browser console

## Expected Behavior

All broadcasts should have valid pkey values (strings containing the record ID).

## Actual Behavior

An additional broadcast is generated with `pkey: false` (boolean), causing:
- SQL error: `operator does not exist: character = boolean`
- Grid fails to reload
- Application becomes unusable

## Root Cause

When child records are inserted via triggers and the parent table has subtables with reverse relation conditions, Genropy generates a broadcast to notify about the relationship change. However, the pkey in this broadcast is set to `false` instead of the actual parent record ID.

## Proposed Fix

Add validation in `onExternalChange` method to skip broadcasts with invalid pkey values before processing them.

See PR: https://github.com/genropy/genropy/pull/XXX

## Related Issues

- Original issue in production application: https://github.com/dgpaci/anaci_projects/issues/181

## Environment

- Genropy version: latest (master branch)
- Database: PostgreSQL
- Browser: Chrome/Firefox (reproduced in both)

## Additional Context

This issue was discovered through extensive debugging of a production application where modifying grid records would trigger the SQL error intermittently. The workaround in the fix prevents the application from crashing while the root cause in broadcast generation should be investigated further.

dgpaci:
## Description

This PR fixes issue #483 where invalid broadcasts with `pkey: false` cause SQL errors when grids attempt to reload.

## Problem

When inserting child records via triggers on tables with subtables defined with reverse relations, Genropy generates an invalid broadcast on the parent table with `pkey: false` instead of a valid ID. This causes:

```
psycopg2.errors.UndefinedFunction: operator does not exist: character = boolean
LINE: ... WHERE ... AND ( "t0"."id" IN (true) )
```

## Solution

Added validation in `onExternalChange` method to:
1. Check if pkey is a valid string before processing
2. Skip broadcasts with invalid pkey values (`false`, `null`, `undefined`, empty string)
3. Log a warning when skipping invalid broadcasts for debugging

## Changes

- Modified `gnrjs/gnr_d11/js/genro_components.js`
- Added pkey validation before processing broadcast changes
- Added console warning for debugging invalid broadcasts

## Testing

Tested with:
- Parent table with subtable using reverse relation condition
- Child table with foreign key and onDelete='cascade'
- Insert operations via trigger_onInserted
- Grid reload scenarios

**Before fix:** SQL error on grid reload with `store_chpkeys=[false]`
**After fix:** Invalid broadcast is skipped, grid reloads successfully

## Related

- Fixes #483
- Original issue: https://github.com/dgpaci/anaci_projects/issues/181

## Notes

This fix is a defensive measure that prevents the application from crashing. The root cause of why `pkey: false` is generated in the broadcast should be investigated further in the broadcast generation mechanism.


gnrjs/gnr_d11/js/genro_components.js
```diff
@@ -7396,12 +7396,20 @@ dojo.declare("gnr.stores.Selection",gnr.stores.AttributesBagRows,{
             return;
         }
         var isExternalDict = {};
+        console.log('[DGPACI DEBUG] onExternalChange - changelist:', changelist);
```
cgabriel, line 7399: Maybe a less personalized debug string?

cgabriel (changes requested):
Fine for me, just fix the log strings with some less personalized ones :-)

dgpaci:
Actually unuseful since originated from a very specific case. See parent issue for more details

Issue #549, opened 2026-02-20
bug(sqlite): IS NOT TRUE rewrite breaks NULL handling in draft filter


genro:
## Bug Description

The SQLite adapter's `prepareSqlText` method in `gnrsqlite.py` rewrites `IS NOT TRUE` to `!=1` and `IS NOT FALSE` to `!=0`. This breaks NULL handling.

**In PostgreSQL:** `NULL IS NOT TRUE` → `TRUE` (correct — NULL is not TRUE)
**In SQLite after rewrite:** `NULL != 1` → `NULL` (incorrect — NULL compared with `!=` yields NULL, which is falsy in WHERE)

## Impact

Any table with `draftField=True` where records have `__is_draft = NULL` (the default) returns **zero rows** on SQLite, because the `excludeDraft` filter adds `$__is_draft IS NOT TRUE` which gets rewritten to `"__is_draft" != 1`.

## Root Cause

`gnrsqlite.py` line 120:

```python
sql = re.sub(" +IS +(NOT +)?(TRUE|FALSE)", self._booleanSubCb, sql, flags=re.I)
```

With callback:

```python
def _booleanSubCb(self, m):
    op = '!=' if m.group(1) else '='
    val = '1' if m.group(2).lower() == 'true' else '0'
    return ' %s%s ' % (op, val)
```

This produces `!=1` for `IS NOT TRUE`, but `NULL != 1` is `NULL` in SQL, not `TRUE`.

## Expected Fix

The rewrite should produce `(col IS NULL OR col != 1)` for `IS NOT TRUE` and `(col IS NULL OR col != 0)` for `IS NOT FALSE`, to correctly handle NULL values as PostgreSQL does.

## Steps to Reproduce

```python
from gnr.app.gnrapp import GnrApp

# On a table with draftField=True and __is_draft = NULL
app = GnrApp('<sqlite_instance_path>')
db = app.db
# Default query includes excludeDraft=True → adds IS NOT TRUE filter
count = db.table('pkg.table_with_draft').query().count()
# Returns 0 instead of expected row count
```

PR #550, opened 2026-02-20
fix(sqlite): IS NOT TRUE rewrite handles NULL correctly

genro:
## Summary

- Fix `_booleanSubCb` in `gnrsqlite.py` to preserve three-valued logic when rewriting `IS [NOT] TRUE/FALSE`
- The old rewrite (`IS NOT TRUE` → `!=1`) silently dropped all rows where the column is NULL, breaking `excludeDraft` on tables with `draftField=True`
- New rewrite: `IS NOT TRUE` → `(expr IS NULL OR expr !=1)`, matching PostgreSQL semantics

## Test plan

- [x] 16-test suite (`test_sqlite_boolean_rewrite.py`) covering all four IS variants, old broken behavior, corrected rewrite, and end-to-end SQL execution on SQLite
- [x] flake8 clean on both modified files

Ref #549

### DIFF
--- a/gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
+++ b/gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
@@ -11,14 +11,13 @@ verify the fix end-to-end.
 """
 
 import os
-import sys
-import subprocess
 import tempfile
 
 import pytest
 
 from gnr.app.gnrapp import GnrApp
 from core.common import BaseGnrTest
+from .common import get_pg_config
 
 DRAFT_MARKER = '__bool_rewrite_test__'
 
@@ -47,28 +46,7 @@ def db_sqlite():
 @pytest.fixture(scope='module')
 def db_pg():
     _ensure_base_gnr_test()
-    if sys.platform == 'win32':
-        pytest.skip('testing.postgresql not available on Windows')
-    pg_instance = None
-    if 'GITHUB_WORKFLOW' in os.environ:
-        pg_conf = dict(host='127.0.0.1', port='5432',
-                       user='postgres', password='postgres')
-    elif 'GNR_TEST_PG_PASSWORD' in os.environ:
-        pg_conf = dict(
-            host=os.environ.get('GNR_TEST_PG_HOST', '127.0.0.1'),
-            port=os.environ.get('GNR_TEST_PG_PORT', '5432'),
-            user=os.environ.get('GNR_TEST_PG_USER', 'postgres'),
-            password=os.environ.get('GNR_TEST_PG_PASSWORD'),
-        )
-    else:
-        try:
-            from testing.postgresql import Postgresql
-        except ImportError:
-            pytest.skip('testing.postgresql not installed')
-        subprocess.run(['pkill', '-f', 'postgres.*tmp'], capture_output=True)
-        pg_instance = Postgresql()
-        dsn = pg_instance.dsn()
-        pg_conf = dict(host=dsn['host'], port=dsn['port'], user=dsn['user'])
+    pg_conf, pg_instance = get_pg_config()
     dbname = pg_conf.pop('database', 'test_bool_rewrite')
     try:
         app = GnrApp('test_invoice', db_attrs=dict(
@@ -81,7 +59,7 @@ def db_pg():
     except Exception:
         pytest.skip('PostgreSQL not available')
     finally:
-        if pg_instance:
+        if pg_instance is not None:
             pg_instance.stop()
 
 
--- a/gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
+++ b/gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
@@ -13,22 +13,13 @@ verify the fix end-to-end.
 import pytest
 from gnr.app.gnrapp import GnrApp
 
-INSTANCE_PATH_PG = (
-    '/Users/gporcari/Sviluppo/Genropy/genropy'
-    '/projects/test_invoice/instances/test_invoice_pg'
-)
-INSTANCE_PATH_SQLITE = (
-    '/Users/gporcari/Sviluppo/Genropy/genropy'
-    '/projects/test_invoice/instances/test_invoice'
-)
-
 DRAFT_MARKER = '__bool_rewrite_test__'
 
 
 @pytest.fixture(scope='module')
 def db_pg():
     try:
-        app = GnrApp(INSTANCE_PATH_PG)
+        app = GnrApp('test_invoice_pg')
         return app.db
     except Exception:
         pytest.skip('PostgreSQL instance not available')
@@ -37,7 +28,7 @@ def db_pg():
 @pytest.fixture(scope='module')
 def db_sqlite():
     try:
-        app = GnrApp(INSTANCE_PATH_SQLITE)
+        app = GnrApp('test_invoice')
         return app.db
     except Exception:
         pytest.skip('SQLite instance not available')
--- a/gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
+++ b/gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
@@ -1,199 +1,173 @@
-"""Test for SQLite boolean rewrite in prepareSqlText.
+"""Test for SQLite boolean rewrite (bug #549).
 
-Verifies that IS [NOT] TRUE/FALSE rewrites handle NULL correctly,
-matching PostgreSQL's three-valued logic semantics.
+The excludeDraft filter generates ``$__is_draft IS NOT TRUE``.
+On PostgreSQL this correctly includes rows where __is_draft is NULL
+(three-valued logic).  The old SQLite rewrite turned IS NOT TRUE
+into ``!=1``, which loses NULL rows because ``NULL != 1`` is NULL.
 
-Bug #549: the old rewrite turned ``IS NOT TRUE`` into ``!=1``,
-but ``NULL != 1`` is ``NULL`` in SQL (not TRUE as in PostgreSQL).
+These tests use the real model (customer with draftField=True) and
+run actual queries through the adapter on both PG and SQLite to
+verify the fix end-to-end.
 """
 
-import re
-import sqlite3
 import pytest
+from gnr.app.gnrapp import GnrApp
+
+INSTANCE_PATH_PG = (
+    '/Users/gporcari/Sviluppo/Genropy/genropy'
+    '/projects/test_invoice/instances/test_invoice_pg'
+)
+INSTANCE_PATH_SQLITE = (
+    '/Users/gporcari/Sviluppo/Genropy/genropy'
+    '/projects/test_invoice/instances/test_invoice'
+)
+
+DRAFT_MARKER = '__bool_rewrite_test__'
 
 
 @pytest.fixture(scope='module')
-def conn():
-    """In-memory SQLite database with a test table."""
-    c = sqlite3.connect(':memory:')
-    cur = c.cursor()
-    cur.execute('CREATE TABLE t (id INTEGER, flag BOOLEAN)')
-    cur.execute('INSERT INTO t VALUES (1, NULL)')
-    cur.execute('INSERT INTO t VALUES (2, 0)')
-    cur.execute('INSERT INTO t VALUES (3, 1)')
-    c.commit()
-    return c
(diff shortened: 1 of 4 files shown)


gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
cgabriel, line 1: This test case doesn't test the actual fix on the adapter - it just works directly with an in-memory sqlite and provide a local implementation of the prepare, so it's not helpful - if the codebase changes, the tests is indipendent and it will continue to work, so there's no point of having it, since it's just a "demo". Therefore, it should rather 1/ create a model having the correct draft attributes 2/ test a query against it verifying the fix in the adapter.

cgabriel (changes requested):

genro:
@cgabriel You're absolutely right — the previous tests were just a standalone demo against raw sqlite3, completely decoupled from the actual adapter code. Apologies for that.

Rewrote the tests in `cbdcf73ef`: they now use the real `customer` model (with `draftField=True`) and run `excludeDraft` queries through the actual SQLite adapter end-to-end. There's also a PG vs SQLite parity check to make sure both return the same results.

gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
genro: Fixed in ba6be599b: also replaced absolute instance paths with name-based discovery (`GnrApp('test_invoice')`).

gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
```diff
@@ -1,199 +1,173 @@
...
-        1    IS NOT TRUE  -> FALSE
-        1    IS FALSE     -> FALSE
-        1    IS NOT FALSE -> TRUE
-    """
+def db_pg():
+    try:
+        app = GnrApp(INSTANCE_PATH_PG)
+        return app.db
+    except Exception:
+        pytest.skip('PostgreSQL instance not available')
```
cgabriel: I actually get *all* of the tests being skipped, since this approach assumes that the instance has a database. Also the CI skips all the test, because since the database is not getting created, it will always fails, and then skipped, so it's not useful. 

THe test must use the test base class "BaseGnrSqlTest" which will provide all the needed to execute the tests against a database.

Having all these tests to be always skipped doesn't bring any value.

cgabriel (dismissed):

genro:
@cgabriel @mbertoldi @fporcari You're right that the current approach (skipping when the instance is unavailable) doesn't work in CI.

The underlying issue is broader than this PR. The current SQL test infrastructure (`BaseGnrSqlTest` + `configureDb`) builds the model programmatically — a few tables (video, movie, cast) defined by hand in `common.py`. This works well for basic SQL operations but is insufficient for testing structural features like multi-level FK chains, aliasColumns, formulaColumns, joinColumns, subtables, pyColumns, etc.

The problem is architectural: in a real Genropy application, the model goes through a specific loading pipeline — `GnrApp` loads the packages, registers table mixins via `tableMixin()`, then `model.build()` calls `config_db()` on each table class. Replicating this programmatically in `configureDb` for a rich model (test_invoice has 17 tables with complex relations) would mean duplicating all the `config_db` logic by hand and keeping it in sync — which is not practical.

An infrastructure already exists in `tests/core/common.py` (`BaseGnrTest` + `InstanceMaker`) that creates a temporary Genropy environment and allows using `GnrApp`. But this is not wired into the SQL test infrastructure.

We need to find a solution that lets the SQL tests start from a `GnrApp` instance — the app serves as a test bench where we test the DB layer against a real, rich model. This would make the tests much closer to the real use case and avoid the maintenance burden of a hand-written programmatic model.

I'd like to discuss how to approach this — whether by extending `BaseGnrSqlTest` to optionally bootstrap from a `GnrApp`, or by creating a shared fixture that both SQL and app tests can use.

genro:
Following up with a concrete proposal for discussion.

The `test_invoice` project is already in the repo under `projects/`, with 17 tables covering FK chains, aliasColumns, formulaColumns, joinColumns, subtables, pyColumns, etc. The only reason it doesn't work in CI is that `GENRO_HOME` is not set, so `GnrApp('test_invoice')` can't find the project.

The change would be minimal and non-invasive:

**1. In `.github/workflows/tests.yml`** — add `GENRO_HOME` pointing to the checkout root:
```yaml
    - name: Test with pytest
      env:
        GNR_LOCALE: en_GB
        LC_ALL: en_GB
        LANG: en_GB
        GENRO_HOME: ${{ github.workspace }}
      run: |
        cd gnrpy/tests && pytest core sql web app xtnd
```

**2. In `tests/sql/common.py`** — add a helper to bootstrap a `GnrApp` for tests that need a rich model:
```python
def get_app_db(instance_name='test_invoice'):
    """Return a db from a GnrApp instance.
    Sets GENRO_HOME automatically if not already set."""
    genro_home = os.environ.get('GENRO_HOME')
    if not genro_home:
        # derive from this file's location: tests/sql/ -> repo root
        genro_home = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..', '..'))
        os.environ['GENRO_HOME'] = genro_home
    from gnr.app.gnrapp import GnrApp
    return GnrApp(instance_name).db
```

**3. Tests needing the rich model** use this helper instead of `GnrSqlDb` + `configureDb`. Existing tests using the programmatic approach remain untouched.

The idea is that the app serves as a test bench for the DB layer — the model goes through the full real pipeline (`tableMixin` → `build()` → `config_db`), making the tests much closer to production behavior.

Would this approach work? @cgabriel @mbertoldi @fporcari

genro:
Il sistema di test è stato adattato per rispettare l'esigenza segnalata nella review: i test non dipendono più da istanze preconfigurate.

Ora le fixture creano i database al volo (sia sqlite che postgres) usando il nuovo parametro `db_attrs` introdotto nella PR #553. Questo permette ai test di girare in CI senza bisogno di istanze preesistenti.

Questa PR dipende dalla #553 — una volta mergiata quella, questa può essere tolta da draft.

cgabriel:
#550 has been merged

genro:
@cgabriel The PR has been completely rewritten since your previous review. Here's what changed:

**Previous issues (now resolved):**
- Tests were a standalone demo against raw sqlite3 — **now they use the real adapter** (`tbl.query(excludeDraft=True)`) end-to-end
- Tests depended on pre-existing instances and were always skipped in CI — **now they create the database on the fly** using `db_attrs` (from #553, now merged) + `model.check(applyChanges=True)`

**Current state (3 files, clean diff vs develop):**
- `gnrsqlite.py`: the actual fix — `IS NOT TRUE` rewrite handles NULL correctly
- `test_sqlite_boolean_rewrite.py`: 4 tests that insert records with `__is_draft` = NULL/FALSE/TRUE and verify the filter through the real adapter. Also includes PG vs SQLite parity check
- `customer.py`: adds `draftField=True` to the test model

CI is green on all 4 builds — all 4 tests **pass** (0 skipped).

gnrpy/tests/sql/test_sqlite_boolean_rewrite.py
```diff
@@ -0,0 +1,209 @@
...
+    app = GnrApp('test_invoice', db_attrs=dict(
+        implementation='sqlite',
+        dbname=os.path.join(tempdir, 'testing'),
+    ))
+    app.db.model.check(applyChanges=True)
+    return app.db
+
+
+@pytest.fixture(scope='module')
+def db_pg():
```
cgabriel, line 47: Why reimplement all this from sql/common.py ? Wouldn't it be better to have just one definition of testing database for testing.postgresql and avoid duplicate code?

cgabriel (dismissed):

genro:
@cgabriel @mbertoldi You're right — the PG discovery logic is duplicated. Here's a detailed analysis and a proposed fix.

## The problem

The logic that determines **how to connect to PostgreSQL** in tests is copy-pasted in three places. It handles three scenarios:

| Scenario | Condition | Behavior |
|---|---|---|
| GitHub CI | `GITHUB_WORKFLOW` in env | `host=127.0.0.1, user=postgres, password=postgres` |
| Remote PG | `GNR_TEST_PG_PASSWORD` in env | reads host/port/user/password from env vars |
| Local | neither of the above | starts `testing.postgresql`, uses its dsn() |

This logic is duplicated in:

1. **`tests/sql/common.py`** — `BaseGnrSqlTest.setup_class()` (lines 59-81)
2. **`tests/sql/test_sqlite_boolean_rewrite.py`** — fixture `db_pg` (this PR, #550)
3. **`tests/sql/conftest.py`** — fixture `db_pg` (PR #551)

## Why we can't just inherit from BaseGnrSqlTest

`BaseGnrSqlTest` is not a simple PG config provider. It:
- Imports `Postgresql` **at module level** (line 6 of `common.py`) — if `testing.postgresql` is not installed, the entire file fails to import
- Stores `pg_conf` as a **class attribute** — requires inheritance
- Manages `pg_instance` lifecycle in `teardown_class`
- Includes `CONFIG`, `SAMPLE_XMLDATA`, `mysql_conf` — things the new tests don't use
- Is decorated with `@excludewin32` — skips everything on Windows

The new tests (PR #550 and #551) use `GnrApp` + `db_attrs` (model from filesystem), not `GnrSqlDb` + `configureDb` (programmatic model). Inheriting from `BaseGnrSqlTest` would pull in unnecessary dependencies and a different structure (class with setup_class vs pytest fixtures).

## Proposed fix

Extract the discovery logic into a **standalone function** in `common.py`:

```python
def get_pg_config():
    """Determine PostgreSQL connection parameters.

    Returns (pg_conf, pg_instance) where pg_instance is not None
    only when a temporary testing.postgresql was started (caller
    must call pg_instance.stop() when done).

    Raises pytest.skip if pg is not available.
    """
    if sys.platform == 'win32':
        pytest.skip('testing.postgresql not available on Windows')
    if 'GITHUB_WORKFLOW' in os.environ:
        return dict(host='127.0.0.1', port='5432',
                    user='postgres', password='postgres'), None
    if 'GNR_TEST_PG_PASSWORD' in os.environ:
        return dict(
            host=os.environ.get('GNR_TEST_PG_HOST', '127.0.0.1'),
            port=os.environ.get('GNR_TEST_PG_PORT', '5432'),
            user=os.environ.get('GNR_TEST_PG_USER', 'postgres'),
            password=os.environ.get('GNR_TEST_PG_PASSWORD'),
        ), None
    try:
        from testing.postgresql import Postgresql
    except ImportError:
        pytest.skip('testing.postgresql not installed')
    subprocess.run(['pkill', '-f', 'postgres.*tmp'], capture_output=True)
    pg_instance = Postgresql()
    dsn = pg_instance.dsn()
    return dict(host=dsn['host'], port=dsn['port'], user=dsn['user']), pg_instance
```

Then:
- `BaseGnrSqlTest.setup_class()` calls `get_pg_config()` internally
- The `db_pg` fixtures in `test_sqlite_boolean_rewrite.py` and `conftest.py` call `get_pg_config()`
- The logic exists in **one place only**

## Impact

| File | Change |
|---|---|
| `common.py` | Add `get_pg_config()`, refactor `BaseGnrSqlTest.setup_class()` to use it |
| `test_sqlite_boolean_rewrite.py` (PR #550) | `db_pg` fixture uses `get_pg_config()` |
| `conftest.py` (PR #551) | `db_pg` fixture uses `get_pg_config()` — to be aligned after merge |

**Note on the module-level import**: currently `common.py` line 6 does `from testing.postgresql import Postgresql` at module level. This causes an import failure if the package is not installed. The proposed `get_pg_config()` imports `Postgresql` **lazily** (inside a try/except), which is more robust. Ideally the import in `BaseGnrSqlTest` should be moved too, but that's a separate concern.

mbertoldi:
@genro, please try to implement the proposed solution using get_pg_config. please remember to stop local postgres instance if it's not None during teardown. testing.postgresql is installed as requirement if genropy is installed as 'developer' profile, as needed to run tests.

genro:
@mbertoldi Done — `get_pg_config()` is now in `common.py` and used by both `BaseGnrSqlTest.setup_class()` and the `db_pg` fixture in this PR. The `pg_instance` teardown is handled correctly: `pg_instance.stop()` is called only when it's not None. CI is green on all 4 builds.

mbertoldi (approved):

cgabriel (approved):

PR #553, opened 2026-02-20
feat(app): add db_attrs parameter to GnrApp.init()

genro:
## Summary

- Add `db_attrs` parameter to `GnrApp.__init__()` and `GnrApp.init()` that overrides db connection attributes read from instanceconfig
- When `db_attrs` is provided, `model.check(applyChanges=True)` is called after startup to create/update the schema
- `forTesting=True` is kept for backward compatibility but emits a `DeprecationWarning` — it internally converts to the equivalent `db_attrs`
- This unblocks #550 and all subsequent PRs that need to test complex scenarios with real app-generated models on different backends (sqlite, postgres)

## Motivation

The old `forTesting=True` flag hardcoded a temporary sqlite database with no way to override. Tests that need to verify behavior on different backends (e.g. the SQLite boolean rewrite fix in #550) had no clean mechanism to specify db connection parameters.

With `db_attrs`, tests can now do:

```python
app = GnrApp('test_invoice', db_attrs=dict(
    implementation='sqlite',
    dbname=os.path.join(tempdir, 'testing'),
))
```

or target postgres:

```python
app = GnrApp('test_invoice', db_attrs=dict(
    implementation='postgres',
    host='localhost',
    dbname='test_db',
))
app.db.model.check(applyChanges=True)
```

## Test plan

- [x] 6 tests in `tests/app/test_gnrapp_db_attrs.py`
  - `db_attrs` creates db, creates tables, supports insert/query, overrides implementation
  - `forTesting` emits `DeprecationWarning` and still works
- [x] flake8 clean

### DIFF
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -975,7 +975,9 @@ class GnrApp(object):
 
         dbattrs = dict(self.config.getAttr('db') or {})
         dbattrs['implementation'] = dbattrs.get('implementation') or 'sqlite'
-        if dbattrs.get('dbname') == '_dummydb':
+        if db_attrs:
+            dbattrs.update(db_attrs)
+        elif dbattrs.get('dbname') == '_dummydb':
             pass
         elif self.remote_db:
             rdb = self.config.get(f"remote_db")#.{self.remote_db}")
@@ -991,9 +993,6 @@ class GnrApp(object):
             if not os.path.isabs(dbname):
                 dbname = self.realPath(os.path.join('..','data',dbname))
             dbattrs['dbname'] = dbname
-
-        if db_attrs:
-            dbattrs.update(db_attrs)
 
         dbattrs['application'] = self
         self.db = GnrSqlAppDb(debugger=getattr(self, 'sqlDebugger', None), **dbattrs)
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -20,9 +20,6 @@
 #License along with this library; if not, write to the Free Software
 #Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
-import tempfile
-import atexit
-import shutil
 import locale
 import sys
 import re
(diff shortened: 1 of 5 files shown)


gnrpy/gnr/app/gnrapp.py
```diff
@@ -1006,13 +1009,11 @@ def removeTemporaryDirectory():
             self.config['menu'] = self.config['menu']['#0']
         #if self.instanceMenu:
         #    self.config['menu']=self.instanceMenu
-            
+
         self.localizer = AppLocalizer(self)
-        if forTesting:
-            # Create tables in temporary database
+        if db_attrs:
             self.db.model.check(applyChanges=True)
```
mbertoldi: Please, let the user to it explicitly in the calling code, as in the example code in PR description.

genro:
Addressed @mbertoldi's review: removed the automatic `model.check(applyChanges=True)` from `init()`. The caller now does it explicitly:

```python
app = GnrApp('test_invoice', db_attrs=dict(implementation='sqlite', dbname=path))
app.db.model.check(applyChanges=True)
```

See commit 325437f.

mbertoldi (dismissed):

gnrpy/gnr/app/gnrapp.py
```diff
@@ -1006,14 +1007,8 @@ def removeTemporaryDirectory():
             self.config['menu'] = self.config['menu']['#0']
         #if self.instanceMenu:
         #    self.config['menu']=self.instanceMenu
-            
+
         self.localizer = AppLocalizer(self)
-        if forTesting:
```
cgabriel, line 1011: Since forTesting has been left as a parameter in the class interface raising a deprecation warning (for me, it should be dropped anyway, since it's for internal use only), the loadTestingData invocation should be left as well. So it's either dropping everything o leaving everything with a deprecation warning.

cgabriel (changes requested):

gnrpy/gnr/app/gnrapp.py
mbertoldi: i'm for dropping and correcting the only two tests using it.

genro:
Done — dropped `forTesting` entirely as suggested by @mbertoldi and @cgabriel:

- Removed `forTesting` parameter from `__init__` and `init()`
- Removed `loadTestingData()`
- Removed unused imports (`atexit`, `shutil`, `tempfile`)
- Converted the 2 tests that used `forTesting=True` (`gnrapp_test.py`, `gnrlocalization_test.py`) to `db_attrs=dict(...)`
- Removed `TestForTestingDeprecation` test class

CI green on all 4 builds. See commit 99c175f.

cgabriel (approved):

PR #565, opened 2026-02-24
fix(grouplets): improvements after grouplet usage

fporcari:
## Summary
- In `gr_loadGrouplet`, use `getPublicMethod('remote', handlername)` instead of bare `getattr` for RPC handler resolution, ensuring only decorated public methods can be invoked
- In `_resolveDialogRoot`, add missing `lightbutton` to the skip tags and refactor the long `||` chain into a `Set` lookup for readability

## Test plan
- [x] flake8 passes with zero errors
- [x] All tests pass (1143 passed, 146 skipped)
- [x] Verified correct handler resolution for both RPC and resource-based paths
- [x] Verified dialog root resolution skips lightbutton nodes correctly


fporcari:
Superseded by new PR with single commit

Issue #548, opened 2026-02-19
refactor(sql): RelationTreeResolver has redundant double cache and unsafe locking


genro:
## Problema

`RelationTreeResolver` in `gnrsqlmodel/resolvers.py` ha due livelli di cache ridondanti:

1. **`__fields`** — attributo privato con double-checked locking (`self._lock.acquire/release`)
2. **`_cache`** — cache del `BagResolver` base (`cacheTime=-1` → permanente)

Entrambe fanno la stessa cosa: cachare il risultato di `_fields()` permanentemente.

### Problemi

- **Lock non sicuro**: `self._lock.acquire()` / `self._lock.release()` senza `try/finally`. Se `_fields()` solleva un'eccezione, il lock resta acquisito → deadlock.
- **Doppia cache inutile**: `__fields` e `_cache` sono entrambe permanenti e ridondanti.
- **Memory leak**: `cacheTime=-1` tiene in memoria la Bag delle relations per tutta la durata del processo, anche per tabelle non più usate.
- **Non componibile**: la cache permanente impedisce di iniettare contenuti temporanei (es. RuntimeModel).

### Soluzione proposta

Spostare la cache delle relations nel `currentEnv` (thread-local, per-request):

- La property `relations` in `table.py` cacha in `currentEnv['_relations'][fullname]`
- `RelationTreeResolver.load()` semplificato: solo `return self._fields(...)`, senza `__fields` né lock
- Thread-safe gratis (currentEnv è per-thread)
- Niente memory leak (currentEnv si pulisce a fine request)
- Invalidazione banale: `del currentEnv['_relations'][fullname]`

### File coinvolti

- `gnrpy/gnr/sql/gnrsqlmodel/table.py` — property `relations` con cache in currentEnv
- `gnrpy/gnr/sql/gnrsqlmodel/resolvers.py` — semplificazione `RelationTreeResolver.load()`

genro:
## Analysis Report: RelationTreeResolver cache strategy and performance

### 1. Bug confirmation: deadlock on exception

The original `RelationTreeResolver.load()` used `self._lock.acquire()` / `self._lock.release()` without `try/finally`. If `_fields()` raised an exception, the lock was never released, causing a **deadlock** on the next call from a different thread.

**Test**: `test_relations_thread_safe.py::test_lock_not_released_after_exception`
- Creates a resolver with a mocked `_fields()` that raises `RuntimeError`
- First call from main thread: raises as expected
[...]

PR #571, opened 2026-02-24
Test code duplication refactor

cgabriel:
These PR includes:

* refactor of duplicated code to setup the test environment
* introduce fixtures at module level to take care of correct temporary directory deletion (otherwise it will populate the test folder and they can't be easily ignored by git)

* cleanup runtime imports
* move common code ancestor class to create the GnrApp using new db_attrs, using class attributes (allowing customization in specific tests if needed)
* avoid module level fixture for db reference, using a proper setup method, reducing the test method footprints.


genro (approved):
Good cleanup. The `sqlite_temp_dir` fixture with `shutil.rmtree` in teardown solves the orphan temp directories issue.

One minor observation: `setup_module` / `teardown_module` are global module-level functions — they work fine but are slightly more fragile than fixtures if someone adds another conftest or changes loading order. Not a blocker, just worth noting.

No conflicts with the `fix/relations-thread-safe` branch (#548) — our benchmark and compiler simulation tests depend on `db_sqlite` which keeps the same public interface.

PR #578, opened 2026-02-25
fix(#548): thread-safe relation tree with currentEnv cache

genro:
## Summary

- Remove unsafe `threading.RLock` from `RelationTreeResolver.load()` — if `_fields()` raised, the lock was never released causing deadlock (confirmed by test)
- Move relation tree cache to `currentEnv` (thread-local, per-request) — thread-safe by design, no memory leak, no stale data
- Set `cacheTime=-1` on child resolvers so deep dotted-path navigation (e.g. `@invoice_id.@customer_id.@state.name`) is cached within the same request
- Invalidate `_relations` cache after `model.build()`

## Files changed

- `gnrsqlmodel/resolvers.py` — simplified `load()`, removed lock and `__fields` cache
- `gnrsqlmodel/table.py` — `_get_relations` property with `currentEnv` cache + `cacheTime=-1`
- `gnrsqlmodel/model.py` — invalidate cache after `build()`

## Performance

Realistic compiler simulation (5 tables, 100 paths): ~2 ms build, ~0.3 ms cached — negligible in context of a full RPC call. See detailed analysis on #548.

## Test plan

- `test_relations_thread_safe.py` (2 tests) — deadlock reproduction
- `test_relations_benchmark.py` (4 tests) — exhaustive walk, multi-table comparison
- `test_compiler_simulation.py` (4 tests) — realistic compiler pattern

All 1394 tests pass (67 skipped).

Ref #548


cgabriel (dismissed):

cgabriel (approved):
I've fixed the tests, the rest was already reviewed in the upstream issue.

mbertoldi (approved):

PR #580, opened 2026-02-25
fix: preserve unique constraint on columns in composite primary keys (closes #576)

### LINKED ISSUE #576: [BUG] fields uniqueness for compositecolumn as primary key
### Problem Description

It seems impossible to create a uniqueness key for a field when it is used in a compositeColumn and this last one is set as primary-key.

Look at the field_b in this example example

```
class Table:
    def config_db(self, pkg):
        tbl = pkg.table('test', pkey='uniquepkey')
        self.sysFields(tbl, user_ins=False, user_upd=False, ins=False, upd=False,ldel=False)
        
        tbl.compositeColumn('uniquepkey', columns='field_a,field_b')
        tbl.column("field_a", size="5")
        tbl.column("field_b", size="5", unique=True)

        tbl.column("field_c", size="5", unique=True)
```
produce
```
CREATE TABLE "fcrm"."fcrm_test"(
 "id" character(22),
 "field_a" character(5) NOT NULL,
 "field_b" character(5) NOT NULL,
 "field_c" character(5),
 PRIMARY KEY (field_a,field_b)
);
ALTER TABLE "fcrm"."fcrm_test"
ADD CONSTRAINT "cst_3a92cc63" UNIQUE ("field_c");
```

Note: adding a unique constraint to the composite column does not help "field_b"; on the contrary another unique constraint is added for it, but I think it is useless:
```
 PRIMARY KEY (field_a,field_b),
 CONSTRAINT "cst_637003d9" UNIQUE ("field_a", "field_b")
```

The problem seems to be the usage of the compositeColumn as primary key.
Here the second example
```
class Table:
    def config_db(self, pkg):
        tbl = pkg.table('test', pkey='id')
        self.sysFields(tbl, user_ins=False, user_upd=False, ins=False, upd=False,ldel=False)
        
        tbl.compositeColumn('uniquepkey', columns='field_a,field_b')
        tbl.column("field_a", size="5")
        tbl.column("field_b", size="5", unique=True)

        tbl.column("field_c", size="5", unique=True)
```
produces:
```
CREATE TABLE "fcrm"."fcrm_test"(
 "id" character(22) NOT NULL,
 "field_a" character(5),
 "field_b" character(5),
 "field_c" character(5),
 PRIMARY KEY (id)
);
ALTER TABLE "fcrm"."fcrm_test"
ADD CONSTRAINT "cst_5d11174d" UNIQUE ("field_b");  # <=== here we are!
ALTER TABLE "fcrm"."fcrm_test"
ADD CONSTRAINT "cst_3a92cc63" UNIQUE ("field_c");
```

Note: adding a unique constraint to the composite key resolves the configuration issue, since both <field_a, field_b> and <field_b> are unique. However, we are forced to introduce an extra "id" column.

### Reproduction Steps

A simple ```...\model/test.py``` can be write and tested with  ```gnr db migrate -v-c ...```

_No response_

### Environment

master and develop

### Additional Material

_No response_

fporcari:
## Summary
- Fix bug where `unique=True` on a column was silently ignored when the column was part of a `compositeColumn` used as the table's primary key
- The ORM extractor now only removes redundant `unique` for single-column PKs (where PK already implies uniqueness), preserving it for composite PK columns
- Added test verifying that a column in a composite PK retains its individual UNIQUE constraint

## Linked Issue
Closes #576

## Test plan
- [x] New test `test_05c1_composite_pkey_with_unique_column` verifies the fix end-to-end against a real PostgreSQL database
- [x] All 112 migration tests pass (both postgres adapters)
- [x] Full test suite passes (1305 passed, 147 skipped)
- [x] Existing test `test_05d_create_table_with_pkey_explicit_unique` confirms single-column PK still correctly drops redundant unique

### DIFF
--- a/gnrpy/gnr/sql/gnrsqlmigration/orm_extractor.py
+++ b/gnrpy/gnr/sql/gnrsqlmigration/orm_extractor.py
@@ -209,10 +209,14 @@ class OrmExtractor:
         pkeys = table_json['attributes']['pkeys']
 
         # Primary key columns are automatically NOT NULL
-        # and don't need a separate index or UNIQUE constraint
+        # and don't need a separate index (PK already creates one).
+        # For single-column PKs, unique is also redundant (PK implies uniqueness).
+        # For composite PKs, individual columns may still need their own
+        # UNIQUE constraint (issue #576).
         if pkeys and (column_name in pkeys.split(',')):
             attributes['notnull'] = '_auto_'
-            attributes.pop('unique', None)
+            if ',' not in pkeys:
+                attributes.pop('unique', None)
             attributes.pop('indexed', None)
 
         column_entity = new_column_item(
--- a/gnrpy/tests/sql/test_gnrsqlmigration.py
+++ b/gnrpy/tests/sql/test_gnrsqlmigration.py
@@ -189,6 +189,26 @@ class BaseGnrSqlMigration(BaseGnrSqlTest):
         check_value = 'CREATE TABLE "alfa"."alfa_recipe_row" ("recipe_code" character varying(12) NOT NULL , "recipe_line" bigint NOT NULL , "description" text , "ingredient_id" bigint , PRIMARY KEY (recipe_code,recipe_line));'
         self.checkChanges(check_value)
 
+    def test_05c1_composite_pkey_with_unique_column(self):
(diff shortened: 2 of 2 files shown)


fporcari:
@dedalus2000 check this out

cgabriel (approved):

Issue #585, opened 2026-02-26
fix(sql): RE_SQL_PARAMS matches PostgreSQL :: cast operator as named parameter


genro:
## Bug

The `RE_SQL_PARAMS` regex used in `prepareSqlText()` across all SQL adapters matches `:name` inside PostgreSQL's `::type` cast syntax.

For example, `CAST(x AS vector)` can also be written as `x::vector` in PostgreSQL. The regex `r":(\S\w*)(\W|$)"` matches `:vector` as a named parameter, causing the query to fail with a missing parameter error.

This affects any raw SQL or formulaColumn that uses the `::` cast shorthand — not just pgvector but also `::text`, `::integer`, `::jsonb`, etc.

### Current workaround

`execute.py` (line ~119) has an escape mechanism: `\:` is replaced with `chr(1)` before preprocessing, then restored after. So `\::vector` survives. But this requires callers to know about the escape, and it doesn't help with `db.table().query()` where the SQL is generated by the compiler.

## Root cause

The regex does not check what precedes the `:`:

```python
# gnrpostgres3.py, gnrpostgres.py, gnrfourd.py:
RE_SQL_PARAMS = re.compile(r":(\S\w*)(\W|$)")

# gnrpostgres8000.py, gnrmysql.py, gnrmssql.py, gnrdb2_400.py:
RE_SQL_PARAMS = re.compile(r":(\w*)(\W|$)")
```

## Proposed fix

Add a negative lookbehind `(?<!:)` to exclude `::` casts:

```python
# Before:
RE_SQL_PARAMS = re.compile(r":(\S\w*)(\W|$)")

# After:
RE_SQL_PARAMS = re.compile(r"(?<!:):(\S\w*)(\W|$)")
```

`(?<!:)` means "the `:` must NOT be preceded by another `:`". This way:
- `:name` → matches (GenroPy named parameter)
- `::vector` → the second `:` is preceded by `:` → **no match**
- `::text`, `::jsonb`, `::integer` → same, no match

### Affected adapters

All 7 adapters define their own `RE_SQL_PARAMS`. The fix applies to all of them:
- `gnrpostgres3.py` (line 37)
- `gnrpostgres.py` (line 49)
- `gnrpostgres8000.py` (line 35)
- `gnrfourd.py` (line 32)
- `gnrmysql.py` (line 44)
- `gnrmssql.py` (line 39)
- `gnrdb2_400.py` (line 35)

[...]

PR #586, opened 2026-02-26
fix(sql): RE_SQL_PARAMS skips PostgreSQL :: cast syntax

genro:
## Summary

- Add negative lookbehind `(?<!:)` and lookahead `(?!:)` to the `RE_SQL_PARAMS` regex in all 7 SQL adapters
- The regex now matches `:name` (named parameters) but skips `::type` (PostgreSQL cast syntax)
- Handles edge cases like `:param::vector` where a parameter is immediately followed by a cast

## Motivation

`prepareSqlText()` converts `:name` → `%(name)s` for psycopg2. The regex `r":(\S\w*)(\W|$)"` also matches `:vector` inside `::vector`, breaking any query that uses PostgreSQL cast shorthand (`::text`, `::integer`, `::jsonb`, `::vector`, etc.).

The existing workaround (`\:` → `chr(1)` escape in `execute.py`) requires callers to know about it and doesn't help with compiler-generated SQL.

## Test plan

- [x] 24 unit tests (both regex variants × 12 cases): named params, `::` casts, mixed expressions
- [x] Edge case: `:target::vector` correctly extracts only `target`

Ref #585


gnrpy/tests/sql/test_re_sql_params.py
cgabriel, line 1: This test verifies a locally defined regex, it doesn't verify the actual code used in the adapters, which can diverge. So the value of the test is pretty low, it doesn't ensure that the adapter works properly, it only provide a demonstration that such regexp works as expected, but providing locals one. If the adapter gets changed, the test will never fail.

cgabriel (changes requested):

gnrpy/tests/sql/test_re_sql_params.py
genro: You're absolutely right — testing a local copy of the regex provides no real safety net.

Replaced with `TestPrepareSqlTextDoubleColon` in `test_gnrbaseadapter.py`: it calls `prepareSqlText` on real `gnrpostgres.SqlDbAdapter` and `gnrmysql.SqlDbAdapter` instances (the two regex variants), verifying the actual output SQL.

Verified that with the old regex the tests fail (`:target::vector` produces `%(target)s:%(vector)s`), and with the fix they pass (`::vector` is preserved).

Removed `test_re_sql_params.py`.

cgabriel (approved):

PR #612, opened 2026-02-26
feat(deps): replace webob with werkzeug exceptions

mbertoldi:
Migrate exception handling from webob to werkzeug across WSGI site implementation. Updates all HTTP exception classes including NotFound, InternalServerError, BadRequest, Forbidden, and Unauthorized. Adjusts exception constructor parameters to use werkzeug's API with description parameter instead of separate message/comment arguments.


cgabriel (approved):
I've added a small patch to remove 2 imports referencing webob exceptions in gnr.web.gnrwebpage_plugins.

I've not added, as discussed, a mixin to raise deprecation warning using old names while raising werkzeug's related exception, since they're not used in any repo. I've only found direct references to webob in colto, which has webob as a direct dependency, so it's not an issue and there are no immediate advantages by doing so.

Issue #626, opened 2026-03-02
fix: indexed method lost for TSV columns (always generates btree instead of gin)


genro:
## Bug

When a column has `indexed=dict(method='gin')`, the `method` parameter is lost during migration and a **btree** index is generated instead of **gin**. This causes migration failure on `tsvector` columns because btree cannot index rows larger than 8191 bytes.

## Root cause

Two problems in `gnrsqlutils.py` (the old `SqlModelChecker` path used by `model.check()` / `applyModelChanges()`):

### 1. `indexed` dict is coerced to boolean

In `gnrpy/gnr/sql/gnrsqlmodel/columns.py` line 224:

```python
indexed = boolean(self.attributes.get('indexed'))
```

When `indexed=dict(method='gin')`, `boolean()` returns `True` and the dict is lost. Then line 227:

```python
self.table._indexedColumn[self.name] = {'columns': self.name, 'unique': unique}
```

Only `columns` and `unique` are saved — `method` is discarded.

### 2. `_buildIndex` / `createIndex` don't accept `method`

In `gnrsqlutils.py`, `_buildIndex()` and `_gnrbaseadapter.createIndex()` have no `method` parameter. They always generate btree indexes.

## Expected behavior

1. `indexed=dict(method='gin')` should preserve the method and generate `CREATE INDEX ... USING gin (...)`.
2. **Additionally**: columns with `dtype='TSV'` (tsvector) should **automatically** use `method='gin'` even when `indexed=True` (without explicit `method='gin'`). A btree index on a tsvector column is never useful and will fail on large documents.

## Reproduction

```python
# In any model:
tbl.column('my_tsv', dtype='TSV', indexed=dict(method='gin'))
# Or via ext_ltx:
tbl.column('document', ext_ltx=True)
```

Then `model.check()` + `applyModelChanges()` generates:

```sql
CREATE INDEX ... ON "schema"."table" ("my_tsv");
-- ERROR: index row requires 8560 bytes, maximum size is 8191
```

Instead of:

```sql
CREATE INDEX ... ON "schema"."table" USING gin ("my_tsv");
```

## Files involved

- `gnrpy/gnr/sql/gnrsqlmodel/columns.py` — preserve `method` from `indexed` dict
- `gnrpy/gnr/sql/gnrsqlutils.py` — pass `method` in `_sqlTableIndexes` / `_buildIndex`
- `gnrpy/gnr/sql/adapters/_gnrbaseadapter.py` — accept `method` in `createIndex`
[...]

PR #628, opened 2026-03-02
fix(#626): auto-GIN index for TSV columns

genro:
## Summary

- Add `DTYPE_INDEX_CONFIG` dict that maps dtype to index configuration (method, required)
- For `TSV`: `method='gin'` and `required=True` — GIN indexes are always created, btree is never used
- The config acts via `setdefault`, so explicit `indexed=dict(method=...)` from the model always wins
- Two new tests: `test_04e` (explicit GIN) and `test_04f` (auto-GIN for TSV)

## Problem

A `dtype='TSV'` column with `indexed=True` generated a btree index, which is useless for tsvector and fails on large documents with `index row requires 8560 bytes, maximum size is 8191`.

## Test plan

- [x] `test_04e_add_column_with_gin_index` — `indexed=dict(method='gin')` generates `USING gin`
- [x] `test_04f_tsv_indexed_true_auto_gin` — `indexed=True` on TSV auto-generates GIN
- [x] Full SQL test suite: 297 passed, 0 failed
- [x] flake8: zero errors on modified files

Ref #626

### DIFF
--- a/gnrpy/gnr/sql/gnrsqlmigration.py
+++ b/gnrpy/gnr/sql/gnrsqlmigration.py
@@ -27,6 +27,10 @@ COL_JSON_KEYS = ("dtype","notnull","sqldefault","size","unique","extra_sql","gen
 
 GNR_DTYPE_CONVERTER = {'X':'T', 'Z':'T', 'P':'T'}
 
+DTYPE_INDEX_CONFIG = {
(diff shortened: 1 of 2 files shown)


genro:
Chiusa: branch basato su master per errore. Riaperta da develop come fix/indexed-method-gin-v2.

Issue #630, opened 2026-03-02
feat: auto-switch model.check() to SqlMigrator for Postgres


genro:
## Feature

When `model.check()` / `applyModelChanges()` is called on a Postgres database, it should automatically use the new `SqlMigrator` path (`gnrsqlmigration`) instead of the old `SqlModelChecker` path (`gnrsqlutils`).

## Motivation

The old `SqlModelChecker` path has known limitations:
- `indexed=dict(method='gin')` is lost because `boolean()` coerces the dict to `True` (see #626)
- `_buildIndex` / `createIndex` don't accept `method` parameter
- No support for advanced index options (with, tablespace, where)

The new `SqlMigrator` already handles all of this correctly. Rather than fixing the old code, we can redirect Postgres users to the new path transparently.

## Proposed approach

In `gnrsqlmodel.py`, method `check()`:

```python
def check(self, applyChanges=False):
    if self.db.implementation in ('postgres', 'postgres3'):
        migrator = SqlMigrator(self.db)
        migrator.prepareMigrationCommands()
        changes = migrator.getChanges()
        self.modelChanges = [changes] if changes else []
        self.modelBagChanges = Bag()
        if applyChanges:
            migrator.applyChanges()
    else:
        checker = SqlModelChecker(self.db)
        self.modelChanges = checker.checkDb()
        self.modelBagChanges = checker.bagChanges
        if applyChanges:
            self.applyModelChanges()
    return bool(self.modelChanges)
```

## Benefits

- Zero impact on SQLite users (old path unchanged)
- Postgres users get all migration features automatically
- Old `SqlModelChecker` code can be gradually deprecated
- Closes the remaining part of #626 (old path bugs become irrelevant)

Ref #626

PR #631, opened 2026-03-02
feat(#630): redirect gnrdbsetup to gnr db migrate for Postgres

genro:
## Summary

- When `gnrdbsetup` detects a Postgres database, it prints a deprecation warning and delegates to `gnr db migrate`
- SQLite path remains unchanged
- This makes the old `SqlModelChecker` bugs (like #626) irrelevant for Postgres users

## Test plan

- [x] flake8: zero errors
- [ ] Manual: `gnrdbsetup <postgres_instance>` should print warning and run migrate
- [ ] Manual: `gnrdbsetup <sqlite_instance>` should work as before

Ref #630

### DIFF
--- a/gnrpy/gnr/app/cli/gnrdbsetup.py
+++ b/gnrpy/gnr/app/cli/gnrdbsetup.py
@@ -184,6 +184,11 @@ def main():
 
     
     app, storename = get_app(options)
+    if app.db.implementation in ('postgres', 'postgres3'):
+        logger.warning('gnrdbsetup is deprecated for Postgres. Redirecting to gnr db migrate.')
+        from gnr.db.cli.gnrmigrate import main as migrate_main
+        migrate_main()
+        return
     errordb = []
     if storename == '*':
         stores = [None] + sorted(app.db.dbstores.keys())


cgabriel (changes requested):
I don't see the point of this approach. The 2 commands have different options, so delegating the execution to a different one may cause problems, since the argument parsing occurs before the delegation.

It could make sense to just deprecate 'gnr db setup' and doing nothing, but automatic delegation is not really a good choice.  

'gnrdbsetup' raise a deprecation warning only because of its name, it doesn't change the implementation used. 

A better approach would be, IMHO, to have just 1 command for all the implementations, acting differently on adapter capabilities, and if a command has to be deprecated, it gets deprecated. 

If the solutions is only caused by an agent not executing the correct command, to me it's a problem of the agent itself.

genro:
@cgabriel agreed. The delegation approach is wrong for the reasons you described.

The plan now is to go in the opposite direction: remove `gnr db migrate` entirely and keep `gnrdbsetup` as the single canonical command. It has been the standard command for years and works correctly — there's no real reason to replace it with a different one.

I'll rework this PR accordingly: deprecate and remove `gnr db migrate`, keep `gnrdbsetup` as-is.

genro:
@mbertoldi @fporcari what do you think? Should we drop `gnr db migrate` and keep `gnrdbsetup` as the single canonical command?

cgabriel:
@genro I'll surely not keep gnrdbsetup, since it's already deprecated.

fporcari:
I think @genro was referring to the gnr db setup (not gnrdbsetup) so “setup” instead of “migrate”.

cgabriel:
> I think @genro was referring to the gnr db setup (not gnrdbsetup) so “setup” instead of “migrate”.

When asking for a choice between two, better be sure what the proposal is.

cgabriel:
I'm converting this to Draft until it's reworked after a decision on which tool has to be deprecated.

My suggestion: create a single cli tool which handle migration for postgres and old setup of adapter not supporting migrations (depending on the adapter of the instance), and then have an alias name with the other name, marked as deprecated. For example, if we choose "gnr db setup" as the final choice, move all the setup AND migration login into a module (for testing), add a CLI tool with such name, and then add a CLI tool named 'gnr db migrate' which executes the same login, but with a deprecation warning.

PR #645, opened 2026-03-03
fix(menu): propagate menuLineBadge to package branches and cleanup

fporcari:
## Summary
- Pop `menuLineBadge` from kwargs in `nodeType_tableBranch` to prevent unexpected keyword arguments
- Remove dead mobile code path in `frameplugin_menu.py` (condition was always false due to `&& false`)
- Fix badge update logic: hide badge when count is 0 (set to `null`) instead of keeping a stale value
- Fix 3-space indentation to 4-space in `PackageMenuResolver` and `DirectoryMenuResolver` (PEP 8)

## Test plan
- [x] flake8 passes with zero errors on modified files
- [x] All 1340 tests pass (147 skipped)
- [x] No unused imports


cgabriel:
Moved to Draft due to commits coming in after PR openining

cgabriel (approved):

Issue #564, opened 2026-02-24
gnrcore/[sys,adm] package decoupling


cgabriel:
We need to remove the needed inside the framework (or control the existance) from gnrcore's sys and adm packages, in order to move such packages in a different repository (among others gnrcore package), which will then become a possible dependency (this will be a following step in changes)

genro:
## Analysis & proposed approach: Capability Registry

### Current state

A scan of the framework core (`gnrpy/gnr/`) found ~90 direct references to `adm` and `sys` tables (preference, userobject, counter, htmltemplate, service, task, error, etc.). These hard-coded `self.db.table('adm.xxx')` calls prevent using the framework without gnrcore packages.

### Proposed solution: Capability Registry via pkgBroadcast

Instead of hard-coding package names, the framework maintains an **empty** `capabilities` dict. During startup, a `pkgBroadcast('registerCapabilities', app)` lets each package register what it provides. Packages are called in load order — **last one wins**, so a custom package can override any capability.

```python
# GnrApp.__init__
self.capabilities = {}
# ... after packages are loaded ...
self.pkgBroadcast('registerCapabilities', self)
```

```python
# adm/main.py
def registerCapabilities(self, app):
    app.capabilities['preference'] = self
    app.capabilities['userobject'] = self
    app.capabilities['counter'] = self
    app.capabilities['htmltemplate'] = self
    app.capabilities['audit'] = self
```

```python
# sys/main.py
def registerCapabilities(self, app):
    app.capabilities['service_registry'] = self
    app.capabilities['task_scheduler'] = self
    app.capabilities['error_log'] = self
    app.capabilities['record_lock'] = self
```

The framework replaces hard-coded calls:

```python
# Before (hard-coded)
self.db.table('adm.preference').getPreference(path, pkg=self.name)

# After (capability-based)
provider = self.application.capabilities.get('preference')
if provider:
    provider.getPreference(path, pkg=self.name)
```

### Key properties

- **No predefined list** — the dict starts empty, packages fill it
- **Load order wins** — packages are broadcast in order, so a later package can override an earlier one
- **Replaceable** — a custom package can provide its own `preference` implementation
- **Graceful degradation** — if capability is `None`/missing, framework skips the feature
- **No `adm`/`sys` references** left in framework core
- **Inspectable** — at runtime you know exactly who provides what

### Dependencies found in framework core

| Area | Tables | References |
|------|--------|-----------|
[...]

PR #652, opened 2026-03-03
feat: capability registry for package decoupling

genro:
## Summary

This PR introduces a **capability registry** pattern to decouple the GenroPy framework core (`gnrpy/gnr/`) from optional packages like `adm` and `sys`.

The goal (ref #564) is to allow using the framework without `adm`/`sys` packages — applications that don't need preferences, user objects, counters, etc. should work without those packages being installed.

## Approach: Capability Registry

Instead of hard-coding references like `self.db.table('adm.preference')` throughout the framework, we introduce:

1. **`app.capabilities`** — a dict populated at startup via `pkgBroadcast('registerCapabilities')`
2. **`app.addCapability(name, provider)`** — packages register themselves as providers for named capabilities
3. **`app.hasCapability(name)`** — guard check for optional features
4. **`@capability(name)` decorator** — applied to base class methods (body is `...`), delegates the call to the registered provider. Raises `GnrException` if no provider is registered.

### How it works

```python
# Base class (gnrdbo.py) — declares the interface
class GnrDboPackage:
    @capability('preference')
    def getPreference(self, path, dflt=None, mandatoryMsg=None):
        ...

# Provider (adm/main.py) — implements the capability
class Package(GnrDboPackage):
    def registerCapabilities(self, app):
        app.addCapability('preference', self)
        app.addCapability('user_preference', self)

    def getPreference(self, caller, path, dflt=None, mandatoryMsg=None, pkg=None):
        pkg = pkg or getattr(caller, 'name', None)
        return self.db.table('adm.preference').getPreference(path, pkg=pkg, ...)
```

The `@capability` decorator intercepts the call, finds the provider, and delegates — passing the original `self` (caller) as first argument so the provider knows who's calling.

For call sites that aren't on decorated methods (e.g. `GnrApp`, `GnrWsgiSite`, `GnrWebPage`), we use `hasCapability` guards:

```python
if self.application.hasCapability('preference'):
    provider = self.application.capabilities['preference']
    provider.envPreferences(self, username=self.user)
```

## What this PR implements

**Preference capability** as a complete proof-of-concept:
- `@capability('preference')` on `GnrDboPackage.getPreference` / `setPreference`
- `hasCapability('preference')` guards in `GnrApp`, `GnrWsgiSite`, `GnrWebPage`
- `user_preference` as separate capability (uses `adm.user` table)
- Full provider implementation in `adm/main.py`

## Remaining capabilities to migrate

If this approach is approved, the same pattern will be applied to all remaining dependencies (45 call sites across 11 files):

| Capability | Call sites | Files | Package |
|--|--|--|--|
| **userobject** | 9 | `gnrapp.py`, `gnrdbo.py`, `gnrwebpage.py`, `developer.py`, `btcbase.py`, `get_selection.py` | adm |
| **task_manager** | 8 | `gnrtask.py`, `gnrtask_new.py` | sys |
| **htmltemplate** | 5 | `gnrbaseclasses.py`, `gnrtablescript.py`, `gnrtablescript_new.py` | adm |
| **user_auth** | 5 | `gnrapp.py`, `gnrdbo.py`, `gnrwebpage.py`, `gnrwebapp.py` | adm/sys |
| **batch_log** | 4 | `btcbase.py`, `btcmail.py` | sys/adm |
| **counter** | 4 | `gnrdbo.py` | adm |
| **connection_log** | 3 | `gnrwsgisite.py` | adm |
| **error_log** | 2 | `gnrwsgisite.py` | sys |
| **data_retention** | 2 | `gnrapp.py` | sys |
| **audit** | 1 | `gnrwebapp.py` | adm |
| **shortcut** | 1 | `gnrwebpage.py` | adm |
| **tblinfo** | 1 | `gnrwebpage.py` | adm |

## Request for review

@mbertoldi @cgabriel @fporcari — please evaluate this approach before we proceed with the remaining capabilities. Key questions:

1. Is the **capability registry + decorator** pattern acceptable for decoupling?
2. Is the **provider receives `caller` as first argument** convention clear enough?
3. Any concerns about the `hasCapability` guard pattern for non-decorated call sites?
4. Should any capability groupings be changed (e.g. merge `preference` + `user_preference`)?

Ref #564

### DIFF
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -984,6 +984,8 @@ class GnrApp(object):
         #    self.config['menu']=self.instanceMenu
 
         self.localizer = AppLocalizer(self)
+        self.capabilities = {}
+        self.pkgBroadcast('registerCapabilities', self)
         self.onInited()
 
     def addPackage(self,pkgid,pkgattrs=None,pkgcontent=None):
@@ -1186,9 +1188,25 @@ class GnrApp(object):
         """Hook method called before the :ref:`instance <instances>` initialization"""
         pass
 
+    def addCapability(self, name, provider, replace=False):
+        """Register a capability provider.
+
+        Packages call this during ``registerCapabilities`` to declare
+        the services they provide (e.g. ``'preference'``, ``'userobject'``).
+        Packages are broadcast in load order — last one wins when
+        *replace* is ``True``.
+
+        :param name: capability identifier (e.g. ``'preference'``)
+        :param provider: the package instance that provides this capability
+        :param replace: if ``False`` (default), raise on duplicate
+        """
+        if name in self.capabilities and not replace:
+            raise KeyError(f"Capability '{name}' is already registered")
+        self.capabilities[name] = provider
+
     def onInited(self):
         """Hook method called after the instance initialization is complete.
-        
+
         By default, it will call the :meth:`onApplicationInited()
         <gnr.app.gnrapp.GnrPackage.onApplicationInited>` method of each package"""
         self.pkgBroadcast('onApplicationInited')
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -1204,6 +1204,10 @@ class GnrApp(object):
             raise KeyError(f"Capability '{name}' is already registered")
         self.capabilities[name] = provider
 
+    def hasCapability(self, name):
+        """Return ``True`` if a provider is registered for *name*."""
+        return name in self.capabilities
+
     def onInited(self):
         """Hook method called after the instance initialization is complete.
 
@@ -1240,12 +1244,14 @@ class GnrApp(object):
         return (found_locale or 'en-GB').replace('_','-')
 
     def setPreference(self, path, data, pkg):
-        if self.db.package('adm'):
-            self.db.table('adm.preference').setPreference(path, data, pkg=pkg)
+        provider = self.capabilities.get('preference')
+        if provider:
+            provider.setPreference(self, path, data, pkg=pkg)
 
     def getPreference(self, path, pkg=None, dflt=None, mandatoryMsg=None):
-        if self.db.package('adm'):
-            return self.db.table('adm.preference').getPreference(path, pkg=pkg, dflt=dflt, mandatoryMsg=mandatoryMsg)
+        provider = self.capabilities.get('preference')
+        if provider:
+            return provider.getPreference(self, path, pkg=pkg, dflt=dflt, mandatoryMsg=mandatoryMsg)
     
     def getResource(self, path, pkg=None, locale=None):
         """TODO
--- a/gnrpy/gnr/app/gnrdbo.py
+++ b/gnrpy/gnr/app/gnrdbo.py
@@ -12,7 +12,7 @@ from collections import defaultdict
 from gnr.core.gnrlang import boolean
 from gnr.core.gnrbag import Bag
 from gnr.core.gnrstring import splitAndStrip,templateReplace,fromJson,slugify
-from gnr.core.gnrdecorator import public_method,extract_kwargs
+from gnr.core.gnrdecorator import public_method,extract_kwargs,capability
 from gnr.core.gnrdict import dictExtract
 
 from gnr.app import logger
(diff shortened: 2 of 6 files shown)


mbertoldi:
I like it, what if `@capability('preference')` is declared multiple times?

cgabriel:
> I like it, what if `@capability('preference')` is declared multiple times?

the decorator is for calling as far as I've understood the proposal, while addCapabitilites introduces a similar problem. They can be overriden using 'replace', while IMHO using a prefix would be helpful to identify the specific functionality, and overrides are more explicits.

Apart from your comment, it seems to me that this whole proposal introduces a new layer of complexity in order to change from this:

```
            if not self.db.package('adm'):
```

to this:

```
            if not self.gnrapp.hasCapability('preference'):
```

while the original point was the problem that the framework relies on a package which relies on the framework itself. From my point of view the whole usage of any genropy packages code should just be removed from anything in 'gnr' namespace.

My 2¢.

genro:
@cgabriel I understand your point, but the capability registry is not just a rename of the current check — it's an **intermediate step** designed to manage the transition gradually without introducing a tight dependency.

The key difference is: with capabilities, **the framework declares what it needs, not where to find it**. This opens up several possibilities:

1. **Alternative implementations at framework level**: for example, preferences could have a default filesystem-based implementation directly in the framework, while the `adm` package overrides it with the current DB-based approach. Same for user objects — the framework could provide a basic storage-node/filesystem/S3 implementation, and `adm` enhances it with its full-featured version.

2. **Graceful degradation**: whoever uses the framework can choose not to provide certain capabilities. An application without `adm` still works — it just won't have DB-backed preferences. If a capability documents its contract clearly, anyone can implement it, not just `adm`/`sys`.

3. **GUI integration**: the capability system is designed to also control the UI — for instance, the grid configurator could declare that it requires the `userobject` capability, and if it's not available, the webstructs simply don't build that piece of interface. This is something you cannot do cleanly with `if self.db.package('adm')`.

So the end goal is exactly what you describe — removing all `gnr` → package dependencies. But capabilities give us a **gradual migration path**: we decouple the contract (what the framework needs) from the provider (who implements it), document each capability independently, and allow both framework-level defaults and package-level overrides.

Going from tight coupling directly to "no code from packages in gnr" in one shot would be a much riskier refactor. Capabilities let us do it incrementally, one feature at a time, while keeping everything working.

cgabriel:
> @cgabriel I understand your point, but the capability registry is not just a rename of the current check — it's an **intermediate step** designed to manage the transition gradually without introducing a tight dependency.


I see, I can understand the point, I was just suggesting that, since we're cleaning up the framework with its internal implementation, it could be an occasione to find a definitive solution for the issue without introducing a complex layer

> The key difference is: with capabilities, **the framework declares what it needs, not where to find it**. 

This was straightforward.

> This opens up several possibilities:
> 1. **Alternative implementations at framework level**: for example, preferences could have a default filesystem-based implementation directly in the framework, while the `adm` package overrides it with the current DB-based approach. Same for user objects — the framework could provide a basic storage-node/filesystem/S3 implementation, and `adm` enhances it with its full-featured version.

So when you say "directly in the framework", you're saying you'd like to move the implementation logic provided by the package inside the framework itself, which is already a decoupling, right?

> 2. **Graceful degradation**: whoever uses the framework can choose not to provide certain capabilities. An application without `adm` still works — it just won't have DB-backed preferences. If a capability documents its contract clearly, anyone can implement it, not just `adm`/`sys`.
> 3. **GUI integration**: the capability system is designed to also control the UI — for instance, the grid configurator could declare that it requires the `userobject` capability, and if it's not available, the webstructs simply don't build that piece of interface. This is something you cannot do cleanly with `if self.db.package('adm')`.

Point 3 is nice, please just keep in mind that the reference to `if self.db.package('adm')` was just an example of the implementation change, switch to check for package to capabilities, to explain my point.

> So the end goal is exactly what you describe — removing all `gnr` → package dependencies. But capabilities give us a **gradual migration path**: we decouple the contract (what the framework needs) from the provider (who implements it), document each capability independently, and allow both framework-level defaults and package-level overrides.
> Going from tight coupling directly to "no code from packages in gnr" in one shot would be a much riskier refactor. Capabilities let us do it incrementally, one feature at a time, while keeping everything working.

If you want to go gradual yes, that's the way.

PR #656, opened 2026-03-04
feat: xmltransform service for XML to HTML conversion via XSLT

dgpaci:
## Summary

- Adds a new `xmltransform` service type (`gnrpy/gnr/lib/services/xmltransform.py`)
- Implements `Main` service class that applies an XSL stylesheet to transform XML into HTML using `lxml`
- Adds `ServiceParameters` UI component to upload/delete the XSL file from the admin interface (`sys/resources/services/xmltransform/service.py`)

## How it works

1. Create a service of type `xmltransform` in the system services admin
2. Upload an XSL stylesheet via the UI (drag & drop)
3. The service exposes `xml_to_html(xml, **kwargs)` which applies the XSL transform and returns HTML bytes

## Test plan

- [x] Tested locally with FPA12, FPR12 and FoglioStileAssoSoftware stylesheets against a sample FatturaPA XML — all produce valid HTML output
- [ ] Verify XSL upload/delete UI in admin
- [ ] Verify service loads correctly on application startup

## Notes

This PR is a **required dependency** for [erpyready#feature/ftel_service](https://github.com/softwellsrl/erpyready) — that branch delegates XML→HTML conversion to this service and will not work without it.


projects/gnrcore/packages/sys/resources/services/xmltransform/service.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env pythonw
+# -*- coding: utf-8 -*-
+#
+#  Created by Davide Paci on 2024-10-21.
+#  Copyright (c) 2024 Softwell. All rights reserved.
```
cgabriel: s/2024/2026

projects/gnrcore/packages/sys/lib/services/xmltransform.py
cgabriel, line 1: Since we want to reduce the dependencies of the framework from other packages, I'm suggesting to move this file into projects/gnrcore/packages/sys/lib/services/ to keep the implementation and its UI componenets all together.

cgabriel (changes requested):

projects/gnrcore/packages/sys/resources/services/xmltransform/service.py
dgpaci: Sad but true, it was actually created in 2024...

projects/gnrcore/packages/sys/lib/services/xmltransform.py
dgpaci: I agree, I moved service to suggested path

projects/gnrcore/packages/sys/resources/services/xmltransform/service.py
dgpaci: Anyway, I'm updating year since feature is being closed now

projects/gnrcore/packages/sys/tests/xmltransform_test.py
cgabriel, line 1: Since the implementation has been moved into 'sys' package, the test should also go into packages/sys/tests/, for the same reason we moved the files from the framework into the package.

cgabriel (changes requested):
The tests should belong with the implementation

cgabriel:
I'm fixing the tests as discussed with @dgpaci, then I'll merge

cgabriel (dismissed):

cgabriel (approved):

PR #650, opened 2026-03-03
feat(sql): complete addMacro registry with all SQL macros (#617)

genro:
## Summary

Completes the `db.addMacro()` macro registration infrastructure (issue #617) so that **every SQL macro** is listed in a single registry (`_macro_registry`) on the db instance.

- **Phase 1** built the core infrastructure: `GnrSqlDb.addMacro()`, `_macro_registry` dict, `MacroExpander.register()`/`replace()`
- **Phase 2** (this PR) registers the remaining app-level macros (`#PREF`, `#THIS`, `#BAG`, `#BAGCOLS`) and enables package-level registration via `pkgBroadcast`

This is **registration only** — the old hard-coded expansion in `compiler.py` remains untouched. The next step will replace the old `sub()` calls with the registry-based mechanism.

`#ENV` is excluded per issue #618 (zero usages, deprecated in favour of `:env_*` bind parameters).

## Registration hierarchy

```
GnrSqlDb.registerMacros()          →  IN_RANGE, PERIOD
  └─ adapter.registerMacros(db)    →  TSQUERY, TSRANK, TSHEADLINE, VECQUERY, VECRANK  (Postgres)
GnrSqlAppDb.registerMacros()       →  PREF, THIS, BAG, BAGCOLS
  └─ pkgBroadcast('registerMacros') →  package-defined macros (e.g. UPPERCASE in test_invoice)
```

## API

```python
# Register a new macro
db.addMacro('MY_MACRO', compiled_regex, callback_or_None)

# Duplicate detection (raises KeyError)
db.addMacro('MY_MACRO', regex, cb)  # KeyError: already registered

# Explicit override
db.addMacro('MY_MACRO', new_regex, new_cb, replace=True)

# Package-level registration (in package main.py)
class Package(GnrDboPackage):
    def registerMacros(self, db):
        db.addMacro('MY_MACRO', regex, callback)
```

## Files changed

| File | Change |
|------|--------|
| `gnrpy/gnr/sql/gnrsql/db.py` | `_macro_registry` dict, `addMacro()` with duplicate detection, `registerMacros()` |
| `gnrpy/gnr/sql/adapters/_gnrbaseadapter.py` | `MacroExpander`: `register()`, `replace()`, `context` dict |
| `gnrpy/gnr/sql/adapters/_gnrbasepostgresadapter.py` | Adapter-level `registerMacros()` for TS*/VEC* |
[...]

Issue #617, opened 2026-02-28
Refactor: make SQL compiler macros extensible via db.addMacro()


genro:
## Context

The SQL compiler (`gnrsqldata/compiler.py`) currently handles macros in two different ways:

1. **Adapter-level macros** (`MacroExpander`): `#TSQUERY`, `#TSRANK`, `#TSHEADLINE`, `#VECQUERY`, `#VECRANK` — extensible, registered per-adapter via class-level `macros` dict.

2. **Hardcoded closures** in `getFieldAlias`: `#PREF`, `#ENV`, `#THIS`, `#BETWEEN`, `#PERIOD`, `#BAG`, `#BAGCOLS` — each implemented as a local function with its own regex.

## Problem

`#PREF` resolves package preferences via `curr_tblobj.pkg.getPreference()`, which depends on the app layer (`GnrSqlAppDb` / `GnrApp`). This creates an architectural coupling: the SQL compiler depends on the app layer.

`#ENV` accesses `db.currentEnv` which is pure SQL layer — it belongs in the compiler.

## Proposal

Add a `db.addMacro(name, regex, callable)` API that allows registering custom macros at runtime:

```python
# In GnrSqlAppDb (app layer) or during package setup:
db.addMacro('PREF', PREFFINDER_REGEX, expand_pref_callable)
```

The compiler would then use the same `MacroExpander.replace()` mechanism for all macros, removing the hardcoded closures for app-layer concerns.

### Steps

1. Extend `MacroExpander` to support instance-level macro registration (currently macros are class-level)
2. Add `db.addMacro()` API on `GnrSqlDb`
3. Move `#PREF` registration to `GnrSqlAppDb` or package setup
4. Evaluate which other hardcoded macros could benefit from this pattern

## Related

- Discovered during 100% coverage work on compiler.py (branch `feature/compiler_100_coverage`)
- `expandPref` closure: compiler.py:315-319
- `MacroExpander` base: `_gnrbaseadapter.py:42-61`
- `MacroExpander` postgres: `_gnrbasepostgresadapter.py:13-44`

PR #660, opened 2026-03-04
fix(sql): remove premature macro registry-to-expander copy

genro:
## Summary

- Remove two lines in `compiler.py` that copied all registered macros (including those with `callback=None`) from `db._macro_registry` into `MacroExpander._registered_macros`
- These lines were introduced in PR #650 but belong to the usage phase, not the preparation phase
- Since registered macros take precedence over class-level ones in `replace()`, macros with `None` callback (PREF, THIS, BAG, BAGCOLS, TSQUERY, TSRANK, etc.) caused `TypeError` when invoked

## Test plan

- [x] All 666 SQL tests pass (1 pre-existing failure on develop unrelated to this change)
- [ ] Verify that adapter macros (TSQUERY, TSRANK, TSHEADLINE) expand correctly on Postgres queries
- [ ] Verify that app-level macros (PREF, THIS, BAG, BAGCOLS) expand correctly

Ref #617

### DIFF
--- a/gnrpy/gnr/sql/gnrsqldata/compiler.py
+++ b/gnrpy/gnr/sql/gnrsqldata/compiler.py
@@ -236,8 +236,6 @@ class SqlQueryCompiler(object):
         self.aliasPrefix = aliasPrefix or 't'
         self.locale = locale
         self.macro_expander = self.db.adapter.macroExpander(self)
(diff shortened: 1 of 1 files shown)


fporcari (commented):
it fix the regression

fporcari (approved):

Issue #666, opened 2026-03-06
feat: GnrListener — declarative event handling via PostgreSQL LISTEN/NOTIFY


genro:
## Context

GenroPy now has (after PR #664):

- `adapter.notify(channel, payload)` — transactional NOTIFY
- `_dbNotify()` in `write.py` — automatic NOTIFY on tables with `notify=True|'field1,field2'`
- Single channel `dbevent` with JSON payload: `{table, pkey, event, fields?}`

### Problem with current `adapter.listen()`

The existing `adapter.listen()` in `gnrpostgres.py` (line 231) has significant limitations:

1. **Blocking** — uses `select()` in a tight loop, blocks the entire process
2. **Single channel** — accepts only one channel name per call
3. **Tied to the adapter** — lives at the wrong abstraction level (SQL adapter vs application)
4. **No filtering** — the caller must parse and filter events manually
5. **No signal handling** — no clean shutdown on SIGTERM/SIGINT
6. **Legacy psycopg2 2.0 compat code** — dead branches for a version nobody uses

This method should be **deprecated** and eventually removed in favor of a proper application-level listener.

## Proposal: GnrListener

A new class `GnrListener` and a `GnrApp.listen()` method that allows packages to register event handlers declaratively.

### Usage (two lines to start)

\`\`\`python
from gnr.app.gnrapp import GnrApp
app = GnrApp('sourcerer')
app.listen()  # blocking — runs as a separate process (systemd service)
\`\`\`

### How app.listen() works

1. Creates a `GnrListener` instance
2. Broadcasts `register_listeners(listener)` to all packages
3. Calls `listener.run()` (infinite loop)

### Default behavior of register_listeners in GnrPackage

- Scans its own tables that have the `notify` attribute
- For those that have an `on_notify` method on the Table class → auto-registers the handler

### Package override (for custom handlers, cross-package tables, different channels)

\`\`\`python
# packages/sourcerer/main.py
class Package(GnrPackage):
    def register_listeners(self, listener):
        super().register_listeners(listener)  # default: own tables with on_notify
        listener.register('dbevent', self.on_commit, table='gnrgh.commit')
        listener.register('embedding_changed', self.on_embedding)

    def on_commit(self, payload):
        # git pull + reindex
        ...

    def on_embedding(self, payload):
        # regenerate NULL embeddings
        ...
\`\`\`

### Handler on table

\`\`\`python
# packages/gnrmcp/model/skill.py
class Table(object):
[...]

PR #669, opened 2026-03-06
feat(listener): GnrListener with @listen decorator and autodiscovery

genro:
## Summary

Declarative event handling system built on top of PostgreSQL LISTEN/NOTIFY (PR #664). Table methods decorated with `@listen()` are auto-discovered and dispatched by a long-running `GnrListener` process.

### New components

- **`@listen()` decorator** (`gnrdecorator.py`) — marks table methods as NOTIFY handlers. `@listen()` listens on `dbevent` (default), `@listen('topic')` on a custom channel. Always with parentheses for future extensibility.
- **`tblobj.notify(channel, **kwargs)`** (`gnrdbo.py`) — shortcut for sending custom NOTIFY from any table method. Payload includes `table`, `user`, `page_id`, `ts` (UTC timestamp) plus caller kwargs.
- **`GnrListener`** (`gnrlistener.py`) — multi-channel PostgreSQL LISTEN dispatcher with filter-based routing, signal handling (SIGTERM/SIGINT), coalescing, and optional thread pool.
- **`GnrAppListener`** (`gnrapplistener.py`) — composition wrapper: accepts a `GnrApp` instance (or instance name string), auto-discovers all `@listen` methods across all table classes, and starts the event loop.
- **`gnrapplisten` CLI** (`gnr/app/cli/gnrapplisten.py`) — entry point for running the listener from the command line with `--workers`, `--timeout`, `--coalesce` options.
- **`adapter.listen_connection(channels)`** — opens a dedicated AUTOCOMMIT connection for LISTEN. Base adapter returns `None` (unsupported).
- **`adapter.poll_notifications(conn)`** — drains pending notifications from a connection. Base adapter returns `[]`. Postgres adapter calls `conn.poll()` + drains `conn.notifies`.

### Changes since review (addressing all 7 review comments)

1. **`toTypedJSON`/`fromTypedJSON` instead of `json.dumps`/`json.loads`** — all payload serialization now uses GenroPy's typed JSON for proper round-trip of `datetime`, `Decimal`, `date` etc.
2. **Enriched notify payloads** — both `tblobj.notify()` and `_dbNotify()` now include `user`, `page_id` (from `self.db.currentPage`) and `ts` (UTC timestamp) in every payload.
3. **Safe SQL parameter binding** — `adapter.notify()` uses `:payload` named parameter instead of string escaping. Handles `payload=None` case separately.
4. **`GnrAppListener`** — extracted from `GnrApp.listen()` into its own module for clean composition.
5. **`gnrapplisten` CLI** — dedicated entry point using `GnrCliArgParse`.
6. **`from gnr.app import logger`** — replaced `logging.getLogger()` with the project-standard logger throughout.
7. **`adapter.poll_notifications(conn)`** — polling logic delegated to adapter layer, keeping `GnrListener` adapter-agnostic.

#### 1. Enable automatic NOTIFY on a table (PR #664)

```python
def config_db(self, pkg):
    tbl = pkg.table('invoice', notify=True, ...)  # emits on dbevent
    # or notify='total,status' to include specific fields in the payload
```

#### 2. React to changes on your own table

```python
# In model/invoice.py
from gnr.core.gnrdecorator import listen

class Table(object):
    @listen()
    def on_change(self, payload):
        """Called on every insert/update/delete of this table."""
        print(f"Invoice {payload['pkey']} changed: {payload['event']}")
```

#### 3. Send and receive custom notifications

```python
# In model/invoice.py — trigger a custom event
class Table(object):
        if record['total'] > 50000:
            self.notify('fatturona', pkey=record['id'], total=str(record['total']))

# In model/alerts.py — react to it
class Table(object):
    @listen('fatturona')
    def on_big_invoice(self, payload):
        self.sendAlert(payload['pkey'], payload['total'])
```

#### 4. Run the listener

```bash
# Via CLI entry point
gnrapplisten myinstance
gnrapplisten myinstance --workers 4 --timeout 10 --coalesce 2

# Via GnrAppListener
from gnr.app.gnrapplistener import GnrAppListener
GnrAppListener('myinstance', workers=4).run()
```

### Notify payload structure

Every notification (both automatic `_dbNotify` and manual `tblobj.notify()`) includes:

```python
{
    'table': 'pkg.tablename',     # always present
    'pkey': '...',                 # record primary key
    'event': 'I'|'U'|'D',         # insert/update/delete (auto only)
    'user': 'username',            # from currentPage (None if no page)
    'page_id': '...',              # from currentPage (None if no page)
    'ts': datetime(UTC),           # timestamp, round-trips via toTypedJSON
    'fields': {...},               # only when notify='field1,field2'
}
```

### Thread pool (workers parameter)

By default (`workers=1`) handlers run synchronously in the main event loop. When `workers > 1`, `GnrListener` creates a `ThreadPoolExecutor(max_workers=N)` for parallel I/O-bound handler execution. Each handler is wrapped in `_safe_call()` for error isolation. The pool shuts down cleanly on SIGTERM/SIGINT.

### Deployment as a service

**supervisord** (recommended):
```ini
[program:gnr-listener]
command=gnrapplisten myinstance --workers 4
autostart=true
autorestart=true
stderr_logfile=/var/log/gnr-listener.err.log
stdout_logfile=/var/log/gnr-listener.out.log
```

**systemd**:
```ini
[Unit]
Description=GnrListener for myinstance
After=postgresql.service

[Service]
ExecStart=gnrapplisten myinstance --workers 4
Restart=always

[Install]
WantedBy=multi-user.target
```

## Test plan

- [x] 21 tests in `test_gnrlistener.py` — decorator, register, dispatch, thread pool, notify (psycopg2+psycopg3), autodiscovery (psycopg2+psycopg3), GnrAppListener init, CLI parser, entry point
- [x] 12 tests in `test_db_notify.py` — base adapter no-op, SQLite no-op, Postgres notify, LISTEN/NOTIFY integration, `_dbNotify` payload construction (insert, delete, update changed, update no-change)
- [x] CI green on Python 3.11, 3.12, 3.13, 3.14

## Future work

The old `adapter.listen(msg, timeout, onNotify, onTimeout)` method still exists on all adapters. It is a legacy single-channel blocking listener with callback-based dispatch. Now that GnrListener provides a superior multi-channel declarative alternative, the old method can be removed in a future cleanup pass.

Ref #666


gnrpy/gnr/app/gnrdbo.py
```diff
@@ -1534,6 +1534,21 @@ def _sequencesOnLoading(self,newrecord,recInfo=None):
...
+        """Send a custom NOTIFY on the given channel.
+
+        The payload always includes ``table`` (this table's fullname).
+        Any extra keyword arguments are merged into the payload.
+
+        Args:
+            channel: The PostgreSQL NOTIFY channel name.
+            **kwargs: Additional key-value pairs for the JSON payload.
+        """
+        import json
```
cgabriel: Please avoid runtime imports.

gnrpy/gnr/app/gnrapp.py
```diff
@@ -1838,9 +1838,36 @@ def hostedBy(self):
     def gnrdaemon(self):
         if not getattr(self,'_gnrdaemon',None):
             from gnr.web.gnrdaemonhandler import GnrDaemonProxy
-            self._gnrdaemon = GnrDaemonProxy(use_environment=True).proxy() 
+            self._gnrdaemon = GnrDaemonProxy(use_environment=True).proxy()
         return self._gnrdaemon
 
+    def listen(self, timeout=5, coalesce=1, workers=1):
```
cgabriel: I'm suggesting to have this implementation to be outside of the gnrapp module.

gnrpy/gnr/sql/adapters/gnrpostgres.py
cgabriel, line 1: I'd put the implementation in the gnr/sql/adapter/_gnrbasepostgresadapter.py in orde r to benefit both postgres and postgres3 adapters.

gnrpy/tests/sql/test_gnrlistener.py
```diff
@@ -0,0 +1,216 @@
...
+        assert results[0] != threading.current_thread().name
+
+
+# -- tblobj.notify() on Postgres ----------------------------------------------
+
+class TestTableNotify:
+    """tblobj.notify() sends NOTIFY with table and kwargs in payload."""
+
+    def test_notify_sends_payload(self, db_pg):
+        import psycopg2
```
cgabriel: Please avoid runtime imports

gnrpy/gnr/sql/adapters/gnrpostgres.py
cgabriel, line 1: Test for psycopg3 are missing.

cgabriel (changes requested):
See specific comments.

I'd also suggest to provide a CLI command to start the listener, to hide the underlying package/module organization allowing easier internal changes.

gnrpy/gnr/app/gnrlistener.py
```diff
@@ -0,0 +1,176 @@
...
+from __future__ import annotations
+
+import json
+import logging
+import select
+import signal
+import time
+from concurrent.futures import ThreadPoolExecutor
+
+log = logging.getLogger('gnr.listener')
```
cgabriel: Please use gnr.app.logger

gnrpy/gnr/app/gnrlistener.py
```diff
@@ -0,0 +1,176 @@
...
+
+        conn = self._listen_connection()
+        log.info('GnrListener started — channels: %s', ', '.join(sorted(self.channels)))
+
+        try:
+            while self._running:
+                if select.select([conn], [], [], self.timeout) != ([], [], []):
+                    conn.poll()
+                    processed = 0
+                    while conn.notifies:
```
cgabriel: Beware that psycopg3 uses conn.notifies(), it's not a list but a callable. This should be implemented at adapter level

gnrpy/gnr/app/gnrdbo.py
genro: Done in ed03f1f — moved `import json` to module level. In the latest commit (8fa3db9) we also replaced `json` entirely with `toTypedJSON`/`fromTypedJSON` from `gnr.core.gnrstring` for proper type round-trip (datetime, Decimal, etc.).

Additionally, `tblobj.notify()` now enriches payloads with `user`, `page_id`, and `ts` (UTC timestamp) from `self.db.currentPage` when available — this allows downstream listeners to correlate notifications with the originating user session.

gnrpy/gnr/app/gnrapp.py
genro: Done in ed03f1f — created `GnrAppListener` in `gnr/app/gnrapplistener.py` using composition: it wraps a `GnrApp` instance, performs autodiscovery of `@listen` handlers, and starts the listener. The `listen()` method has been removed from `GnrApp`.

Also added a CLI entry point `gnrapplisten` (in `gnr/app/cli/gnrapplisten.py`) that inherits `--loglevel`, `--debug`, `--version` from `GnrCliArgParse` and adds `-w/--workers`, `-t/--timeout`, `-c/--coalesce`.

gnrpy/gnr/sql/adapters/gnrpostgres.py
genro: Done in ed03f1f — `listen_connection()` moved to `gnr/sql/adapters/_gnrbasepostgresadapter.py` so both psycopg2 and psycopg3 adapters inherit it.

gnrpy/tests/sql/test_gnrlistener.py
genro: Done in ed03f1f — all runtime imports moved to module level in the test file.

gnrpy/gnr/sql/adapters/gnrpostgres.py
genro: Done in ed03f1f — added `TestTableNotifyPsycopg3` and `TestAutodiscoveryPsycopg3` test classes using a `db_pg3` fixture. The latest commit (8fa3db9) also adds `TestGnrAppListener` and `TestGnrAppListenCli` for coverage of the new composition wrapper and CLI entry point.

gnrpy/gnr/app/gnrlistener.py
genro: Good point — will check what `gnr.app.logger` provides and switch to it. Currently using `logging.getLogger('gnr.listener')` which is consistent with the gnr namespace but may not leverage the app-level logging configuration.

gnrpy/gnr/app/gnrlistener.py
genro: Acknowledged — the current `GnrListener.run()` loop uses the psycopg2 pattern (`select()` + `poll()` + `conn.notifies` list). For psycopg3 this should be delegated to the adapter level since `conn.notifies()` is a generator, not a list. Will address in a follow-up.

gnrpy/gnr/app/gnrlistener.py
genro: Done in 0d0822f — replaced `logging.getLogger('gnr.listener')` with `from gnr.app import logger`, following project convention. Also removed unused `logging` import from `gnrapplistener.py`.

gnrpy/gnr/app/gnrlistener.py
genro: Done in 0d0822f — added `poll_notifications(conn)` to the adapter layer (`_gnrbaseadapter.py` returns `[]`, `_gnrbasepostgresadapter.py` does `conn.poll()` + drains `conn.notifies`). `GnrListener.run()` now calls `self.db.adapter.poll_notifications(conn)` instead of directly accessing connection internals. If psycopg3 needs different polling logic it can override this method in its adapter.

genro:
All 7 review comments addressed across 3 commits:

### `ed03f1f` — initial review feedback
1. **Runtime imports** → moved to module level (`gnrdbo.py`, test file)
2. **Listener outside gnrapp** → created `GnrAppListener` in `gnrapplistener.py` (composition) + `gnrapplisten` CLI entry point
3. **listen_connection in shared adapter** → moved to `_gnrbasepostgresadapter.py`
4. **psycopg3 tests missing** → added `TestTableNotifyPsycopg3` and `TestAutodiscoveryPsycopg3` with `db_pg3` fixture

### `8fa3db9` — enriched payloads
- Replaced `json.dumps`/`json.loads` with `toTypedJSON`/`fromTypedJSON` for proper type round-trip
- Enriched all notify payloads with `user`, `page_id` (from `self.db.currentPage`) and `ts` (UTC timestamp)
- Used `:payload` named SQL parameter in `adapter.notify()` for safe escaping
- 33 tests total (21 listener + 12 db_notify)

### `0d0822f` — logger and poll delegation
5. **`from gnr.app import logger`** → replaced `logging.getLogger()` throughout
6. **`adapter.poll_notifications(conn)`** → polling logic delegated to adapter layer, `GnrListener` is now fully adapter-agnostic

CI green on all 4 Python versions (3.11, 3.12, 3.13, 3.14). Ready for re-review @cgabriel.

mbertoldi (dismissed):

gnrpy/tests/sql/test_gnrlistener.py
cgabriel: There are still plenty of runtime imports left in the module (the following is just a summary):

```
        import threading
        from gnr.app.gnrapplistener import GnrAppListener
        from gnr.core.cli import GnrCliArgParse
        import gnr.app.cli.gnrapplisten as cli_mod
        import importlib
```

gnrpy/pyproject.toml
cgabriel, line 1: This list contains the deprecated CLI commands, to support the migration from the old scheme. This new CLI is now 'gnr app applisten', so I'd 1/ renamed it to just 'gnrlisten.py', to obtain a non-redundant name ('gnr app listen') and 2/ remove from this list, since we are trying to not encourage the usage of the old script naming convention.

cgabriel (commented):
I've review the latest changes, now I have come architecture related comments:

1/ There is no locking mechanism nor an acknowledge of the operation executed triggered by the notification. thus

2/ Being there no lock, I can run multiple listen and have the operation executed twice (or more). If running a single listener/executor is mandatory, a locking system is needed. And if this is the case, then

3/ The solution can't scale horizontally. Everything must be executed on a single listen/executor process/host, which can't scale up

4/ Loop protection: what if a notification handler executes a change that triggers a new notification?

5/ Since the feature is supported only by a specific adapter (postgres), I suggest to add a new capability in gnr.sql.AdapterCapabilities, in order to have a simple method to verify if the instance can support certain operations (for example, when running 'gnr app listen <instance>' on an instance using an adapter not supporting the feature, can exit gracefully)

6/ the notification/execution queue is in memory, if the process get restart/crashes, the queue is lost.

gnrpy/tests/sql/test_gnrlistener.py
genro: Done — all runtime imports moved to module level. The test file now imports everything at the top:

```python
import threading
from concurrent.futures import ThreadPoolExecutor
from gnr.core.cli import GnrCliArgParse
from gnr.app.gnrapplistener import GnrAppListener
import gnr.app.cli.gnrlisten as cli_mod
import importlib
```

gnrpy/pyproject.toml
genro: Done — renamed `gnrapplisten.py` to `gnrlisten.py`, so the command is now `gnr app listen`. Removed the legacy entry point from `[project.scripts]` in pyproject.toml.

genro:
@cgabriel Thanks for the thorough architectural review. Here are our thoughts on each point:

**1/ & 2/ Locking and duplicate execution**

By design, the notification pattern is "wake up, there's work to do" — not "execute this task exactly once". The notification is a bell, not a command. Each listener autonomously decides what to do based on the current database state, not on the notification payload alone. This means duplicate notifications are harmless: the handler checks the DB and acts only if needed.

That said, running multiple listeners on the same instance should be documented as unsupported (single-listener-per-instance). A future enhancement could add a PG advisory lock to enforce this at runtime, but for now documentation is sufficient.

**3/ Horizontal scaling**

The current design is intentionally single-process for simplicity. Future evolution paths include:
- Topic-based routing (`channel@listener_id`) where each listener only processes messages addressed to it
- An orchestrator pattern where a master listener distributes work
- Worker-based partitioning by package or table

These are natural extensions that don't require architectural changes — just richer registration logic in `GnrAppListener`.

**4/ Loop protection**

This is the handler's responsibility. A handler that triggers a change on its own table would indeed cause re-notification. The coalesce parameter already helps (batching notifications), but explicit loop guards (e.g. checking a flag or timestamp in the record) should be documented as a best practice. The framework doesn't need to enforce this — it's a domain-level concern.

**5/ AdapterCapabilities**

Good suggestion. We'll add this in a follow-up. The current `listen_connection()` returning `None` for unsupported adapters already provides a runtime check, but a declarative capability flag would be cleaner for the CLI to check upfront before starting the event loop.

**6/ In-memory queue**

This is inherent to PostgreSQL LISTEN/NOTIFY — notifications are ephemeral by design. If the listener crashes, unprocessed notifications are lost. But since the pattern is "something changed, go check" rather than "execute this command", the listener simply re-evaluates the current state on restart. No work is truly lost — at worst, some changes are processed with a delay.

This is documented in the PR description: the feature is a lightweight alternative to polling, not a durable message queue. For guaranteed delivery, a proper queue (Celery, RabbitMQ) would be needed, but that's a different use case entirely.

genro:
@cgabriel @mbertoldi gentle ping — review is pending on this PR. All feedback from the previous round has been addressed. Ready when you are.

gnrpy/gnr/app/gnrlistener.py
cgabriel: Yes it needs a different logic, and must be implemented otherwise it won't get overridden from _gnrbasepostgresadapter.py and the wrong one will be in use when using the postgres3 adapter.

Basically, rather than using the connection attributes 'notifies' as in:

```
conn.poll()

while conn.notifies:
    notify = conn.notifies.pop(0)
    do_whatever_you_please(notify)
```

it's a generator:

```
for notify in conn.notifies():
    do_whatever_you_please(notify)
```

cgabriel:
> @cgabriel Thanks for the thorough architectural review. Here are our thoughts on each point:
> 
> **1/ & 2/ Locking and duplicate execution**
> 
> By design, the notification pattern is "wake up, there's work to do" — not "execute this task exactly once". The notification is a bell, not a command. Each listener autonomously decides what to do based on the current database state, not on the notification payload alone. This means duplicate notifications are harmless: the handler checks the DB and acts only if needed.
> 
> That said, running multiple listeners on the same instance should be documented as unsupported (single-listener-per-instance). A future enhancement could add a PG advisory lock to enforce this at runtime, but for now documentation is sufficient.
> 
> **3/ Horizontal scaling**
> 
> The current design is intentionally single-process for simplicity. Future evolution paths include:
> 
> * Topic-based routing (`channel@listener_id`) where each listener only processes messages addressed to it
> * An orchestrator pattern where a master listener distributes work
> * Worker-based partitioning by package or table
> 
> These are natural extensions that don't require architectural changes — just richer registration logic in `GnrAppListener`.
> 
> **4/ Loop protection**
> 
> This is the handler's responsibility. A handler that triggers a change on its own table would indeed cause re-notification. The coalesce parameter already helps (batching notifications), but explicit loop guards (e.g. checking a flag or timestamp in the record) should be documented as a best practice. The framework doesn't need to enforce this — it's a domain-level concern.
> 
> **5/ AdapterCapabilities**
> 
> Good suggestion. We'll add this in a follow-up. The current `listen_connection()` returning `None` for unsupported adapters already provides a runtime check, but a declarative capability flag would be cleaner for the CLI to check upfront before starting the event loop.
> 
> **6/ In-memory queue**
> 
> This is inherent to PostgreSQL LISTEN/NOTIFY — notifications are ephemeral by design. If the listener crashes, unprocessed notifications are lost. But since the pattern is "something changed, go check" rather than "execute this command", the listener simply re-evaluates the current state on restart. No work is truly lost — at worst, some changes are processed with a delay.
> 
> This is documented in the PR description: the feature is a lightweight alternative to polling, not a durable message queue. For guaranteed delivery, a proper queue (Celery, RabbitMQ) would be needed, but that's a different use case entirely.

Looking forward for the followups and the documentation then!

cgabriel (approved):

PR #678, opened 2026-03-10
fix(decorator): preserve __name__ in @customizable via functools.wraps

### LINKED ISSUE #614: @customizable decorator does not preserve __name__, breaking compatibility with @public_method and .remote()
## Summary

When `@public_method` and `@customizable` are stacked on the same method, calling that method via `.remote()` fails because GenroPy's RPC dispatcher cannot resolve the method name.

## Root Cause

In `gnrpy/gnr/core/gnrdecorator.py`, the `customizable` decorator wraps the original function in `newFunc` but **does not preserve `__name__`**:

```python
def customizable(func):
    def newFunc(page, *args, **kwargs):
        oncalling_result = customize(page, '%s_oncalling_' % func.__name__, *args, **kwargs)
        if oncalling_result is False:
            return
        result = func(page, *args, **kwargs)
        kwargs['_original_result'] = result
        customize(page, '%s_oncalled_' % func.__name__, *args, **kwargs)
        return result
    return newFunc  # newFunc.__name__ == 'newFunc', not func.__name__
```

`newFunc.__name__` is literally `'newFunc'`, not the original function name. The GenroPy RPC dispatcher looks up public methods by `__name__`, so when the client requests e.g. `convegnoPreferenze`, the server cannot match it to the wrapped method.

Compare with `@autocast` and `@deprecated` in the same file, which correctly preserve `__name__`:

```python
# autocast
wrapper.__name__ = func.__name__
if hasattr(func, 'is_rpc'):
    wrapper.is_rpc = func.is_rpc

# deprecated
newFunc.__name__ = func.__name__
newFunc.__doc__ = func.__doc__
newFunc.__dict__.update(func.__dict__)
```

`@customizable` does none of this.

## Reproduction

```python
@public_method
@customizable
def convegnoPreferenze(self, pane):
    tc = pane.tabContainer(...)
    ...
```

```python
# In the form builder:
tc.contentPane(title='Configurazione').remote(self.convegnoPreferenze)
```

Clicking the tab triggers the RPC call, which fails because `convegnoPreferenze.__name__ == 'newFunc'`.

## Expected Behavior

`@customizable` should preserve `__name__` (and ideally all function metadata) so it is transparent to the RPC system and compatible with `@public_method`.

## Proposed Fix

Use `functools.wraps` in the `customizable` decorator:

```python
from functools import wraps

def customizable(func):
    def customize(page, name, *args, **kwargs):
        ...

    @wraps(func)  # preserves __name__, __doc__, __dict__, __wrapped__, etc.
    def newFunc(page, *args, **kwargs):
        ...

    return newFunc
```

Or at minimum, manually copy the essential attributes:

```python
newFunc.__name__ = func.__name__
newFunc.__doc__ = func.__doc__
newFunc.__dict__.update(func.__dict__)  # also propagates is_rpc
```

Note: propagating `__dict__` would make the decorator order irrelevant — both `@public_method` + `@customizable` and `@customizable` + `@public_method` would work correctly.

dgpaci:
Fixes #614

## Root cause

The `@customizable` decorator in `gnrpy/gnr/core/gnrdecorator.py` wrapped the original function without preserving its metadata. As a result, `wrapper.__name__` was literally `'wrapper'` instead of the decorated function's name.

The GenroPy RPC dispatcher resolves public methods by `__name__`, so when `@public_method` and `@customizable` were stacked on the same method, the server could not match the incoming RPC call to the correct handler — causing failures when calling via `.remote()`.

## Fix

Added `@wraps(func)` (from `functools`) to the inner `wrapper` function inside `customizable`. This preserves `__name__`, `__doc__`, `__dict__` (including `is_rpc`) and `__wrapped__`, making the decorator fully transparent to the RPC system regardless of the stacking order of `@public_method` and `@customizable`.

The change is two lines:

```python
from functools import wraps   # added import

def customizable(func):
    ...
    @wraps(func)              # added decorator
    def wrapper(page, *args, **kwargs):
        ...
```

This approach is consistent with how `@deprecated` already handles metadata preservation in the same file.

## Testing

Tested locally on the scenario described in the issue (`@public_method` + `@customizable` stacked, called via `.remote()`): the RPC call now resolves correctly. Previously it failed with an unresolvable method name.

## Notes

- Zero risk of regression: `functools.wraps` is stdlib, the change is minimal, and the wrapper logic is untouched.
- Both decorator orderings (`@public_method` + `@customizable` and `@customizable` + `@public_method`) now work correctly because `__dict__` propagation includes `is_rpc`.

### DIFF
--- a/gnrpy/gnr/core/gnrdecorator.py
+++ b/gnrpy/gnr/core/gnrdecorator.py
@@ -21,6 +21,7 @@
 #Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 import warnings
+from functools import wraps
 from gnr.core.gnrdict import dictExtract
 
 def metadata(**kwargs):
@@ -179,6 +180,7 @@ def customizable(func):
             if result is False:
                 return result
 
+    @wraps(func)
     def wrapper(page,*args,**kwargs):
         oncalling_result = customize(page,'%s_oncalling_' %func.__name__,*args,**kwargs)
         if oncalling_result is False:


cgabriel (dismissed):
Looking good to me.

cgabriel (approved):
Looking good to me

fporcari (approved):
Minimal and correct fix. @wraps(func) is the standard way to preserve metadata in decorators, consistent with how @deprecated already works in the same file. LGTM. One note: the commit message includes a Co-Authored-By reference that should be removed before merging — please squash or amend to clean it up.

PR #679, opened 2026-03-10
fix(batch): show scheduler dialog when no table_script_parameters_pane defined (#659)

### LINKED ISSUE #659: [BUG] Schedulable batch not schedulable without script parameters pane
### Problem Description

To be schedulable, a batch needs to specify:

````
class Main(BaseResourceAction):
    batch_prefix = 'XXX'
    batch_title = caption
    batch_schedulable = True
```

But that's not enough! A table script parameters pane (even empty) needs to be defined:

````
 def table_script_parameters_pane(self, pane, **kwargs):
        pass
````

If not so, dialog does not appear and therefore batch is actually not schedulable.

But it's not obvious and frustrating in my opinion since I forget every time and need some time to realize what's the issue there...

### Reproduction Steps

Add batch_schedulable True to your batch
Nothing happens!
Add an empty table_script_parameters_pane
Magic happens!

_No response_

### Environment

genropy@latest

### Additional Material

_No response_

### PR DESCRIPTION
**Pull request by dgpaci**
## Problem

When a batch is marked as schedulable (via `batch_schedulable`) but does not define a `table_script_parameters_pane`, the Schedule button was never shown. This is because the parameters dialog was only built when `table_script_parameters_pane` existed — so the `@oncalled` override in the sys package had no dialog footer to attach the Schedule button to.

## Solution

Introduce `showParametersDialog` in `table_script_dialogs`:

```python
showParametersDialog = hasParameters or bool(schedulable)
```

This ensures the parameters dialog (including its Cancel/Confirm footer) is always built for schedulable batches, regardless of whether `table_script_parameters_pane` is defined. The `@oncalled` override in the sys package then attaches the Schedule button to the existing footer normally.

The call to `table_script_parameters_pane` is now guarded separately by `hasParameters`, so it is only invoked when the method actually exists.

## Side effect (known and accepted)

`BaseResourceBatch` sets `batch_schedulable = 'admin'` as the class default. As a result, **admin users will now see a confirm/cancel dialog for any batch that does not explicitly set `batch_schedulable = False` and has no `table_script_parameters_pane`**.

This is considered acceptable because:
- Asking for confirmation before running a batch is always preferable to running immediately
- Previously, the only way to trigger this confirm dialog was to define an empty `table_script_parameters_pane` — a workaround that is now unnecessary
- Non-admin users are completely unaffected: the permission check on `'admin'` fails and `showParametersDialog` stays `False`, so those batches still run immediately without any dialog

To suppress the dialog for a specific batch, set `batch_schedulable = False` explicitly.

## Files changed

- `resources/common/gnrcomponents/batch_handler/batch_handler.py` — introduce `showParametersDialog`, guard `table_script_parameters_pane` call with `hasParameters`
- `projects/gnrcore/packages/sys/resources/gnrcomponents/batch_handler/batch_handler.py` — remove redundant `if hasParameters` guard; `parametersDialog.footerNode` is now always available for schedulable batches

Fixes #659


cgabriel (approved):

fporcari (dismissed):
Clean fix. The separation of showParametersDialog from hasParameters is well thought out — the dialog is built for schedulable batches while the parameters_pane call remains guarded. Side effect is acceptable and well documented. LGTM.

cgabriel (changes requested):
Please solve conflicts with develop merge.

genro (dismissed):
Logic is sound. Minor note: sys/batch_handler dropped the return value from _scheduler_footer — verify it was unused upstream.

cgabriel (approved):

fporcari (approved):

genro (approved):

PR #685, opened 2026-03-11
correctly evaluate in the correct order the debug option. 

### LINKED ISSUE #351: [BUG] Debug mode always on
### Problem Description

When starting the wsgi server, the "debug mode: on" is always on, even with --nodebug, with debug="False::B" in siteconfig.

In gnrwebpage.init_options[338] che 
self.options.__dict__[option] = site_option or wsgi_options.get(option) or envopt.get(option)
 
where wsgi_options is a dict with default value where debug=True. So any debug=False option could overrule this.

### Reproduction Steps

See above

_No response_

### Environment

master branch

### Additional Material

_No response_

cgabriel:
We start from wsgi default options, then site configuration, then enviroment.

The previous failed when a "debug=False" was put in the site configuration, since the logic was based on the option value and not on the option itself, so the option value played a master role in the evaluation.

close #351

### DIFF
--- a/gnrpy/gnr/web/serverwsgi.py
+++ b/gnrpy/gnr/web/serverwsgi.py
@@ -333,9 +333,17 @@ class Server(object):
         options = self.options.__dict__
         envopt = dictExtract(os.environ,'GNR_WSGI_OPT_')
         for option in list(options.keys()):
+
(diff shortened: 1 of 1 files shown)


fporcari (commented):
Looks fine

mbertoldi (approved):

PR #687, opened 2026-03-11
The DateTimeTextBox becomes empty on blur

dedalus2000:
The populated DateTimeTextBox becomes empty as soon as I click elsewhere, even if I haven't changed anything. The first lines of the patch fix this issue.
Line 3850 fixes the issue where value was not being reassigned to datesplit when match was set.
From line 3907 I keep the time part at the previous value instead of resetting it. I'm not sure whether it's useful, but it seems consistent to me.

### DIFF
--- a/gnrjs/gnr_d11/js/genro_widgets.js
+++ b/gnrjs/gnr_d11/js/genro_widgets.js
@@ -3827,6 +3827,14 @@ dojo.declare("gnr.widgets.DateTextBox", gnr.widgets._BaseTextBox, {
 
     patch_parse:function(value,constraints){
         if(value && !this._focused){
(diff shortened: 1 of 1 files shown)


fporcari (dismissed):
LGTM. The main fix (early return when formatted value matches displayed value on blur) is correct and well guarded with instanceof Date + isNaN check. The datesplit reassignment fix is also sound. Note: d20 has no JS files, so no sync needed. Minor note: the time-preservation behavior (keeping previous hours/minutes/seconds instead of resetting to midnight) is a subtle change — works well for DateTimeTextBox use cases.

fporcari (approved):
looks fine

PR #688, opened 2026-03-12
Transitory refactoring  of daemon package and modules

cgabriel:
Transitory refactoring of daemon package and modules and its internal references to rearrange the daemon files in a sub-package ('gnr.web.daemon'), which will allow easier plugging via entrypoints of external packages.


genro (commented):
## Review Notes

### 1. `sys.modules` mutation in `__init__.py`

The entrypoint mechanism in `gnr/web/daemon/__init__.py` directly mutates `sys.modules`:

```python
sys.modules[__name__] = _mod
```

This pollutes Python's shared space at module level. If the entrypoint module fails to load or is malformed, the entire `gnr.web.daemon` namespace becomes corrupted with no recovery path. Also, any code that has already imported symbols from the original module before this replacement will hold stale references.

Consider an alternative approach — for example, a factory or proxy pattern that doesn't replace the module itself in `sys.modules`.

### 2. Test coverage

- `tests/web/gnrdaemon.py` only does `import gnr.web.daemon` — it doesn't verify the entrypoint mechanism works (or doesn't break when no entrypoints are registered).
- `gnrdaemonprocesses_test.py` was deleted without a replacement test for `daemon.processes`.

### 3. Minor

- The file `gnrdaemon.py` in tests doesn't follow the `*_test.py` naming convention used by the deleted tests.

Overall the refactoring direction makes sense — having a sub-package for easier plugging via entrypoints is a good idea. The concern is mainly around the `sys.modules` replacement pattern.

mbertoldi (approved):

cgabriel:
> ## Review Notes
> ### 1. `sys.modules` mutation in `__init__.py`
> The entrypoint mechanism in `gnr/web/daemon/__init__.py` directly mutates `sys.modules`:
> 
> ```python
> sys.modules[__name__] = _mod
> ```
> 
> This pollutes Python's shared space at module level. If the entrypoint module fails to load or is malformed, the entire `gnr.web.daemon` namespace becomes corrupted with no recovery path. Also, any code that has already imported symbols from the original module before this replacement will hold stale references.
> 
> Consider an alternative approach — for example, a factory or proxy pattern that doesn't replace the module itself in `sys.modules`.

I'm aware of this, but as per title it's transitory - it just allows to develop a daemon separately until merged, and if the new one is not installed, there are no problem at all.

> ### 2. Test coverage
> * `tests/web/gnrdaemon.py` only does `import gnr.web.daemon` — it doesn't verify the entrypoint mechanism works (or doesn't break when no entrypoints are registered).
> * `gnrdaemonprocesses_test.py` was deleted without a replacement test for `daemon.processes`.

> ### 3. Minor
> * The file `gnrdaemon.py` in tests doesn't follow the `*_test.py` naming convention used by the deleted tests.


Both test were empty, and the file naming follows the convetion with all the other tests.
 
> Overall the refactoring direction makes sense — having a sub-package for easier plugging via entrypoints is a good idea. The concern is mainly around the `sys.modules` replacement pattern.

Thank you Claude!

PR #698, opened 2026-03-14
fix: allow multiple ext_ handlers on the same column

genro:
## Summary
- The `ext_kwargs` loop in `DbModelSrc.column()` exited with `return result` after processing the first `ext_` handler, preventing subsequent handlers from running
- A column with both `ext_ltx` and `ext_emb` (e.g. `tbl.column('docstring', ext_ltx='*', ext_emb=True)`) would only get the first extension configured
- Fix: remove the premature `return result` so all ext_ handlers are executed

## Test plan
- [ ] Verify a column with two `ext_*` attributes gets both extensions configured
- [ ] Verify existing single-ext columns are unaffected


cgabriel (approved):
Nice catch, it also prevents localization configuration if an ext handler is found.

PR #700, opened 2026-03-16
refactor: extract auth verifiers from gnrwebpage into dedicated module

fporcari:
## Summary
- Extract Basic Auth and Bearer token verification logic from `gnrwebpage.py` into a new `gnr.web.verifiers` module
- Introduce `GnrVerifier` base class with `AuthorizationBearerVerifier` and `AuthorizationBaseTagsVerifier` subclasses
- Add verifier dispatch in `getPublicMethod` supporting both `handler.verifier` attribute and legacy `handler.tags` fallback
- Add protection against malformed Authorization headers (split limits, base64 decode try/except)
- Remove unused `b64decode` import and `basicAuthenticationTags` method from `gnrwebpage.py`

## Test plan
- [x] flake8 passes with zero errors on modified files
- [x] All 1368 tests pass (260 skipped)
- [x] Verified no regression: `externalcall.py` catches both `GnrUserNotAllowed` and `GnrBasicAuthenticationError`
- [ ] Manual test: Basic Auth protected endpoint still works
- [ ] Manual test: handler without auth still works


gnrpy/gnr/web/verifiers.py
cgabriel, line 1: I'd avoid redefining __call__, but rather use __call__ = verify, and use this name to be subclassed. This will make the subclasses implementation more readable, and if the logic in __call__ changes for whatever reason, we can do that in the base class only.

Also, I'd decor verify in GnrVerifier with abstractmethod, in order to force subclassing.

gnrpy/gnr/web/verifiers.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def get_bearer(self):
+        """Extract the bearer token from the Authorization header.
+        Returns the token string or None if missing/invalid."""
+        auth_header = self.page.request.headers.get('Authorization', '')
+        if not auth_header.startswith('Bearer '):
+            return None
+        return auth_header[7:]
+
+    def get_token_to_verify(self):
```
cgabriel, line 58: This must be decorated with abstractmethod too. Otherwise, en empty token will work if the subclass forgets to implement it. By decorating it, the usage is prohibited, which guarantees to avoid the above problem.

cgabriel (changes requested):
See pending comments.

gnrpy/gnr/web/verifiers.py
fporcari: Done in e800fbd.

GnrVerifier now inherits from ABC. `verify()` is the abstract method that subclasses override, and `__call__` delegates to `self.verify(**kwargs)`.

Note: a plain `__call__ = verify` alias at class level wouldn't provide polymorphic dispatch (Python resolves `__call__` via MRO, so it would always find the base class function rather than the subclass override). The thin `__call__` wrapper ensures correct delegation.

gnrpy/gnr/web/verifiers.py
fporcari: Done — `get_token_to_verify()` is now decorated with `@abstractmethod`. Any subclass that forgets to implement it will fail at instantiation time.

cgabriel (approved):

PR #716, opened 2026-03-20
fix: _th_mixinResource safeMode and multidomain error URLs

fporcari:
## Summary
- Add `safeMode` parameter to `_th_mixinResource` so that tolerant callers (`th_tableViewer`, `th_tableEditor`, `th_thFormHandler`) can proceed with framework defaults when no table-specific `th_*.py` mixin is found. This fixes a regression from `2222f99db` where lookup tables without explicit th_ files failed with `GnrMixinNotFound`.
- Fix error detail URLs in multidomain apps: `current_domain` is now saved only when multidomain is active, and the `detail_url` formula uses `/_main_/` prefix since errors are centralized in the primary database.

## Test plan
- [ ] Verify lookup table picker works without explicit `th_*.py` (e.g. `dialogTableHandler` with picker on a lookup table)
- [ ] Verify error detail URL in multidomain app includes `/_main_/` prefix
- [ ] Verify error detail URL in single-domain app remains unchanged
- [ ] All automated tests pass (1368 passed, 260 skipped)

### DIFF
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -1016,7 +1016,8 @@ class GnrWsgiSite(object):
                 kwargs['request_uri'] = request.url
                 kwargs['request_host'] = request.host_url
 
-        kwargs.setdefault('current_domain', self.currentDomain)
+        if self.multidomain:
+            kwargs.setdefault('current_domain', self.currentDomain)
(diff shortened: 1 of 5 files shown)


cgabriel (approved):

genro (approved):
LGTM — safeMode fix for lookup tables and multidomain error URL logic are clean.

PR #719, opened 2026-03-23
feat: exact row-height calculation via AFM font metrics

### LINKED ISSUE #718: calcRowHeight: exact row height calculation for wrapping text
## Problem

The `calcRowHeight` method in `print_gridres.py` uses a character-count approximation to estimate how many lines a text will occupy:

```python
math.ceil(len(txt) / float(v))
```

This is inherently inaccurate because:
- Characters have different widths (proportional fonts like Helvetica)
- The available width depends on margins, cell padding, and column widths
- There is no way to account for word-wrap boundaries precisely

The result is that text is either **cut off** (under-estimated height) or there is **excess blank space** (over-estimated height). Both have been observed in practice, sometimes within the same document.

## Proposed solution

Use **AFM (Adobe Font Metrics)** character width data for standard PDF fonts (Helvetica, Times, Courier) to simulate exact word-wrap and count lines precisely — the same data that `reportlab` uses internally, but without the dependency.

The implementation would consist of:

1. **`gnr/utils/fonts.py`** — AFM width tables for standard fonts + `string_width(text, font_name, font_size)` utility
2. **`gnr/core/gnrstring.py`** — `stringWidth(text, fontName, fontSize)` as a drop-in for `reportlab.pdfbase.pdfmetrics.stringWidth`
3. **`TableScriptToHtml`** — new method `calcRowsNumber(text, width_mm, font_name, font_size)` that performs exact word-wrap simulation
4. **`print_gridres.py`** — tiered `calcRowHeight`:
   - subclass overrides `calcRowHeight` → used as-is (existing behaviour)
   - subclass overrides `getRowWrapFields()` → `calcRowHeight` uses `calcRowsNumber` for exact measurement
   - neither → returns `grid_row_height` (existing default)

## Why not Pillow / fonttools?

Both are available as transitive dependencies of `weasyprint`, but both require a font **file path** at runtime. Font file locations vary across OS and server configurations and cannot be assumed. The AFM approach is self-contained.

## Why not reportlab?

`reportlab` was removed as a genropy dependency in commit 1a3914506500ba8b05ec615579dc10b3f6da4e32.

## Notes

- AFM data for the standard 14 PDF fonts is in the public domain (published by Adobe)
- The approach has been validated against a real-world ticket report: exact line counts match PDF rendering for Helvetica 10pt
- `Courier` is a monospaced font so all character widths are 600 units — trivially exact

dgpaci:
## Summary

- Adds `gnr/utils/fonts.py` with lazy-loaded AFM width data from `resources/common/fonts/afm_widths.json`
- Adds `gnr.core.gnrstring.stringWidth()` — drop-in replacement for `reportlab.pdfbase.pdfmetrics.stringWidth`, no external dependencies
- Centralizes `calcRowsNumber()` in `GnrHtmlBuilder` (eliminates copy-paste between `TableScriptToHtml` and `RecordToHtmlPage`)
- Adds `font_family` and `font_size` class attributes to `GnrHtmlBuilder`
- Handles explicit `\n` newlines in row-count calculation
- Replaces magic number `2.8346` with named constant `MM_TO_PT = 72 / 25.4`
- Keeps `self.calcRowsNumber()` on report classes as a backward-compatible delegation wrapper

## How it works

Set `font_family = 'Helvetica'` on a report class to:
1. Apply `body { font-family: Helvetica }` to the document automatically (single source of truth)
2. Enable exact line-count calculation via AFM word-wrap simulation in `calcRowsNumber()`

Override `getRowWrapField()` to return the text of the wrapping field — `calcRowHeight` will use it automatically. For complex reports, call `self.calcRowsNumber(text)` directly inside a custom `calcRowHeight`.

## Changes vs previous version (review feedback addressed)

1. `calcRowsNumber` centralized in `GnrHtmlBuilder` — no duplication between `TableScriptToHtml` and `RecordToHtmlPage`
2. AFM data moved from inline Python dict to `resources/common/fonts/afm_widths.json`, loaded lazily
3. Explicit `\n` newlines now handled correctly (previously treated as whitespace)
5. `MM_TO_PT = 72 / 25.4` constant replaces hardcoded `2.8346`
6. `font_size` defaults to `self.font_size` (builder attribute) instead of hardcoded `10`
7. Tests added for `stringWidth` and `calcRowsNumber`
8. Branch rewritten without AI attribution in commits
9. `self.calcRowsNumber()` delegation wrapper added on report classes for backward compatibility

## Test plan

- [ ] Set `font_family = 'Helvetica'` on a report with wrapping text, verify no text clipping and no excess blank space
- [ ] Verify `stringWidth('Hello', 'Helvetica', 10)` returns expected value
- [ ] Verify `font_family = None` (default) leaves existing report behaviour unchanged
- [ ] Test with `Times-Roman` and `Courier` font variants
- [ ] Test a report subclass that calls `self.calcRowsNumber()` directly

Closes #718


cgabriel (dismissed):

genro (commented):
## Review — PR #719: exact row-height calculation via AFM font metrics

Good feature — the existing `len(txt)/v` heuristic is clearly inadequate for proportional fonts. The AFM approach is sound and performant. However, there are several issues to address before merging.

---

### 1. Code duplication

`calcRowsNumber()`, `getRowWrapField()`, `font_family`, `text_width_mm` are identical copy-paste between `TableScriptToHtml` (gnrbaseclasses.py) and `RecordToHtmlPage` (gnrtablescript.py). `calcRowHeight` with AFM logic is triplicated (gnrtablescript.py, print_gridres.py, gnrbaseclasses.py).

Both class hierarchies are separate but both use `GnrHtmlBuilder` as `self.builder`. The measurement logic should live as a method on `GnrHtmlBuilder` — both hierarchies can then call `self.builder.calcRowsNumber(...)` without any duplication.

### 2. Hardcoded AFM tables in Python code

`fonts.py` embeds 119 lines of character width dictionaries. This is a fragile dependency — hard to update, inspect, or extend to new fonts.

**Proposal**: move the data to a JSON file in common resources (e.g. `resources/common/fonts/afm_widths.json`), loaded at runtime. This separates data from code and makes it maintainable.

### 3. Explicit newlines not handled

`text.split()` treats `\n` as whitespace — a text with explicit line breaks is measured as a continuous line, underestimating the row count.

**Proposal**: the centralized method should use a generator that yields lines for both `\n` breaks and width overflow, then count:

```python
def wrap_lines(text, avail_pt, font_name, font_size):
    for line in text.split('\n'):
        current = 0.0
        space_pt = stringWidth(' ', font_name, font_size)
        has_words = False
        for word in line.split():
            w = stringWidth(word, font_name, font_size)
            if not has_words:
                current = w
                has_words = True
            elif current + space_pt + w <= avail_pt:
                current += space_pt + w
            else:
                yield current
                current = w
        yield current if has_words else 0.0
```

### 4. HTML content in cells

The calculation assumes plain text. Cells may contain HTML with mixed font-sizes, bold/italic spans, or even sub-grids — `text.split()` produces wrong measurements for all of these.

This doesn't need to be solved in this PR, but centralizing the logic in `GnrHtmlBuilder` makes it possible to add a strategy switch later: AFM for plain text, weasyprint render for HTML content (detected via regex `<[a-zA-Z/]`).

### 5. Magic constant `2.8346`

The mm-to-pt conversion factor is used inline without a name, and it's a rounded approximation.

**Proposal**: extract as a named constant with the exact value: `MM_TO_PT = 72 / 25.4`. The centralized method in `GnrHtmlBuilder` could also accept a `um` (unit of measure) parameter and handle conversion internally, supporting `mm`, `pt`, `px`, `cm`, `in`.

### 6. `font_size` hardcoded to 10

The default `font_size=10` in `calcRowsNumber()` is not connected to any class attribute. If the report uses a different font-size (set in CSS), the calculation silently uses the wrong size.

**Proposal**: the centralized method should take `font_size` from the builder/report context, with an explicit override parameter.

### 7. No tests

`calcRowsNumber` and `stringWidth` are pure functions — easy to test. The PR should include at least basic tests: empty string, single line, multi-line wrap, known character widths.

### 8. CLAUDE.md violation

The commit message contains `Co-Authored-By: Claude Sonnet 4.6` and the PR body contains "Generated with Claude Code". The repo rules (CLAUDE.md) explicitly forbid any LLM/AI/Claude references in commits and PRs. This needs a new branch with a clean commit.

---

### Suggested prompt for the author

Here is a prompt you can give to your Claude to implement the suggested changes:

```
Fix PR #719 based on review feedback. Here is what needs to change:

1. CENTRALIZE in GnrHtmlBuilder (gnrpy/gnr/core/gnrhtml.py):
   - Add a `calcRowsNumber(text, width, font_name=None, font_size=None, um='mm')` method
   - It should handle unit conversion (mm, pt, px, cm, in) via a dict of conversion factors to pt
   - font_size should default to a class attribute (add `font_size = 10` to GnrHtmlBuilder)
   - font_name should default to a class attribute (add `font_family = None`)
   - Use a generator internally that yields lines for both '\n' splits AND width overflow
   - Load AFM data from a JSON file in resources/common/fonts/afm_widths.json (create it)

2. REMOVE DUPLICATION:
   - Remove calcRowsNumber, getRowWrapField, font_family, text_width_mm from both
     TableScriptToHtml (gnrbaseclasses.py) and RecordToHtmlPage (gnrtablescript.py)
   - Both should call self.builder.calcRowsNumber(...)
   - getRowWrapField() stays as an override point on the report classes
   - calcRowHeight in both classes and print_gridres.py should call self.builder.calcRowsNumber()

3. REMOVE fonts.py — replace with resources/common/fonts/afm_widths.json
   - Same data, JSON format, loaded once at GnrHtmlBuilder init or on first use

4. EXTRACT CONSTANT: MM_TO_PT = 72 / 25.4 (put in gnrhtml.py)

5. ADD TESTS for stringWidth and calcRowsNumber:
   - Empty string -> 0
   - Short text fitting in one line -> 1
   - Text wrapping to N lines -> N
   - Text with explicit \n
   - Known character widths match expected values

6. NEW BRANCH: create a fresh branch without any Claude/AI attribution
   in commits or PR body. The repo CLAUDE.md forbids this strictly.
```

fporcari (approved):
Good work — the AFM approach is a clear improvement over the `len(txt)/v` heuristic, and Giovanni's review points have all been addressed cleanly.

JS centralisation in `GnrHtmlBuilder`, lazy JSON loading, explicit `\n` handling, `MM_TO_PT` constant, backward-compat wrappers, tests — all solid.

One minor observation for a follow-up: `TableScriptToHtml` does not override `calcRowHeight` — it inherits the base `BagToHtml` version which just returns `self.grid_row_height`. So a direct subclass of `TableScriptToHtml` with `font_family` set would **not** benefit from the AFM calculation unless it overrides `calcRowHeight` manually. Only `RecordToHtmlPage` and `print_gridres.Main` have the AFM-aware override.

This is probably fine for now (direct `TableScriptToHtml` subclasses typically override `calcRowHeight` already), but worth keeping in mind — a one-liner override on `TableScriptToHtml` itself would close the gap.

genro (approved):
Thorough review of the full diff. Clean design: calcRowsNumber centralised in GnrHtmlBuilder, AFM data lazy-loaded from JSON, stringWidth as drop-in replacement for reportlab, backward-compat wrappers on both report classes, zero impact on existing reports (font_family=None by default). Relative path for afm_widths.json resolves correctly. Tests cover edge cases well. Import cleanup is a welcome bonus.

cgabriel (approved):

PR #721, opened 2026-03-23
add werkzeug's middleware ProxyFix 

cgabriel:
Added werkzeug's middleware ProxyFix (initially, only under K8S execution) to correctly use x-forwarder-for, x-forwarded-proto etc header, in order to correctly compute the external url.

Since 'external_host' must be removed from instanceconfig configuration file (to be able to use an instance on different hostname, the name depends on how the application is deployed), the correct external computation is a must when behing a reverse proxy / ssl terminator. Otherwise links will be generated using 'http://' (schema used between reverse proxy and application backend). This is usually not noticed by the user, because we have a direct from http to https, but if sensitive information are in the QUERY_INDEX (like tokens, keys etc) they are transferred via http.

### DIFF
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -18,6 +18,7 @@ from werkzeug.utils import redirect
 from werkzeug.exceptions import (HTTPException, InternalServerError,
                                   NotFound, Forbidden, PreconditionFailed,
                                   BadRequest, Unauthorized)
+from werkzeug.middleware.proxy_fix import ProxyFix
 
 from gnr.core.gnrbag import Bag
 from gnr.core import gnrstring
@@ -1570,6 +1571,14 @@ class GnrWsgiSite(object):
                 wsgiapp = SentryWsgiMiddleware(wsgiapp)
             except Exception as e:
                 logger.error(f"Sentry support has been disabled due to configuration errors: {e}")
+
+        # when the application is executed being a reverse proxy / ssl terminator,
+        # werkzeug needs this middleware to compute the correct external host
+        # which is used by externalUrl in the Site object when the value is
+        # computed using the request data.
+        # Limited to Kubernetes environment for initial staging testing
+        if os.environ.get("KUBERNETES_SERVICE_HOST", None):
+            wsgiapp = ProxyFix(wsgiapp, x_for=1, x_proto=1, x_host=1)
         return wsgiapp
 
     def build_gnrapp(self, options=None):


genro (commented):
## Review

The change is correct and safe — `ProxyFix` is the standard werkzeug solution for this problem. Code is clean, well-placed (wraps after Sentry), and the parameters `x_for=1, x_proto=1, x_host=1` are appropriate for a single-proxy setup.

### Points to discuss

1. **K8s-only limitation**: The `KUBERNETES_SERVICE_HOST` check works for initial staging, but the problem is identical for any reverse-proxy deployment (nginx, traefik, HAProxy in Docker/compose). A siteconfig flag (e.g. `wsgi?proxy_fix=true`) would be more flexible and future-proof. Could be a follow-up.

2. **`x_prefix` parameter**: `ProxyFix` also supports `x_prefix=1` for `X-Forwarded-Prefix`. Worth considering if the app can be mounted under a subpath behind the proxy. Not critical.

3. **No opt-out in K8s**: If `KUBERNETES_SERVICE_HOST` is set but there's no actual proxy in front (e.g. local minikube without ingress), ProxyFix could produce incorrect values. A config-based opt-out would help.

### Verdict

Mergeable as-is with no risk. The main suggestion (config-based activation instead of K8s-only) can be addressed in a follow-up PR.

cgabriel:
> ## Review
> The change is correct and safe — `ProxyFix` is the standard werkzeug solution for this problem. Code is clean, well-placed (wraps after Sentry), and the parameters `x_for=1, x_proto=1, x_host=1` are appropriate for a single-proxy setup.
> 
> ### Points to discuss
> 1. **K8s-only limitation**: The `KUBERNETES_SERVICE_HOST` check works for initial staging, but the problem is identical for any reverse-proxy deployment (nginx, traefik, HAProxy in Docker/compose). A siteconfig flag (e.g. `wsgi?proxy_fix=true`) would be more flexible and future-proof. Could be a follow-up.

The limitation is currently introduce to allow building images of the develop branch and deploy in the k8s testing enviroment, without affecting other deployments. Since the middleware doesn't introduce the proxy as a requirements, the goals is to just use it without any kind of activation flags. It has been tested with nginx ssl-termination without problems, but i'd like to first test it in the kubernetes test cluster without impacting others for more thorough tests.

> 2. **`x_prefix` parameter**: `ProxyFix` also supports `x_prefix=1` for `X-Forwarded-Prefix`. Worth considering if the app can be mounted under a subpath behind the proxy. Not critical.

The omission of x_prefix is volountary, since the "multidomain" stuff recently introduced is based on the prefix to determine the tenants, conflicting with this approach.

> 3. **No opt-out in K8s**: If `KUBERNETES_SERVICE_HOST` is set but there's no actual proxy in front (e.g. local minikube without ingress), ProxyFix could produce incorrect values. A config-based opt-out would help.

As stated above, the goal is to have the middleware always present, in order to be functional in every enviroment.

> ### Verdict
> Mergeable as-is with no risk. The main suggestion (config-based activation instead of K8s-only) can be addressed in a follow-up PR.

Thank you.

Issue #722, opened 2026-03-23
[BUG] AttachManager creates DB record before file upload — orphan records on storage failure


genro:
## Problem Description

When uploading a file via AttachManager, the attachment record is inserted and committed to the database **before** the file is actually saved to storage. If the storage write fails (permission denied, S3 auth error, network timeout, disk full, etc.), the record remains in the database pointing to a non-existent file.

Additionally, the error is not properly communicated to the client — the user sees a generic toast or no feedback at all, while the DB contains an orphan record.

## Root Cause Analysis

The upload flow involves three layers, each with its own issue:

### Layer 1: `onUploadingAttachment` commits before file save

In `resources/common/gnrcomponents/attachmanager/attachmanager.py` (line 559-577):

```python
def onUploadingAttachment(self, kwargs):
    # ... setup ...
    attachment_tblobj.insert(record)       # line 574
    kwargs['attachment_id'] = record['id']
    self.db.commit()                        # line 576 — COMMITTED
    self.clientPublish('inserted_attachment', ...)
```

This is a **pre-upload** callback (`onUploadingMethod`). It runs before the file is saved. Its primary purpose is to compute `uploadPath` and `filename`, but it also inserts and commits the record.

### Layer 2: `rpc_upload_file` has no error handling around file save

In `gnrpy/gnr/web/gnrwebpage_proxy/rpc.py` (line 123-179):

```python
def rpc_upload_file(self, *args, **kwargs):
    onUploading = self.page.getPublicMethod('rpc', onUploadingMethod)
    onUploading(kwargs)                     # line 127 — record committed here
    # ...
    file_path, file_url = site.uploadFile(  # line 143 — file save here
        file_handle=file_handle, ...)       # if THIS fails, record is orphaned
    # ...
    if onUploadedMethod:
        handler(...)                        # line 168 — post-upload callback
```

The call sequence is:
1. `onUploading()` → inserts + commits record
2. `site.uploadFile()` → saves file to storage (may fail)
3. `onUploaded()` → post-processing

There is no try/except around step 2, and no rollback of the record created in step 1 if it fails.

### Layer 3: `S3File.__exit__` doesn't check for existing exceptions

In `gnrpy/gnr/web/gnrwsgisite_proxy/gnrcloudservices.py` (line 63-69):

```python
def __exit__(self, exc, value, tb):
    if self.write_mode:
        self.file.seek(0)
        self.s3.upload_fileobj(self.file, self.bucket, self.key)  # line 66
[...]

Issue #720, opened 2026-03-23
raiseIfDeveloper bypasses isDeveloper() check — non-developer users get 500 instead of error_id


genro:
## Problem

When a non-developer user triggers an exception in an RPC call, they see a generic 500 error instead of the error_id with a link to the sys_error detail page.

## Root cause

`raiseIfDeveloper()` in `gnrwsgisite.py:1169` uses `force_debug` which bypasses the `isDeveloper()` check:

```python
def raiseIfDeveloper(self, exception=None):
    page = self.currentPage
    if self.debug and ((page and page.isDeveloper()) or self.force_debug):
        if exception is not None:
            raise exception
        raise
```

`force_debug` is always `True` when `self.debug` is `True` (set in `setDebugAttribute` line 548-549). This means the `isDeveloper()` check is never evaluated — the exception is always re-raised for ALL users.

As a result, the error handling flow in `_rpcDispatcher` (gnrwebpage.py:629-648) that saves the error to `sys.error` and returns the error_id to the user is never reached.

### Related: #351 / PR #685

PR #685 fixed the debug option evaluation order, but did not address `force_debug` bypassing `isDeveloper()`.

### `force_debug` is redundant

`force_debug` is only written in `setDebugAttribute` and was only read in `raiseIfDeveloper`. It is always equal to `self.debug` — there is no code path where `debug=True` and `force_debug=False`. The `isDeveloper()` check was effectively dead code.

## Proposed fix

Remove `force_debug` from the condition:

```python
def raiseIfDeveloper(self, exception=None):
    page = self.currentPage
    if self.debug and page and page.isDeveloper():
        if exception is not None:
            raise exception
        raise
```

### Behavior after fix

| Scenario | Before (broken) | After (fix) |
|----------|-----------------|-------------|
| Developer + debug on | Werkzeug debugger | Werkzeug debugger (unchanged) |
| Non-developer + debug on | **500 generic** | **error_id shown** |
| Anyone + debug off (gunicorn) | error_id shown | error_id shown (unchanged) |

## Branch

`fix/error-handler-non-developer` on genropy repo — tested locally, fix confirmed working.

PR #724, opened 2026-03-23
fix: error handling for non-developer users and upload error reporting

genro:
## Summary

- **raiseIfDeveloper** simplified: removed `force_debug`, exceptions are re-raised only when `debug=True` AND `page.isDeveloper()`. Non-developer users now receive a proper `error_id` instead of a raw 500 traceback.
- **Upload error reporting**: `uploadMultipart_oneFile` (d11 + d20) now parses the RPC response envelope and routes errors through `handleRpcError`. Previously, the raw XHR `onload` callback bypassed the standard RPC result handler entirely, so upload failures were silently ignored on the client side.
- **Traceback in toast**: server captures `traceback.format_exc()` and sends it as `error_traceback` in the envelope. The toast copy button supports **Shift+click** to copy the full traceback instead of the short error message.
- **cleanup() moved** from `serve_page` finally to `dispatcher` finally, ensuring it always runs even on errors outside serve_page.
- **Visual debug indicator**: `gnr_debug` CSS class added to body when `site.debug` is active; `.pbl_root_bottom` turns red in debug mode.

## Review requests

**@fporcari** — please review the **client-side JS changes**, in particular:
- `genro_rpc.js` (d11 + d20): the new `sender.onload` handler that parses the envelope and calls `handleRpcError` on upload errors
- `genro_dev.js`: extraction of `error_traceback` from envelope for the toast `copyDetail`
- `genro_toast.js`: Shift+click on copy button for detailed text

**@cgabriel** — please review the **server-side Python changes**:
- `gnrwsgisite.py`: removal of `force_debug`, simplified `raiseIfDeveloper`, `cleanup()` relocation
- `gnrwebpage.py`: traceback capture in `_rpcDispatcher`, `gnr_debug` CSS class

## Related issues

- Refs #720 (raiseIfDeveloper bypass)
- Refs #722 (AttachManager orphan records on upload failure — this PR fixes the error reporting side; the record creation order is a separate fix)

## Test plan

- [ ] Trigger a server exception as non-developer user → should see error toast with error_id (not 500)
- [ ] Trigger a server exception as developer in debug mode → should re-raise with traceback
- [ ] Upload a file to an invalid/read-only path → should see error toast (not silent failure)
- [ ] Shift+click copy button on error toast → should copy full traceback
- [ ] Verify debug mode shows red bottom bar
- [ ] Verify normal (non-debug) mode is unaffected


gnrpy/gnr/web/gnrwebpage.py
```diff
@@ -628,6 +628,8 @@ def _rpcDispatcher(self, *args, **kwargs):
             result = str(e)
         except Exception as e:
             self.site.raiseIfDeveloper()
+            import traceback as _tb
```
cgabriel: Please avoid runtime imports

cgabriel (changes requested):

gnrpy/gnr/web/gnrwebpage.py
genro: Fixed — moved to module-level import in f5604947ce.

fporcari (approved):
Good improvements overall. The upload error reporting via envelope parsing and the cleanup() relocation are well designed.

Single required change: move `import traceback` to module level (see inline comment). Once fixed, this is ready to merge.

gnrpy/gnr/web/gnrwebpage.py
fporcari: The `import traceback as _tb` must be at module level — runtime imports are not allowed per project convention (as cgabriel already noted). `traceback` is a stdlib module, just add it to the top-level imports.

gnrpy/gnr/web/gnrwebpage.py
```diff
@@ -637,6 +639,7 @@ def _rpcDispatcher(self, *args, **kwargs):
                 rpc_kwargs=Bag(kwargs)
             )
```
fporcari, line 640: Nice use of the existing `envelope_` convention in `result_bag()` — no plumbing needed, the traceback flows through to the client automatically.

gnrpy/gnr/web/gnrwsgisite.py
```diff
@@ -1162,11 +1158,12 @@ def dispatcher(self, environ, start_response):
                 % (environ.get('SCRIPT_NAME'), environ.get('PATH_INFO')))
             return exc(environ, start_response)
         finally:
+            self.cleanup()
```
fporcari, line 1165: Good move. In the previous code, `cleanup()` ran in `_dispatcher`'s finally block — BEFORE `setResultInResponse()`, which still accesses `page.sql_time`, `page.xml_size`, etc. Moving it here ensures the page is alive during response building.

cgabriel (approved):

gnrpy/gnr/web/gnrwebpage.py
genro: Already fixed — the import is at module level in the current version. @fporcari please resolve this thread.

gnrpy/gnr/web/gnrwebpage.py
genro: Thanks! @fporcari please resolve this thread.

gnrpy/gnr/web/gnrwsgisite.py
genro: Thanks for the analysis. @fporcari please resolve this thread so the PR can be merged.

PR #726, opened 2026-03-24
fix(#722): move attachment record creation from pre-upload to post-upload

genro:
## Summary

- **onUploadingAttachment** (pre-upload): no longer inserts or commits the record. Only computes `uploadPath`, `filename` and prepares record data (`_atc_description`, `_atc_filepath`) in kwargs.
- **onUploadedAttachment** (post-upload): now handles both new insert and update cases. Database writes happen only after the file is successfully saved to storage.
- **onUploadingAttachmentUpd** (pre-upload update): same treatment — only path computation, the actual `recordToUpdate` + commit moved to `onUploadedAttachment` (detected via `pkey` in kwargs).

Previously, if an upload failed (permission denied, S3 auth error, disk full, network timeout), an orphan attachment record was left in the database pointing to a non-existent file. The `clientPublish('inserted_attachment')` also fired prematurely, causing the UI to update as if the upload succeeded.

## Cross-repo impact (discovered via Sourcerer)

Using Sourcerer `code_search_code` across all indexed repositories, we found that `onUploadingAttachment` is used in 8 repos:

| Repo | Impact |
|------|--------|
| **writers, colto, genropynet, frigel_cont, icond** | Use inherited `self.onUploadingAttachment` — **automatically fixed** |
| **erpylight, icond** (th_pt_distinta) | Commented out (inactive) — no impact |
| **teamset** (`ordine_acquisto_component.py`) | Has a **custom** `onUploadingFattura` that replicates the same antipattern, plus updates the parent `ordine_acquisto` record in the pre-upload. **Requires separate fix.** Opened [softwellsrl/teamset#1](https://github.com/softwellsrl/teamset/issues/1) with detailed analysis and fix proposal. |

This cross-repo discovery would not have been possible by searching only in the genropy repository.

## Test plan

- [ ] Upload an attachment to a valid path → record created, file saved, UI updated
- [ ] Upload an attachment to a read-only/invalid path → no record in DB, error toast shown, UI not updated
- [ ] Update an existing attachment (re-upload) → record updated only after file saved
- [ ] Verify `inserted_attachment` clientPublish fires only after successful save

Refs #722


cgabriel (commented):
Can't approve - now the record doesn't get saved at all.

I've tried to upload an attachemnt to an s3-backed storage, the file gets uploaded, but the record is not created. No errors reported to the user neither in the logs.

FTR, the attachment table is gm_base.paziente_atc from softwellsrl/genromed.

genro:
@cgabriel Thanks for testing. I've traced the full flow and can't find an obvious logic error — the kwargs chain from onUploading → rpc_upload_file → onUploaded should carry all the data needed for the insert.

Could you share the server log output from the failed upload attempt? I need to see if there's a silent exception being caught by _rpcDispatcher. Specifically:

1. Any traceback or error in the server console during the upload
2. The wsgi debug log (if debug=True) — the `RESOURCE` line for the upload POST

If you're running with debug=False, could you temporarily enable it and try again? The error might be swallowed by the generic except in _rpcDispatcher without visible feedback (which is exactly what PR #724 improves).

Also, was this test done from the attachment **form** (editing an existing record) or from the **view** (dropping a new file on the grid/drop zone)? This matters because the form passes `rpc_pkey=#FORM.pkey` which would route to the update path instead of insert.

cgabriel:
> @cgabriel Thanks for testing. I've traced the full flow and can't find an obvious logic error — the kwargs chain from onUploading → rpc_upload_file → onUploaded should carry all the data needed for the insert.
> 
> Could you share the server log output from the failed upload attempt? I need to see if there's a silent exception being caught by _rpcDispatcher.

As reported before, nothing in log output.

> Specifically:
> 
> 1. Any traceback or error in the server console during the upload

Nothing, as stated.

> 2. The wsgi debug log (if debug=True) — the `RESOURCE` line for the upload POST
 
> If you're running with debug=False, could you temporarily enable it and try again? The error might be swallowed by the generic except in _rpcDispatcher without visible feedback (which is exactly what PR #724 improves).

It's running with --debug, and GNR_LOGLEVEL=debug
 
> Also, was this test done from the attachment **form** (editing an existing record) or from the **view** (dropping a new file on the grid/drop zone)? This matters because the form passes `rpc_pkey=#FORM.pkey` which would route to the update path instead of insert.

In the View.

cgabriel (commented):
Nothing has changed in the behaviour since my previous review.

fporcari (changes requested):
## Root cause found for cgabriel's bug

The three `dropUploader` instances in the **View** (grid drop zone) do **not** pass `onUploadedMethod`:

- **Line 292** (`th_top` with `linkedTo`)
- **Line 357** (`th_top` gallery mode)  
- **Line 461** (`th_top` simple mode)

They all set `onUploadingMethod=self.onUploadingAttachment` but **never** set `onUploadedMethod`.

### Why the old code worked

All DB writes (insert, commit, `clientPublish`) were inside `onUploadingAttachment` (pre-upload), which is always called regardless.

### Why this PR breaks it

DB writes are now in `onUploadedAttachment` (post-upload), but `rpc_upload_file` only calls `onUploadedMethod` when it's explicitly provided. Without it, `rpc_upload_file` falls through to the `elif uploaderId:` branch (rpc.py:170-178) which looks for `onUploaded_<uploaderId>` — a method that doesn't exist for attachments. So the record is silently never created.

### Fix

Add `onUploadedMethod=self.onUploadedAttachment` to all three View dropUploaders (lines 292, 357, 461).

### Additional notes

- The `attachment_id` parameter in `onUploadedAttachment` signature is now dead code (always `None`) — should be removed
- `onUploadingAttachmentUpd` sets `_atc_filepath` but not `_atc_description` — asymmetric with the new-insert path

genro:
@fporcari All three points addressed in 56a30b4:

1. **onUploadedMethod added** to all three View dropUploaders (lines 297, 363, 467) — this was the root cause of cgabriel's bug
2. **attachment_id removed** from `onUploadedAttachment` signature (dead parameter)
3. **_atc_description added** to `onUploadingAttachmentUpd` for symmetry with the new-insert path

@cgabriel Could you re-test the upload from the View (grid drop zone) with this fix? The record should now be created correctly after the file is saved.

cgabriel (approved):
Tried with the latest changes, the upload to s3 + record creation (and view reload) works fine for me in the previous testing enviroment used for my latest review/comments.

fporcari (approved):

PR #560, opened 2026-02-24
feat(grouplets): add grouplet system for modular form fragments

fporcari:
## Summary
- Add Grouplet system for dynamically loading modular UI form fragments from table resources, with menu-driven selection and dialog-based editing (memory, document, record stores)
- Refactor form store layer: extract `_BaseForm` base class from BoxForm/FrameForm with shared `_load_rpc`/`_save_rpc` logic, add `load_record`/`save_record` store handlers and `saveRecord` RPC
- Add `GroupletHandler` Python component for resource discovery and remote form building, with `Grouplet`/`GroupletForm` JS widgets and animated dialog builders (`memoryDataEditor`, `documentDataEditor`, `recordDataEditor`)

## Test plan
- [x] Manual testing of grouplet dialogs (memory, document, record modes)
- [x] Verify grouplet menu discovery and nested grouplet categories
- [x] Test existing BoxForm/FrameForm forms still work after _BaseForm refactor
- [x] Automated tests pass (1040 passed)

PR #734, opened 2026-03-25
fix(#686): restore custom rpcmethod in document store load

### LINKED ISSUE #686: [BUG] IsADirectoryError when opening Localization Editor (storageNode receives directory path)
## Description

Opening the Localization Editor raises an `IsADirectoryError` crash.

## Traceback

```
File "gnrpy/gnr/web/gnrwsgisite.py", line 1316, in _dispatcher
    result = page()
File "gnrpy/gnr/web/gnrwebpage.py", line 585, in __call__
    result = self._call_handler(*args, **kwargs)
File "gnrpy/gnr/web/gnrwebpage.py", line 614, in _rpcDispatcher
    result = self.rpc(method=method, _auth=auth, **kwargs)
File "gnrpy/gnr/web/gnrwebpage_proxy/rpc.py", line 38, in __call__
    result = handler(**kwargs)
File "gnrpy/gnr/web/gnrwebpage.py", line 2785, in getSiteDocument
    with snode.open('rb') as f:
File "gnrpy/gnr/lib/services/storage.py", line 314, in open
    return self.service.open(self.path, **kwargs)
File "gnrpy/gnr/lib/services/storage.py", line 704, in open
    return open(self.internal_path(*args), **kwargs)
IsADirectoryError: [Errno 21] Is a directory: '/path/to/genropy'
```

## Root Cause

The Localization Editor calls `this.form.goToRecord(attr.folderPath)` where `folderPath` is the `destFolder` of the selected localization block. For the `core` block, `destFolder = getGenroRoot()` — the genropy root directory (e.g. `/path/to/genropy`).

This triggers `getSiteDocument(path='/path/to/genropy')` on the server. The method calls `self.site.storageNode(path)` which, through `LegacyStorageHandler._adapt_path`, resolves to the `_raw_` storage service (`base_path='/'`). The resulting internal path is the genropy root directory itself.

`getSiteDocument` then checks `snode.exists` (True — directory exists), `snode.ext` (empty — no extension), and falls into the `elif snode.exists:` branch, which tries to `open()` the directory, raising `IsADirectoryError`.

## Relevant code

**`gnrpy/gnr/web/gnrwebpage.py` — `getSiteDocument`:**
```python
snode = self.site.storageNode(path)
if not snode.exists:
    content = defaultContent
else:
    if snode.ext == 'xml':
        with snode.open('rb') as f:
            content = Bag(f)
    elif snode.exists:          # always True here — dead else below
        with snode.open('rb') as f:
            content = f.read()  # crashes if path is a directory
    else:
        content = ''            # unreachable
```

Note: there is also a secondary bug — the inner `else: content = ''` is unreachable code, since `elif snode.exists` is always `True` inside the outer `else` block.

**`resources/common/services/storage/raw.py`:**
```python
class Service(BaseLocalService):
    def __init__(self, parent=None, base_path=None, **kwargs):
        super(Service, self).__init__(parent=parent, base_path='/', **kwargs)
```

The `_raw_` storage service has `base_path='/'`, so any path without a `:` prefix resolves as an absolute filesystem path.

## Likely regression

The bug is likely related to the storage handler refactoring introduced in PR #331 (`refactor(storage): extract storage handling to dedicated proxy handler`), which moved `storageNode` logic to the new `LegacyStorageHandler`.

## Expected behavior

`getSiteDocument` should not attempt to `open()` a path that resolves to a directory. Either:
- Add a `snode.isdir` guard so directory paths are treated like non-existing documents (return `defaultContent`), or
- Fix the Localization Editor to avoid passing filesystem directory paths as pkeys to the document store.

## Steps to reproduce

1. Open a Genropy application with multiple localization blocks
2. Navigate to **System → Localization Editor**
3. The crash occurs immediately on page load (or when selecting the `core` block)

genro:
## Summary

Fixes #686 — `IsADirectoryError` when opening the Localization Editor.

## Root cause

The grouplets refactoring (PR #560, commit `d98a4cf56f`) rewrote `load_document` in `genro_frm.js` and hardcoded `'getSiteDocument'` as the RPC method:

```javascript
// NEW (broken) — ignores custom handler
kw.rpcmethod = kw.rpcmethod || 'getSiteDocument';
```

The **original** code respected the custom `rpcmethod` set via `form.store.handler('load', rpcmethod=...)`:

```javascript
// OLD (working) — custom handler first, default as fallback
this.handlers.load.rpcmethod = loader.rpcmethod || 'getSiteDocument';
```

The Localization Editor uses `store='document'` with a custom handler (`loadLocalizationFile`) that knows how to handle directory paths (it does `os.path.join(path, 'localization.xml')`). After the regression, `getSiteDocument` was called instead, which tried to `open()` a directory path — crash.

## Fix

Restore the original priority chain in `load_document`:

```javascript
kw.rpcmethod = kw.rpcmethod || this.handlers.load.rpcmethod || 'getSiteDocument';
```

Also reverts the symptomatic `snode.isdir` guard on `getSiteDocument` (commit `aed62de316`) which masked the real cause without fixing it.

## Files changed

- `gnrjs/gnr_d11/js/genro_frm.js` — restore handler rpcmethod priority
- `gnrjs/gnr_d20/js/genro_frm.js` — same fix for dojo_20
- `gnrpy/gnr/web/gnrwebpage.py` — revert isdir workaround

## Test plan

- [ ] Open a Genropy app with localization (e.g. any app with `adm` package)
- [ ] Navigate to **System → Localization Editor**
- [ ] Select the **core** block — should load without error
- [ ] Select package blocks — should also load correctly
- [ ] Verify that other `store='document'` forms still work (e.g. doc_handler)

@dgpaci please test on the environment where the bug was originally reported.

### DIFF
--- a/gnrjs/gnr_d11/js/genro_frm.js
+++ b/gnrjs/gnr_d11/js/genro_frm.js
@@ -2759,7 +2759,7 @@ dojo.declare("gnr.formstores.Base", null, {
         const that = this;
         const form = this.form;
         kw = kw || {};
-        kw.rpcmethod = kw.rpcmethod || 'getSiteDocument';
+        kw.rpcmethod = kw.rpcmethod || this.handlers.load.rpcmethod || 'getSiteDocument';
         kw.path = form.getCurrentPkey();
 
         const onResult = function(result){
--- a/gnrjs/gnr_d20/js/genro_frm.js
+++ b/gnrjs/gnr_d20/js/genro_frm.js
@@ -2759,7 +2759,7 @@ dojo.declare("gnr.formstores.Base", null, {
         const that = this;
         const form = this.form;
         kw = kw || {};
-        kw.rpcmethod = kw.rpcmethod || 'getSiteDocument';
+        kw.rpcmethod = kw.rpcmethod || this.handlers.load.rpcmethod || 'getSiteDocument';
         kw.path = form.getCurrentPkey();
 
         const onResult = function(result){


dgpaci (commented):
Provided solution doesn't seem to work since this error raises when opening Localization Editor

````
During handling of the above exception, another exception occurred:

  File "/Users/dgpaci/sviluppo/gnrenv/lib/python3.12/site-packages/gnr/web/gnrwsgisite.py", line 1011, in writeException
    return self.db.table('sys.error').writeException(description=str(exception),
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SqlTable' object has no attribute 'writeException'
```

dgpaci (changes requested):
Still not working, no errors but localization rows are not loaded, everything's empty

Image

PR #738, opened 2026-03-26
fix: S3 md5hash returns None for multipart ETags

fporcari:
## Summary
- S3 multipart upload ETags contain a hyphen (e.g. `d41d8cd98f00b204e9800998ecf8427e-2`) and are not valid MD5 hashes
- `md5hash()` now checks for the hyphen and returns `None` instead of the invalid value
- Explicit `return None` added for clarity when the object does not exist

## Test plan
- [ ] Verify `md5hash()` returns a valid hash for single-part uploads
- [ ] Verify `md5hash()` returns `None` for multipart uploads (ETag contains `-`)
- [ ] Verify `md5hash()` returns `None` for non-existent objects


projects/gnrcore/packages/sys/resources/services/storage/aws_s3.py
```diff
@@ -176,7 +176,10 @@ def versions(self,*args):
     def md5hash(self,*args):
         bucket = self._head_object(*args)
         if bucket:
-            return bucket['ETag'][1:-1]
+            etag = bucket['ETag'][1:-1]
+            if '-' not in etag:
```
cgabriel: I'd use 

```
if len(etag) == 32:
    return etag
```

which is O(1) rather than using in which is O(n).

cgabriel (dismissed):
I've checked the etag documentation, for multipart it's not an md5 so the check is fine. I've left a simple optimization fix comment, which would be nice to have.

projects/gnrcore/packages/sys/resources/services/storage/aws_s3.py
fporcari: Good catch, applied.

cgabriel (approved):

projects/gnrcore/packages/sys/resources/services/storage/aws_s3.py
cgabriel: Nice, TY

genro (approved):
Clean fix, correct approach with len check. LGTM.

Issue #743, opened 2026-03-26
[BUG] Operator precedence bug in dataRpc routes application errors to _onError instead of _onResult


fporcari:
## Problem Description

In `gnrjs/gnr_d11/js/gnrdomsource.js:407`, a JavaScript operator precedence bug causes **application-level errors** returned by Python RPC methods via `dict(error='...')` to be routed to `_onError` instead of `_onResult`. This makes any `result.error` handling inside `_onResult` callbacks **dead code**.

### The bug

```javascript
// gnrdomsource.js:407 — current (broken)
error = error || result ? result.error : null;
```

Due to operator precedence, this parses as:
```javascript
error = (error || result) ? result.error : null;
```

Since `result` is almost always truthy, this **always** evaluates `result.error` — extracting application errors and routing them to `_onError`, making the `if (error) / else` branch mutually exclusive with `_onResult`.

### Expected behavior

Two distinct error channels:
- **`_onError(error, kwargs)`** — transport/infrastructure errors (server down, timeout, HTTP 500, unhandled Python exceptions)
- **`_onResult(result, kwargs, old)`** — receives the full result, including `result.error` for business logic errors

### Actual behavior

Both error types are conflated into `_onError`. When `_onError` is not defined (the vast majority of dataRpc calls), the error is **silently swallowed** — no user feedback at all.

### Commit history

**Commit `860556c`** — March 9, 2016:
```javascript
// Added to extract result.error — worked correctly but crashed on null result
error = error || result.error;
```

**Commit `1b8ca01`** — March 10, 2016 ("fix small regression"):
```javascript
// Intended to guard against null result, but missing parentheses changed semantics
error = error || result ? result.error : null;
```

The fix intended to write `error = error || (result ? result.error : null)` but omitted parentheses, fundamentally changing the behavior.

### Why it went unnoticed for 10 years

1. **`_onError` is rarely defined** — only 5 places in the entire GenroPy codebase define it. Most dataRpc calls have neither `_onError` nor `result.error` handling.
2. **The generic form save covers most cases** — `_gnrbasewebpage.py:395` has its own `_onError`. Most business errors come as unhandled Python exceptions (transport-level), not `dict(error=...)`.
3. **Where `dict(error=...)` is used with `_onError`, it works "by accident"** — e.g., `login.py` defines `_onError` which catches the message. The `result.error` branch in `_onResult` is dead code but the error is shown anyway.
[...]

PR #744, opened 2026-03-26
fix: convert login_newUser from frameForm to boxForm for auto-sizing

fporcari:
## Summary
- Replace `frameForm` with fixed dimensions (26em x 20em) with `boxForm` + `loginboxPars()` pattern in `login_newUser` dialog
- Move header/footer outside the form to the outer `box` div, matching the pattern used by all other login dialogs (`loginDialog`, `login_lostPassword`, etc.)
- Switch error display from `form.publish("message",...)` to `genro.dlg.floatingMessage()` since boxForm doesn't use FramePane regions
- Sanitize exception message in `login_createNewUser` to avoid exposing internal error details

## Context
After the formbuilder-to-formlet migration, the `login_newUser` dialog was the only login dialog still using `frameForm` with hardcoded dimensions. Formlet uses top labels by default, which caused the content to overflow the fixed 26em x 20em frame. All other login dialogs already use the auto-sizing `loginboxPars()` pattern.

## Related
- #743 — documents a pre-existing operator precedence bug in `gnrdomsource.js:407` that routes `dict(error=...)` to `_onError` instead of `_onResult` (not introduced by this PR)

## Test plan
- [x] Open new user registration dialog — verify it auto-sizes correctly with top labels
- [x] Submit with missing fields — verify validation error message appears
- [x] Submit with duplicate username — verify `validate_nodup` error
- [x] Submit with valid data but email failure — verify error message shown
- [x] Submit with valid data and working email — verify success message and page reload
- [x] Test with `closable=True` — verify close button works
- [x] Test "Login" link in footer — verify navigation back to login dialog

### DIFF
--- a/projects/gnrcore/packages/adm/resources/login.py
+++ b/projects/gnrcore/packages/adm/resources/login.py
@@ -464,39 +464,39 @@ class LoginComponent(BaseComponent):
                             subscribe_openNewUser='this.widget.show(); genro.formById("newUser_form").newrecord($1);',
                             subscribe_closeNewUser='this.widget.hide();')
 
-        kw = self.loginboxPars()
-        kw['width'] = '26em'
-        kw['height'] = '20em'
-        kw.update(kwargs)
-        form = dlg.frameForm(frameCode='newUser',datapath='new_user',store='memory',**kw)
+        box = dlg.div(**self.loginboxPars())
         if closable:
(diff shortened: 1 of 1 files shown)


genro (approved):
Clean fix, consistent with the existing login dialog pattern. LGTM.

One minor suggestion: consider adding a `logger.error()` call before the sanitized error return in `login_createNewUser` — currently the original exception is silently lost.

PR #751, opened 2026-03-30
fix: add missing H↔HZ type conversions to migration system

### LINKED ISSUE #750: fix: missing H→HZ type conversion in migration system
## Description

The `TYPE_CONVERSIONS` dictionary in `gnr/sql/adapters/_gnrbaseadapter.py` is missing entries for `H` (time without time zone) ↔ `HZ` (time with time zone) conversions.

When attempting a migration that changes a column from `H` to `HZ` (or vice versa) on a non-empty table, the migration system raises:

```
gnr.sql.gnrsql_exceptions.GnrSqlException: !!Error code GNRSQL-001 : Incompatible data type change in a non-empty column. Column <table>.<column> H HZ.
```

## Root Cause

`DH` ↔ `DHZ` (timestamp) conversions are already handled:

```python
('DH', 'DHZ'): None,
('DHZ', 'DH'): None,
```

But the symmetric entries for `H` ↔ `HZ` (time) are missing.

## Fix

Add the missing entries to `TYPE_CONVERSIONS`:

```python
('H', 'HZ'): None,
('HZ', 'H'): None,
```

PostgreSQL supports `ALTER COLUMN TYPE ... USING column::time with time zone` as a simple cast, so `None` (direct ALTER) is correct — same as `DH↔DHZ`.

## Steps to Reproduce

1. Define a column with `dtype='H'`
2. Change it to `dtype='HZ'` in the model
3. Run `gnr db migrate` on a non-empty table
4. Migration fails with `GNRSQL-001`

dgpaci:
## Summary

- Add `('H', 'HZ'): None` and `('HZ', 'H'): None` to `TYPE_CONVERSIONS` in `_gnrbaseadapter.py`
- Mirrors the existing `('DH', 'DHZ')` / `('DHZ', 'DH')` entries which were already handled

## Problem

Migrating a non-empty column from `H` (time without time zone) to `HZ` (time with time zone) raises:

```
GnrSqlException: GNRSQL-001 : Incompatible data type change in a non-empty column. Column <table>.<column> H HZ.
```

## Fix

PostgreSQL handles `time without time zone` → `time with time zone` as a direct `ALTER COLUMN TYPE` cast — same as `DH→DHZ`. Setting the conversion value to `None` triggers a simple alter with no `USING` clause.

## Test plan

- [ ] Define a model column with `dtype='H'`
- [ ] Change to `dtype='HZ'`, run `gnr db migrate` on a non-empty table → should succeed
- [ ] Reverse: change back to `'H'`, run migrate → should succeed

Closes #750

### DIFF
--- a/gnrpy/gnr/sql/adapters/_gnrbaseadapter.py
+++ b/gnrpy/gnr/sql/adapters/_gnrbaseadapter.py
@@ -149,6 +149,8 @@ class SqlDbAdapter(object):
         ('DH', 'DHZ'): None,
         ('DHZ', 'D'): None,
         ('DHZ', 'DH'): None,
+        ('H', 'HZ'): None,
+        ('HZ', 'H'): None,
     }
 
 


dgpaci:
Closing in favour of #752 which correctly targets the develop branch.

PR #755, opened 2026-03-30
fix: include required_columns in print resource grid queries

fporcari:
## Summary
- Print resources using `fieldcell` for pyColumns with `required_columns` were missing those columns in the SQL query, causing errors or empty values in printed output
- `gridColumnsFromStruct()` now extracts the `required_columns` attribute from struct cells
- `grid_sqlcolumns` property now collects and includes `required_columns` in the SQL column set, matching the behavior of JS grids (`columnsFromStruct()` in `genro_grid.js`)

## Test plan
- [ ] Use a print resource with a `fieldcell` referencing a pyColumn that has `required_columns`
- [ ] Verify the printed output includes correct computed values
- [ ] Verify JS grids continue to work normally


genro (approved):
Clean and correct fix. The Python-side `grid_sqlcolumns` now mirrors what the JS grid already does in `columnsFromStruct()` (genro_grid.js:86-89): split `required_columns` by comma and add each to the SQL column set.

The change is minimal and well-scoped — one file, one attribute propagation in `gridColumnsFromStruct`, and the collection loop in `grid_sqlcolumns`.

Issue #756, opened 2026-03-31
feat: evaluate mypantry for optional dependency management


genro:
## Summary

Evaluate adopting [mypantry](https://pypi.org/project/mypantry/) (genro-pantry) as a unified runtime registry for optional dependencies in GenroPy.

## Context

During the review of PR #748 (encrypted columns), the introduction of mypantry as a dependency was flagged as out of scope. The reviewer raised valid concerns:

- New external project to maintain for something the standard library can handle
- With pantry, all imports become runtime — pre-shipping tests won't detect problems
- Static code analyzers are limited in their activity
- Expanded surface for SCA/APTs
- Python lazy imports are coming in mainstream Python

At the same time, GenroPy has several optional dependencies today (psycopg2, mysqlclient, cryptography, and potentially more in the future) that are each handled ad-hoc with try/except or importlib checks. mypantry provides a consistent API for this:

- `pantry.has("package")` — clean boolean check
- `pantry["package"]` — import or raise with clear message
- `@pantry("package")` — decorator to guard functions
- `pantry.report()` — diagnostic summary

## Decision needed

Should GenroPy adopt mypantry for managing optional dependencies, or continue with ad-hoc checks (and eventually leverage Python's native lazy imports when available)?

This is decoupled from PR #748, which will use `cryptography` as a mandatory dependency regardless of the outcome here.

## References

- PR #748 review discussion
- genro-pantry issues: genropy/genro-pantry#2, genropy/genro-pantry#3

PR #748, opened 2026-03-27
feat: encrypted columns (R/Q/X modes)

genro:
## Summary

Adds transparent column-level encryption with three modes:

- **R** (Reversible): Fernet AES-128-CBC — for secrets (API tokens, service passwords)
- **Q** (Querable): AES-SIV deterministic — for searchable personal data (tax IDs, IBANs)
- **X** (one-shot): SHA-256 hash — for tokens you issue (irreversible)

```python
tbl.column('bank_account', dtype='T', encrypted='R')
tbl.column('registration_id', dtype='T', encrypted='Q')
tbl.column('access_token', dtype='T', encrypted='X')
```

Key from siteconfig (`<encryption secret_key="..."/>`) or env var `GNR_ENCRYPTION_KEY`.
`cryptography` is a mandatory dependency (already in the transitive deptree via paramiko / pywebpush / fido2 / oic).

## Implementation

- **Write**: single point in `prepareRecordData()` — covers insert, update, insertMany, raw_insert, raw_update
- **Read**: `compiled.encryptedColumns` dict populated by compiler, decrypted via `ColumnEncryptor.decrypt_row()` in fetch, _dofetch, loadRecord, out_dict
- **Search**: Q columns auto-encrypted in whereTranslator; `encrypted` attribute propagated in querybag from JS
- **UI visibility**: X excluded from all field trees, R excluded from QTREE (search filters), Q visible everywhere

## Test plan

- [x] 37 tests (22 unit + 15 integration with SQLite) — all passing locally
- [x] CI green on Python 3.11, 3.12, 3.13, 3.14
- [x] flake8 clean on all modified files
- [x] Branch rebased on develop (linear history)
- [x] Encrypted values in raw DB are prefixed (`$R$`, `$Q$`, `$X$`)
- [x] Transparent decryption on fetch, selection, record bag/dict
- [x] Q mode searchable with encrypted value
- [x] X mode `verify()` works, `decrypt()` returns None
- [x] Null values pass through unchanged

## Review history

Initial review (2026-03-31) requested changes on dependency management (mypantry vs mandatory `cryptography`) and env var naming. Addressed in 1373c76 (2026-05-06): mypantry removed entirely, `cryptography` moved to main `dependencies`, `GNR_ENCRYPTION_KEY` in place. All 8 inline review threads resolved. Ready for re-review.

Refs #747

Issue #11, opened 2024-10-14
itemsMaxWidth


dgpaci:
Choose if it is a valid solution (and in such case, extend it) or deprecate it

PR #757, opened 2026-03-31
feat: adopt mypantry for optional dependency management

genro:
## Summary

- Add `mypantry` to genropy dependencies
- Include proposal document (`docs/development/pantry_adoption.md`) with full analysis

## Context

Extracted from PR #748 (encrypted columns) following review feedback that dependency
management and encrypted columns should be separate concerns.

The proposal document covers:
- **20 conditional imports** found in genropy codebase (via Sourcerer cross-repo search)
- **7 inconsistent patterns** currently in use (HAS_X, module=False, print, cascade, pass, fallback, re-raise)
- **10 dead try/except** for packages already mandatory in pyproject.toml
- **Side-by-side comparison** of current approach vs pantry API
- **Open questions** from PR #748 review with responses grounded in current codebase state
- **PEP 690/810 status**: PEP 690 rejected (2022), PEP 810 targets Python 3.15 (Oct 2026) but solves deferred loading, not availability checking

## Related

- Issue #756
- PR #748 review discussion

## Decision

To be discussed at technical committee meeting (2026-04-01).


genro:
The issues raised against the `genro-pantry` repository have been addressed:

- genropy/genro-pantry#2 (multiple `top_level` packages) — resolved in v0.5.1
- genropy/genro-pantry#3 (installed context detection) — resolved in v0.5.0
- genropy/genro-pantry#4 (pipreqs compatibility) — documented

Adopting pantry in genropy would provide a consistent, uniform approach to managing runtime optional dependencies — replacing the 7 different patterns currently scattered across the codebase.

Like any tool, it does not cover 100% of cases, but it contributes to a coherent and maintainable style across the project. Should there be remaining concerns, we would appreciate them being grounded in specific, documented issues.

fporcari:
Good analysis — the document is thorough and the idea behind pantry is elegant: a single, clean API to replace 7 inconsistent patterns is the right direction. A few notes:

**1. Dead try/except cleanup**

The 10 try/except blocks for packages already mandatory in pyproject.toml (#11–#20) are clearly unnecessary. This cleanup is valuable and uncontroversial — it could go ahead immediately in a dedicated PR, regardless of the pantry decision.

**2. mypantry maturity**

The project is one month old and genropy/genro-pantry#2 (module name resolution for packages with multiple top_level modules) is still open. This is a real case in genropy — e.g. psycopg2cffi exposes its module as `psycopg2`. It would be better to resolve at least this issue before merging, to avoid surprises during phase 2 migration.

**3. Minor doc correction**

The document states mypantry has "zero external dependencies", but it actually depends on `packaging`. Not a problem in itself (packaging is already in the dependency tree), but worth correcting for accuracy.

genro:
Closing this PR.

After discussion at the 2026-04-01 technical committee meeting, the development team decided not to adopt mypantry as a dependency management tool for genropy. Reasons aligned with the concerns raised in the review of PR #748:

- Adds an external project to maintain for a problem partially addressed by the standard library, with PEP 810 (Python 3.15, Oct 2026) on the horizon for deferred imports.
- Shifts imports to runtime, reducing the effectiveness of pre-shipping checks and static analyzers.
- Expands the surface for supply-chain risks for limited concrete benefit.
- For the genropy docker base image (single flavour) the optional installation profile providing cryptography is mandatory anyway, so the "optional" dependency would be optional only on paper.

For the immediate need (encrypted columns), the agreed path is to make cryptography a direct mandatory dependency. That change is now part of PR #748.

If a future need re-opens the question of optional dependency management, this can be revisited as a fresh discussion.

Issue #759, opened 2026-04-01
fix: ZipFile not exception-safe in gnrwsgisite.zipFiles — potential file descriptor leak


genro:
## Problem

`gnrwsgisite.py:zipFiles()` (line 1951) creates a `zipfile.ZipFile` without a `with` statement. If any exception occurs during the file iteration loop (lines 1952-1963) — e.g., file not found, I/O error, permission denied — `zip_archive.close()` at line 1964 is never called, leaking the file descriptor.

In a long-running server with repeated failures, this can lead to **"too many open files"** (EMFILE) errors.

## Affected code

**`gnrpy/gnr/web/gnrwsgisite.py:1941-1964`** — `zipFiles()` method:

```python
zip_archive = zipfile.ZipFile(zipresult, mode='w', ...)
for fpath in file_list:
    # ... operations that can raise ...
    zip_archive.write(local_path, newname)
zip_archive.close()  # never reached on exception
```

## Other occurrences (low risk — in-memory StringIO)

These use the same `close()` pattern but operate on `io.StringIO`, so no OS file descriptor is consumed:

- `gnrpy/gnr/web/gnrwebpage_proxy/utils.py:192`
- `gnrpy/gnr/xtnd/gnrhtmltables.py:80`
- `gnrpy/gnr/core/gnrstring.py:914, 927`

## Proposed fix

Convert to `with` statement:

```python
with zipfile.ZipFile(zipresult, mode='w', compression=zipfile.ZIP_DEFLATED, allowZip64=True) as zip_archive:
    for fpath in file_list:
        ...
        zip_archive.write(local_path, newname)
```

Optionally convert the StringIO-based occurrences too for consistency.

## Reference

`gnrpy/gnr/db/cli/gnrmigrate.py:98` already uses the correct `with` pattern.

PR #760, opened 2026-04-01
fix: close leaked file descriptor in S3TemporaryFilename

genro:
## Summary

- `S3TemporaryFilename.__exit__()` never called `os.close(self.fd)` on the file descriptor returned by `tempfile.mkstemp()`. When processing thousands of S3 files (e.g. expense reports with attachments via `zipFiles()`), the leaked fds accumulate and cause EMFILE ("too many open files") errors.
- Added `os.close(self.fd)` inside a `try/finally` block so the fd is always released, even if `upload_file()` fails.

## Test plan

- [x] 7 new regression tests in `gnrpy/tests/web/test_s3_temporary_filename.py`
  - fd closed after normal exit
  - fd closed after exception in with-block
  - fd closed when upload fails
  - temp file removed (keep=False) / kept (keep=True)
  - no fd leak over 200 iterations (simulates zipFiles loop)
  - fd closed on S3 download failure
- [x] Full test suite passes (1516 passed, pre-existing failures unchanged)
- [x] flake8 clean

Refs #759

### DIFF
--- a/gnrpy/tests/web/test_s3_temporary_filename.py
+++ b/gnrpy/tests/web/test_s3_temporary_filename.py
@@ -1,6 +1,7 @@
 import importlib.util
 import os
 
+import botocore.exceptions
 import pytest
 
 # S3TemporaryFilename lives under projects/gnrcore/ which is not on the
@@ -40,7 +41,6 @@ class FakeS3Client:
 
     def download_file(self, bucket, key, dest):
         if self.fail_download:
-            import botocore.exceptions
             raise botocore.exceptions.ClientError(
                 {'Error': {'Code': '404', 'Message': 'Not Found'}},
                 'GetObject',
--- a/gnrpy/tests/web/test_s3_temporary_filename.py
+++ b/gnrpy/tests/web/test_s3_temporary_filename.py
@@ -1,0 +1,160 @@
+import importlib.util
+import os
+
+import pytest
+
+# S3TemporaryFilename lives under projects/gnrcore/ which is not on the
+# default test sys.path.  We load it directly from the file system so the
+# test can run without requiring a full GenroPy site setup.
+_s3_module_path = os.path.join(
+    os.path.dirname(__file__), os.pardir, os.pardir, os.pardir,
+    'projects', 'gnrcore', 'packages', 'sys', 'resources',
+    'services', 'storage', 'aws_s3.py',
+)
+_s3_module_path = os.path.normpath(_s3_module_path)
+
+
+def _get_s3_temporary_filename_class():
+    """Import S3TemporaryFilename from the aws_s3 module."""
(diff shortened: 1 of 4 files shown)


gnrpy/tests/web/test_s3_temporary_filename.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def test_fd_closed_when_upload_fails(self):
+        """fd must be closed even if upload_file raises during __exit__."""
+        client = FakeS3Client(content=b'original', fail_upload=True)
+        ctx = _make_ctx(client, mode='w')
+
+        with pytest.raises(OSError, match='upload failed'):
+            with ctx as local_path:
+                fd = ctx.fd
+                with open(local_path, 'wb') as f:
```
cgabriel, line 109: This must simulate the OSError, doing a raise (like the previous test).

On my host, the test fails:

```python
tests/web/test_s3_temporary_filename.py ..F....                 [100%]

============================== FAILURES ===============================
___ TestS3TemporaryFilenameFdLeak.test_fd_closed_when_upload_fails ____

self = <test_s3_temporary_filename.TestS3TemporaryFilenameFdLeak object at 0x7fa13d536d70>

    def test_fd_closed_when_upload_fails(self):
        """fd must be closed even if upload_file raises during __exit__."""
        client = FakeS3Client(content=b'original', fail_upload=True)
        ctx = _make_ctx(client, mode='w')
    
>       with pytest.raises(OSError, match='upload failed'):
E       Failed: DID NOT RAISE <class 'OSError'>

tests/web/test_s3_temporary_filename.py:108: Failed
======================= short test summary info =======================
FAILED tests/web/test_s3_temporary_filename.py::TestS3TemporaryFilenameFdLeak::test_fd_closed_when_upload_fails - Failed: DID NOT RAISE <class 'OSError'>
===================== 1 failed, 6 passed in 0.47s =====================
(env) ✘-1 ~/genropy/genropy/gnrpy [hotfix/s3-fd-leak|✚ 1…2] 
```

cgabriel (changes requested):
the change is fine, but the tests need fixes.

gnrpy/tests/web/test_s3_temporary_filename.py
genro: Fixed in 8aba0a7 and 7b6e8a5:

1. The test now forces mtime change with os.utime() so the upload path is always exercised, regardless of filesystem mtime resolution (the test was passing on APFS with nanosecond resolution but failing on ext4/HFS+ with 1-second resolution).

2. Moved import botocore.exceptions to module level.

cgabriel (approved):
LGTM

fporcari (approved):
Approved

PR #762, opened 2026-04-01
feat(email): mail proxy integration for async email delivery

fporcari:
## Summary
- Integrate Genropy email package with external async-mail-service (genro-mail-proxy) for asynchronous SMTP delivery with rate limiting, deferred delivery, and retry logic
- Add `mailproxy` service with tenant registration, account sync, and webhook-based message lifecycle (queue → proxy → SMTP → delivery report)
- Add `proxy_sync` and `proxy_get_attachments` endpoints for proxy callbacks with `_MAILPROXY_` tag-based authentication
- Add proxy dashboard UI (accounts grid, message queue, run-now/cleanup controls) shared via `MailProxyDashboard` component
- Extend `email.message` with `proxy_ts`, `proxy_priority`, `batch_code` fields; extend `email.account` with proxy rate-limit columns
- Graceful fallback: when proxy is not configured, existing direct-SMTP batch actions continue to work unchanged
- Improve login password reset flow with proper validation (`validate_notnull`, `_onCalling` checks)

## Test plan
- [ ] Verify direct-send flow still works when mailproxy service is absent
- [ ] Verify `send_mail` batch action is a no-op (not crash) when proxy exists but is not registered
- [ ] Register a tenant via the dashboard and verify account sync
- [ ] Send a message and verify proxy_sync webhook processes delivery reports
- [ ] Verify attachment download via proxy_get_attachments endpoint
- [ ] Verify proxy dashboard loads correctly in both mp_controls and sending_dashboard pages
- [ ] Verify queue layout still works in sending_dashboard when proxy is disabled


genro (commented):
## Review

Well-structured PR with good separation of concerns: service (`mailproxy.py`), webhook endpoint (`mp_endpoint.py`), dashboard UI (`mailproxy_dashboard.py`), and incremental changes to existing models and batch actions. The graceful fallback is well designed.

### Positive points
- Clean architecture: proxy is optional, direct SMTP flow is not broken
- `_immediate_message_parameters()` is a good pattern for login flow fallback
- `deferAfterCommit` in `message_to_send.trigger_onInserted` — correct, notifies proxy only after commit
- Dashboard reuses `bagGrid` — consistent with Genropy patterns
- README is detailed and useful

### Issues

**1. `_immediate_message_parameters` duplicated** (login.py and prefhandler.py)
Same identical method in two files. Should be extracted to a shared mixin or to the mail service.

**2. `externalcall.py` — `json` import added at module level**
The import is in the base module shared by all external calls. The change in `rootPage` to return JSON on auth errors is correct in concept, but touches a critical framework point. If `convert_result = False` is set by another page for different reasons, that page will also receive JSON on auth errors instead of XML.

**3. `mp_endpoint.py` — no rate limiting on `proxy_get_attachments` endpoint**
The proxy can call this endpoint to download attachments. There is no throttling. A compromised proxy could use it to exfiltrate files from storage.

**4. `_ensure_mailproxy_user` — password regenerated on every `activateService` call**
Every time `activateService` is called, the system user password is overwritten with `secrets.token_urlsafe(32)`. If the proxy still has the old password cached, callbacks will fail until re-registration.

**5. `message_to_send.py` — `trigger_onInserted` calls `proxyRunNow` for every single message**
If 1000 messages are inserted in a batch, 1000 `deferAfterCommit` calls are scheduled. The fixed `_deferredId` ensures only one fires, but `deferAfterCommit` with the same id overwrites — verify that the call is not lost if a commit happens mid-batch.

**6. `queue_mismatch` formula — `$proxy_ts IS NULL` added but `message_to_send` formula not updated**
The `queue_mismatch` formula now excludes messages with non-null `proxy_ts`. But the `message_to_send` formula (model line 82) was not updated — a message with `proxy_ts` set and `send_date` NULL still shows `message_to_send = true`.

**7. `deactivateService` — `delete_tenant` fails silently**
```python
try:
    self.delete_tenant()
except Exception:
    pass
```
If communication with the proxy fails, the tenant remains registered on the proxy but locally is marked as unregistered. This creates an inconsistency requiring manual intervention on the proxy.

**8. `ServiceParameters` — typo: `numberTextbox` vs `numberTextBox`** (line ~1683 in diff)
```python
fb.numberTextbox('^.db_max_waiting', ...)    # lowercase 'b'
fb.numberTextBox('^.batch_size', ...)         # uppercase 'B'
```
Verify which is the correct widget name.

**9. `_storage_node` used in `_attachment_entry_from_path` but not defined**
The method `_attachment_entry_from_path` calls `self._storage_node(path)` but this method is not defined in the class. Should probably be `self.site.storageNode(path)`.

### Minor suggestions
- `batch_code` size `':22'` — is this sufficient for UUID v4 (36 characters)?
- `proxy_limit_behavior` has only two values (`defer`/`drop`) — could use a `values` constraint
- The dashboard makes 3 separate API calls (overview, accounts, messages) when the proxy could expose a single aggregated endpoint

fporcari:
Thanks for the thorough review. Here's what was addressed and what was not, with reasoning.

## Fixed

**#1 — `_immediate_message_parameters` duplicated**
Resolved differently than suggested: instead of a mixin, added an `immediate` parameter to `sendmail` in the mail service. The callers now simply pass `immediate=True` and the service decides internally whether to force direct send (no proxy) or let the message go through the queue. The duplicated methods have been removed from both `login.py` and `prefhandler.py`.

**#6 — `message_to_send` formula not updated with `proxy_ts`**
Added `AND $proxy_ts IS NULL` to the `message_to_send` formula, aligning it with `queue_mismatch`. In practice the queue removal was the real gatekeeper, but the formula was conceptually wrong.

**#7 — `deactivateService` fails silently**
Replaced `except: pass` with `logger.warning(...)`. The deactivation flow should not block on proxy communication failure, but the error should be logged.

**#8 — `numberTextbox` typo**
Fixed to `numberTextBox`.

**#9 — `_storage_node` not defined**
Fixed to `self.site.storageNode(path)`. This was a real bug on the `weak_attachments` code path — dormant because no messages in production use that field yet.

## Not fixed (with reasoning)

**#2 — `externalcall.py` JSON on auth errors**
Currently only `mp_endpoint.py` sets `convert_result = False`. The risk is low and the change touches a critical framework point. Will revisit if other endpoints adopt `convert_result = False`.

**#3 — Rate limiting on `proxy_get_attachments`**
The endpoint requires Basic Auth with `_MAILPROXY_` tag. If the proxy credentials are compromised, rate limiting is the least of the concerns. Reasonable hardening but not blocking for this PR.

**#4 — Password regenerated on every `activateService`**
This is by design. The password is effectively a shared token: `_ensure_mailproxy_user` generates it and `register_tenant` sends it to the proxy in the same `activateService` call. They are never out of sync. The proxy receives the new password every time it re-registers.

**#5 — `proxyRunNow` called for every inserted message**
The fixed `_deferredId` is intentional deduplication. Only one `proxyRunNow` fires after commit, and it processes all pending messages in the queue — not just the one that triggered it. No calls are lost.

## Minor suggestions

- `batch_code` size `:22` — this field does not store UUIDs, the size is adequate for its purpose.
- `proxy_limit_behavior` values constraint — noted, can be added later.
- Dashboard 3 API calls — the proxy does not currently expose an aggregated endpoint. Can be optimized in a future iteration.

Commit: `17e5a67` — fix(email): address PR review findings for mail-proxy integration

genro (approved):
All critical findings addressed in commit 17e5a67. The reasoning for deferred items is sound. Approved.

PR #769, opened 2026-04-03
sped up the initial loading

dedalus2000:
I’m trying to speed up the initial loading.

I replaced all ```<style>@import url(...);</style>``` blocks with direct ```<link rel="stylesheet">``` tags. Now the browser can download all CSS in parallel instead of discovering them one by one during the parsing of ```@import```.

For CSS with media queries (e.g. print, mobile), I moved the media attribute directly onto the ```<link media="...">``` tag.

I added defer only to the scripts in js_requires. Not to dojo.js, dijitImport, or genroJsImport, because many scripts depend on others being loaded; by working only on js_requires I’m reasonably confident I won’t introduce issues.

For now I don’t have a speed test comparing before and after applying the patch. From a quick impression, the initial loading speed doesn’t seem to have changed much—likely because I’m testing on pages with few CSS files or very small ones.

### DIFF
--- a/gnrjs/gnr_d11/tpl/gnr_header.tpl
+++ b/gnrjs/gnr_d11/tpl/gnr_header.tpl
@@ -67,7 +67,7 @@
 % endfor
 
 % for jsname in js_requires:
-        <script type="text/javascript" src="${jsname}"></script>
+        <script type="text/javascript" src="${jsname}" defer></script>
 % endfor
         % if logo_url:
             <style type="text/css">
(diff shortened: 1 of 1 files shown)


genro (changes requested):
Thank you for the analysis — the observation about `@import` vs `<link>` for parallel CSS loading is correct.

However, a comprehensive revision of the base page preparation system is currently in progress, which already takes into account some of these observations. The final version will incorporate these improvements.

For this reason we are closing this PR, but your input is appreciated and will be reflected in the upcoming changes.

genro:
Closing as explained in the review — a comprehensive revision of the base page preparation system is in progress and will incorporate these improvements. Thank you for the contribution.

PR #773, opened 2026-04-06
improved CLI arg validation for IP address (interface) where to

cgabriel:
improved CLI arg validation for IP address (interface) where tobind the server to.

Update server startup log to display the current binding and provide a valid URL where to connect to, avoiding issues on widely used and distributed operating systems.

### DIFF
--- a/gnrpy/gnr/web/serverwsgi.py
+++ b/gnrpy/gnr/web/serverwsgi.py
@@ -2,6 +2,8 @@ import sys
 import os
 import atexit
 import webbrowser
+import ipaddress
+import argparse
 
 from werkzeug.serving import make_server, is_running_from_reloader
 from werkzeug._reloader import run_with_reloader
@@ -127,6 +129,16 @@ class GnrDebuggedApplication(DebuggedApplication):
 class ServerException(Exception):
     pass
 
+def ip_address(value):
+    """
+    Type validator for command line
+    """
+    try:
+        ipaddress.ip_address(value)
+        return value
+    except ValueError:
+        raise argparse.ArgumentTypeError(f"Invalid IP address: {value}")
+    
 class Server(object):
     description = "This command serves a genropy web application."
 
@@ -154,19 +166,16 @@ class Server(object):
                             dest='open_browser',
                             action='store_true',
                             help="Automatically open the browser to this application")
-        
         parser.add_argument('-c', '--config',
                             dest='config_path',
                             help="gnrserve directory path")
-
         parser.add_argument('-p', '--port',
                             dest='port',
                             help="Sets server listening port (Default: 8080)")
-
         parser.add_argument('-H', '--host',
                             dest='host',
+                            type=ip_address,
                             help="Sets server listening address (Default: 0.0.0.0)")
-
         parser.add_argument('--restore',
                             dest='restore',
                             help="Restore from path")
@@ -314,8 +323,9 @@ class Server(object):
 
     @property
     def app_url(self):
-        return f'{self.app_scheme}://{self.app_host}:{self.app_port}'
-    
+        connect_host = '127.0.0.1' if self.app_host == '0.0.0.0' else self.app_host
+        return f'{self.app_scheme}://{connect_host}:{self.app_port}'
+        
     def serve(self):
         self.app_port = int(self.options.port)
         self.app_host = self.options.host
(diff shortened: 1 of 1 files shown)


genro (commented):
## Review of PR #773

### Good

- The split into two log lines ("Started server on" + "Connect at") is a clear improvement — separates binding info from the usable connection URL.
- The `0.0.0.0` → `127.0.0.1` conversion in `app_url` is correct and useful: `0.0.0.0` is not a valid connection target on most OSes.
- IP validation at CLI level gives an early, clear error instead of a cryptic socket error later.

### Issues

#### 1. `ip_address()` validator rejects hostnames (blocking)

`ipaddress.ip_address()` only accepts numeric IP addresses. Passing `-H localhost` or `-H myhost.local` will raise an error:

```
error: argument -H/--host: Invalid IP address: localhost
```

This is a regression — currently any hostname is accepted and passed to `make_server()`, which delegates to the OS for resolution. The `--host` help text says "address" but hostnames are valid bind targets.

**Suggestion**: either use `socket.getaddrinfo()` to validate (accepts both IPs and hostnames), or rename the flag semantics explicitly and document the restriction.

#### 2. `import argparse` is redundant

`GnrCliArgParse` already extends `argparse.ArgumentParser` (defined in `gnr.core.cli`), so `argparse` is already imported there. But more importantly, the `ip_address()` function only needs `argparse.ArgumentTypeError` — you could instead just raise a plain `ValueError`, because argparse's `type=` argument catches `ValueError` and `TypeError` automatically and converts them to a user-friendly error message. This would eliminate the `import argparse` entirely:

```python
def ip_address(value):
    try:
        ipaddress.ip_address(value)
        return value
    except ValueError:
        raise ValueError(f"Invalid IP address: {value}")
```

(Or even simpler — just let the original `ValueError` from `ipaddress.ip_address()` propagate.)

#### 3. Validation gap: `type=` only applies to CLI args

The `type=ip_address` validator runs only when `--host` is passed on the command line. Values coming from `siteconfig.xml` (`wsgi?host`) or environment variables (`GNR_WSGI_OPT_host`) are applied in `init_options()` and bypass argparse validation entirely. Not necessarily a problem, but worth being aware of — invalid hosts from config will still cause a runtime error.

#### 4. Minor: trailing whitespace

Line after `raise argparse.ArgumentTypeError(...)` has trailing whitespace (visible in the diff).

### Summary

The logging improvement and `0.0.0.0` conversion are solid. The IP validation needs rework to support hostnames — this is a breaking change for anyone using `-H localhost` or similar. Suggest fixing point 1 before merge.

cgabriel:
> ### Issues
> #### 1. `ip_address()` validator rejects hostnames (blocking)
> `ipaddress.ip_address()` only accepts numeric IP addresses. Passing `-H localhost` or `-H myhost.local` will raise an error:
> 
> ```
> error: argument -H/--host: Invalid IP address: localhost
> ```
> 
> This is a regression — currently any hostname is accepted and passed to `make_server()`, which delegates to the OS for resolution. The `--host` help text says "address" but hostnames are valid bind targets.
> 
> **Suggestion**: either use `socket.getaddrinfo()` to validate (accepts both IPs and hostnames), or rename the flag semantics explicitly and document the restriction.

You're absolutely right! This has been changed to accepts hostnames too in the validator, using socket.getaddrinfo() as suggested.
 
> #### 2. `import argparse` is redundant
> `GnrCliArgParse` already extends `argparse.ArgumentParser` (defined in `gnr.core.cli`), so `argparse` is already imported there. But more importantly, the `ip_address()` function only needs `argparse.ArgumentTypeError` — you could instead just raise a plain `ValueError`, because argparse's `type=` argument catches `ValueError` and `TypeError` automatically and converts them to a user-friendly error message. This would eliminate the `import argparse` entirely:
> 
> ```python
> def ip_address(value):
>     try:
>         ipaddress.ip_address(value)
>         return value
>     except ValueError:
>         raise ValueError(f"Invalid IP address: {value}")
> ```
> 
> (Or even simpler — just let the original `ValueError` from `ipaddress.ip_address()` propagate.)

Removed, since it's not needed anymore
 
> #### 3. Validation gap: `type=` only applies to CLI args
> The `type=ip_address` validator runs only when `--host` is passed on the command line. Values coming from `siteconfig.xml` (`wsgi?host`) or environment variables (`GNR_WSGI_OPT_host`) are applied in `init_options()` and bypass argparse validation entirely. Not necessarily a problem, but worth being aware of — invalid hosts from config will still cause a runtime error.

Out of scope.

> #### 4. Minor: trailing whitespace
> Line after `raise argparse.ArgumentTypeError(...)` has trailing whitespace (visible in the diff).

Got it!

cgabriel:
Approved by @genro verbally.

PR #774, opened 2026-04-07
feat(security): add XSS sanitization for grid cells and dataTemplate rendering

fporcari:
## Summary
- Add `stripJsFromHtml` utility function that strips `<script>` tags, `on*` event handlers, and `javascript:` URIs from HTML strings at rendering time
- Protect grid cell rendering (`structFromBag_cellFormatter`) and `dataTemplate` rendering with sanitization gated by site-level `sanitize_js` switch in `instanceconfig.xml` (default OFF)
- Exclude developer-controlled HTML from sanitization: `template`, `js`, `apply`, `isbutton`, `showlinks`, `_customGetter`, `dtype='B'`/`'P'`
- Pass site switches from server to client via `page.data('gnr.switches', ...)` following the existing `api_keys` pattern

## Test plan
- [x] Regex unit tests (20 cases) — script tags, event handlers, javascript URIs, safe HTML preserved, non-string passthrough, combined vectors
- [x] Grid rendering test with malicious data and exclusion columns (isbutton, dtype B, template, js, checkboxcolumn, checkboxcell)
- [x] dataTemplate rendering test with XSS payloads
- [x] flake8 zero errors on all modified Python files
- [x] Full test suite: 1375 passed, 0 failed


gnrjs/gnr_d11/js/gnrlang.js
```diff
@@ -379,7 +395,8 @@ function dataTemplate(str, data, path, showAlways,kw) {
                                     if(formats[as_name]){
                                         value = gnrformatter.asText(value,{format:formats[as_name],dtype:dtype});
                                     }
-                                    if(editpars){                                  
+                                    if(editpars){
+                                        value = genro.safeHtmlContent(value);
```
cgabriel, line 399: Shouldn't this 2 lines be inverted in the order? value is assigned to safeHtmlContent, but after that is completely overriden by a new non-validated html

cgabriel (changes requested):

fporcari:
@cgabriel Good catch — you were right that the sanitization flow in the `editpars` block had an issue.

The `safeHtmlContent` call was correctly placed *before* the wrapper HTML (sanitize user value, then embed in dev-controlled markup), but the real problem was that `masks` and `_formattedValue` checks at lines 403-407 ran unconditionally *after* the `editpars` block, potentially overwriting the already-wrapped inline widget HTML with unsanitized values.

Fixed by merging the two separate `if(editpars)` / `if(!editpars)` blocks into a single `if/else`, so masks and formatting only apply in the non-editpars branch.

Could you re-review when you get a chance? Thanks!

gnrjs/gnr_d11/js/gnrlang.js
```diff
@@ -379,14 +395,16 @@ function dataTemplate(str, data, path, showAlways,kw) {
...
                                         value = '<div class="gnrinlinewidget_container"><div class="gnreditabletext" ondblclick="inlineWidget(event)" varname="'+as_name+'" >'+(isNullOrBlank(value)?'&nbsp':value)+'</div></div>';
```
cgabriel, line 400: The issue is still there, like 399 and line 400 shouldn't be inverted?

cgabriel (changes requested):

gnrjs/gnr_d11/js/gnrlang.js
fporcari: @cgabriel the order at lines 399-400 is intentional and correct. Here's why:

1. **Line 399** sanitizes `value` (user data) — strips `<script>`, `on*` handlers, `javascript:` URIs
2. **Line 400** embeds the *already sanitized* value into the developer-controlled wrapper HTML

If we inverted them, `safeHtmlContent` would strip the `ondblclick="inlineWidget(event)"` from the wrapper div, breaking inline editing entirely.

The pattern is: **sanitize user data first, then embed it in trusted markup** — not the other way around.

The actual bug I fixed in the last commit was a different issue: the `masks` and `_formattedValue` checks were running *after* the `editpars` block unconditionally, potentially overwriting the already-wrapped HTML. That's now correctly gated inside an `else` branch.

cgabriel: Thank you for the clarification!

cgabriel (approved):

genro (commented):
## Review of PR #774

### Architecture

The design is solid:
- `stripJsFromHtml()` — pure function with 4 regex patterns (script tags, on* handlers, javascript: URIs in href/src)
- `genro.safeHtmlContent()` — wrapper that checks the `sanitize_js` switch once (lazy-cached in `_sanitize_js`), delegates to `stripJsFromHtml` when active
- Switch propagated from server via `page.data('gnr.switches', Bag(self.application.config['switches']))` — follows existing `api_keys` pattern
- Default OFF — zero impact on existing deployments until explicitly enabled

### Sanitization placement

**Grid cells** (`genro_grid.js`): sanitization applied after all formatting (js, template, apply) but before zoom/draggable HTML wrapper. Exclusion flags (`_skipSanitize`) pre-computed once in closure scope — good for performance. Correctly skips developer-controlled content: `template`, `js`, `apply`, `isbutton`, `showlinks`, `_customGetter`, `dtype B/P`.

**dataTemplate** (`gnrlang.js`): 
- Bag branch with `editpars`: sanitize *before* wrapping in `gnrinlinewidget_container` div (preserves `ondblclick="inlineWidget(event)"` on trusted wrapper). Correct.
- Bag branch without `editpars`: sanitize *after* masks/formattedValue. Correct.
- Bag branch with sub-templates/df_templates: no sanitization (developer-controlled templates). Correct.
- Non-Bag branch: sanitize `gnrformatter.asText()` output. Correct.

The `if(editpars)/else` refactor (last commit) properly prevents masks/formattedValue from overwriting already-wrapped editpars HTML. cgabriel's concern was valid and is now correctly addressed.

### Issues

#### 1. d20 not modified (blocking)

The sanitization changes are applied only to `gnr_d11/js/`:
- `gnrlang.js` — `stripJsFromHtml` + `_JS_EXEC_PATTERNS` + dataTemplate changes
- `genro.js` — `safeHtmlContent`  
- `genro_grid.js` — `_skipSanitize` + grid cell sanitization

The corresponding `gnr_d20/js/` files are **not modified**. Sites using d20 will have `sanitize_js` enabled in config but no actual sanitization running — a false sense of security.

#### 2. Regex edge cases (non-blocking, for awareness)

The regex approach covers the main XSS vectors but has known limitations:
- **Encoded payloads**: `&#106;avascript:`, `java&#x0A;script:` will bypass the `javascript:` URI patterns
- **data: URIs**: `alert(1)</script>">` is not caught
- **SVG/MathML**: `<svg onload="alert(1)">` is caught by the `on*` pattern, but `<svg><animate onbegin="...">` edge cases may vary
- **Mutation XSS (mXSS)**: browser HTML re-parsing can reconstruct stripped patterns from encoded fragments

These are acceptable trade-offs for a regex-based approach (vs. a DOM-based sanitizer like DOMPurify). The current patterns cover the most common attack vectors. Worth documenting that this is a first layer of defense, not a complete XSS prevention.

#### 3. `Bag(self.application.config['switches'])` may throw if no switches section (minor)

If `instanceconfig.xml` has no `<switches>` element, `self.application.config['switches']` returns `None`, and `Bag(None)` should handle it gracefully (returns empty Bag). Verified this is safe — just noting for completeness.

#### 4. Test page is thorough

20 regex unit tests + grid rendering tests + dataTemplate tests, all client-side. Good coverage of attack vectors and exclusion logic.

### Summary

Clean architecture, correct sanitization placement, good test coverage. The **d20 gap** needs to be fixed before merge — the same changes must be applied to `gnr_d20/js/gnrlang.js`, `gnr_d20/js/genro.js`, and `gnr_d20/js/genro_grid.js`.

genro (approved):
LGTM — solid architecture, correct sanitization placement, good test coverage. Note: d20 JS files still need the same changes (gnrlang.js, genro.js, genro_grid.js) — tracked in review comment.

PR #775, opened 2026-04-07
fix(login): enable keyboard Tab navigation in login dialogs

fporcari:
## Summary
- Add `focusOnTab=True` to all 7 submit buttons across login-related dialogs (login, lost password, new password, OTP, confirm user, new user, screen lock) so they are reachable via Tab key
- Exclude secondary lightbuttons (Lost password, New User) from the tab flow with `tabindex=-1` to keep the main login sequence focused: Username → Password → Enter
- Add CSS focus indicators for confirm buttons (`.dijitFocused`) and option links (`:focus`) for visual feedback

## Test plan
- [ ] Open login dialog, verify Tab order: Username → Password → Enter
- [ ] Verify focus ring appears on the Enter button when focused
- [ ] Verify Enter key triggers login when button is focused
- [ ] Verify Shift+Tab navigates backwards correctly
- [ ] Verify Lost password / New User links are still clickable (just not in tab flow)
- [ ] Verify other login dialogs (lost password, new password, OTP, screen lock) have tabbable submit buttons

### DIFF
--- a/projects/gnrcore/packages/adm/resources/login.css
+++ b/projects/gnrcore/packages/adm/resources/login.css
@@ -65,6 +65,18 @@
     font-size: .95em;
 }
 
+.login_confirm_btn.dijitFocused .dijitButtonNode{
+    outline: 2px solid var(--accent-color, #4a90d9);
+    outline-offset: 2px;
+    box-shadow: 0 0 0 3px color-mix(in srgb, var(--accent-color, #4a90d9) 25%, transparent);
+}
+
+.login_option_btn:focus{
+    outline: 2px solid var(--accent-color, #4a90d9);
+    outline-offset: 2px;
+    border-radius: 2px;
+}
+
 .login_option_btn{
     color: var(--text-secondary);
     font-size: .8em;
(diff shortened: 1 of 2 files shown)


genro (commented):
## Review of PR #775

### How it works

GenroPy buttons by default get `tabindex=32767` (effectively unreachable via Tab), unless `focusOnTab=True` is set — which keeps the default tabindex and makes them part of the normal tab flow.

### Changes

**login.py** (9 lines changed):
- 7 submit buttons across all login dialogs get `focusOnTab=True`: Enter, Recover, Send, Confirm, Send Email
- 2 secondary links (`Lost password`, `New User`) get `tabindex=-1` to exclude them from tab flow

**login.css** (12 lines added):
- Focus indicator for `.login_confirm_btn.dijitFocused` — outline + box-shadow using `--accent-color`
- Focus indicator for `.login_option_btn:focus` — outline with border-radius

### Assessment

Clean and consistent. The tab flow Username → Password → Enter is the right default. Secondary actions remain clickable but don't clutter keyboard navigation.

Two minor notes:

1. **`.login_option_btn:focus` rule may be unreachable** — these buttons now have `tabindex=-1`, so they can only be focused programmatically, not via Tab. The CSS rule isn't harmful but won't fire in normal keyboard navigation. It could be useful if someone clicks them (click gives focus), so it's fine to keep.

2. **CSS `color-mix()` compatibility** — `color-mix(in srgb, ...)` requires Safari 16.2+, Chrome 111+, Firefox 113+. Should be fine for any modern deployment, just noting it.

No issues found. Good to merge.

genro (approved):
LGTM — clean keyboard navigation fix for all login dialogs, correct tab flow and focus indicators.

PR #778, opened 2026-04-07
ensure a menu is displayed

### LINKED ISSUE #430: Menu not shown when 'menu package' not specified in instanceconfig
## Issue Description

When an instance does not specify `<menu package="*"/>` in `instanceconfig.xml`, the application menu is not displayed, showing an empty menu instead.

## Current Behavior

In `gnr/web/gnrmenu.py`, the `indexMenu` property (lines 245-255) has this logic:

```python
@property
def indexMenu(self):
    if self._page.userMenu:
        return self._page.userMenu
    instanceMenu = self.getInstanceMenu()
    if instanceMenu:
        return instanceMenu
    pkgMenus = self.app.config['menu?package']  # ← Reads from instanceconfig.xml
    if pkgMenus:
        return self.legacyMenuFromPkgList(pkgMenus)
    return self.mainPackageMenu(self._page.package.name)  # ← Falls back to mainPackage only!
```

**Problem**: When `menu?package` is not specified in instanceconfig, it falls back to `mainPackageMenu()` which:
1. Only searches for menu in the mainPackage
2. If no ApplicationMenu is explicitly defined, returns an empty menu
3. Does not iterate through all packages to build the menu

## Expected Behavior

The default behavior should be `package='*'` (all packages), equivalent to:

```xml
<menu package="*"/>
```

**Proposed logic**:
- **Default**: Show menus from all packages (as if `package='*'` was specified)
- **Only if explicitly specified**: Use the specific package list (e.g., `package="erpy_base,erpy_fatt"`)
- **Only if ApplicationMenu exists**: Use the ApplicationMenu from mainPackage

## Why This Matters

Most instances have multiple packages (e.g., `gnrcore:sys`, `gnrcore:adm`, `gnr_it:glbl`, custom packages) and expect to see menu items from all of them by default. The current behavior requires explicitly adding `<menu package="*"/>` to every instanceconfig, which is not intuitive.

## Suggested Fix

Change line 252-255 in `gnrmenu.py` from:

```python
pkgMenus = self.app.config['menu?package']
if pkgMenus:
    return self.legacyMenuFromPkgList(pkgMenus)
return self.mainPackageMenu(self._page.package.name)
```

To:

```python
pkgMenus = self.app.config.get('menu?package', '*')  # Default to '*'
return self.legacyMenuFromPkgList(pkgMenus)
```

Or preserve mainPackageMenu as a special case:

```python
pkgMenus = self.app.config.get('menu?package')
if pkgMenus is None:
    # Try mainPackage first, fallback to all packages
    mainPkgMenu = self.mainPackageMenu(self._page.package.name)
    if mainPkgMenu and len(mainPkgMenu) > 0:
        return mainPkgMenu
    pkgMenus = '*'
return self.legacyMenuFromPkgList(pkgMenus)
```

## Workaround

Until this is fixed, add to `instanceconfig.xml`:

```xml
<menu package="*"/>
```

## Environment

- Genropy version: Latest (2025)
- File: `gnr/web/gnrmenu.py` lines 245-255

cgabriel:
check if the main package has a menu, and if not, fallback to using '*' as menu?package, which will show the menu arrange by package list from the instance.

This maintains compatibility with <menu package=..> usage in instanconfig.xml, the menu (if present) from the main package is used, but at least it will show a menu if the main package doesn't have one and <menu> is not configured.

Otherwise, no menu will be shown. Useful for new main packages for an application which doesn't have a menu yet.

close #430

### DIFF
--- a/gnrpy/gnr/web/gnrmenu.py
+++ b/gnrpy/gnr/web/gnrmenu.py
@@ -249,10 +249,15 @@ class MenuResolver(BagResolver):
         instanceMenu = self.getInstanceMenu()
         if instanceMenu:
             return instanceMenu
-        pkgMenus = self.app.config['menu?package']
-        if pkgMenus:
-            return self.legacyMenuFromPkgList(pkgMenus)
-        return self.mainPackageMenu(self._page.package.name)
+
+        pkgMenus = self.app.config.get('menu?package')
+        if pkgMenus is None:
+            # Try mainPackage first, fallback to all packages
+            mainPkgMenu = self.mainPackageMenu(self._page.package.name)
+            if mainPkgMenu and len(mainPkgMenu) > 0:
+                return mainPkgMenu
+            pkgMenus = '*'
+        return self.legacyMenuFromPkgList(pkgMenus)
     
     def mainPackageMenu(self,pkg):
         result = self.pkgMenu(pkg,className=getattr(self._page,'menuClass',None))


fporcari (approved):
Clean fallback logic. The cascade mainPackage → '*' makes sense for new packages without a menu. Good catch on using .get() instead of direct key access.

PR #779, opened 2026-04-08
log an error when a virtual column is used in a compositeColumn.

### LINKED ISSUE #256: Warning message if a compositeColumn is defined using virtual columns
compositeColumns need real columns to be used in their definition: using *virtual columns* leads to undesired effects such those described in #254.

At the moment, there is no check if virtual columns are used in a compositeColumn definition. We should log a WARNING or an ERROR to tell clearly the developer that something's wrong.

cgabriel:
close #256

### DIFF
--- a/gnrpy/gnr/sql/gnrsqlmodel/model.py
+++ b/gnrpy/gnr/sql/gnrsqlmodel/model.py
@@ -1073,6 +1073,8 @@ class DbModelSrc(GnrStructData):
                 val = f""" '"' ||  ${column} || '"' """
             elif dtype not in ('L', 'F', 'R', 'B'):
                 val = rf""" '"' ||  ${column} || '\:\:{dtype}"' """
+            if column in self['virtual_columns']:
+                logger.error("Composite column %s use virtual column %s, a real column is required", name, column)
             composed_of.append(column)
             chunks.append(val)
         composed_of_str = ','.join(composed_of)
--- a/gnrpy/gnr/sql/gnrsqlmodel/model.py
+++ b/gnrpy/gnr/sql/gnrsqlmodel/model.py
@@ -1073,7 +1073,7 @@ class DbModelSrc(GnrStructData):
                 val = f""" '"' ||  ${column} || '"' """
             elif dtype not in ('L', 'F', 'R', 'B'):
                 val = rf""" '"' ||  ${column} || '\:\:{dtype}"' """
-            if column in self['virtual_columns']:
+            if column in self.get('virtual_columns', []):
                 logger.error("Composite column %s use virtual column %s, a real column is required", name, column)
             composed_of.append(column)
             chunks.append(val)


dgpaci (approved):
LGTM. The 2-line check correctly warns developers when a `compositeColumn` is defined over a virtual column, addressing #256. Use of `self.get('virtual_columns', [])` safely handles tables without any virtual columns.

## Tests performed

Tested in a sandbox project against the `prodotto` table.

**Negative case (regression):** added a `formulaColumn` (wrapper on `virtual_column`) plus a `compositeColumn` referencing it:

```python
tbl.formulaColumn('descrizione_upper',
                  sql_formula='upper($descrizione)',
                  dtype='T', name_long='Descrizione maiuscola')
tbl.compositeColumn('codice_descrizione_key',
                    columns='codice,descrizione_upper')
```

On app startup, the expected error log was produced:

```
ERROR: gnr.sql/model - Composite column codice_descrizione_key use virtual column descrizione_upper, a real column is required
```

**Positive case (no false positive):** replaced the `compositeColumn` with one using only real columns:

```python
tbl.compositeColumn('codice_descrizione_key',
                    columns='codice,descrizione')
```

No error log emitted — confirming the check does not fire on well-formed definitions.

genro (changes requested):
The check is correct but logging an error and continuing is not enough — the resulting SQL formula will be broken and fail at runtime anyway.

Since `compositeColumn` runs at model load (startup), this is the right place to fail hard. A virtual column as source of a compositeColumn is a model definition error and should raise immediately.

Suggested change — replace the `logger.error` with a raise:

```python
if column in self.get('virtual_columns', []):
    raise GnrSqlException(
        f"compositeColumn '{name}': column '{column}' is virtual, a real column is required"
    )
```

Also, move the check right after the column name normalization (after the `$` strip at line 1070), before computing `dtype` and `val` — no point doing that work on a column that will be rejected.

genro:
**Note:** this review is still in progress — do not consider it complete yet.

Additional point we are evaluating: the check should not test whether the column is virtual, but rather whether it is a **real** column. A column that is neither real nor virtual (e.g. a typo) would pass the current check silently. The correct guard is to verify that the column exists as a physical column in the table. More details to follow.

cgabriel:
> The check is correct but logging an error and continuing is not enough — the resulting SQL formula will be broken and fail at runtime anyway.
> 
> Since `compositeColumn` runs at model load (startup), this is the right place to fail hard. A virtual column as source of a compositeColumn is a model definition error and should raise immediately.
> 
> Suggested change — replace the `logger.error` with a raise:
> 
> ```python
> if column in self.get('virtual_columns', []):
>     raise GnrSqlException(
>         f"compositeColumn '{name}': column '{column}' is virtual, a real column is required"
>     )
> ```
> 
> Also, move the check right after the column name normalization (after the `$` strip at line 1070), before computing `dtype` and `val` — no point doing that work on a column that will be rejected.

Your point is valid, but please consider that this is something out of context of the original request #256

genro:
After further analysis, we propose a different approach on branch `fix/compositeColumn-validation`.

The current fix checks `virtual_columns` inside the `compositeColumn` loop during `config_db`. This has two limitations:

1. **Order-dependent**: if a source column is declared *after* the `compositeColumn` in `config_db`, the check would fail even though the column is perfectly valid
2. **Incomplete validation**: checking "is virtual" misses the case of a non-existent column (e.g. a typo), which would pass silently

The proposed solution defers the formula construction to the building phase (`deferOnBuilding`), after all `config_db` methods have run. This gives:

- **Order independence**: source columns can be declared in any order relative to the `compositeColumn`
- **Explicit raise**: validates that each source column is a physical column (`columns_bag.getNode(column)`), catching both virtual columns and typos with a clear `GnrSqlException`
- **No breaking change**: the virtual column node is created immediately so `.relation()` chaining keeps working

Branch: `fix/compositeColumn-validation`

genro:
Understood — but once a problem surfaces, it's in our interest to investigate thoroughly rather than limit ourselves to a fix that only partially covers the issue.

Every intervention has a cost in terms of management, and if we have the opportunity to touch an area and improve it properly, it's preferable not to limit ourselves to the minimum requested. The deferred approach on branch `fix/compositeColumn-validation` addresses the root cause (order-independent validation, catches both virtual and non-existent columns) with minimal additional complexity.

dgpaci:
The branch `fix/compositeColumn-validation` mentioned above is not pushed to origin (verified via `git ls-remote`, only the `256-warning-...` branch is present). Could you open a new PR from `fix/compositeColumn-validation` so we can review the proposed `deferOnBuilding` approach side-by-side with the current fix and make a call? Thanks.

cgabriel:
> The branch `fix/compositeColumn-validation` mentioned above is not pushed to origin (verified via `git ls-remote`, only the `256-warning-...` branch is present). Could you open a new PR from `fix/compositeColumn-validation` so we can review the proposed `deferOnBuilding` approach side-by-side with the current fix and make a call? Thanks.

@genro your call here, the branch with your proposal is missing

genro:
A clarification on my earlier comment referencing branch `fix/compositeColumn-validation`: at the time of that comment the branch had not actually been pushed to origin — I apologise for the confusion that caused, as @dgpaci correctly flagged.

The branch is now pushed: https://github.com/genropy/genropy/tree/fix/compositeColumn-validation

It implements the deferred approach discussed, by splitting `compositeColumn()` into two phases:

1. **Synchronous (at `config_db` time)** — normalise the column names, create the virtual-column node with `sql_formula=None`, and register a deferred callback. The node is returned immediately so `.relation()` chaining on the result keeps working.
2. **Deferred (in `runOnBuildingCb`, after all `config_db` methods have run)** — validate every source column against the now-complete table definition:
   - physical column → use its dtype to build the formula fragment
   - virtual column → `raise GnrSqlException` (was silently producing broken SQL)
   - missing column → `raise GnrSqlException` (was silently creating an empty physical column via `self.column()` side effect)

   Then assembles the final `sql_formula` and assigns it on the virtual-column node.

This addresses both limitations raised in the previous review:
- **Order-independence**: source columns can be declared before or after the `compositeColumn()` call in `config_db`
- **Typo detection**: a non-existent source column now fails fast at build time instead of producing a broken formula at query time

### Verification

- Lint: `flake8` clean on the modified file
- Tests: full `gnrpy/tests/` suite — `1575 passed`, `67 skipped`, `22 failed` (all pre-existing locale/date-related failures on clean develop, identical count with and without the patch)
- Dedicated `test_composite_column.py`: 14/14 passed
- The existing `compositeColumn` usages in `test_invoice` (`price_year`, `price_year_note`) still work end-to-end, including `.relation()` chaining and navigation through composite keys

### Proposal

Please consider this branch as an alternative to the current PR. Happy to open a separate PR if the team prefers reviewing it side-by-side, or to close this PR and open a new one — whichever is more convenient. No strong preference here, the technical direction is what matters.

genro:
@cgabriel @mbertoldi @dgpaci — fyi, see the comment above regarding the alternative branch `fix/compositeColumn-validation`.

dgpaci:
Thanks for pushing the branch, @genro. Since the `fix/compositeColumn-validation` approach is technically divergent from this PR (raises vs logs, deferred vs synchronous, and additionally catches typos / missing source columns), it's cleaner to review and merge it on its own merits. Could you open a dedicated PR against `develop` from that branch?

We'll close this PR #779 in favor of the new one -> #817

cc @cgabriel @mbertoldi

PR #782, opened 2026-04-09
fix validation error indicator not showing on readonly fields (checkboxtext)

fporcari:
ReadOnly CSS selectors (.dijitReadOnly.dijitTextBox) had higher specificity
than the Error selector (.dijitError), so error border/background was masked
on readonly fields like checkboxtext. Added ReadOnly+Error rules with
matching specificity to ensure validation indicators display correctly.

https://claude.ai/code/session_01AWZwSfSgsT4T4Y5oY3PuqZ

PR #783, opened 2026-04-09
fix validation error indicator not showing on readonly fields

fporcari:
## Summary
- ReadOnly CSS selectors (.dijitReadOnly.dijitTextBox) had higher specificity than the Error selector (.dijitError), so error border/background was masked on readonly fields like checkboxtext
- Added ReadOnly+Error rules with matching specificity to ensure validation indicators display correctly
- Applied to both gnr_d11 and gnr_d20

Supersedes #782


genro (approved):
ok

PR #787, opened 2026-04-10
feat(#786): add WSGI gzip compression middleware

### LINKED ISSUE #786: feat: add WSGI gzip compression middleware
## Problem

When enabling gzip compression in the reverse proxy (Nginx/Traefik), a Content-Length mismatch can occur: Werkzeug sets Content-Length on the uncompressed body, but after proxy compression the actual body size differs. This can cause truncated responses or client errors depending on the proxy configuration.

Currently GenroPy has no application-level compression and relies entirely on the reverse proxy.

## Solution

Add a WSGI `GzipMiddleware` that compresses responses at the application level, making the app independent from reverse proxy compression configuration.

The middleware:
- Checks `Accept-Encoding: gzip` from the client
- Only compresses compressible content types (text/*, application/json, application/javascript, application/xml)
- Respects a minimum response size threshold (default 500 bytes)
- Only uses compressed version if it's actually smaller
- Correctly recalculates `Content-Length` after compression
- Sets `Content-Encoding: gzip` and `Vary: Accept-Encoding` headers

## Configuration

Enable via siteconfig:
```xml
<wsgi compression="true" compression_min_size="500" compression_level="6"/>
```

## Files

- `gnrpy/gnr/web/compression.py` — new GzipMiddleware
- `gnrpy/gnr/web/gnrwsgisite.py` — middleware integration in `build_wsgiapp()`

genro:
## Summary

- Add `GzipMiddleware` WSGI middleware that compresses HTTP responses at the application level
- Makes GenroPy independent from reverse proxy compression (Nginx/Traefik)
- Solves Content-Length mismatch when reverse proxy compresses responses that already have Content-Length set by Werkzeug

## Configuration

Enable via siteconfig:
```xml
<wsgi compression="true" compression_min_size="500" compression_level="6"/>
```

## Details

The middleware buffers the response and compresses when all criteria are met:
- Client sends `Accept-Encoding: gzip`
- Content-Type is compressible (text/*, application/json, application/javascript, application/xml)
- Response size >= minimum_size (default 500 bytes)
- Compressed size < original size
- Response is not already encoded

After compression, `Content-Length` is recalculated, `Content-Encoding: gzip` and `Vary: Accept-Encoding` are added.

## Test plan

- [ ] Start dev server with `<wsgi compression="true"/>` in siteconfig
- [ ] Verify HTML pages return `Content-Encoding: gzip` header
- [ ] Verify `Content-Length` matches the compressed body size
- [ ] Verify small responses (< 500 bytes) are not compressed
- [ ] Verify requests without `Accept-Encoding: gzip` get uncompressed responses
- [ ] Verify binary responses (images, zip) are not compressed

Closes #786

### DIFF
--- a/gnrpy/gnr/web/compression.py
+++ b/gnrpy/gnr/web/compression.py
@@ -1,0 +1,143 @@
+# Copyright 2026 Softwell S.r.l.
+# Licensed under the Apache License, Version 2.0
+
+"""WSGI middleware for gzip compression of HTTP responses.
+
+Compresses responses when all criteria are met:
+    - Client accepts gzip (Accept-Encoding header)
+    - Response size >= minimum_size
+    - Content-Type is compressible (text/*, application/json, etc.)
+    - Compressed size < original size
+
+Adds Content-Encoding: gzip and Vary: Accept-Encoding headers.
+Updates Content-Length to compressed size.
+
+Enable via siteconfig::
+
+    <wsgi compression="true" compression_min_size="500" compression_level="6"/>
+"""
+
+import gzip
+import io
+
+
+COMPRESSIBLE_TYPES = (
+    'text/',
+    'application/json',
+    'application/javascript',
+    'application/xml',
+    'application/xhtml+xml',
+)
+
+
+class GzipMiddleware:
+    """WSGI middleware that applies gzip compression to responses.
+
+    Wraps the WSGI app, intercepts responses, and compresses them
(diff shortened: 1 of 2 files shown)


cgabriel (approved):
Verified cases with local wsgi+werkzeug, nginx http termination (via tcp, not unix socket) with gzip enable and gzip disabled.

PR #798, opened 2026-04-15
Fix XLSX importer column shift and update_only+sql_mode

### LINKED ISSUE #797: Fix Excel importer: column shift with empty headers and update_only+sql_mode handling
## Description

The Excel (XLSX) importer has multiple bugs that cause incorrect data mapping and import failures when:
1. The source file contains columns with empty or non-alphanumeric headers (e.g., system columns like `_protectionStatus`)
2. The "Update only" import mode is used together with "SQL Mode"

## Bug 1: Column value shift with empty/whitespace headers in XlsxReader

**Root cause**: Two issues in `XlsxReader` combine to produce a column alignment shift:

### 1a. `_sheetlines` ignores cell position (`cell.column`)
In `read_only=True` mode, openpyxl may omit leading empty cells from data rows if those cells were never written in the XML. The `_sheetlines` method simply appends values in iteration order (`result.append(value)`) without using `cell.column` to place them at the correct position. This causes data rows to shift left when leading cells are missing.

### 1b. `addSheet` checks for empty header before `slugify`, not after
Headers like `' '` (space) or `_protectionStatus` pass the `if not header` check (they're truthy), but `slugify()` converts them to `''`. The empty string is then excluded from `colindex`, causing the corresponding position to be filtered out of every data row — shifting all subsequent values by one position.

**Result**: `row['id']` returns the value from the next column (e.g., `importo`), causing wrong data to be used as keys or inserted into wrong fields.

**Reproduction**:
1. Export a table that includes a system column (like `_protectionStatus`) as XLSX
2. Re-import the file using the Match Columns importer
3. Values are mapped to the wrong columns

## Bug 2: `update_only` import mode ignored when `sql_mode` is active

**Root cause**: In `defaultMatchImporterXls`, the `if sql_mode` branch is checked before `elif import_mode=='update_only'`. When both are active, the code always executes `insertMany` instead of `recordToUpdate`, causing:
- `UniqueViolation` errors (trying to INSERT existing records)
- The update logic is never reached

## Bug 3: `KeyError: '_updater_keyfield'` in `adaptedRecords`

**Root cause**: The `match_index` dict contains a special metadata key `_updater_keyfield` (set by the JS client for update mode). The dict comprehension in `adaptedRecords` iterates all `match_index` items and tries `row[k]` with `k='_updater_keyfield'`, which doesn't exist as a column in the Excel row.

## Files affected

- `gnrpy/gnr/core/gnrlist.py` — `XlsxReader._sheetlines` and `XlsxReader.addSheet`
- `gnrpy/gnr/web/gnrwebpage_proxy/utils.py` — `defaultMatchImporterXls` and `adaptedRecords`

dgpaci:
## Summary

Fixes #797. The column shift and the `update_only` handling turned out to be two
independent defects with a single symptom each:

- **`XlsxReader._process_firstline`** (`gnrpy/gnr/core/flatfiles.py`): slugify the
  header **before** the empty check, so a header that slugifies to `''` (whitespace,
  `#`, `_protectionStatus`) becomes `gnr_emptycol_N` instead of dropping out of
  `colindex`. That drop is the whole shift: the header kept its slot in `index` while
  `__call__` filtered its value out of every row, so `row['id']` returned the next
  column. `_sheetlines` needed no change — appending the values is already aligned,
  because the read_only reader pads every row from column A.
- **`defaultMatchImporterXls`** (`gnrpy/gnr/web/gnrwebpage_proxy/utils.py`): check
  `import_mode=='update_only'` before `sql_mode`, so the combination updates instead
  of going to `insertMany` and dying on `UniqueViolation`.
- **`adaptedRecords`**: assigning a pkey to the keyless rows is now the explicit
  `assign_pkey` parameter, and it is off in the update paths. It exists for the raw
  `insertMany`, and in an update `raw_update` would write it over the pkey of the
  record being updated (`SET pkey=<generated> WHERE pkey=<current>`, triggerless),
  orphaning every fkey pointing at it.
- **`adaptedRecords`**: filter the client's `_updater_keyfield` metadata entry out of
  the `match_index` comprehension, which otherwise looks it up as a row column and
  raises `KeyError`.

## Tests

`gnrpy/tests/web/table_importer_test.py` (new) drives `defaultMatchImporterXls` over a
real xlsx file against the `gnrtest` instance and asserts on the records left in the
database — no mocks. On `develop` 4 of its 5 tests fail with
`KeyError: '_updater_keyfield'`; `test_update_only_sql_mode_keeps_the_pkey` also fails
on this branch before the last commit, where it caught the pkey being rewritten.

`gnrpy/tests/core/flatfiles_test.py` covers the reader: blank header before named
columns, header left empty by slugify, blank leading/interior/trailing cells, and
empty rows under `compressEmptyRows`/`allEmptyRows`. The first two fail on `develop`.

`pytest tests --ignore=tests/sql` → 1264 passed. flake8 clean on the four files.

## Manual test plan

- [ ] Import an XLSX that has a leading empty or system column (e.g. `_protectionStatus`) — values map to the right columns
- [ ] Import with "Update only" + "SQL Mode", keying on a business code — records are updated **and keep their pkey**
- [ ] Import with "SQL Mode" alone — insert still works, rows still get a pkey
- [ ] Import without "SQL Mode" — standard insert/update unchanged


cgabriel (approved):

dgpaci:
<html><head></head><body><p>Pushed <code>8cc68048</code> on top of the rebase: the positional rewrite of <code>_sheetlines</code> was broken, and the tests that would have caught it were missing.</p>
<p><strong>The <code>AttributeError</code></strong></p>
<p>In <code>read_only</code> mode openpyxl pads rows with <code>openpyxl.cell.read_only.EmptyCell</code>, whose <code>__slots__</code> is empty — it exposes <code>value</code> but has no <code>column</code>. So both <code>line[-1].column</code> and <code>cell.column</code> blow up on any sheet that contains a blank cell:</p>
<pre><code>row: ['id', 'importo', 'note']
AttributeError: 'EmptyCell' object has no attribute 'column'
</code></pre>
<p>(header <code>id | importo | note</code>, one data row with <code>A2</code> left empty.)</p>
<p><code>gnrpy/tests/core/data/test.xlsx</code> is two rows with no blank cell at all, which is why the suite stayed green.</p>
<p><strong>The shift is not where the fix was</strong></p>
<p>The root cause of #797 is entirely in <code>_process_firstline</code>. On <code>develop</code> a whitespace-only header is truthy, so it is not renamed, and <code>slugify</code> then turns it into <code>''</code>; it stays in <code>headers</code>/<code>index</code> but drops out of <code>colindex</code>, while <code>__call__</code> filters the values by <code>colindex</code>. With header row <code>['   ', 'id', 'importo']</code> and data row <code>['note', 'K1', 100]</code>:</p>
<pre><code>headers:  ['', 'id', 'importo']
index:    {'': 0, 'id': 1, 'importo': 2}
colindex: {1: True, 2: True}
row['id'] = 100          # expected 'K1'
</code></pre>
<p>That is exactly the reported symptom, and the <code>_process_firstline</code> change fixes it on its own. I could not produce a case where openpyxl omits leading cells from <code>sheet.rows</code>, so the positional rewrite was not what fixed #797 — it only added the crash.</p>
<p><strong>What <code>8cc68048</code> does</strong></p>
<p><code>_sheetlines</code> now iterates with <code>sheet.iter_rows(min_col=1)</code>. openpyxl then pads every row from column A up to the sheet width, so appending the values preserves the alignment without ever reading a cell coordinate. It also covers a file whose declared dimension does not start at column A, where cells outside <code>[min_col, max_col]</code> would be dropped silently.</p>
<p>Plus the four regression tests that were missing:</p>

  | develop | previous head | 8cc68048
-- | -- | -- | --
blank header before named columns | fail (shift) | pass | pass
header made only of punctuation | fail | pass | pass
rows with blank leading / interior / trailing cell | pass | fail (AttributeError) | pass
empty rows honour compressEmptyRows / allEmptyRows | pass | fail (AttributeError) | pass


<p><code>tests/core</code>: 343 passed, 1 pre-existing network-dependent failure (<code>gnrbag_test.py::test_fillFromUrl</code>). flake8 clean on the modified files.</p>
<p>The <code>update_only</code> before <code>sql_mode</code> reorder and the <code>_updater_keyfield</code> filter in <code>adaptedRecords</code> were already correct and are untouched. Still no coverage for those two on the web side — <code>defaultMatchImporterXls</code> needs the db fixture, so I left it out of this commit.</p>
<p>@cgabriel ready for another look.</p></body></html>

fporcari (changes requested):
The diagnosis of the column shift is right, and it is the good kind of right — I
rebuilt the mechanism from the code rather than the description and it lands
exactly where you say. `_build_index` builds the header→position index from the
**full** `headers` list, while `__call__` filters the values through `colindex`
(`gnr/core/flatfiles.py:202`):

```python
yield GnrNamedList(s['index'], [c for i, c in enumerate(line) if i in s['colindex']])
```

With `headers = ['', 'id', 'importo']` and `colindex = {1, 2}`, the index says
`id`→1 but the filtered list has two elements, so `row['id']` returns `importo`.
The unwritten invariant is *every position in `headers` must be in `colindex`*, and
keeping the column under a generated name restores it. Worth noting that
`XlsReader._process_firstline` (`flatfiles.py:245-250`) satisfies the same
invariant the opposite way — it filters `headers` too, so it stays aligned but
drops the column entirely. Same file, two behaviours for the same input; not a
defect of this PR, but someone will trip on it.

Verified on `8cc680487`: `pytest tests/core/flatfiles_test.py` → 47 passed, and
flake8 clean on the three files with the project config.

Requesting changes for one reason, plus a test gap and a stale description.

## Blocking: `update_only` + `sql_mode` rewrites the primary key of every updated record

Before this PR that combination went to `insertMany` and died on
`UniqueViolation` — broken, but loud. With the branches reordered it reaches the
update path, and there:

1. `adaptedRecords(sql_mode=True)` generates a pkey when the row has none
   (`gnrwebpage_proxy/utils.py:405-408`), through `newPkeyValue` →
   `getUuid()`, or `max($pkey)+1` for numeric keys (`gnrsqltable/utils.py:75-90`).
2. The update branch pops **only** the updater key: `key = r.pop(_updater_keyfield)`
   (`utils.py:370`). If `_updater_keyfield` is not the pkey, `r` still carries the
   generated one.
3. `rec.update(r)` then exits `RecordUpdater` with `raw=True`, which calls
   `raw_update(self.record, self.oldrecord, pkey=self.pkey)` where `self.pkey` is
   the **old** pkey (`gnrsqltable/helpers.py`, `__exit__`). That writes
   `SET <pkey>=<new uuid> WHERE <pkey>=<old>`, raw, with no triggers.

And `_updater_keyfield` is not the pkey in the normal case: the client sets it from
whichever column the user ticks as `is_key` in the match grid
(`gnrjs/gnr_d11/js/genro_components.js:1522`), which is typically a business code.
So the main use of the feature is the case that corrupts, silently, and every
foreign key pointing at those records is orphaned.

The fix is to not generate a pkey when the mode is an update — the
`if sql_mode: r[tpkey] = newPkeyValue(r)` block exists for the `insertMany` path
and is actively wrong for `recordToUpdate`. No test covers this branch: all four
new tests are `XlsxReader` only.

## The 1a fix is not pinned by any test

I ran the four new tests against `develop`'s `flatfiles.py`: **2 failed, 6 passed**.
Only the two header tests fail without the fix.
`test_XlsxReader_rows_with_blank_cells` and
`test_XlsxReader_empty_rows_are_skipped_by_default` pass on `develop` too, so the
`iter_rows(min_col=1)` half — the original traceback — has nothing holding it.

That is not fixable with another assertion: `_write_xlsx` writes the file *with
openpyxl*, which always emits a complete `dimension`, so it cannot produce the file
the change targets (a sheet whose dimension does not start at A, or which omits
leading cells). It needs a committed xlsx fixture.

## The description describes the abandoned solution

The first bullet says "use `cell.column` to position values at the correct index".
The diff does the opposite, and its own new docstring says so: *"Cells are never
inspected for their position: in read_only mode the padding cells are EmptyCell
instances, which expose `value` but have no `column` attribute."* The commit
message agrees — "read xlsx rows with min_col=1 **instead of** cell.column".
`min_col=1` is the better solution; the body just still promises the other one, and
#797 §1a says the same. Both worth correcting, since the next reader will go
looking for `cell.column` and not find it. #797 also names
`gnrpy/gnr/core/gnrlist.py`, while the code now lives in `flatfiles.py`.

## Minor

- The docstring of `test_XlsxReader_header_slugifying_to_empty_is_renamed` says
  punctuation-only headers "slugify to `''` and get a generated name", but its own
  assertions show `'---'` → `'_'` — kept, not renamed. The docstring contradicts
  the test.
- Consequence of the same point: two different punctuation headers can both
  slugify to `'_'`, giving duplicate header names.
- The `_updater_keyfield` filter in `adaptedRecords` is redundant in the
  `update_only` path itself: `adaptedRecords` is a generator and
  `match_index.pop` (`utils.py:366`) runs before it is consumed. It does earn its
  place in the `sql_mode`/`else` branches, where the client can send the key
  anyway. Worth a comment, because as written the correctness of that path rests
  on generator laziness.

## Process

The PR has no assignee. Since the merge belongs to the assignee, nobody is on the
hook for closing this one — which is plausibly how it reached four months. The only
review on it is an approval that was dismissed on 15 April, the day it opened, so
it has never actually had a standing review either.

dgpaci:
@fporcari all three points addressed, in `65a9da534` and `8ca5ea5d9`. Your reading of the pkey rewrite is right down to the last step, and the test I added reproduces it before it is fixed.

**Blocking: the pkey of every updated record**

Fixed in `65a9da534`. Assigning the pkey is now an explicit `assign_pkey` parameter of `adaptedRecords`, off whenever `import_mode=='update_only'`; it defaults to `sql_mode` so the `importer` call site keeps its behaviour. The chain you described is exactly what happens — `raw_update(record, oldrecord, pkey=self.pkey)` takes the WHERE from `old_record` and puts every key of the row in the SET (`_gnrbaseadapter.py:864-876`), so the generated key landed on the record it was updating.

It is covered now, against the database rather than a mock: `gnrpy/tests/web/table_importer_test.py` builds a real xlsx, runs `defaultMatchImporterXls` on the `gnrtest` instance and reads the records back. On the branch as it stood:

```
assert record['id'] == self.record['id']
AssertionError: assert 'TC1PqUypNBaTRzxgNlBD_Q' == 'dhlr9CZXOKywW83NBQ5_lw'
```

Five tests: pkey preserved under `update_only`+`sql_mode`, the same import without `sql_mode`, an unknown key reported and not inserted, `sql_mode` insert still getting its pkey (so the fix cannot over-reach into `insertMany`), and `_updater_keyfield` never read as a column. Four of the five fail on `develop` with `KeyError: '_updater_keyfield'` — including the `update_only`+`sql_mode` one, which never even reaches the `UniqueViolation` #797 reports.

**The 1a fix is not pinned by any test — because there is no 1a fix**

You are right that no assertion can pin it, and the reason is stronger than the fixture problem: `iter_rows(min_col=1)` **is** `sheet.rows`. `ReadOnlyWorksheet.rows` delegates to `iter_rows()`, which does `min_col = min_col or 1` itself, and the `if self._current_row == 0 and not any([...])` early return — the only thing the explicit argument could change — cannot fire on a read-only sheet: `ReadOnlyWorksheet.__init__` sets `_current_row = None` and nothing ever assigns it again (openpyxl 3.1.5, `worksheet/_read_only.py:39`; `_current_row = 0` belongs to the writable `Worksheet`).

So I built the fixture anyway to be sure — hand-written sheet XML, header in A1:C1, data row with `<c r="A2">` omitted, under three dimensions (`A1:C2`, `B1:C2`, none at all). All three read identically through `sheet.rows` and through `iter_rows(min_col=1)`: `[['id','importo','note'], [None,'100','x']]`. `_get_row` places each value at `cell['column'] - min_col` and fills the gaps with `EmptyCell`, so a row that omits leading cells is already positionally correct. No committed file can distinguish the two, which is also why the padding cells were there to break `cell.column` in the first place.

`8ca5ea5d9` therefore takes `_sheetlines` back to what `develop` has, and keeps the finding in the docstring where the next reader will meet it. The shift is entirely `_process_firstline`, exactly as you concluded, and the two header tests pin it. Nothing is left unpinned.

**The description described the abandoned solution**

Rewritten. It now says slugify-before-the-empty-check, states that `_sheetlines` needs no change and why, and documents `assign_pkey`. Happy to correct #797 §1a and the `gnrlist.py` path the same way — say the word and I will, though I would rather add a comment than rewrite the original report, since your review and this discussion both quote it.

**Minor**

- Docstring of `test_XlsxReader_header_slugifying_to_empty_is_renamed` fixed: `'#'` slugifies to `''` and is renamed, `'---'` slugifies to `'_'` and is kept, which is what the assertions say.
- Duplicate names from two punctuation-only headers: noted in that docstring as a known limitation of naming columns after slugify. It behaves the same on `develop` — a header made only of punctuation was truthy there too, so it was never renamed — so I left it out of this PR rather than widen it. Worth its own issue if you agree.
- Comment added on the `_updater_keyfield` filter, saying that `update_only` pops it before the generator is consumed and that the filter earns its place in the other paths. The passing test on `develop` is precisely `test_update_only_updates_by_business_key`, for that reason.

**Process**

The PR is assigned now. Checks: `pytest tests --ignore=tests/sql` → 1264 passed; `tests/core tests/web` → 1036 passed, `gnrbag_test.py::test_fillFromUrl` included this time; flake8 clean on the four files.

fporcari (approved):
The blocking point is closed, and closed at the right level. Making the pkey assignment an explicit `assign_pkey` parameter rather than special-casing the update branch means the rule lives where the decision is made, and defaulting it to `sql_mode` keeps every other caller exactly as it was. Approving.

I re-derived the mechanism rather than taking it from your reply. `raw_update` builds the WHERE from `old_record`'s pkeys and the SET from every key of `record_data` (`_gnrbaseadapter.py:865-876`), so a pkey carried by the row genuinely lands in the SET against a WHERE holding the old one — `SET pkey=<generated> WHERE pkey=<current>`, raw, triggerless. That is the corruption, and `assign_pkey=bool(sql_mode) and import_mode!='update_only'` at the call site removes its only source.

The `_updater_keyfield` filter is right for a reason worth writing down, because it looks redundant: the update path pops the entry from `match_index` *before* consuming the generator, so there the filter is a no-op — it earns its place on the paths that never pop, which is where the `KeyError` was.

And `_process_firstline`: slugifying before the empty check restores the invariant that every position in `headers` is in `colindex`. That is the whole of #797, and it is in the one place that owns it.

**The ratchet, with a distinction that matters.** 52 passed on `tests/web/table_importer_test.py` + `tests/core/flatfiles_test.py` at 8ca5ea5d9a. Restoring `utils.py` and `flatfiles.py` from develop, 6 of the new tests fail — but `test_update_only_sql_mode_keeps_the_pkey` fails there with `KeyError: '_updater_keyfield'`, not with the rewrite, because develop never reaches the update path at all. The rewrite reproduces against **8cc680487d**, the head I reviewed, where the branches were already reordered and `assign_pkey` did not exist yet:

```
assert record['id'] == self.record['id']
AssertionError: assert '0CijU_14OhuNLimLjIX2aA' == 'bUz1WeqLNWSz0v2z0DFLjA'
```

Worth stating because "6 red on develop" on its own would imply the test pins the rewrite against develop, and it does not — it pins it against the state this PR passed through. Which is the honest and the more useful thing for it to pin.

**One note, not an objection.** `adaptedRecords` has a second caller: `tableImporterRun` at `utils.py:323`, the branch that hands the rows to a per-table `importer` named in `importerStructure`. It does not pass `assign_pkey`, so it keeps the default `sql_mode` — correct, since changing it would alter behaviour this PR has no business altering. But that means a project importer which *updates* rather than inserts, with `sql_mode` on, still gets a generated pkey and still has the hazard you just fixed here. The parameter now gives it the way out; nothing tells its authors that it needs one. Worth an issue rather than a line in this diff.

Not verified: no run against PostgreSQL — the fixture is the sqlite `gnrtest` instance — and no pass through the actual import UI, so the client side that sets `_updater_keyfield` from the ticked `is_key` column is read from `genro_components.js:1522`, not observed.

PR #800, opened 2026-04-16
Pr attachment filename

dedalus2000:
A batch returns a file path, allowing Genropy to serve it as a download.
Unfortunately, when the path contained special characters (a comma?), it stopped working.

I found that the filename value in the Content-Disposition header was missing quotes:

```
filename=name.pdf  => wrong
filename="name.pdf"  => ok
```

I also added UTF-8 encoding support:
```
filename*=UTF-8''name.pdf
```

Also added ```urllib.parse.quote``` for non-ASCII characters. 
Note that ```.quote``` has the argument ```safe=''``` to escape the slash, but I can assume that Genropy only passes the filename to it


cgabriel (approved):
I've just cleaned up some formatting and spacing.

PR #803, opened 2026-04-16
feat(adm): add api_token table for Bearer authentication

### LINKED ISSUE #802: Add API bearer token table to adm package
## Summary

Add a persistent API token table to the `adm` package for machine-to-machine authentication via Bearer tokens. This fills a gap in GenroPy's auth system which currently only supports human users via browser sessions.

## Motivation

The `data_api` package (gnrdbextra) exposes OData/GraphQL endpoints that external clients (Power BI, Excel, partner integrations) need to access. These clients cannot use browser-based login — they need Bearer tokens.

The same mechanism will be useful for any future API endpoint, webhook receiver, or external integration.

## Two authentication channels

| Who | How | Result |
|-----|-----|--------|
| Internal user (employee, analyst) | OIDC or username/password → existing Avatar | auth_tags from avatar |
| External client (Power BI, partner) | `Authorization: Bearer <token>` | auth_tags from token record |

Both produce the same output: a set of **auth_tags** that downstream code uses for authorization.

## Table: `adm.api_token`

```
adm.api_token
├── id              (pkey, size='22')
├── token           (A, size='64', unique, indexed) — the Bearer value
├── description     (A) — "Power BI production", "Partner XYZ"
├── auth_tags       (A) — comma-separated tags (e.g. "read_crm,read_orders")
├── is_active       (B, default=True) — enable/disable without deleting
├── created_ts      (DHZ) — creation timestamp
├── expires_ts      (DHZ, nullable) — optional expiry, null = never expires
├── last_used_ts    (DHZ, nullable) — updated on each use
├── created_by      (A, size='22') — FK to adm.user who created it
├── notes           (T) — free text notes
```

## Token generation

- Token value should be a secure random string (e.g. `secrets.token_urlsafe(48)`)
- Token is shown ONCE at creation time, then only the last 4 chars are visible
- No way to recover a lost token — revoke and create a new one

## Validation method

Add to the adm package a method that other packages can call:

```python
# In adm package main.py or lib/
def validate_bearer_token(self, token_value):
    """Validate a Bearer token and return auth info.
    
    Returns:
        dict with 'auth_tags', 'description', 'token_id' if valid
        None if invalid, expired, or inactive
    """
    record = self.db.table('adm.api_token').record(
        where='$token=:t AND $is_active=:a', t=token_value, a=True
    )
    if not record:
        return None
    if record['expires_ts'] and record['expires_ts'] < now():
        return None
    # Update last_used_ts
    ...
    return {'auth_tags': record['auth_tags'], ...}
```

### Table Handler
- `resources/tables/api_token/th_api_token.py` — standard TH with:
  - View: description, auth_tags, is_active, created_ts, expires_ts, last_used_ts
  - Form: all fields, token shown only on creation
  - Generate token button
  - Revoke (set is_active=False) button

### Menu entry
- Add under Administration menu: "API Tokens" (tags='admin')

### Preference (optional)
- `adm.bearer_auth_enabled` (Boolean) — global on/off for Bearer auth

## Security notes

- Tokens must be stored as-is (not hashed) because validation needs exact match for O(1) lookup. If hashing is desired, use indexed hash column + verify.
- Token creation requires admin role
- Expired tokens should be auto-cleaned periodically (optional batch)
- Rate limiting on failed token lookups (optional, future)

## Branch

`feature/bearer_token` from `develop`

## Related

- genropy/genropy#801 — data_api adapter
- genropy/genro-data-api#1 — OData/GraphQL protocol
- softwellsrl/gnrdbextra `packages/data_api/` — consumer of this auth mechanism

genro:
## Summary

Adds `adm.api_token` table for persistent API Bearer token authentication. Enables machine-to-machine access (Power BI, Excel, partner integrations) alongside existing user-based auth.

Closes #802

## Schema

**`adm.api_token`** (model/api_token.py):
- `token` (unique, indexed) — **SHA-256 hash** of the secret; the plain-text token is returned only once at creation and never persisted
- `description` — human-readable name
- `group_code` (FK adm.group, mode=foreignkey) — group inherited by the token
- `is_active` (bool, default True) — enable/disable without deleting
- `expires_ts` (DHZ) — optional expiry
- `last_used_ts` (DHZ) — updated by `validate_token` (via deferToCommit)
- `created_by` (FK adm.user.id, onDelete=setnull) — audit FK; user deletion preserves the record
- `notes` (text) — free-form notes
- `token_hint` — last 4 chars (`...XXXX`), shown in lists for identification
- `all_tags` (formulaColumn, dtype='A') — comma-joined authorization tags from `api_token_tag` M2M

**`adm.api_token_tag`** (model/api_token_tag.py): M2M between `api_token` and `htag` (authorization tag), allowing multiple tags per token without comma-separated strings.

## Flow

- **Create**: admin clicks "Create New Token", enters a description; `create_api_token()` generates `secrets.token_urlsafe(48)`, persists its **SHA-256 hash**, and returns the full plain-text token + record id. A modal dialog shows the full token **once** (not stored on the form record). `token_hint` is the only thing kept visible afterwards.
- **Disable**: toggle `is_active=False` on the form (no separate "Revoke" button).
- **Validate**: `Package.validate_bearer_token(token_value)` → `adm.api_token.validate_token(token_value)`. Returns `{token_id, auth_tags, group_code, description}` if active and not expired, else `None`. `last_used_ts` is updated via `deferToCommit` so the write piggy-backs on the request's own commit.

## How other packages use it

\`\`\`python
auth = self.db.package('adm').validate_bearer_token(bearer_token)
if not auth:
    return forbidden()
# auth['auth_tags'], auth['group_code'], auth['token_id'] available
\`\`\`

## UI placement

- Menu entry: Administration > Users setup > "API Tokens" (\`tags='admin'\`)
- View: list with description, group, active flag, token hint, tags, created_by, expiry, last used
- Form: editable description / group / is_active / expires_ts / notes + inline \`api_token_tag\` table for tag assignment
- "Create New Token" button on the View bottom toolbar — RPC subscriber and result dialog live on the View (always mounted, work also with empty list)
- \`generateApiToken\` is gated with \`tags='admin'\`

## Test plan

- [ ] \`gnr db migrate\` creates \`api_token\` and \`api_token_tag\` tables
- [ ] Open \`/adm/api_token\` as admin; "Create New Token" with empty list shows the dialog with the token
- [ ] Same with non-empty list still works
- [ ] Toggle \`is_active=False\` → \`validate_bearer_token\` returns None
- [ ] Set \`expires_ts\` in the past → returns None
- [ ] Try to call \`generateApiToken\` as a non-admin user → permission denied
- [ ] Validate that DB writes from \`last_used_ts\` happen at request commit time, not synchronously

### DIFF
--- a/projects/gnrcore/packages/adm/main.py
+++ b/projects/gnrcore/packages/adm/main.py
@@ -88,6 +88,10 @@ class Package(GnrDboPackage):
     def onExternalUser(self,externalUser=None):
         self.db.table('adm.user').syncExternalUser(externalUser)
         
+    def validate_bearer_token(self, token_value):
+        """Validate a Bearer token. Called by other packages (e.g. data_api)."""
+        return self.db.table('adm.api_token').validate_token(token_value)
+
     def newUserUrl(self):
         return 'adm/new_user'
 
--- a/projects/gnrcore/packages/adm/menu.py
+++ b/projects/gnrcore/packages/adm/menu.py
@@ -29,6 +29,7 @@ class Menu(object):
         root.thpage(u"!!Access groups", table="adm.access_group",tags='_DEV_,superadmin')
         root.thpage(u"!!Help documents", table="adm.group_helpdoc")
         root.webpage('!!User preferences',filepath='/adm/user_preference')
+        root.thpage(u"!!API Tokens", table="adm.api_token", tags="admin")
 
     def utilitySubmenu(self,utility,**kwargs):
         utility.thpage(u"!!Letterheads", table="adm.htmltemplate")
--- a/projects/gnrcore/packages/adm/model/api_token.py
+++ b/projects/gnrcore/packages/adm/model/api_token.py
@@ -15,6 +15,8 @@ class Table(object):
         tbl.column('token', size='64', unique=True, indexed=True,
                    name_long='!!Token', _sendback=True)
         tbl.column('description', name_long='!!Description')
+        tbl.column('group_code', size=':15', name_long='!!Group').relation(
+                   'group.code', relation_name='api_tokens', mode='foreignkey')
         tbl.column('is_active', dtype='B', default=True, name_long='!!Active')
         tbl.column('expires_ts', dtype='DHZ', name_long='!!Expires')
         tbl.column('last_used_ts', dtype='DHZ', name_long='!!Last Used')
@@ -61,7 +63,7 @@ class Table(object):
         Returns None if invalid, expired, or inactive.
         """
         records = self.query(
-            columns='$id,$description,$expires_ts,$is_active,$all_tags',
+            columns='$id,$description,$expires_ts,$is_active,$all_tags,$group_code',
             where='$token=:t AND $is_active=:a',
             t=token_value, a=True
         ).fetch()
@@ -76,5 +78,6 @@ class Table(object):
         return {
             'token_id': record['id'],
             'auth_tags': record.get('all_tags', ''),
+            'group_code': record.get('group_code', ''),
             'description': record['description']
         }
--- a/projects/gnrcore/packages/adm/model/api_token.py
+++ b/projects/gnrcore/packages/adm/model/api_token.py
@@ -1,0 +1,76 @@
+# encoding: utf-8
+
+import secrets
+from datetime import datetime, timezone
+from gnr.core.gnrdecorator import public_method
+
+
+class Table(object):
+    def config_db(self, pkg):
+        tbl = pkg.table('api_token', pkey='id',
+                        name_long='!!API Token',
+                        name_plural='!!API Tokens')
+        self.sysFields(tbl)
+        tbl.column('token', size='64', unique=True, indexed=True,
+                   name_long='!!Token', _sendback=True)
+        tbl.column('description', name_long='!!Description')
+        tbl.column('auth_tags', name_long='!!Auth Tags')
+        tbl.column('is_active', dtype='B', default=True, name_long='!!Active')
+        tbl.column('expires_ts', dtype='DHZ', name_long='!!Expires')
+        tbl.column('last_used_ts', dtype='DHZ', name_long='!!Last Used')
+        tbl.column('created_by', size='22', group='_',
+                   name_long='!!Created By').relation('adm.user.id',
+                   relation_name='api_tokens', onDelete='cascade')
+        tbl.column('notes', dtype='T', name_long='!!Notes')
+        tbl.column('token_hint', size='8', name_long='!!Token Hint')
+
+    def generate_token(self):
+        """Generate a new secure token and return the full value."""
+        return secrets.token_urlsafe(48)
+
+    @public_method
+    def create_api_token(self, description=None, auth_tags=None,
+                         expires_ts=None, notes=None, created_by=None):
+        """Create a new API token. Returns (record_id, full_token).
+
+        The full token is returned ONLY at creation time.
+        """
+        token_value = self.generate_token()
+        record = self.newrecord(
+            token=token_value,
+            description=description or '',
+            auth_tags=auth_tags or '',
+            is_active=True,
+            expires_ts=expires_ts,
+            notes=notes or '',
+            created_by=created_by,
+            token_hint='...' + token_value[-4:]
+        )
+        self.insert(record)
+        self.db.commit()
+        return record['id'], token_value
+
+    def validate_token(self, token_value):
+        """Validate a Bearer token.
+
+        Returns dict with auth_tags, description, token_id if valid.
+        Returns None if invalid, expired, or inactive.
+        """
+        records = self.query(
+            columns='$id,$auth_tags,$description,$expires_ts,$is_active',
+            where='$token=:t AND $is_active=:a',
+            t=token_value, a=True
+        ).fetch()
+        if not records:
+            return None
+        record = records[0]
+        if record['expires_ts'] and record['expires_ts'] < datetime.now(timezone.utc):
+            return None
+        self.raw_update(record={'id': record['id'],
+                                'last_used_ts': datetime.now(timezone.utc)})
+        self.db.commit()
+        return {
+            'token_id': record['id'],
+            'auth_tags': record['auth_tags'],
+            'description': record['description']
+        }
--- a/projects/gnrcore/packages/adm/model/api_token.py
+++ b/projects/gnrcore/packages/adm/model/api_token.py
@@ -13,7 +13,7 @@ class Table(object):
                         rowcaption='$description')
         self.sysFields(tbl)
         tbl.column('token', size='64', unique=True, indexed=True,
-                   name_long='!!Token', _sendback=True)
+                   name_long='!!Token')
         tbl.column('description', name_long='!!Description')
         tbl.column('group_code', size=':15', name_long='!!Group').relation(
                    'group.code', relation_name='api_tokens', mode='foreignkey')
--- a/projects/gnrcore/packages/adm/model/api_token.py
+++ b/projects/gnrcore/packages/adm/model/api_token.py
@@ -9,12 +9,12 @@ class Table(object):
     def config_db(self, pkg):
         tbl = pkg.table('api_token', pkey='id',
                         name_long='!!API Token',
-                        name_plural='!!API Tokens')
+                        name_plural='!!API Tokens',
+                        rowcaption='$description')
         self.sysFields(tbl)
         tbl.column('token', size='64', unique=True, indexed=True,
                    name_long='!!Token', _sendback=True)
         tbl.column('description', name_long='!!Description')
-        tbl.column('auth_tags', name_long='!!Auth Tags')
         tbl.column('is_active', dtype='B', default=True, name_long='!!Active')
         tbl.column('expires_ts', dtype='DHZ', name_long='!!Expires')
         tbl.column('last_used_ts', dtype='DHZ', name_long='!!Last Used')
@@ -23,14 +23,19 @@ class Table(object):
                    relation_name='api_tokens', onDelete='cascade')
         tbl.column('notes', dtype='T', name_long='!!Notes')
         tbl.column('token_hint', size='8', name_long='!!Token Hint')
+        tbl.formulaColumn('all_tags',
+                          select=dict(table='adm.api_token_tag',
+                                      columns='@tag_id.authorization_tag',
+                                      where='$api_token_id=#THIS.id'),
+                          dtype='A', name_long='!!Tags')
 
     def generate_token(self):
         """Generate a new secure token and return the full value."""
         return secrets.token_urlsafe(48)
 
     @public_method
-    def create_api_token(self, description=None, auth_tags=None,
-                         expires_ts=None, notes=None, created_by=None):
+    def create_api_token(self, description=None, expires_ts=None,
+                         notes=None, created_by=None):
         """Create a new API token. Returns (record_id, full_token).
 
         The full token is returned ONLY at creation time.
@@ -39,7 +44,6 @@ class Table(object):
         record = self.newrecord(
             token=token_value,
             description=description or '',
-            auth_tags=auth_tags or '',
             is_active=True,
             expires_ts=expires_ts,
             notes=notes or '',
@@ -57,7 +61,7 @@ class Table(object):
         Returns None if invalid, expired, or inactive.
         """
         records = self.query(
-            columns='$id,$auth_tags,$description,$expires_ts,$is_active',
+            columns='$id,$description,$expires_ts,$is_active,$all_tags',
             where='$token=:t AND $is_active=:a',
             t=token_value, a=True
         ).fetch()
@@ -71,6 +75,6 @@ class Table(object):
         self.db.commit()
         return {
             'token_id': record['id'],
-            'auth_tags': record['auth_tags'],
+            'auth_tags': record.get('all_tags', ''),
             'description': record['description']
         }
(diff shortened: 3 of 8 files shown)


fporcari (dismissed):
Good implementation, clean model design that reuses existing htag/group infrastructure correctly.

Two things to fix before merge:

1. **Remove `_sendback=True` from the `token` column** (api_token.py line 14). This causes the full token to be sent to the client on every form load. The token should only be returned once at creation time via `create_api_token` — after that, only `token_hint` should be visible.

2. **Remove test_invoice changes** (instanceconfig.xml and menu.py). The `data_api` package addition is unrelated to this PR.

Minor note for future consideration: `raw_update` + `commit` on every `validate_token` call could become a bottleneck under high traffic. Fine for now, but worth keeping in mind for async/batch update of `last_used_ts` later.

fporcari (dismissed):
Solid foundation and the M2M `htag` integration is a clear improvement over the comma-separated `auth_tags` column proposed in #802. A few items need to be addressed before merge.

## Blocking

### 1. PR description doesn't match the implementation
The description still refers to the design from #802 (`auth_tags` column, "Revoke" button) but the actual code:
- has a `all_tags` formulaColumn fed by a new `api_token_tag` M2M table (not mentioned)
- has no Revoke button — `is_active` is just a checkbox on the form
- adds `group_code` and `notes` columns not listed in the description

Please rewrite the description so downstream consumers (data_api, future readers) know what the schema actually looks like.

### 2. "Create New Token" flow is fragile
The button lives in `th_bottom_custom` (view toolbar) and publishes `generateApiToken`. The subscriber (`dataRpc subscribe_generateApiToken=True`) and the result dialog are inside `th_form`, which is only mounted when a record is open. The dialog's `connect_show` also reads `#FORM.newTokenResult.token`, which depends on the form datapath being live.

Please verify in the browser that creating a token from an empty list works. If it doesn't, the RPC + dialog need to live in a component that's always mounted (e.g. `th_top_custom` or page-level), not inside `th_form`.

### 3. `validate_token` calls `self.db.commit()` on every API request
```python
self.raw_update(record={'id': record['id'], 'last_used_ts': datetime.now(timezone.utc)})
self.db.commit()
```
This runs on every request that uses Bearer auth. Two problems:
- it commits whatever pending state the caller had, with no visibility for the caller
- it serializes on the token row under load

Use `deferToCommit` (see how `user_tag.setUserAuthTags` does it) so the update piggy-backs on the request's own commit.

### 4. Timezone-aware vs naive comparison on `expires_ts`
```python
if record['expires_ts'] and record['expires_ts'] < datetime.now(timezone.utc):
```
DHZ values can come back naive depending on DB config. If they do, this raises `TypeError: can't compare offset-naive and offset-aware datetimes` on every validation that has an expiry set. Please confirm empirically what DHZ returns in this project and normalize accordingly.

## Should fix before merge

### 5. `created_by` with `onDelete='cascade'`
Deleting a user wipes the audit trail of every token they created. `SET NULL` is the right choice for an audit-only FK.

### 6. `@public_method generateApiToken` is not permission-gated
Menu visibility (`tags='admin'`) only hides the entry — the RPC endpoint is reachable by any authenticated user who knows the path. Add `tags='admin'` (or equivalent) on the `public_method` itself, otherwise non-admin users can mint tokens.

### 7. No automated tests
This is security-sensitive code (token creation, expiry check, revoke). Even minimal tests covering valid / expired / inactive / unknown tokens would be appropriate. The current test plan is fully manual.

## Minor

- `validate_token` uses `query(...).fetch()[0]` — `record(where=...)` (as in #802) is more idiomatic for a single-record lookup.
- `formulaColumn('all_tags', select=dict(...))` with `dtype='A'`: please confirm the join produces the comma-joined string for multiple tags, not just the first row. If not, an aggregate is needed.
- `relation('group.code', ...)` works in-package, but `'adm.group.code'` is safer.
- `'...' + token_value[-4:]` → f-string is a touch cleaner.

fporcari (changes requested):
Solid foundation, M2M with `htag` is a good improvement over the comma-separated `auth_tags` from #802. Four blocking items before merge.

### 1. PR description doesn't match the implementation
The description still describes the #802 design (`auth_tags` column, "Revoke" button), but the actual code:
- has an `all_tags` formulaColumn fed by a new `api_token_tag` M2M table (not mentioned)
- has no Revoke button — `is_active` is just a checkbox on the form
- adds `group_code` and `notes` columns not listed

Please rewrite the description so downstream consumers (data_api, future readers) know what the schema actually is.

### 2. "Create New Token" flow is fragile
The button is in `th_bottom_custom` (view toolbar) and publishes `generateApiToken`. The subscriber (`dataRpc subscribe_generateApiToken=True`) and the result dialog live inside `th_form`, which is only mounted when a record is open. The dialog's `connect_show` also reads `#FORM.newTokenResult.token`, which depends on the form datapath being live.

Please verify in the browser that creating a token from an empty list works. If it doesn't, RPC + dialog need to live in an always-mounted component (e.g. `th_top_custom` / page-level), not inside `th_form`.

### 3. `validate_token` calls `self.db.commit()` on every API request
```python
self.raw_update(record={'id': record['id'], 'last_used_ts': datetime.now(timezone.utc)})
self.db.commit()
```
This runs on every Bearer-authenticated request. It commits any pending state the caller had (silent side effect) and serializes on the token row under load. Use `deferToCommit` (see `user_tag.setUserAuthTags`) so the update piggy-backs on the request's own commit.

### 4. Timezone-aware vs naive comparison on `expires_ts`
```python
if record['expires_ts'] and record['expires_ts'] < datetime.now(timezone.utc):
```
DHZ can return naive datetimes depending on DB config. If it does, this raises `TypeError: can't compare offset-naive and offset-aware datetimes` on every validation with an expiry set. Please confirm empirically and normalize.

genro:
@fporcari thanks for the thorough review. The 4 blocking points and the 2 "should fix" ones are addressed in commit `2fc137ea`:

1. **PR description** — rewritten to match the actual implementation (`api_token_tag` M2M, `all_tags` formulaColumn, `is_active` checkbox instead of Revoke, `group_code` and `notes` columns documented).
2. **Create New Token flow** — `dataRpc` subscriber, result dialog and `generateApiToken` RPC moved from `Form` to `View.th_top_custom`, so they are always mounted and the flow works with an empty list.
3. **`validate_token` synchronous commit** — replaced `raw_update` + `db.commit()` with `self.db.deferToCommit(self.raw_update, ..., _deferredId=record['id'])`, piggy-backing on the caller's commit.
4. **`expires_ts` tz-aware vs naive** — normalize to UTC when `tzinfo is None` before comparing with `datetime.now(timezone.utc)`.
5. **`created_by` cascade** — switched to `onDelete='setnull'` to preserve the audit trail when a user is deleted.
6. **`generateApiToken` permission gate** — added `tags='admin'` on the `@public_method` decorator so the RPC endpoint itself is gated, not just the menu entry.

**Minor items** — partially addressed in `8567be7c`:

- `token_hint` now uses an f-string (`8567be7c`).
- `query(...).fetch()[0]` → `record(where=...)`: left as-is. Switching would require passing `virtual_columns='\$all_tags'` (formulaColumns aren't loaded by default through `record()`) plus `mode='dict'` + `ignoreMissing=True` to keep the current return-shape, so it's not a clean cosmetic swap on the most security-sensitive function of this PR. Kept the explicit `query()` to avoid silent behavior changes.
- `relation('group.code', ...)` → `'adm.group.code'`: left as-is for consistency. The other 5 occurrences in `adm` (`user.py`, `user_tag.py`, `user_config.py`, `user_group.py`, `group_helpdoc.py`) all use the in-package short form; changing it only here would diverge from the package convention.

**Point #7 (automated tests)** — not addressed in this PR. No `gnrcore` package currently ships pytest tests (automated tests live only in `gnrpy/tests/`), so defining a test convention for `gnrcore/packages/*` is a discussion of its own. I'd rather not invent a one-off pattern just for `api_token`; happy to follow up once the convention is set.

The `formulaColumn('all_tags', dtype='A')` behavior with multiple tags is worth confirming empirically — works in my local check but the test plan in the PR description covers it manually.

Could you take another look when you have a moment? CI is green on all 4 Python versions.

fporcari:
Thanks for the round of fixes — the four blocking points and the two should-fix items from the previous review are properly addressed. A few new items came out from a closer pass on the model file.

## 1. Token storage at rest (security)

`token` is persisted as plaintext in `adm.api_token.token` (`api_token.py` line 14). A DB dump — backup theft, replica leak, SQL audit log, debugger session, even a routine support archive — hands an attacker total API access. `adm.user` already hashes passwords (`user.py:102-103`); we shouldn't introduce a new persistent-secret column going the other way.

The encrypted-fields PR doesn't really solve this case: for Bearer tokens **issued by us**, we never need to read the plaintext back — the client sends it, we transform and compare. So **hash, not encrypt**.

Concrete proposal — ~5 lines, no data migration since the table is new:

```python
import hashlib

def _hash_token(self, token_value):
    return hashlib.sha256(token_value.encode('utf-8')).hexdigest()  # 64 char hex
```

- Rename column `token` → `token_hash` (size `64`, unique, indexed) — same shape, different semantics
- In `create_api_token`: persist `self._hash_token(token_value)`, return the plaintext `token_value` to the caller (shown once in the dialog, as today)
- In `validate_token`: lookup with `where='$token_hash=:h'`, `h=self._hash_token(token_value)`

No salt needed: with ~256 bits of entropy from `secrets.token_urlsafe(48)`, the input is already "salted" by entropy — rainbow tables don't apply. No bcrypt/argon2 needed either: those are slow on purpose to defend low-entropy human passwords, here they would only add ~100 ms of CPU per Bearer-authenticated request for zero security gain.

SHA-256 is what GitHub, Stripe and AWS use for the same use case. MD5 would also be technically sufficient on entropy grounds, but it's a flag at every security audit and SHA-256 costs the same.

## 2. `expires_ts` tz normalization is defensive programming (`api_token.py:73-76`)

```python
# DHZ may return a naive datetime depending on DB config; normalize to UTC.
if expires_ts.tzinfo is None:
    expires_ts = expires_ts.replace(tzinfo=timezone.utc)
```

The comment itself admits we don't know what DHZ guarantees. One of the two:

- If DHZ is guaranteed to return tz-aware, this is a dead branch and should be removed — the function becomes a single comparison.
- If DHZ can actually return naive, the bug is in the DHZ adapter, not here. Patching it locally spreads the same workaround across every future call site that compares DHZ with `datetime.now(tz=...)`.

The current shape — silent normalize with a dubitative comment — is the worst option: the DHZ bug stays hidden and we have a local patch that contradicts the project's "no defensive programming" rule.

If we want a belt-and-suspenders guard, at least make it loud:

```python
assert expires_ts.tzinfo is not None, "DHZ returned naive datetime — fix the adapter"
```

so the upstream bug surfaces instead of being absorbed.

## 3. `create_api_token` still calls `self.db.commit()` (`api_token.py:55`)

```python
self.insert(record)
self.db.commit()
return record['id'], token_value
```

`validate_token` correctly moved away from synchronous commit via `deferToCommit` — the same reasoning applies symmetrically here:

- the commit silently flushes whatever pending state the caller had (hidden side effect)
- it anticipates the commit relative to the `@public_method` RPC boundary

Worst-case comparison:

- **With explicit commit (current)**: if the response fails to reach the client (serialization error, timeout, browser closed), the token is committed in DB but the user doesn't know it exists → orphan token in DB, only removable by admin.
- **Without explicit commit (proposed)**: either the user sees the token AND it's in DB, or rollback and nothing. The user retries, no residue.

The second is strictly better. Suggest removing `self.db.commit()` and letting the `@public_method` RPC boundary commit on successful return.

---

Once these three are in, this is ready to ship. Nothing else is blocking.

genro:
@fporcari new commit `a4b3564d` addresses a security gap that the review didn't catch (and that I missed earlier too): tokens were stored in plain-text in `adm.api_token.token`. Anyone with read access to the table (DB backup, replica, SQL injection) had every Bearer token ready to use.

The fix:

- `create_api_token` persists `SHA-256(token_value)` instead of the plain token.
- Plain-text token is returned to the caller exactly once at creation and never written to DB.
- `validate_token` hashes the incoming Bearer token before the equality lookup.
- No salt: `secrets.token_urlsafe(48)` already carries 384 bits of entropy, rainbow tables are not a threat, and deterministic hashing keeps `validate_token` at O(1) per request.
- Column size unchanged: 64 chars matches SHA-256 hex length exactly, no schema change.

I considered using `encrypted='X'` from #748 instead, but that mode adds a per-row random salt (`$X$<salt>$<hash>`), which forces `validate_token` to scan all active tokens and call `verify()` on each — O(n) per authenticated request. For Bearer tokens the standard practice (GitHub / Stripe / AWS) is salt-less SHA-256: lookup O(1), and the salt would protect nothing since the input space is already random.

Could you take another look? CI is green.

genro:
@fporcari commit `22df756e` addresses the remaining two points from your review:

**#2 (`expires_ts` tz normalization)** — branch removed. DHZ is treated as tz-aware; if it ever returns a naive datetime, the comparison will raise loudly and the bug surfaces in the DHZ adapter rather than being absorbed here. Folded into a single inline check:

```python
expires_ts = record['expires_ts']
if expires_ts and expires_ts < datetime.now(timezone.utc):
    return None
```

**#3 (`create_api_token` explicit commit)** — `self.db.commit()` removed between `insert()` and `return`. Symmetric to the `deferToCommit` fix already applied to `validate_token`: the `@public_method` RPC boundary handles the commit, the caller's pending state is no longer flushed silently, and the orphan-token scenario (response lost after explicit commit) goes away.

CI is green on all 4 Python versions. Per our convention I'm not re-requesting review automatically — ping me when you've had a chance to look.

fporcari (approved):
Both fixes land cleanly — thanks. My open points are addressed.

Side note: any code comparing a DHZ field with `datetime.now(timezone.utc)` (including `validate_token`) currently raises `TypeError` on SQLite, because the SQLite adapter strips tz on write (`gnrsqlite.py:562-564`). That's an adapter bug — DHZ's contract is tz-aware, the consumer is right to trust it. Opening a separate issue against the adapter.

PR #804, opened 2026-04-16
check self.page.wsk to not be None before invoking its methods.

cgabriel:
self.site.wsk (used as a default when 'asyncServer' is not defined) is None, so invoking methods on it causes errors.

### DIFF
--- a/gnrpy/gnr/web/gnrwebpage_proxy/gnrpdb.py
+++ b/gnrpy/gnr/web/gnrwebpage_proxy/gnrpdb.py
@@ -45,7 +45,7 @@ class GnrPdbClient(GnrBaseProxy):
             debugger_page_id = connectionStore.getItem('_dev.gnride_page_id')
         except Exception as e:
             pass
-        if not debugger_page_id or self.page.page_id==debugger_page_id:
+        if not debugger_page_id or self.page.page_id==debugger_page_id or not self.page.wsk:
             return
         breakpoints = connectionStore.getItem('_pdb.breakpoints')
         bp = 0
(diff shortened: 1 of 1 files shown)


genro (dismissed):
The approach of adding `if self.page.wsk:` guards in individual methods is defensive — it hides the real problem instead of preventing it.

If `wsk` is `None`, the remote debugger cannot work at all (no WebSocket = no communication with the debugger page). The correct fix is to prevent `GnrPdb` from being instantiated in the first place.

**Suggested change** — in `GnrPdbClient.onPageStart()`, line 48, add the `wsk` check to the existing early-return condition:

```python
if not debugger_page_id or self.page.page_id==debugger_page_id or not self.page.wsk:
    return
```

This way `GnrPdb` is never created when the WebSocket handler is unavailable, and the scattered guards become unnecessary.

Please replace the current changes with this single-line fix.

cgabriel:
> The approach of adding `if self.page.wsk:` guards in individual methods is defensive — it hides the real problem instead of preventing it.
> 
> If `wsk` is `None`, the remote debugger cannot work at all (no WebSocket = no communication with the debugger page). The correct fix is to prevent `GnrPdb` from being instantiated in the first place.
> 
> **Suggested change** — in `GnrPdbClient.onPageStart()`, line 48, add the `wsk` check to the existing early-return condition:
> 
> ```python
> if not debugger_page_id or self.page.page_id==debugger_page_id or not self.page.wsk:
>     return
> ```
> 
> This way `GnrPdb` is never created when the WebSocket handler is unavailable, and the scattered guards become unnecessary.
> 
> Please replace the current changes with this single-line fix.

The changes have been updated as you requested.

genro (approved):
Fix applied as requested — the `wsk` check in `onPageStart` prevents `GnrPdb` from being instantiated when WebSocket is unavailable. Clean and minimal. LGTM.

mbertoldi (approved):

PR #806, opened 2026-04-16
fix: initialize rootenv in pageStore at page registration

fporcari:
## Summary

- Initialize `rootenv` as an empty `Bag()` in `_register_new_page` so it is never `None` in the pageStore
- `rpc_main` overwrites it with the full rootenv via `update()` as before

## Problem

When a new page is registered, the pageStore `data` does not include `rootenv`. It only gets populated later in `rpc_main` via `getStartRootenv()`. The `rootenv` property (`gnrwebpage.py:430`) caches the first read — if anything accesses `self.rootenv` before `rpc_main` completes and the store returns `None`, that `None` is cached permanently, causing `TypeError` crashes in consumer code.

## Root cause

`_register_new_page` creates the initial page data with `pageArgs`, `class_info`, `init_info`, `page_info` but no `rootenv`. The gap between page registration and `rpc_main` initialization leaves a window where `rootenv` is `None`.

## Fix

One line added: `data['rootenv'] = Bag()` in `_register_new_page`, ensuring rootenv is always at least an empty Bag in the pageStore.

## Test plan

- [x] flake8 passes
- [x] All 1386 tests pass (260 skipped)
- [ ] Test with anaci webapp: open the socio page when pageStore has no rootenv yet — verify no TypeError

### DIFF
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -312,7 +312,8 @@ class GnrWebPage(GnrBaseWebPage):
             
         self.page_id = page_id or getUuid()
         page_info = dict([(k,getattr(self,k,None)) for k in ATTRIBUTES_SIMPLEWEBPAGE])
-        data = Bag()   
+        data = Bag()
+        data['rootenv'] = Bag()
         data['pageArgs'] = kwargs
         data['class_info'] = class_info
         data['init_info'] = init_info
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -313,6 +313,7 @@ class GnrWebPage(GnrBaseWebPage):
         self.page_id = page_id or getUuid()
         page_info = dict([(k,getattr(self,k,None)) for k in ATTRIBUTES_SIMPLEWEBPAGE])
         data = Bag()
+        data['rootenv'] = Bag()
         data['pageArgs'] = kwargs
         data['class_info'] = class_info
         data['init_info'] = init_info
@@ -429,7 +430,7 @@ class GnrWebPage(GnrBaseWebPage):
         
     @property
     def rootenv(self):
-        if not hasattr(self,'_rootenv') or self._rootenv is None:
+        if not hasattr(self,'_rootenv'):
             self._rootenv = self.pageStore().getItem('rootenv')
         return self._rootenv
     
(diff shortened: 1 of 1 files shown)


cgabriel (commented):
Why not use the proposed solution with the cache invalidation proposed in the original issue?

fporcari:
We evaluated both approaches:

**A) Cache invalidation in the property** (`or self._rootenv is None`): fixes the stale cache but costs an extra daemon read each time rootenv is None.

**B) Initialize rootenv as `Bag()` in `_register_new_page`**: rootenv is never None in the pageStore, so the property always finds a valid value on first read. Zero extra daemon reads. `rpc_main` overwrites it with the full rootenv via `update()` as before.

The current commit implements **B** — the approach that avoids the extra read. The issue's proposed fix (cache invalidation in `anc_cnv_webapp.py`) would be an application-level workaround; this fix addresses the root cause at the framework level so all consumers benefit.

@dgpaci please test with the socio webapp page to confirm the TypeError is resolved.

cgabriel (approved):

dgpaci (approved):
Approved. Tested against the anaci webapp with temporary WARNING loggers instrumented at the three relevant points in `gnrwebpage.py` (`_register_new_page`, the `rootenv` property getter, and `rpc_main`) to trace the pageStore state during the race window the PR closes.

### Test scenarios

Two full sessions executed against the anaci instance:

1. **Login as an administrator user (`dpaci`)** — navigation through the backoffice pages.
2. **Login as a member user (`SE00000`)** — navigation through the member-facing webpages.

Plus the pre-login landing page in both sessions.

### Observations

- **`_register_new_page`** always initializes `data['rootenv']` as an empty `Bag` — confirmed for every registered page.
- **Early reads of `self.rootenv` before `rpc_main` completes** were observed on multiple pages (reading `keys=0` or `keys=1` from the pageStore before the full rootenv is populated). This is exactly the race window this PR targets: previously these would have cached `None` permanently on the page instance.
- **No `None` was ever returned** by `pageStore().getItem('rootenv')` with the fix applied — the instrumented code was set to emit a distinctive WARNING log if it occurred, and it never fired.
- **`rpc_main` populates the rootenv correctly** (observed 11/18/26 keys depending on avatar/context) and all subsequent reads see the fully populated rootenv.

### Conclusion

The one-line initialization in `_register_new_page` successfully closes the race window between page registration and `rpc_main`. The pageStore now always contains a well-formed (possibly empty) `rootenv` Bag, so any consumer accessing `self.rootenv` during that window gets a valid Bag instead of `None`, preventing the `TypeError` crash and the permanent caching of `None`.

No regressions observed in the navigated pages.

PR #808, opened 2026-04-17
fix(#805): restore normal flow for grid edit widgets

### LINKED ISSUE #805: Grid row height collapses during inline cell editing
## Description

When double-clicking to edit a cell in an inline editable grid (`inlineTableHandler`), the row height collapses dramatically (to ~7.77px or even disappears completely). This makes the editing experience unusable as the row becomes nearly invisible while typing.

## Root cause

In `startEdit()` (`genro_wdg.js`), `cellNode.innerHTML = null` clears the cell content to make room for the edit widget. However, the edit widget uses `position: absolute` (via `.widgetInCell` CSS class), so it does not contribute to the cell's height. Since cells have `height: 0px` in CSS (content-driven height), the edited cell collapses. If the remaining cells in the row have little or no content, the entire row collapses.

This was worsened by commit `99cdcc51d` which changed `.widgetInCell` from `top:0; bottom:0; left:0; right:0` (filling the cell) to `top:1px; left:3px; right:3px` without `bottom`, so the widget no longer stretches to fill the cell height.

## Reproduction

Using a simple `inlineTableHandler` with `ViewEditable` struct, three cases demonstrate the issue:

### Case 1 — Two columns with explicit width: OK
```python
def th_struct(self, struct):
    r = struct.view().rows()
    r.fieldcell('tipo_addebito_codice', edit=True, width='20em')
    r.fieldcell('tipo_addebito_codice', edit=True, width='20em')
```
**Result:** Row height is correct during editing. The non-edited cell has enough content to maintain the row height.

Image

### Case 2 — Single column: row disappears
```python
def th_struct(self, struct):
    r = struct.view().rows()
    r.fieldcell('tipo_addebito_codice', edit=True, width='20em')
```
**Result:** On double-click, the row **completely disappears**. The only cell is the one being edited, its content is cleared, and the absolute-positioned widget provides no height.

Image

### Case 3 — Two columns, one without explicit width: row shrinks to ~7.77px
```python
def th_struct(self, struct):
    r = struct.view().rows()
    r.fieldcell('tipo_addebito_codice', edit=True, width='20em')
    r.fieldcell('importo', edit=True)
```
**Result:** On double-click, the row height shrinks to approximately **7.77px**. The `importo` cell without explicit width is too narrow to maintain a proper row height.

Image

## Suggested fix

Preserve the cell height during editing by setting `min-height` on the cell before clearing its content:

**In `startEdit()`** (before `cellNode.innerHTML = null`):
```javascript
cellNode.style.minHeight = cellNode.offsetHeight + 'px';
cellNode.innerHTML = null;
```

**In `endEdit()`** (after restoring content):
```javascript
editingInfo.cellNode.innerHTML = contentText;
editingInfo.cellNode.style.minHeight = '';
```

The `min-height` is also automatically cleaned up when `updateRow()` rebuilds the entire row DOM.

## Affected files

- `gnrjs/gnr_d11/js/genro_wdg.js` — `startEdit()` and `endEdit()` in `gnr.GridEditor`
- `gnrjs/gnr_d20/js/genro_wdg.js` — same (dojo 2.0 copy)

fporcari:
## Summary

- Remove `position: absolute` from `.widgetInCell` — edit widgets now participate in normal document flow, preventing row height collapse when editing cells
- Replace `objectPop(attr, 'width')` with `attr.width = 'auto'` for cleaner inline style management without CSS `!important`
- Add `cell_onStartEdit` hook in widget handler lifecycle for per-widget customization after widget creation
- SimpleTextarea uses `position: fixed` overlay positioned at cell coordinates with scroll-tracking, avoiding the `overflow: hidden` limitation of grid cells

Closes #805

## Test plan

- [ ] Navigate to `/test/components/grid/grid_edit_widgets`
- [ ] Test all widget types (test_0): TextBox, NumberTextBox, DateTextBox, CheckBox, FilteringSelect, SimpleTextarea, Description
- [ ] Test single column editing (test_2): row must not disappear on double-click
- [ ] Test mixed width columns (test_3): row height must stay stable on double-click
- [ ] Test textarea: must float over cell, reposition on grid scroll, close cleanly on blur
- [ ] Test checkbox: must still center correctly in cell
- [ ] Verify narrow columns (test_1): widgets must not overflow

### DIFF
--- a/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
+++ b/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
@@ -57,7 +57,6 @@
     padding: 0px 0px;
 }
 
-._common_d11 .dojoxGrid-cell:has(> .widgetInCell),
 ._common_d11 .dojoxGrid-cell:has(> .gnrcheckbox_wrapper) {
     position: relative;
 }
@@ -80,21 +79,16 @@
 }
 ._common_d11 .widgetInCell {
     color: var(--text-color);
-    position: absolute;
-    top: 1px;
-    left: 3px;
-    right: 3px;
-    width: auto !important;
-    margin: 0 !important;
-    display: flex;
-    align-items: center;
-    justify-content: center;
+    margin: 1px 1px 0 1px;
     padding: 2px 1px;
     border: 1px solid var(--border-light);
     border-radius: var(--radius-sm);
     box-shadow: 1px 1px 2px var(--text-secondary) inset;
     box-sizing: border-box;
     overflow: hidden;
+    display: flex;
+    align-items: center;
+    justify-content: center;
 }
 ._common_d11 .widgetInCell > *{
     flex: 1;
@@ -102,10 +96,16 @@
     min-width: 0;
 }
 
-._common_d11 .widgetInCell:has(.dijitTextArea),
-._common_d11 .widgetInCell.dijitTextArea {
-    bottom: 1px;
-    height: auto !important;
+textarea.cellEditFixed {
+    padding: 2px 1px;
+    border: 1px solid var(--border-light);
+    border-radius: var(--radius-sm);
+    box-shadow: 1px 1px 2px var(--text-secondary) inset;
+    box-sizing: border-box;
+    background: white;
+    outline: none;
+    overflow: auto;
+    resize: none;
 }
 
 /* --- Row colors --- */
--- a/gnrjs/gnr_d11/js/genro_wdg.js
+++ b/gnrjs/gnr_d11/js/genro_wdg.js
@@ -1700,7 +1700,7 @@ dojo.declare("gnr.GridEditor", null, {
         grid.currRenderedRowIndex = lastRenderedRowIndex;
         grid.selection.select(grid.currRenderedRowIndex);
         attr.datapath = this.widgetRootNode.absDatapath('.' + rowLabel);
-        objectPop(attr, 'width'); /* width is managed by CSS via .widgetInCell absolute positioning */
+        attr.width = 'auto';
         /* checkbox centering is handled by CSS via .gnrcheckbox_wrapper flexbox */
         //attr.preventChangeIfIvalid = true;
         if ('value' in attr) {
@@ -1811,6 +1811,10 @@ dojo.declare("gnr.GridEditor", null, {
             editWidgetNode.widget.focus();
         }
         editWidgetNode.grid = gridEditor.grid;
+        var wdghandler = genro.wdg.getHandler(wdgtag);
+        if(wdghandler.cell_onStartEdit){
+            wdghandler.cell_onStartEdit(cellNode,editingInfo,fldDict.attr,editWidgetNode);
+        }
 
     },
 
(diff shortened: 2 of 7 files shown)


genro (commented):
Good fix — the approach of bringing `.widgetInCell` back into normal document flow is clean, and the `cell_onStartEdit` hook is a nice addition to the widget lifecycle.

**Base branch**: this PR targets `master` — should it target `develop` instead?

**Observations:**

1. **Window resize**: `positionTextarea` tracks grid scroll but not window resize. If the browser window is resized while editing a textarea, the overlay will be misaligned. Consider adding a resize listener (same connect/disconnect pattern).

2. **Unused variables**: in `test_2_single_column` and `test_3_mixed_width`, the `frame` variable is assigned but never used — will trigger a linter warning.

genro:
Additional note on the `position: fixed` textarea:

Since the textarea is positioned relative to the viewport (not its container), there are scenarios where it can become misaligned or remain visible when it shouldn't:

1. **Container resize** (splitter drag, collapsible panel, tab resize): the grid moves but `positionTextarea` is only connected to grid scroll — the textarea stays at the old coordinates
2. **Grid hidden** (tab switch, collapsed panel): the grid disappears but the fixed textarea remains visible on screen, since `cell_onDestroying` is only called on edit exit, not on container visibility change

Case 2 is the more problematic one — a floating textarea with no visible grid underneath.

genro:
**Alternative approach to consider**: instead of `position: fixed` on the textarea (which requires scroll/resize tracking and has the visibility issue above), would it be possible to simply set `overflow: visible` on the cell during editing (plus a `z-index` to stay above adjacent cells)?

The textarea would stay in the cell's flow, expand beyond its borders naturally, and follow the grid on scroll, resize, and hide — no listeners needed.

Worth evaluating whether `overflow: visible` on the cell causes visual issues with adjacent cells. If not, it would be a simpler solution.

fporcari:
The `overflow: visible` approach was actually the first thing we tried — we spent hours testing various combinations of `overflow: visible` on the cell, with and without `z-index`, with wrappers, with `position: relative`/`absolute` on parent elements.

The fundamental problem is that `overflow: hidden` on the `<td>` is set by the dojox grid framework itself, and even when overridden with inline styles, the clipping still occurs through the ancestor chain (the `<tr>`, `<table>`, and the grid's scroll container all participate in clipping).

We also tried a wrapper-based approach (similar to how `.gnrcheckbox_wrapper` works) with `position: relative` on the wrapper and `position: absolute` on the textarea — same result, the grid's table layout clips the overflow.

The `position: fixed` overlay was the only approach that reliably escaped the grid's clipping context. The scroll tracking handles the most common case (grid scroll). The container resize and visibility issues you raised are valid edge cases — we can add listeners for those if needed, but they are relatively uncommon during active textarea editing.

fporcari:
To clarify — `position: fixed` was the last resort after exhausting all in-flow alternatives. It was not the first choice.

Regarding window resize and container changes: these are non-issues in practice. When the user interacts with the browser window (resize, splitter drag, tab switch), the textarea loses focus, which triggers the `onBlur` callback that closes the edit widget. The fixed positioning is only active while the user is actively typing in the textarea.

As for grid scroll — which is the one real concern since the user can scroll with the mouse wheel while typing — this is handled by connecting a scroll listener to the grid's scrollbox that repositions the textarea in real time. The textarea follows the cell seamlessly during scroll.

genro (approved):
Thanks for the detailed explanation. The `onBlur` closing the edit widget on focus loss covers the resize/visibility concerns we raised. And the dojox grid clipping chain makes `overflow: visible` unviable — `position: fixed` is the right call here.

LGTM.

Issue #812, opened 2026-04-17
[FEATURE] Empty state indication for tableHandler and query results


dgpaci:
## Summary

Provide a built-in way to signal "query executed, zero results" in `tableHandler` / view grids — e.g. a `messageIfEmpty` attribute, an icon, or a distinctive style — so users can tell an empty result apart from a grid that hasn't loaded yet.

## Problem or Need

Today, when a `tableHandler` (or any view-based grid) returns zero rows, the grid area just renders empty/white. There is no visual cue that:

- the query actually ran,
- the filters/conditions matched nothing,
- this is an expected "no data" state and not a loading/broken widget.

In practice this often leads to confusion: users (and developers during debugging) stare at a blank pane and can't tell whether the grid is still loading, misconfigured, or simply returned zero rows. The same ambiguity exists when writing custom UIs that render query results directly.

A concrete recent example: a convention registration dialog shows a `plainTableHandler` titled "Seleziona l'opzione desiderata". When no quotas match the filter, the title renders but the body is blank — with no indication that the server has already answered with "zero rows". The developer has to add their own `stackContainer` + `dataController` watching `totalrows` to swap in a fallback pane. This is boilerplate that every form ends up reimplementing (often incorrectly).

## Proposed Solution

Add first-class support for an empty-state on `tableHandler` / view / grid widgets. Possible shapes (not mutually exclusive):

1. **`messageIfEmpty` attribute** (preferred, low-effort)
   - String or dict accepted on `tableHandler` / `plainTableHandler` / `inlineTableHandler` / view / grid
   - When the store finishes loading and `totalrows == 0`, render the message in place of (or overlaid on) the empty grid area
   - Dict form could allow `{message, icon, style}` for more customization

2. **Empty-state CSS class**
   - The grid root gets a `gnr-empty-results` (or similar) class when loaded with zero rows, so themes can style it distinctly (dashed border, muted icon, etc.) without each page reimplementing it

3. **Loaded-but-empty state broadcast**
   - A clean path like `.view.store?empty?#v` (or similar) that is `true` only after a successful load with zero rows — distinguishing "still loading" from "loaded, empty". This makes custom controllers trivial:
     ```python
     th.dataController("SET .selectedPage = empty ? 'nessunaQuota' : 'risultati';",
                       empty='^#FORM.data.view.store?empty?#v')
     ```

Ideally (1) would be the default behavior with a sensible default message (localizable), and (2)/(3) available for pages that need custom rendering.

[...]

PR #815, opened 2026-04-17
feat(#812): add empty grid state indicator

genro:
## Summary

- Add `gnr_empty_grid` CSS class toggled on the grid DOM node when `rowCount === 0` in `patch_updateRowCount`
- Show a subtle `no_results.svg` icon (search lens with X) centered in the grid scrollbox as background image when the class is active
- New SVG icon at `resources/common/css_icons/google_icons/no_results.svg`

## Test plan

- [x] Open a tableHandler page (e.g. Products)
- [ ] Search for something that returns zero results → no_results icon visible centered in grid
- [x] Clear the search → icon disappears, data shows normally
- [ ] Initial page load with data → no icon visible
- [ ] d20 not yet updated (CSS and JS changes are d11 only for now)

Refs #812

### DIFF
--- a/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
+++ b/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
@@ -498,3 +498,11 @@
 ._common_d11 .dojoxGrid-scrollbox {
     overflow: auto;
 }
+
+/* --- Empty grid state --- */
+._common_d11 .gnr_empty_grid .dojoxGrid-scrollbox {
+    background-image: url('/_rsrc/common/css_icons/google_icons/no_results.svg');
+    background-repeat: no-repeat;
+    background-position: center center;
+    background-size: 80px 80px;
+}
--- a/gnrjs/gnr_d11/js/genro_components.js
+++ b/gnrjs/gnr_d11/js/genro_components.js
@@ -6616,15 +6616,22 @@ dojo.declare("gnr.stores._Collection",null,{
     runQuery:function(cb,runKwargs){
         var that = this;
         this.runningQuery = true;
-        var result =  this.storeNode.fireNode(runKwargs);
+        var wrappedCb = function(r){
+            cb(r);
+            var empty = r && r.attr && r.attr.totalrows === 0;
+            that.gridBroadcast(function(grid){
+                grid.domNode.classList.toggle('gnr_empty_grid', empty);
+            });
+        };
+        var result = this.storeNode.fireNode(runKwargs);
         if(result instanceof dojo.Deferred){
             result.addCallback(function(r){
                 that.runningQuery = false;
-                cb(r)
+                wrappedCb(r);
             });
         }else{
             that.runningQuery = false;
-            result = cb(result);
+            result = wrappedCb(result);
         }
         return result;
     },
--- a/gnrjs/gnr_d11/js/genro_components.js
+++ b/gnrjs/gnr_d11/js/genro_components.js
@@ -6616,22 +6616,15 @@ dojo.declare("gnr.stores._Collection",null,{
     runQuery:function(cb,runKwargs){
         var that = this;
         this.runningQuery = true;
-        var wrappedCb = function(r){
-            cb(r);
-            var empty = r && r.attr && r.attr.totalrows === 0;
-            that.gridBroadcast(function(grid){
-                grid.domNode.classList.toggle('gnr_empty_grid', empty);
-            });
-        };
-        var result = this.storeNode.fireNode(runKwargs);
+        var result =  this.storeNode.fireNode(runKwargs);
         if(result instanceof dojo.Deferred){
             result.addCallback(function(r){
                 that.runningQuery = false;
-                wrappedCb(r);
+                cb(r)
             });
         }else{
             that.runningQuery = false;
-            result = wrappedCb(result);
+            result = cb(result);
         }
         return result;
     },
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -3178,7 +3178,6 @@ dojo.declare("gnr.widgets.VirtualStaticGrid", gnr.widgets.DojoGrid, {
             //this.updateColumnsetsAndFooters(); makes grids slower
             //},1,this);
         }
-        this.domNode.classList.toggle('gnr_empty_grid', this.rowCount === 0);
     },
     mixin_setSortedBy:function(sortedBy) {
         this.sortedBy = sortedBy;
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -1383,7 +1383,9 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
         dojo.connect(widget, 'onSelected', widget, '_gnrUpdateSelect');
         genro.src.onBuiltCall(dojo.hitch(widget, 'render'));
         dojo.connect(widget, 'modelAllChange', dojo.hitch(sourceNode, this.modelAllChange));
-
+        dojo.connect(widget, 'updateRowCount', function(){
+            this.domNode.classList.toggle('gnr_empty_grid', this.rowCount === 0);
+        });
 
     },
 
@@ -1391,7 +1393,6 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
         if (this.attr.rowcount) {
             this.setRelativeData(this.attr.rowcount, this.widget.rowCount);
         }
-        //this.widget._gnrUpdateSelect();
     },
     mixin_columnNodelist:function(idx, includeHeader) {
         var condition = 'td.dojoxGrid-cell[idx="' + idx + '"]';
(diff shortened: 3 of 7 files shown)


dgpaci (changes requested):
Nice visual result, but the current approach is too invasive — it now shows on every tableHandler, including simple ones (e.g. `arguments`, `tags`) that don't have a real store/search and where a blank grid is already the correct empty state. I'd propose these changes before merging:

1. **Opt-in via attribute** — default should be no indicator. Enable it only when the developer explicitly asks for it:
   ```python
   noResults=True
   # or, customizable:
   noResults=dict(message='!!No results', icon='lens')
   ```
   This also opens the door to per-grid customization of message and icon.

2. **Localization** — the message is currently hardcoded in English inside the SVG. It must be translatable (`!!No results`) and rendered via the standard translation pipeline, not baked into the asset. Icon + message should be separated so the message can go through `localization.xml`.

3. **Timing** — the indicator currently appears *before* the query runs, which looks broken. It should toggle on only:
   - after the query has executed and returned zero rows (if a query exists), or
   - at `_onStart` only if no query is attached.

   Toggling unconditionally inside `updateRowCount` fires too early in the lifecycle.

genro:
Thanks for the review @dgpaci. Addressed your points:

1. **Opt-in / scope** — moved the toggle from `patch_updateRowCount` (which fires on all grids) into the `runQuery` callback. Now the indicator only appears on grids that have a real query/store. Grids without a query (arguments, tags, static grids) will never show the icon. This makes an explicit opt-in attribute unnecessary.

2. **Localization** — removed the text from the SVG entirely. It's now icon-only (search lens with X), no hardcoded strings.

3. **Timing** — the toggle now fires only after `runQuery` completes and the result has `totalrows === 0`. It also clears when the next query returns data. No more premature display during lifecycle init.

Please re-review.

dgpaci (dismissed):

genro:
@fporcari quick review please when you have a moment — dgpaci already approved, just need your sign-off to merge.

fporcari (changes requested):
Good idea and clean implementation overall. A few changes needed:

### 1. Bug: return value lost in `runQuery` (genro_components.js)

The synchronous path previously did `result = cb(result)`, preserving the callback return value. The new `wrappedCb` doesn't return anything, so `result` becomes `undefined`. This can break downstream code that depends on the return value of `runQuery`.

Fix: add `return cb(r);` at the top of `wrappedCb`:
```js
var wrappedCb = function(r){
    var res = cb(r);
    var empty = r && r.attr && r.attr.totalrows === 0;
    that.gridBroadcast(function(grid){
        grid.domNode.classList.toggle('gnr_empty_grid', empty);
    });
    return res;
};
```

### 2. Remove the toggle from `runQuery` — keep it only in `updateRowCount`

The class is toggled in two places with different semantics: `runQuery` uses `totalrows` (server-side total), while `updateRowCount` uses `this.rowCount` (visible rows after client-side filters). These can diverge — e.g. totalrows > 0 but rowCount === 0 after a local filter — creating a race on who "wins".

Since `updateRowCount` is always called after data arrives and covers all cases (server query, client filter, insert/delete, sort), it should be the single owner of the toggle. Remove the `gridBroadcast` toggle from `runQuery` entirely — this also makes point 1 moot.

### 3. Bonus: port CSS + JS changes to d20

The CSS and JS changes are d11 only. Since d11 and d20 must stay in sync, please port the CSS rule and the `updateRowCount` hook to d20 as well. Should be a quick copy.

genro:
Thanks @fporcari, the three points are fair. Before changing anything I want to flag a design choice from the earlier round with @dgpaci that constrains the "move the toggle entirely into `updateRowCount`" approach, then leave the call to you.

### Constraint from the previous iteration

When @dgpaci reviewed the first version, his concern was that `updateRowCount` fires on **every** grid — tableHandler, tags, arguments, static grids, grids still in their init cycle before any data arrives. Showing the "no results" icon on those is wrong: a blank arguments grid or a still-loading grid is not an empty search result.

That's why I moved the toggle into `runQuery` in the second iteration: it only fires on grids with a real query, and only after the query actually completes.

So `updateRowCount` alone is not sufficient — we need a way to distinguish "this grid ran a query and it returned zero rows" from "this grid is empty because it hasn't loaded yet / has no query / is static".

### Options to reconcile both concerns

Three ways I see, in order of my preference:

1. **Default-on for tableHandler main grid, opt-in elsewhere, single toggle in `updateRowCount`.**
   Pass `emptyIndicator=True` by default from `__commonTableHandler` (the one grid per TH that always has a real query). Everywhere else it's `False` by default. `updateRowCount` toggles the class only if `this.attr.emptyIndicator` is truthy. Developer can override both ways: `emptyIndicator=False` on a TH, or `emptyIndicator=True` on a manual grid.
   - Pro: single owner of the toggle (your point 2), no race with client-side filters, fixes the return-value bug (point 1) automatically since there's no `wrappedCb` anymore.
   - Con: requires wiring `emptyIndicator` through `__commonTableHandler` / `tableViewer` / `thFrameGrid` to land on the grid node attrs.

2. **Keep the post-query marker, move the toggle to `updateRowCount`.**
   In `runQuery` set `grid._hasRunQuery = true` (and nothing else). In `updateRowCount`, toggle the class only if `this._hasRunQuery`. `wrappedCb` becomes almost trivial, and the return-value bug disappears because we keep a proper `return cb(r)`.
   - Pro: minimal surface change; no new public attribute.
   - Con: the "which grids opt in" logic is implicit (every grid that ever runs a query opts itself in) — which may be fine, but doesn't address a potential explicit opt-out.

3. **Minimal fix, keep current two-point design.**
   Just add the missing `return cb(r)` and port everything to d20. Accept the race between `totalrows` and `rowCount`.
   - Pro: smallest diff.
   - Con: doesn't address your point 2 at all.

### My ask

I'd rather not redo the design round without your input. Which option would you take? Happy to go with whatever you decide — or if you want to push the changes yourself on this branch, even better. I won't touch the code until you reply.

Point 3 (d20 port) is uncontroversial and I'll do it alongside whichever of 1/2/3 above we land on.

fporcari:
@genro thanks for the thorough analysis — the constraint from dgpaci's earlier review makes sense and I should have considered it.

Option 1 is the right approach. An explicit `emptyIndicator` attribute, defaulting to `True` in `__commonTableHandler` and `False` everywhere else, gives us:

- Single owner of the toggle (`updateRowCount`)
- Clean opt-in/opt-out semantics
- No return-value bug (no `wrappedCb` needed)
- No ambiguity between "empty query result" and "grid hasn't loaded yet"

Go with that. Looking forward to the updated diff + d20 port.

genro:
Pushed the Option 1 implementation as discussed.

**Changes:**

1. **`genro_components.js`** — reverted `runQuery` to the original form. The `wrappedCb` is gone, the return value of `cb(result)` is preserved on the synchronous path as before.

2. **`genro_grid.js`** — single toggle owner in `patch_updateRowCount`, conditioned on `this.sourceNode.attr.emptyIndicator`. No more split semantics between `totalrows` and `rowCount` — this is now the one place that owns the class.

3. **`resources/common/th/th.py`** — `__commonTableHandler` sets `grid_kwargs.setdefault('emptyIndicator', True)`, so the main TH grid gets the indicator by default. Any other grid is opt-in (pass `emptyIndicator=True` explicitly), and TH grids can opt out with `emptyIndicator=False`.

4. **d20 port** — mirrored the JS toggle in `gnrjs/gnr_d20/js/genro_grid.js` and the CSS rule in `gnrjs/gnr_d20/css/gnrbase_css/15_gnr_grid_extended.css`.

Ready for re-review @fporcari @dgpaci.

fporcari:
@genro the Option 1 implementation looks good overall, but there's a leftover toggle in `DojoGrid.created` (genro_grid.js, line ~1387) that bypasses the `emptyIndicator` guard:

```js
dojo.connect(widget, 'updateRowCount', function(){
    this.domNode.classList.toggle('gnr_empty_grid', this.rowCount === 0);
});
```

This fires on **every** DojoGrid (not just VirtualStaticGrid) and doesn't check `emptyIndicator`, so any grid with 0 rows gets the `gnr_empty_grid` class — arguments, tags, static grids included. It defeats the opt-in semantics.

Please remove this `dojo.connect` block from `DojoGrid.created`. The conditioned toggle in `VirtualStaticGrid.patch_updateRowCount` is sufficient and is the single owner as intended.

fporcari:
@genro I've been testing several approaches for this and I don't think we should merge it. Here's the reasoning:

### What we tried

1. **CSS background icon (current PR)** — the `no_results.svg` shown via `gnr_empty_grid` class. Issues: timing (icon appears before any query runs), the toggle in `DojoGrid.created` bypasses the `emptyIndicator` guard so it fires on all grids, and distinguishing "not loaded yet" from "loaded with zero rows" requires adding state tracking (`_queryDone` flag) that adds complexity for marginal benefit.

2. **Toast notification on zero results** — `genro.toast.show()` centered on the grid after `totalrows === 0`. Clean and simple, but in practice a toast popping up every time a search returns nothing becomes annoying fast — especially for users who search frequently. It's the kind of thing clients end up hating.

3. **Hybrid approaches** (opt-in attribute + CSS + JS state tracking) — increasingly complex for what it's solving.

### The real issue

A "no results" indicator for something that happens quite frequently in normal usage ends up looking like an error or a problem — when it's neither. The grid already has the `waiting` spinner during query execution. When the spinner disappears and the grid is empty, the user understands the query ran and found nothing. Adding a visual indicator on top of that draws unnecessary attention to a normal, expected state.

I'd suggest closing this one. If a specific page has a UX need for an explicit empty state (like the convention registration dialog mentioned in #812), that's better handled locally with a `dataController` on `totalrows` — which is already a supported pattern.

genro:
Closing as won't-fix for now.

A broader UI revision is in progress; the empty-grid state will be re-evaluated when the grid components are revisited as part of that effort. The proposal here (and the review feedback from @dgpaci on opt-in API + localization + timing, and from @fporcari on the runQuery return value, single-source toggle and d20 alignment) is preserved in this PR's history and will inform the design when the moment comes. The original need stays tracked in #812.

PR #817, opened 2026-04-20
fix(#256): defer compositeColumn formula build for order-independent validation

### LINKED ISSUE #256: Warning message if a compositeColumn is defined using virtual columns

genro:
## Summary

Alternative approach to the now-closed PR #779. Defers the SQL-formula construction in `compositeColumn()` to the building phase so that source-column validation is **order-independent** and catches both virtual-column and missing-column cases with a hard `GnrSqlException`.

Closes #256.

## Problem

Today `compositeColumn()` builds its `sql_formula` synchronously during `config_db`. This has two issues:

1. **Order-dependent validation**: checking whether a source column is virtual (the approach in PR #779) only works if the source column has already been declared at the point `compositeColumn()` is called. If declared afterwards, the check is silently wrong.
2. **No typo detection**: referencing a non-existent source column silently creates an empty physical column via `self.column(column)` side effect, then produces a broken SQL formula that fails at runtime.

## Approach

Split `compositeColumn()` into two phases:

**Phase A — synchronous (at config_db time)**
- Normalise column names (strip `$` prefixes), warn on non-canonical input (preserved from current behaviour)
- Create the virtual-column node with `sql_formula=None` — returned immediately so `.relation()` chaining on the result keeps working
- Register a deferred callback via `self.root._dbmodel.deferOnBuilding(...)`

**Phase B — deferred (in `runOnBuildingCb`, after every `config_db` has completed)**

New private method `_buildCompositeColumnFormula(name, composed_of)`:
- Physical column → use its dtype to build the formula fragment (same logic as before)
- Virtual column → `raise GnrSqlException` (a broken `sql_formula` would fail at runtime anyway — fail fast at build time)
- Missing column → `raise GnrSqlException` (replaces the silent side-effect column creation)
- Assemble and assign `sql_formula` on the virtual-column node

The `deferOnBuilding` pattern is already established in this codebase (see `gnr/app/gnrdbo.py:406` and `:462` for `linkHierarchicalToMaster` / `addRelXtdRelidxColumn`).

## Changes

Single file, `gnrpy/gnr/sql/gnrsqlmodel/model.py`:
- `compositeColumn()` (lines 1049-1089) reworked to Phase A only
- New `_buildCompositeColumnFormula()` (lines 1091-1128) for Phase B

Made explicit a branch that was previously implicit: for `dtype in ('L', 'F', 'R', 'B')` the formula fragment is the bare `${column}` reference (in the old code this fell through as the initial `val` assignment).

## Test plan

- [x] `flake8` clean on modified file
- [x] `gnrpy/tests/sql/test_composite_column.py` — 14/14 passed (pg + sqlite, covers value format, JOINs, deep navigation, WHERE, ORDER BY)
- [x] Full `gnrpy/tests/` suite — `1575 passed`, `67 skipped`, `22 failed` (all pre-existing locale/date failures on clean `develop`, identical count with or without the patch)
- [x] `test_invoice` usages (`price_year.compositeColumn('product_year_key', ...)` and `price_year_note.compositeColumn('product_year_ref', ...).relation(...)`) work end-to-end, including composite-key JOIN navigation

## Follow-up

- [ ] Verify on real projects: start `test_invoice` server and exercise composite-key flows manually
- [ ] Consider extending the same validation to `bagItemColumn` and other virtual-column helpers (out of scope here, worth a dedicated issue)

### DIFF
--- a/gnrpy/gnr/sql/gnrsqlmodel/model.py
+++ b/gnrpy/gnr/sql/gnrsqlmodel/model.py
@@ -1054,6 +1054,12 @@ class DbModelSrc(GnrStructData):
         Builds a SQL formula that concatenates the given columns into a
         JSON-style array string (``[val1, val2, ...]``).
 
+        The SQL formula and source-column validation are deferred to the
+        building phase (:meth:`DbModel.runOnBuildingCb`) so that source
+        columns can be declared in any order relative to this call, and
+        so that missing or virtual source columns raise at build time
+        rather than producing a broken SQL formula.
+
         Args:
             name: Column name.
             columns: Comma-separated list of source column names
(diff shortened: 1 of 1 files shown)


dgpaci (approved):
LGTM. Verified the deferred-build approach end-to-end on `sandbox` instance (sqlite, `fatt.prodotto` table), commit 26d7515c. All three scenarios from the previous PR #779 review are covered, plus the new typo case.

## Tests performed

**Case 1 — virtual source column (regression for #256)**
```python
tbl.formulaColumn('descrizione_upper', sql_formula='upper($descrizione)',
                  dtype='T', name_long='Descrizione maiuscola')
tbl.compositeColumn('codice_descrizione_key', columns='codice,descrizione_upper')
```
Startup aborts hard with:
```
GnrSqlException: compositeColumn 'codice_descrizione_key' in table fatt.prodotto:
source column 'descrizione_upper' is a virtual column; a physical column is required
```
Fail-fast behaviour is correct — the resulting SQL formula would have been broken at query time anyway.

**Case 2 — real source columns (no false positive)**
```python
tbl.compositeColumn('codice_descrizione_key', columns='codice,descrizione')
```
Build phase completes cleanly, no composite-related error logged or raised.

**Case 3 — typo / missing source column (new coverage vs PR #779)**
```python
tbl.compositeColumn('codice_descrizione_key', columns='codice,descrizione_TYPO')
```
Startup aborts hard with:
```
GnrSqlException: compositeColumn 'codice_descrizione_key' in table fatt.prodotto:
source column 'descrizione_TYPO' does not exist
```
This case previously passed silently (the old `self.column(column)` call created an empty physical column as a side effect, then produced a broken SQL formula). Catching it at build time is a real improvement.

## Notes on the implementation

- The virtual-column node is still created synchronously with `sql_formula=None`, so `.relation()` chaining on the return value remains safe.
- `deferOnBuilding` is already an established pattern in the codebase (`gnrdbo.py:406`, `:462`), so this is consistent.
- Error messages include both the composite name and the enclosing `tblname` — easy to locate the faulty definition in a large model.

Approving.

cgabriel (approved):
LGTM

close #256

PR #822, opened 2026-04-22
fix: pass table parameter through iframe form handler

fporcari:
## Summary
- Allow the iframe form handler to forward the `table` attribute to the form page
- In `public.py`, use the passed `table` parameter (with fallback to `self.maintable`) for `formId`, `datapath`, and `thFormHandler`
- In `th.js`, pass `main_th_table` in iframe attributes so the form page knows which table to use

## Test plan
- [ ] Open an iframe form on a table different from `self.maintable` and verify it loads correctly
- [ ] Verify standard iframe forms (without explicit table override) still work as before

### DIFF
--- a/resources/common/public.py
+++ b/resources/common/public.py
@@ -676,16 +676,17 @@ class TableHandlerMain(BaseComponent):
         formResource = th_kwargs.pop('formResource',None)
         root.attributes.update(overflow='hidden')
         public = boolean(th_kwargs.pop('public',False))
-        formId = th_kwargs.pop('formId',self.maintable.replace('.','_'))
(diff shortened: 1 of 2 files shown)


dgpaci:
My tests on related issue https://github.com/softwellsrl/anaci/issues/321 are positive, this PR solves the problem

genro (approved):
Clean, focused fix. All three uses of `self.maintable` (formId, datapath, thFormHandler) now consistently use the `table` parameter with backward-compatible fallback. dgpaci confirmed it resolves the issue on a real case. LGTM.

**Note**: base branch is `master` — develop might be the intended target.

PR #825, opened 2026-04-22
GnrSqlAppDb now as "application" as mandatory

### LINKED ISSUE #651: GnrSqlAppDb: make application parameter required
## Problem

`GnrSqlAppDb.__init__()` inherits `application=None` as default from `GnrSqlDb`. However, `GnrSqlAppDb` is the app-level subclass — it **always** needs an application instance (set at `gnrapp.py:995`).

The existing test in `tests/app/gnrapp_test.py:151` instantiates it without arguments:

```python
def test_gnrsqlappdb(self):
    a = ga.GnrSqlAppDb()
    assert a.application is None
```

This doesn't reflect real usage — in production `GnrApp` always passes `application=self`.

## Suggestion

1. Make `application` a **required** parameter in `GnrSqlAppDb` (override `__init__` or validate in the body)
2. Update the test to verify that instantiating without `application` **raises an error**

## Context

Found while working on `addMacro` registration (#617). The new `registerMacros()` override in `GnrSqlAppDb` calls `self.application.pkgBroadcast(...)`, which fails when `application is None`. We added a guard (`if self.application:`) as a workaround, but the real fix is making `application` mandatory for this subclass.

@cgabriel — this test was added in Jan 2024 with a `FIXME: maybe this should be moved to gnr.sql`. What do you think about making `application` required?

cgabriel:
GnrSqlAppDb now as "application" as mandatory, raise error if not provided.

Updated tests accordingly.

close #651

### DIFF
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -298,7 +298,12 @@ class DbStoresHandler(object):
 
 class GnrSqlAppDb(GnrSqlDb):
     """TODO"""
-
+    def __init__(self, *args, **kwargs):
+        if not kwargs.get("application", None):
+            raise TypeError("'application' is mandatory for GnrSqlAppDb")
+        
+        super().__init__(*args, **kwargs)
+        
     @property
     def stores_handler(self):
         handler = getattr(self, '_stores_handler', None)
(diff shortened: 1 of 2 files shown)


cgabriel:
I've merged this without a review since the implementation was already requested in the upstream issue.

PR #828, opened 2026-04-23
implemented a storage mover based on db record

cgabriel:
It accepts an instance, an optional column name where to search for the filepath (default to 'filepath', used in attachment table), and a source and destination storage.

Every storage record from the source one are detected, content moved and record updated accordingly.

The classic use case is when a deployed application changes the storage destination during its lifecycle, moving, for example, from local filesystem 'site:' to s3 storages. This script allow to migrate the previous content to the new destination, avoiding a mixup storage, with files on different backends for the same kind.

Useful also to move between s3 service providers, while maintainin service continuity.

This script is a generic tool but initial created to address softwellsrl/anaci#325

### DIFF
--- a/gnrpy/gnr/web/cli/gnrstoragemove.py
+++ b/gnrpy/gnr/web/cli/gnrstoragemove.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python
+# encoding: utf-8
+
+import os
+from datetime import datetime
+
+from gnr.core.cli import GnrCliArgParse
+from gnr.sql.gnrsql_exceptions import GnrSqlMissingColumn, GnrSqlMissingField, GnrSqlMissingTable
+from gnr.web import logger
+from gnr.web.gnrwsgisite import GnrWsgiSite
+
+
+description = "move files from a storage to another, based on db record"
+
+
+class StorageMover(object):
+    def __init__(self, options):
+        self.options = options
+        self.site = GnrWsgiSite(options.site_name)
+        self._log_lines = []
+
+    def _log(self, level, msg, *args):
+        text = msg % args if args else msg
+        ts = datetime.now().strftime('%H:%M:%S')
+        self._log_lines.append('[%s] %s %s' % (level, ts, text))
+        getattr(logger, level.lower())(msg, *args)
+
+    def _write_report(self, started_at, candidates, moved, errors):
+        dry_run = self.options.dry_run
+        ts = started_at.strftime('%Y%m%d_%H%M%S')
+        from_storage = self.options.from_storage
+        to_storage = self.options.to_storage
+        site_name = self.options.site_name
+        filename = '%s_%s_to_%s_%s.log' % (site_name, from_storage, to_storage, ts)
+        filepath = os.path.join(os.getcwd(), filename)
+
+        mode_label = 'DRY RUN' if dry_run else 'LIVE'
+        lines = [
+            'Storage Move Report',
+            '=' * 60,
+            'Site:    %s' % site_name,
+            'From:    %s' % from_storage,
+            'To:      %s' % to_storage,
+            'Column:  %s' % self.options.column,
+            'Tables:  %s' % ', '.join(self.options.tables),
+            'Mode:    %s' % mode_label,
+            'Started: %s' % started_at.strftime('%Y-%m-%d %H:%M:%S'),
+            '',
+            'Detail',
+            '-' * 60,
+        ] + self._log_lines + [
+            '',
+            'Summary',
+            '-' * 60,
+            'Candidates (files found in source): %d' % candidates,
+        ]
+
+        if dry_run:
+            lines.append('Would move:                         %d' % candidates)
+        else:
+            lines.append('Moved:                              %d' % moved)
+
+        lines += [
+            'Errors (source file not found):     %d' % errors,
+            '',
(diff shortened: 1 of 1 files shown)


dgpaci (changes requested):
Nice work on this — the script is compact and the dry-run + report split is a good touch. A few observations before it's good to go.

### ⚠️ Requested change — report is lost on crash

`_log_lines` is buffered in memory and only flushed by `_write_report` at the end of `run()`. If any `move()` raises, if the operator hits Ctrl-C, or if a `commit()` fails partway through, no report is written — and we lose the trace of which files were already moved. For the "service continuity / live migration" use case described in the PR, that's exactly the scenario where the report matters most.

Please wrap the main loop in a `try/finally` that always calls `_write_report`, or write entries incrementally as they happen. Either is fine; the key is that the report survives an aborted run.

### 💡 Suggestion — register the CLI in `pyproject.toml`

Every other CLI under `gnr/web/cli/` is registered under `[project.scripts]` in `gnrpy/pyproject.toml` (see `gnrsyncstorage`, `gnrdaemon`, `gnrtaskscheduler`…). Without an entry like:

```toml
gnrstoragemove = 'gnr.web.cli.gnrstoragemove:main'
```

the command is only runnable as `python -m gnr.web.cli.gnrstoragemove` after install. Not blocking — if that's the intended invocation, fine — but most likely worth adding for consistency with the other tools.

### Minor observations (take or leave)

- **DB/filesystem consistency window.** The current order is `move → raw_update → commit`. If the commit fails after a successful cross-backend move, the file is on the destination but the record still points to the source — an orphan. A safer sequence for cross-backend moves would be `copy → update → commit → delete source`, so a failed commit leaves the original intact. Worth considering given the tool's stated purpose.
- **`_validate_storage` errors bypass the report.** Pre-flight errors go through `logger.error` directly, not `self._log(...)`, so they never appear in the report file. Minor consistency fix.
- **No progress indication.** For large tables or slow S3 backends the tool can look hung. `gnrsyncstorage` uses an optional `progressbar` — easy to replicate if useful.
- **`raw_update` uses the pre-computed `dest_path` string** rather than `dest_node.fullpath` after the move. In practice they coincide (the `child(basename)` branch of `_move_file` only fires when dest is a directory), but it's an implicit assumption worth a comment or a defensive read from the returned node.

cgabriel:
> Nice work on this — the script is compact and the dry-run + report split is a good touch. A few observations before it's good to go.

Thank yor for your review.

> ### ⚠️ Requested change — report is lost on crash
> `_log_lines` is buffered in memory and only flushed by `_write_report` at the end of `run()`. If any `move()` raises, if the operator hits Ctrl-C, or if a `commit()` fails partway through, no report is written — and we lose the trace of which files were already moved. For the "service continuity / live migration" use case described in the PR, that's exactly the scenario where the report matters most.
> 
> Please wrap the main loop in a `try/finally` that always calls `_write_report`, or write entries incrementally as they happen. Either is fine; the key is that the report survives an aborted run.

I've wrapped everything under a try/finally block, to preserve the log generation for unexpected errors.
 
> ### 💡 Suggestion — register the CLI in `pyproject.toml`
> Every other CLI under `gnr/web/cli/` is registered under `[project.scripts]` in `gnrpy/pyproject.toml` (see `gnrsyncstorage`, `gnrdaemon`, `gnrtaskscheduler`…). Without an entry like:
> 
> ```toml
> gnrstoragemove = 'gnr.web.cli.gnrstoragemove:main'
> ```
> 
> the command is only runnable as `python -m gnr.web.cli.gnrstoragemove` after install. Not blocking — if that's the intended invocation, fine — but most likely worth adding for consistency with the other tools.

Not implemented: scripts in pyproject.toml are only the old ones, preserved for backwards compatibility. All new script are automatically collected via 'gnr' command, so there is really no need to add them here.
 
> ### Minor observations (take or leave)
> * **DB/filesystem consistency window.** The current order is `move → raw_update → commit`. If the commit fails after a successful cross-backend move, the file is on the destination but the record still points to the source — an orphan. A safer sequence for cross-backend moves would be `copy → update → commit → delete source`, so a failed commit leaves the original intact. Worth considering given the tool's stated purpose.

You're right, much saver. Replaced with a copy -> commit -> delete sequence.

> * **`_validate_storage` errors bypass the report.** Pre-flight errors go through `logger.error` directly, not `self._log(...)`, so they never appear in the report file. Minor consistency fix.

Done

> * **No progress indication.** For large tables or slow S3 backends the tool can look hung. `gnrsyncstorage` uses an optional `progressbar` — easy to replicate if useful.

Don't work to add a dependency just for this, the log is enough to check progress.

> * **`raw_update` uses the pre-computed `dest_path` string** rather than `dest_node.fullpath` after the move. In practice they coincide (the `child(basename)` branch of `_move_file` only fires when dest is a directory), but it's an implicit assumption worth a comment or a defensive read from the returned node.

Made it explicit.

Please review! Thank you!

dgpaci (approved):
All requested changes addressed in b89866b:

- Main loop wrapped in try/finally so the report is always written, even on aborted runs ✅
- Switched to copy → update → commit → delete source for safer rollback semantics ✅
- Pre-flight and schema errors now routed through `self._log` so they land in the report ✅
- Explicit comment documenting the `dest_path` / `dest_node.path` equivalence ✅

LGTM.

Issue #671, opened 2026-03-09
Remove dead Dojo multi-version code (keep versioning mechanism for future 2.0)


genro:
## Summary

GenroPy's codebase contains support for multiple Dojo versions (1.1, 1.4, 1.5, 1.7, 1.8) but **only version 1.1 is actually used**. The multi-version code is dead weight that should be cleaned up.

The versioning **mechanism** itself must be preserved: in the future there will be a version `2.0` (a Softwell fork of Dojo, initially identical to 1.1). So the architecture must support `dojo_version='11'` (current) and `dojo_version='20'` (future).

## Current state

| Asset | Exists | Used |
| --- | --- | --- |
| `dojo_libs/dojo_11/` | yes | yes |
| `dojo_libs/dojo_18/` | yes | no (1 isolated page) |
| `gnrjs/gnr_d11/` | yes | yes |
| `gnrjs/gnr_d13,14,15,18/` | no | — |
| `frontend/dojo_14.py` | yes | no |
| `frontend/dojo_15.py` | yes | no |
| `frontend/dojo_18.py` | yes | almost no (1-2 pages) |
| `GnrDomSrc_dojo_14/15/18` in gnrwebstruct.py | yes | no (all are `pass`) |
| JS branches on `dojo_version` for 1.5/1.7 | yes | dead code |

### Pages using a version other than 11 (only 3 found)

- `genropy:webpages/html_mixablepage.py` → `dojo_version='18'`
- `genromed:webpages/report/ambulatorio_occupazione.py` → `dojo_version='13'`
- `genropynet:webpages/serverpages/home.py` → `dojo_version='18'`

### Phase 1: Dead Python frontend files

**Delete:**
- `gnrpy/gnr/web/gnrwebpage_proxy/frontend/dojo_14.py`
- `gnrpy/gnr/web/gnrwebpage_proxy/frontend/dojo_15.py`
- `gnrpy/gnr/web/gnrwebpage_proxy/frontend/dojo_18.py`

**Modify `gnrpy/gnr/web/gnrwebstruct.py`:**
- Remove empty classes: `GnrDomSrc_dojo_14`, `GnrDomSrc_dojo_15`, `GnrDomSrc_dojo_18` (all just `pass`, inheriting from `GnrDomSrc_dojo_11`)

### Phase 2: Dead JS version branches

Versions 1.4, 1.5, 1.7 will never exist again. Remove their branches, keep the `dojo_version` variable for future `2.0` checks.

[...]

PR #830, opened 2026-04-24
fix(#826): allow Shift key in pasteOnGrid onkeydown

### LINKED ISSUE #826: [BUG] Shift blocks scrolling in grids with paste support (pasteOnGrid)
## Problem Description

In grids that have the `onpaste` attribute set (i.e. grids flagged as `pasteOnGrid`), holding the **Shift** key blocks scrolling of the grid viewport.

The classic range-selection pattern is broken:

**Expected:** click the first row → hold Shift → scroll down → click the last row → the whole range is selected and can be copied.

**Observed:** while Shift is held, the grid does not scroll. Users must release Shift, scroll to the desired row, press Shift again, and only then click the last row.

This regression appeared after the theme modernization release (PR #671, commit `e225849c0`). The new DOM/CSS structure makes the blocking behavior manifest under normal user interaction.

## Reproduction Steps

The simplest reliable repro is **Windows + mouse wheel**:

1. Open any page with a grid that has the `onpaste` attribute configured (any grid with paste/copy-from-Excel support).
2. Click anywhere inside the grid (or click the first visible row).
3. Hold down the **Shift** key.
4. Scroll down with the **mouse wheel**.

**Observed result:** while Shift is held, the wheel does not scroll the grid viewport — scrolling is suppressed entirely.

**Expected result:** Shift + mouse wheel should scroll normally, so the user can keep Shift held while moving to the target row and complete a range selection with a single interaction.

The same blocking can also be observed via the keyboard (Shift + PageDown / ArrowDown), but the wheel case is the one users hit most often.

## Testing notes — platform / device dependence

The bug is **clearly reproducible on Windows with a regular mouse wheel**.

On **macOS with a trackpad** the bug does **not** appear to manifest: Shift + two-finger scroll keeps scrolling the grid normally. It is not yet clear whether the difference is due to the OS (event dispatch / default action handling) or to the input device (discrete wheel events vs continuous trackpad gestures). A test on macOS with an external mouse wheel would help disambiguate.

This platform asymmetry is important for review and validation: a fix verified only on macOS/trackpad cannot be considered conclusive — it must be tested on Windows + mouse wheel where the bug is consistently visible.

## Root cause hypothesis (to be confirmed)

In `gnrjs/gnr_d20/js/genro_grid.js:797` (and the mirrored `gnrjs/gnr_d11/js/genro_grid.js:849`), the inline `onkeydown` handler attached to the `contenteditable` grid `domNode` only whitelists `metaKey` and `ctrlKey`:

```js
widget.domNode.setAttribute('onkeydown',
  "if(event.metaKey || event.ctrlKey || event.target!=event.currentTarget) return true; return false;");
```

This handler is the obvious culprit for the **keyboard** variant (Shift + PageDown / ArrowDown / Space), which falls into the `return false` branch.

However, for the **mouse-wheel** variant the keydown handler is not in the event path — wheel events do not fire `keydown`. So allowing `shiftKey` in the keydown allow-list, by itself, will not unblock Shift + wheel scrolling. The wheel-blocking behavior likely comes from a different path (e.g. a `wheel`/`mousewheel` handler, or the contenteditable's own focus/selection-extension behavior swallowing the default scroll). This needs to be investigated separately.

## Environment

Genropy 26.04.21 (current `develop`), any browser. Bug confirmed on **Windows with mouse wheel**; not reproduced on **macOS with trackpad**. Regression introduced by PR #671 (style modernization).

fporcari:
## Summary

- Add `event.shiftKey` to the `onkeydown` allow-list on contenteditable grid nodes (pasteOnGrid)
- Applied to both `gnr_d11` and `gnr_d20` versions of `genro_grid.js`

Fixes #826

## Root cause

The `onkeydown` handler on pasteOnGrid grids only allowed `metaKey` and `ctrlKey` through, blocking all Shift-only key events. After the DOM/CSS restructuring in PR #671, the contenteditable node receives keyboard focus more readily, making the Shift-scroll block manifest during normal range-selection interactions.

## Test plan

- [ ] Open a page with a pasteOnGrid grid
- [ ] Click a row, hold Shift, scroll with mouse wheel — scrolling should work
- [ ] Click a row, hold Shift, use PageDown/ArrowDown — navigation should work
- [ ] Complete a Shift+click range selection across scrolled rows
- [ ] Verify Ctrl+C / Cmd+C copy still works
- [ ] Verify paste from Excel still works

### DIFF
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -846,7 +846,7 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
             widget.domNode.setAttribute('contenteditable',"true");
             widget.domNode.setAttribute('onpaste',"return false;");
             widget.domNode.setAttribute('oncut',"return false;");
-            widget.domNode.setAttribute('onkeydown',"if(event.metaKey || event.ctrlKey || event.target!=event.currentTarget) return true; return false;");
+            widget.domNode.setAttribute('onkeydown',"if(event.metaKey || event.ctrlKey || event.shiftKey || event.target!=event.currentTarget) return true; return false;");
             dojo.connect(widget.domNode,'onpaste', funcCreate(savedAttrs.onpaste,'event',widget));
         }
         objectFuncReplace(widget.selection, 'clickSelectEvent', function(e) {
--- a/gnrjs/gnr_d20/js/genro_grid.js
+++ b/gnrjs/gnr_d20/js/genro_grid.js
@@ -794,7 +794,7 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
             widget.domNode.setAttribute('contenteditable',"true");
             widget.domNode.setAttribute('onpaste',"return false;");
             widget.domNode.setAttribute('oncut',"return false;");
-            widget.domNode.setAttribute('onkeydown',"if(event.metaKey || event.ctrlKey || event.target!=event.currentTarget) return true; return false;");
+            widget.domNode.setAttribute('onkeydown',"if(event.metaKey || event.ctrlKey || event.shiftKey || event.target!=event.currentTarget) return true; return false;");
             dojo.connect(widget.domNode,'onpaste', funcCreate(savedAttrs.onpaste,'event',widget));
         }
         objectFuncReplace(widget.selection, 'clickSelectEvent', function(e) {


dgpaci (changes requested):
Tested on **Windows with a regular mouse wheel** (the original repro from #826) and the bug is **still present** with this patch applied: holding Shift + scrolling the wheel down does not scroll the grid viewport. No improvement observed.

## Why the fix doesn't address the symptom

This PR only adds `event.shiftKey` to the `onkeydown` allow-list on the contenteditable grid `domNode`. That branch is the right thing to do for the **keyboard** variant of the bug (Shift + PageDown / ArrowDown / Space), but it has no effect on the mouse-wheel variant — wheel events do not fire `keydown`, so the keydown handler is simply not in the event path. The Shift + wheel block must be coming from a different place (a `wheel`/`mousewheel` handler, or the contenteditable's own default behavior when focused), and that is what actually needs to be fixed.

## Platform / device caveat (please read)

I've updated #826 with a "Testing notes" section explaining this — the bug **does not reproduce on macOS with a trackpad**, only on **Windows with a mouse wheel**. It's not yet clear whether the asymmetry is due to the OS or to the input device (discrete wheel events vs continuous trackpad gestures). The practical consequence for review is:

- A "works for me" check on macOS/trackpad is **not sufficient** to validate this fix.
- Validation must be done on **Windows + mouse wheel**, where the bug is consistently visible.

## What I'd suggest

1. Keep the `shiftKey` addition to the keydown allow-list (it's correct and unblocks the keyboard variant — useful on its own).
2. Investigate the wheel path separately: there is almost certainly a `wheel`/`mousewheel` handler, or a side-effect of the `contenteditable` focused state, that swallows the default scroll when Shift is held. That's the real fix for the headline symptom.
3. Re-test the final patch on Windows + mouse wheel before re-requesting review.

As it stands, the PR title says "fix #826" but the user-visible bug is unchanged on the platform where it actually manifests, so I can't approve this.

dgpaci (approved):
Updating my previous review.

After further investigation directly on Windows + mouse wheel — the platform where #826 reproduces — I'm splitting the validation in two:

## Keyboard variant (this PR) — approved

The `event.shiftKey` addition to the `onkeydown` allow-list on `pasteOnGrid` grids is correct and necessary. It unblocks Shift + PageDown / ArrowDown / Space / Home / End on grids with paste support, which were previously falling into the `return false` branch. This is a clean, narrowly scoped change with no risk of regression on non-pasteOnGrid grids. LGTM as-is.

## Mouse-wheel variant — separate PR (#851)

The headline symptom of #826 (Shift + wheel does not scroll the grid viewport) is **not** caused by the `onkeydown` handler — wheel events do not fire `keydown`. Diagnosis on Windows + Chrome/Edge:

- The browser emits a normal `wheel` event with `deltaY` set (not remapped to `deltaX`) and `defaultPrevented=false`.
- Windows applies a platform-specific default action for `Shift + wheel`: it targets the nearest ancestor with horizontal overflow and **does not** fall back to vertical scroll — that's the actual freeze.
- This is a **general grid regression** (introduced in PR #671 styling), not specific to `pasteOnGrid`: it reproduces on any grid in `gnr_d11` / `gnr_d20`, regardless of whether `onpaste` is set.

The fix that resolves the wheel variant is a `wheel` capture-phase listener on `widget.domNode` for **every** grid that, on `shiftKey`, calls `preventDefault()` and applies `deltaY` manually to `views.views[0].scrollboxNode.scrollTop`. Validated on Windows + mouse wheel; no regression on Ctrl/Cmd + wheel, normal wheel, paste from Excel, or trackpad behavior.

Submitted as #851, also targeting `master` with the `hotfix` label so both ship together. #826 will be closed by #851, since this PR alone does not resolve the user-visible symptom on Windows.

Approving this PR for the keyboard part.

PR #831, opened 2026-04-24
fix(#829): checkbox visibility toggle in _ask dialogs

### LINKED ISSUE #829: [BUG] Checkbox in _ask dialog shows only label when conditional visibility toggles
### Problem Description

In a `dataRpc` `_ask` dialog, a `checkbox` field with conditional visibility (`hidden` bound to another field's value) renders incorrectly when it becomes visible: only the **label** is shown, while the actual checkbox input is missing.

**Observed behavior**: after the controlling field is filled and the checkbox becomes visible, the widget area shows the label text only — no clickable checkbox is rendered.

**Expected behavior**: the checkbox input should appear alongside its label, fully interactive.

### Reproduction Steps

A reproducible example is already available in the gnrcore test suite:

- File: `projects/gnrcore/packages/test/webpages/datastore/datarpc.py`
- Test: `test_5_ask` (around line 96)

Relevant snippet:

```python
def test_5_ask(self, pane):
    fb = pane.formbuilder(cols=1)
    btn = fb.button('How many days left to your birthday?')
    btn.dataRpc('.days_left', self.getTime, _ask=dict(title='When is your next birthday?',
                                fields=[dict(name='birthday', tag='datetextbox', lbl='Date'),
                                        dict(name='confirm', tag='checkbox', lbl='Confirm', hidden='^.birthday?=!#v')]))
    fb.div('^.days_left', lbl='Days left')
```

Steps:
1. Open the `test` package and navigate to the `datarpc` test page.
2. Run test `test_5_ask` — click the button *"How many days left to your birthday?"*.
3. In the ask dialog, the `confirm` checkbox is initially hidden (correct).
4. Fill the `birthday` date field.
5. The `confirm` row becomes visible, but only the label *"Confirm"* is rendered — the checkbox input is missing.

### Environment

Genropy 26.04.22, develop branch, reproducible on the gnrcore `test` package.

### Additional Material

The `hidden` expression used is `^.birthday?=!#v`, which toggles visibility based on whether `birthday` has a value. The issue seems specific to checkboxes inside `_ask` dialog fields — verifying the same conditional pattern on other widget types would help narrow down whether it affects only `checkbox` or the conditional rendering path in the ask dialog in general.

fporcari:
## Summary

- Checkbox `mixin_setHidden` now registers `_gnrcheckbox_wrapper` as hidden target and delegates to `sourceNode.setHidden()`
- This ensures layout recalculation (`fakeResize`) and consistent visibility handling with other widget types
- Applied to both `gnr_d11` and `gnr_d20`

Fixes #829

## Root cause

The checkbox's `mixin_setHidden` had a custom `else` branch that directly set `display` style on the wrapper div via `dojo.style`, bypassing `sourceNode.setHidden()`. When a checkbox with a dynamic `hidden` binding was created inside a hidden container (e.g. in an `_ask` dialog), the layout was never recalculated when visibility toggled, leaving the checkbox input invisible.

## Test plan

- [ ] Open `test` package → `datastore/datarpc` → `test_5_ask` (with checkbox field added per issue description)
- [ ] Click button → dialog opens → checkbox hidden (correct)
- [ ] Fill birthday field → checkbox row becomes visible with label AND checkbox input
- [ ] Clear birthday → checkbox hides again
- [ ] Verify regular checkboxes (without `hidden` binding) still work normally
- [ ] Verify checkboxes with `_hiddenTargets` already set still work

### DIFF
--- a/gnrjs/gnr_d11/js/genro_widgets.js
+++ b/gnrjs/gnr_d11/js/genro_widgets.js
@@ -3698,11 +3698,10 @@ dojo.declare("gnr.widgets.CheckBox", gnr.widgets.baseDojo, {
         }
     },
     mixin_setHidden: function(hidden) {
-        if(this.sourceNode._hiddenTargets){
-            this.sourceNode.setHidden(hidden);
-        }else{
-            dojo.style(this.sourceNode._gnrcheckbox_wrapper, 'display', (hidden ? 'none' : ''));
+        if(!this.sourceNode._hiddenTargets){
+            this.sourceNode._hiddenTargets = [this.sourceNode._gnrcheckbox_wrapper];
         }
+        this.sourceNode.setHidden(hidden);
     },
 
     _getKeeperRoot:function(sourceNode){
--- a/gnrjs/gnr_d11/js/genro_widgets.js
+++ b/gnrjs/gnr_d11/js/genro_widgets.js
@@ -3699,7 +3699,7 @@ dojo.declare("gnr.widgets.CheckBox", gnr.widgets.baseDojo, {
     },
     mixin_setHidden: function(hidden) {
         if(!this.sourceNode._hiddenTargets){
-            this.sourceNode._hiddenTargets = [this.sourceNode._gnrcheckbox_wrapper];
+            this.sourceNode._hiddenTargets = [this.sourceNode._gnrcheckbox_wrapper, this.domNode];
         }
         this.sourceNode.setHidden(hidden);
     },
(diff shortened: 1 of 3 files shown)


dgpaci (approved):
LGTM — fix verified locally, the `_ask` dialog scenario from #829 now works as expected and regular checkboxes are unaffected.

The root-cause analysis is correct: the previous `else` branch bypassed `sourceNode.setHidden()` and skipped the layout recalculation. Always routing through `setHidden()` (and seeding `_hiddenTargets` with both the wrapper and `domNode`) is the right call, and the change is mirrored cleanly in `gnr_d11` and `gnr_d20`.

**Suggestion (non-blocking):** I'd remove `projects/gnrcore/packages/test/webpages/inputfields/checkbox_hidden.py`. It's not a real test (no assertions, just a manual playground) and it duplicates scenarios that already exist elsewhere — keeping it in the repo risks confusing future readers into thinking it's part of the test suite when it isn't. The fix is already validated by the manual test plan in this PR; no need to keep the page around.

genro (approved):

PR #840, opened 2026-04-28
Add LlmService base class with model aliases

fporcari:
## Summary
- Add `LlmService` base class in `gnrpy/gnr/lib/services/llm.py` extending `GnrBaseService`
- Implement model alias system (`low`/`medium`/`max`) so callers can request capability tiers instead of hardcoded model names
- Define abstract `complete()` method as the standard interface for LLM provider implementations

## Test plan
- [ ] Verify `LlmService` can be subclassed by provider-specific implementations (Anthropic, OpenAI, etc.)
- [ ] Verify model alias resolution works: alias → resolved name, unknown → pass-through, None → default
- [ ] Verify service discovery picks up `llm` service type from the standard service directories


genro (commented):
Review of the `LlmService` scaffold. Overall the class is well aligned with the conventions of `gnr.lib.services` (filename = service_type, no `super().__init__()` call, `self.parent = parent`, `**kwargs` accepted). A few points worth addressing before this becomes the contract for provider implementations.

### 1. Bug in `resolve_model` when an alias is requested but not configured

```python
return self._model_aliases.get(model) or model or self.model
```

If a caller asks for `resolve_model('low')` but `model_low` is `None` (or `''` from the config UI), the method silently returns the literal string `'low'`. That string is then passed to the provider API as a model identifier and produces an opaque "unknown model" error far from the actual cause.

Suggested fix — distinguish "alias key but unset" from "literal model name":

```python
if model in self._model_aliases:
    return self._model_aliases[model] or self.model
return model or self.model
```

This way an unconfigured alias falls back to `self.model` instead of leaking the alias key as a model name.

### 2. `service_type` class attribute not set

`GnrBaseService.updateServiceParameters` reads `self.service_type` to persist parameters on `sys.service`. Other services either rely on the discovery mechanism to inject it or set it explicitly. Worth confirming that `LlmService` will get `service_type = 'llm'` populated correctly when instantiated through `getService('llm')`, otherwise add it as a class attribute.

### 3. Docstring style mismatch with the rest of the module

The rest of `gnr/lib/services/` uses Sphinx-style docstrings (`:param X:`, `:return:` — see `mail.py`, `storage.py`). This PR uses Google style (`Args:` / `Returns:`). Minor, but worth aligning for consistency across the module.

### 4. `**kwargs` silently dropped in `__init__`

Extra parameters coming from the service config (e.g. `timeout`, `temperature_default`) are accepted and discarded without trace. Same pattern as other services, so not blocking — but if the intent is for subclasses to consume provider-specific extras, storing `self._extra = kwargs` would make it explicit.

### 5. Return contract of `complete()`

The dict-with-keys contract (`answer`, `model`, `usage`) is fragile (silent typos, no IDE autocomplete). For a base class that defines the interface for every future provider, a `TypedDict` or a small dataclass would harden the contract. Not blocking for the scaffold but worth considering before the first concrete implementation lands.

### 6. No test coverage

`gnrpy/tests/core/gnrbaseservice_test.py` is currently a stub, so this is consistent with the rest of the module — but a minimal unit test for `resolve_model` (covering the alias / passthrough / unset-alias / None cases) would lock down the contract and surface point 1 above.

gnrpy/gnr/lib/services/llm.py
```diff
@@ -0,0 +1,78 @@
...
+        Args:
+            model: A model name, an alias (``low``, ``medium``, ``max``),
+                   or ``None`` to use the service default.
+
+        Returns:
+            str: The resolved model identifier.
+        """
+        if model is None:
+            return self.model
+        return self._model_aliases.get(model) or model or self.model
```
cgabriel: If the service is missing one of the ['low','medium','max'] definitions provided in the constructor, I'll always the the "default" one ('model' passed to the constructor). Wouldn't it be more solid by raising an error that the requested variant is not available? If I configure only 'max', it will be used for trivial tasks. Also, if I pass model='goober' (or any non-existing definition) to resolve_model, it will be used, even if it's not existing. What should happen in such case?

cgabriel (changes requested):

fporcari:
Thanks for the thorough review. Here's a summary of what was addressed and what was left as-is (with rationale).

### Addressed

**1. Bug in `resolve_model` (genro #1 + cgabriel inline)**
Fixed. Now the method explicitly checks `if model in self._model_aliases` — if the alias key exists but the value is not configured, it falls back to `self.model` instead of leaking the alias key as a literal model name.

The design intent: aliases (`low`, `medium`, `max`) represent capability tiers for generic code. A service with a single model configured should work for any alias — all of them fall back to the default. Literal model names are passed through as-is; validation of those is the provider SDK's responsibility at call time.

**3. Docstrings (genro #3)**
Converted to Sphinx-style (`:param:` / `:returns:`) to align with the rest of `gnr/lib/services/`.

**6. Tests (genro #6)**
Added `llmservice_test.py` with 4 test cases covering: default fallback, configured alias, unconfigured alias fallback, and literal model passthrough.

### Left as-is (consistent with framework patterns)

**2. `service_type` (genro #2)**
`BaseServiceType.addService()` injects `service_type` after instantiation — no service in the module sets it as a class attribute. `LlmService` follows the same convention.

**4. `**kwargs` (genro #4)**
All stub services (`PdfService`, `TranslationService`, `QueueService`, `GitService`) accept and discard `**kwargs` in the same way. This is the standard pattern for base classes in this module.

**5. TypedDict/dataclass (genro #5)**
Premature for a base class — no other service defines structured return types. Worth revisiting when concrete implementations stabilize the contract.

gnrpy/gnr/lib/services/llm.py
fporcari: Fixed in the latest commit. The logic now:

- **Alias key** (low, medium, max): if configured, returns the mapped model. If not configured, falls back to the default model. This is intentional — aliases represent capability tiers, and a service with a single model should work for any tier.
- **Literal model name**: passed through as-is. If it doesn't exist, the provider SDK will raise at call time — the base class can't know valid provider model names.

cgabriel (approved):

PR #842, opened 2026-04-28
feat: add genro.cmd unified client capabilities catalog

genro:
## Summary

- Introduce `gnr.GnrCmd`, a single namespace exposing the framework's core client-side capabilities (open page, run batch, show dialog, save form, ...) as ~47 thin wrappers over existing primitives.
- Each command carries an explicit machine-readable schema, so the catalog is introspectable at runtime via `genro.cmd.list()` and `genro.cmd.toolsManifest()`.
- Three equivalent invocation channels are supported.

## Why

Today client-side capabilities are scattered across at least five planes: top-level `genro.*` API, pub/sub events, RPC calls, form/grid handlers, and ad-hoc widget actions. There is no single, machine-readable registry that an external orchestrator (e.g. an agent that needs to translate a user intent into a concrete action) can consume. `genro.cmd` provides that registry while remaining thin: every method is a wrapper over the existing primitive, so no behavior is duplicated.

## Invocation channels

```js
// 1. Direct JS
genro.cmd.openPageInTab({pageName: 'foo', file: 'pkg/foo'});

// 2. Pub/sub
genro.publish('gnrcmd', {command: 'openPageInTab', pageName: 'foo', file: 'pkg/foo'});

// 3. Datachange (useful for server -> client push via setInClientData)
genro.setData('gnr.gnrcmd', {command: 'openPageInTab', pageName: 'foo', file: 'pkg/foo'});
```

## Catalog coverage

~47 commands across 11 categories:
- **navigation** — openPageInTab, openPageInWindow, openPageInDialog, gotoURL, pageReload, pageBack
- **dialog** — alert, ask, prompt, confirm, notify, lockScreen, unlockScreen
- **form** — formNewRecord, formLoadRecord, formSave, formDelete, formAbort, formReload, formGetData
- **grid** — gridRefresh, gridGetSelected, gridSelectRow, gridExport
- **record** — openRecord, newRecord, deleteRecord
- **batch** — runBatch, runPrint, runExport
- **rpc** — serverCall
- **data** — setData, getData, resetData
- **preferences** — setUserPreference, getUserPreference, setAppPreference, getAppPreference, setInStorage, getFromStorage
- **file** — download, viewPDF, recordToPDF
- **misc** — copyToClipboard, focusWidget, refreshBadge

## Files

- `gnrjs/gnr_d11/js/genro_cmd.js` (new) + identical mirror in `gnr_d20`
- `gnrjs/gnr_d11/js/genro.js`: instantiate `this.cmd` in `GenroClient` constructor (mirror in `gnr_d20`)
- `gnrpy/gnr/web/gnrwebpage_proxy/frontend/dojo_{11,20}.py`: register `genro_cmd` module

Out of scope (deliberate): per-package application capabilities, dynamic registration API, server-side endpoint exposing the manifest. These are planned follow-ups.

## Elementi da verificare (PR in draft)

The PR is kept in **draft** until the items below are verified on a real instance. Background: the work was developed against the existing primitives by reading the source (`framedIndexManager.newBrowserWindowPage`, `dlg.iframeDialog`, `publish('table_script_run')`, `subscribeEvent`, `dataSubscribe`, ...), but no live click-through was performed.

- [ ] `genro.cmd` is loaded and reachable on a freshly started page (`--reload`) — both `gnr_d11` and `gnr_d20`
- [ ] `genro.cmd.list()` returns ~47 entries with `{name, category, description, params, returns}`
- [ ] `genro.cmd.toolsManifest()` returns valid Anthropic tool-use shape `{name, description, input_schema}`
- [ ] Direct call: `genro.cmd.notify({message: 'test'})` shows a toast
- [ ] Pub/sub channel: `genro.publish('gnrcmd', {command: 'notify', message: '...'})` triggers the same toast
- [ ] Datachange channel: `genro.setData('gnr.gnrcmd', {command: 'notify', message: '...'})` triggers the same toast
- [ ] Validation: `genro.cmd.openPageInTab({})` throws "missing required param" (no silent fallback)
- [ ] Unknown command: `genro.publish('gnrcmd', {command: 'doesNotExist'})` logs an error and does not crash
- [ ] `runBatch` against a known action in `test_invoice` still triggers the existing batch flow (equivalent to the pre-existing `publish('table_script_run', ...)`)
- [ ] No regression in pages/forms that already use `framedIndexManager`, `dlg`, form handlers (the wrappers are pass-through but worth a smoke test)

## Test plan

- [ ] Open a page of an existing project with `--reload` and in browser console verify:
  - [ ] `genro.cmd` is defined and `genro.cmd.list()` returns ~47 entries
  - [ ] `genro.cmd.toolsManifest()` returns an array of objects with `{name, description, input_schema}` shape
  - [ ] `genro.cmd.notify({message: 'test'})` shows a floating message
  - [ ] `genro.publish('gnrcmd', {command: 'notify', message: 'via publish'})` shows the message
  - [ ] `genro.setData('gnr.gnrcmd', {command: 'notify', message: 'via setData'})` shows the message
  - [ ] `genro.cmd.openPageInTab({})` throws on missing required params
- [ ] Run `runBatch` against a known action in `test_invoice` and verify the existing batch flow still triggers (equivalent to the pre-existing `publish('table_script_run', ...)`)
- [ ] Smoke test on both `gnr_d11` and `gnr_d20` variants

### DIFF
--- a/gnrjs/gnr_d11/js/genro.js
+++ b/gnrjs/gnr_d11/js/genro.js
@@ -224,8 +224,9 @@ dojo.declare('gnr.GenroClient', null, {
 
         this.dom = new gnr.GnrDomHandler(this);
         this.vld = new gnr.GnrValidator(this);
-        this.wsk = new gnr.GnrWebSocketHandler(this,this.websockets_url,{debug:false});     
-        this.som = new gnr.GnrSharedObjectHandler(this);  
+        this.wsk = new gnr.GnrWebSocketHandler(this,this.websockets_url,{debug:false});
+        this.som = new gnr.GnrSharedObjectHandler(this);
+        this.cmd = new gnr.GnrCmd(this);
        //var onerrorcb = function(errorMsg,url,linenumber){
        //    genro.onError(errorMsg,url,linenumber);
        //};
--- a/gnrjs/gnr_d11/js/genro_cmd.js
+++ b/gnrjs/gnr_d11/js/genro_cmd.js
@@ -1,0 +1,897 @@
+/*-*- coding: utf-8 -*-
+ *--------------------------------------------------------------------------
+ * package            : Genropy clientside
+ * module genro_cmd   : Unified client-side command catalog
+ *
+ * Exposes a single namespace `genro.cmd` collecting the framework's
+ * core client-side capabilities (open page, run batch, show dialog,
+ * save form, ...). Every command is a thin wrapper over an existing
+ * primitive and carries an explicit machine-readable schema, so that
+ * the catalog is introspectable at runtime via:
+ *
+ *     genro.cmd.list()           -> array of {name, category, description, params, returns}
+ *     genro.cmd.toolsManifest()  -> Anthropic tool-use compatible array
+ *
+ * Three equivalent invocation channels are supported:
+ *   1. Direct JS:   genro.cmd.openPageInTab({pageName, file})
+ *   2. Pub/sub:     genro.publish('gnrcmd', {command: 'openPageInTab', pageName, file})
+ *   3. Datachange:  genro.setData('gnr.gnrcmd', {command: 'openPageInTab', pageName, file})
+ *--------------------------------------------------------------------------
+ */
+
+dojo.declare('gnr.GnrCmd', null, {
+
+    constructor: function(genro) {
+        this.genro = genro;
+        this._registerDispatchers();
+    },
+
+    _registerDispatchers: function() {
+        var self = this;
+        dojo.subscribe('gnrcmd', function(kw) {
+            self._dispatch(kw || {});
+        });
+        this.genro.dataSubscribe('gnr.gnrcmd', 'gnrcmd_dispatcher', {
+            any: function() {
+                var payload = self.genro.getData('gnr.gnrcmd');
+                if (payload) self._dispatch(payload);
+            }
+        });
+    },
+
+    _dispatch: function(payload) {
+        var command = payload.command;
+        if (!command) {
+            console.error('[gnrcmd] missing "command" key in payload', payload);
+            return;
+        }
+        var fn = this[command];
+        if (typeof fn !== 'function' || !fn.schema) {
+            console.error('[gnrcmd] unknown command:', command);
+            return;
+        }
+        var params = {};
+        for (var k in payload) {
+            if (k !== 'command') params[k] = payload[k];
+        }
+        try {
+            this._validateParams(command, params);
+        } catch (e) {
+            console.error('[gnrcmd] validation error for ' + command + ':', e.message);
+            throw e;
+        }
+        return fn.call(this, params);
+    },
+
+    _validateParams: function(cmdName, params) {
+        var schema = this[cmdName].schema;
+        var spec = schema.params || {};
+        for (var pname in spec) {
+            var pdef = spec[pname];
+            var hasIt = (params[pname] !== undefined && params[pname] !== null);
+            if (pdef.required && !hasIt) {
+                throw new Error('missing required param "' + pname + '"');
+            }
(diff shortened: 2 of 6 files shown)


fporcari (commented):
Reviewed this in full — all 46 declared commands traced to the primitive each one delegates to. The motivation is sound and I think the idea survives: there really is no single machine-readable registry of client capabilities, and an external orchestrator needs one. The schema-per-command shape is the right instinct.

As it stands, though, it cannot run:

1. **It breaks page bootstrap.** The constructor calls `dataSubscribe('gnr.gnrcmd', …)` (`genro_cmd.js:29-40`), and `new gnr.GnrCmd(this)` is instantiated inside `genroInit` (`genro.js:227`). But `genro._data` is created in `start()` (`genro.js:581`), which `genroInit` only schedules at its very end (`genro.js:269`, `dojo.addOnLoad(genro,'start')`). So `dataSubscribe` → `getDataNode` → `this._data.getNode(...)` on `undefined` → TypeError thrown before `addOnLoad` is reached. No page boots.

2. **24 of the 46 wrappers do not match their primitive.** Not a sampling — the full table. Eleven pass positional arguments in the wrong order, which compiles and does the wrong thing: `ask`/`confirm` swap title and message (`genro_dlg.js:387` takes `(title, msg, …)`), `setInStorage`/`getFromStorage` are shifted by one against `(storageType, key, value, nameSpace)` (`genro.js:1924,1934`), `lockScreen(message, timeout)` puts the message into the `locking` boolean (`genro.js:2436`), `viewPDF(url, title)` feeds a URL to a parameter that expects a server-side filename and sets `forcedownload` from the title. Three more pass a string where the primitive reads `kw.something` (`formLoadRecord`, `formDelete`, `formNewRecord` — the last silently drops plain `{field: value}` defaults, which must be `default_<field>`). And **seven target something that does not exist**: `genro.unlockScreen` is nowhere in `gnrjs/`, `gridExport`/`gridSelectRow` have no mixin, and `open_record` / `new_record` / `refreshBadge` are published to zero subscribers.

3. **The schema has already drifted from the wrappers**, which is worse here than elsewhere because the whole point is that a consumer trusts `toolsManifest()`: `lockScreen.timeout` does not exist, `serverCall.async: boolean` reaches a string `mode` slot, `openPageInTab.returns` describes a value the primitive never returns, and three `returns: object|array` are actually `gnr.GnrBag` instances — not JSON-serializable, so unusable as a tool result without conversion. `_validateParams` also skips type checking entirely for `object` and never rejects unknown keys, so a typo'd parameter is silently dropped.

4. **No tests, and none are possible today** — there is no JS test infrastructure in the repo at all, so green CI on four Python versions covers only the two one-line list appends in `dojo_{11,20}.py`, not the 1794 lines of client code.

5. **The branch is unmergeable**: `gnrjs/gnr_d20/js/` was deleted on develop (`d7af14807`), and roughly half this diff re-adds it — `genro_cmd.js` there is byte-identical to the d11 copy, plus a modification to a `genro.js` that no longer exists.

One thing worth saying plainly since it is a design question rather than a bug: `serverCall` is an unrestricted RPC bridge reachable from a `dojo.subscribe` topic and from a datastore path, and `deleteRecord` / `formDelete` / `formSave` are destructive with no confirmation, where the framework path always asks (`genro_frm.js:720-736`). No server-side authorization is bypassed, but the UI guard rails are. If this comes back, those categories want an explicit opt-in rather than being on by default on every page.

Same conclusion as #906: this reads as a proof of concept that made its point and then sat while the ground moved. I would rather **close it and reopen from a fresh branch** when we actually want the catalog, than keep it open pending a rebase nobody will do on spec. If it does come back, the single cheapest quality gate is a smoke check that walks the catalog and asserts every target resolves — that alone catches seven of the findings above.

Your call as author.

PR #844, opened 2026-04-28
feat: add onEnter_shiftNewline option and pass event to onEnter callback

fporcari:
## Summary

Adds the `onEnter_shiftNewline` widget option: when set, Shift+Enter falls through to the browser's default newline instead of blurring the field and firing the `onEnter` callback, so a textarea with `onEnter` can still accept multiline input. Also passes the keyboard event to the `onEnter` callback.

`gnr.widgets.baseHtml` (`gnrjs/gnr_d11/js/genro_widgets.js`) is the only Enter dispatcher for the generic `onEnter` attribute, and `SimpleTextarea` reaches it through `baseDojo`, so the flag applies to textareas as intended.

- `genro_widgets.js:160` — `onEnter_shiftNewline` added to the `objectExtract` whitelist that moves attributes into `savedAttrs`.
- `genro_widgets.js:374-379` — the flag is read into a local and the keydown handler returns early on Shift+Enter, leaving the default behavior in place.
- `genro_widgets.js:382` — `setTimeout(callback, 100)` becomes `setTimeout(function(){ callback(evt); }, 100)`.

## Why

`gnrtools/resources/ai_chat/ai_chat.py:47` already ships `onEnter_shiftNewline=True` on its `simpleTextArea`. The attribute is currently unknown to the framework, so the AI chat drawer sends the message on Shift+Enter and cannot accept a multiline message. This is the missing counterpart of the SimpleTextarea `speech` work.

## Backward compatibility

`funcCreate(savedAttrs.onEnter)` wraps a string handler as `function(){body}`, and JavaScript ignores surplus arguments, so passing `evt` cannot break an existing zero-parameter handler. Every `onEnter=` call site was checked — `adm/resources/login.py`, `docu_components.py`, `resources/common/th/th_view.py`, `foundation/includedview.py`, `foundation/dialogs.py`, the JS-side handlers in `genro_frm.js` / `genro_dlg.js` / `genro_dev.js`, plus the usages in the downstream projects — and none declares a parameter.

Note that the event is only usable from JavaScript function handlers: string handlers would have to reach it via `arguments[0]`, since `funcCreate` is called without a `pars` argument. A follow-up passing `'evt'` as `pars` (the idiom used in `genro_grid.js:846` and `genro_components.js:4295`) would make it usable from string macros too.

## Test plan

- [x] Rebased on current `develop`; the `gnrjs/gnr_d20/` half of the original patch was dropped, that tree having been removed as a dead duplicate (refs #927).
- [x] `node --check gnrjs/gnr_d11/js/genro_widgets.js` passes.
- [x] Manual: on a textarea with `onEnter` + `onEnter_shiftNewline`, Shift+Enter inserts a newline and does not fire the callback; plain Enter still fires it. Worth confirming there is no stray trailing newline on the plain-Enter path, where the handler calls `evt.target.blur()` without `dojo.stopEvent(evt)`.


dgpaci (dismissed):
Tested through the interactive test page `projects/gnrcore/packages/test/webpages/inputfields/onenter.py` added on this branch (commit d4bbfbde3), following the framework convention for widget/JS features:

1. **Event passed to the onEnter callback** — a textbox whose `onEnter` handler reads `$1` and displays `$1.target.tagName` and `$1.shiftKey`. Pressing Enter shows the event data correctly (`$1` was `undefined` before this change), both with and without Shift.
2. **`onEnter_shiftNewline=True`** — a simpleTextArea acting like a chat input: Enter fires the handler (the message is copied to a target path and the field is cleared), while Shift+Enter inserts a newline without triggering `onEnter` and without blurring the field.
3. **Backward compatibility** — the same simpleTextArea without the new option: Shift+Enter still triggers `onEnter` exactly as before, so existing pages are unaffected.

All three scenarios behave as expected. LGTM.

dgpaci (approved):
Tested locally with an interactive test page (framework convention for widget/JS features) exercising three scenarios:

1. **Event passed to the onEnter callback** — a textbox whose `onEnter` handler reads `$1` and displays `$1.target.tagName` and `$1.shiftKey`. Pressing Enter shows the event data correctly (`$1` was `undefined` before this change), both with and without Shift.
2. **`onEnter_shiftNewline=True`** — a simpleTextArea acting like a chat input: Enter fires the handler (the message is copied to a target path and the field is cleared), while Shift+Enter inserts a newline without triggering `onEnter` and without blurring the field.
3. **Backward compatibility** — the same simpleTextArea without the new option: Shift+Enter still triggers `onEnter` exactly as before, so existing pages are unaffected.

All three scenarios behave as expected. LGTM.

PR #847, opened 2026-04-29
845 bug pysftp service fails with ed25519 and other modern private keys

### LINKED ISSUE #845: [BUG] pysftp service fails with Ed25519 (and other modern) private keys
## Problem Description

The SFTP service implementation at `projects/gnrcore/packages/sys/resources/services/ftp/pysftp/service.py` cannot establish a connection when the configured `private_key` is an **Ed25519** key (or ECDSA, or any modern key type other than RSA/DSS).

**Observed behavior:**
Calling the service raises `paramiko.ssh_exception.SSHException: unpack requires a buffer of 4 bytes` during authentication.

**Expected behavior:**
The service should successfully authenticate using any private key format supported by the underlying SSH library (RSA, DSS, ECDSA, **Ed25519**).

### Root cause

`pysftp` 0.2.9 (the only published version, **unmaintained since 2016**) hardcodes authentication to RSA/DSS only. From [`pysftp/__init__.py:145-171`](https://github.com/Yenthe666/pysftp/blob/master/pysftp/__init__.py):

```python
def _set_authentication(self, password, private_key, private_key_pass):
    if password is None:
        ...
        if isinstance(private_key, (AgentKey, RSAKey)):
            self._tconnect['pkey'] = private_key
        else:
            private_key_file = os.path.expanduser(private_key)
            try:                                  # try rsa
                self._tconnect['pkey'] = RSAKey.from_private_key_file(...)
            except paramiko.SSHException:         # if it fails, try dss
                self._tconnect['pkey'] = DSSKey.from_private_key_file(...)
```

When given a path to an Ed25519 key, both `RSAKey.from_private_key_file()` and `DSSKey.from_private_key_file()` raise `SSHException`, and the second exception is propagated to the caller. `paramiko.Ed25519Key` (available since paramiko 2.2) is never tried.

The `isinstance(private_key, (AgentKey, RSAKey))` check at line 158 also blocks the alternative of passing a pre-loaded `Ed25519Key` instance: it would not match the check and would fall into the `else` branch, which calls `os.path.expanduser` on a non-string and crashes.

**Updating paramiko does not fix the issue** — paramiko already supports Ed25519. The bug is in pysftp's hardcoded logic, and pysftp will not be fixed upstream (unmaintained for ~9 years).

## Reproduction Steps

1. Generate an Ed25519 key:
   ```
   ssh-keygen -t ed25519 -f /tmp/test_ed25519 -N ''
   ```
2. Configure a `sys.service` of type `ftp` with `implementation=pysftp/service`, host of any SSH server that accepts that key, and `private_key` set to `/tmp/test_ed25519`.
3. From a Python session in the instance:
   ```python
   site = GnrWsgiSite('<instance>')
   service = site.getService(service_type='ftp', service_name='<svc>')
   with service() as conn:
       print(conn.listdir())
   ```
4. The call to `service()` raises `SSHException: unpack requires a buffer of 4 bytes`.

The same key works fine via the `sftp` CLI or with paramiko's `SSHClient.connect(...)` directly, confirming the key and server are healthy.

## Environment

- genropy: develop (current)
- Python: 3.12
- paramiko: 4.x (any modern version — irrelevant)
- pysftp: 0.2.9 (the only published version)
- OS: macOS / Linux (platform-independent — pure Python issue)

## Resolution Options

Two viable paths. Both require modifying `pysftp/service.py`. The decision is between minimum-diff and proper fix.

### Option 1 — Patch the existing service to pre-load the key with paramiko

In `Service.__call__`, before constructing `pysftp.Connection`:

1. Detect the key file format (read the PEM header — `-----BEGIN OPENSSH PRIVATE KEY-----` for Ed25519/modern, `-----BEGIN RSA PRIVATE KEY-----`, etc.) and load it with the appropriate `paramiko.<Type>Key.from_private_key_file()`.
2. Subclass `pysftp.Connection` (or monkey-patch `_set_authentication`) to accept any `paramiko.PKey` instance, since the upstream `isinstance` check is too narrow.
3. Pass the pre-loaded `PKey` instance to the patched connection.

**Pros**
- Minimum diff. Keeps the existing API and the existing dependency.
- Lower risk of regressions for current callers (they keep getting a `pysftp.Connection`).
- Faster to ship.

**Cons**
- Still depends on an unmaintained library. Future bugs in pysftp will keep biting (e.g. host key handling, transport options, Python 3.13+ compatibility).
- The subclass/monkey-patch is fragile: any pysftp internal change breaks it. Practically not a concern (pysftp is dead) but conceptually ugly.
- Doesn't unlock other pysftp shortcomings (e.g. the existing `cnopts.hostkeys = None` is an explicit security regression compensating for pysftp's awkward host-key API — same constraint stays).

### Option 2 — Rewrite the service on top of paramiko directly

Replace the body of `Service.__call__` with `paramiko.SSHClient` + `client.open_sftp()`. Keep the same public API (`__call__`, `downloadFilesIntoFolder`, `uploadFilesIntoFolder`, `default_path` semantics) so callers are unaffected.

**Pros**
- Removes a dead, single-purpose dependency. paramiko is already a transitive dep via pysftp.
- Native support for Ed25519, ECDSA, encrypted keys, agent keys, all current paramiko features.
- More control over host-key verification (we can stop relying on the `cnopts.hostkeys = None` workaround at line 50 and offer a proper known_hosts policy).
- Future maintenance lands on a healthy library.

**Cons**
- Larger diff. Higher chance of subtle behavior change vs. pysftp (e.g. `pysftp.Connection.put` vs. `paramiko.SFTPClient.put` semantics, default paths, exception types) — needs a careful pass.
- All current callers should be regression-tested. The codebase uses this service in `SftpDirectoryResolver` (`projects/gnrcore/packages/sys/lib/services/ftp.py:54-108`) and in caller-side `with service() as conn: conn.listdir()/.put()/.get()/.stat()` patterns; the wrapper must expose `listdir`, `stat`, `put`, `get`, `isdir`, `close`, and context-manager behavior with the same signatures.
- Slightly more code to maintain in genropy (but in exchange for dropping a third-party dep).

**Recommendation:** Option 2. The cost is a one-time controlled migration; the benefit is removing an abandoned library that already blocks modern key types and will keep producing similar issues.

## Related UI issue — `Private key` field is ambiguous

Independent of the fix above, the `ServiceParameters` form at `projects/gnrcore/packages/sys/resources/services/ftp/pysftp/service.py:98-107` has a UX problem:

```python
fb.textbox(value='^.private_key', lbl='Private key')
```

The label "Private key" plus a single-line `textbox` does not communicate whether the user is supposed to:
- paste the key contents, or
- enter a filesystem path to the key.

Today the field is **only a filesystem path** (because `pysftp.Connection` only accepts a path), but the UI suggests otherwise — and storing a multi-line PEM in a textbox doesn't even round-trip cleanly through XML serialization (newlines collapse to spaces, breaking PEM parsing).

Three options to fix the UI, in increasing order of safety:

1. **Rename** the label to `Private key path` and keep the `textbox`. Smallest change.
2. **Switch to a `textArea`** and accept the key contents inline (the implementation writes it to a tempfile or loads it via `paramiko.<Type>Key.from_private_key(io.StringIO(...))`). This avoids deploying separate key files alongside the instance. Need to be mindful of how the value is serialized (use CDATA or preserve newlines).
3. **Use an uploader** (file picker), as we already do in other services. Stores the contents in the service parameters bag, decoded server-side. Most user-friendly but requires deciding how to represent the binary/PEM content in the bag.

Whichever shape is chosen, the rewrite (Option 2 above) is a good moment to fix the form too — paramiko accepts both paths and file-like objects, so the implementation becomes trivial.

cgabriel:
New implementation of the ftp service using sftp as a transport. This is using just paramiko, and not an external dependency like pysftp (which is abandonware, and doesn't support newer key types out of the box).

Pysftp implementation has been marked with deprecation warning, in order to have user switch to this new implementation.

Paramiko is already a dependency of the framework, so no new external dependencies are added.

### DIFF
--- a/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/component.py
+++ b/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/component.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+
+from gnr.core.gnrbag import DirectoryResolver
+from gnr.core.gnrdecorator import public_method
+from gnr.web.gnrbaseclasses import BaseComponent
+from gnr.web.gnrwebstruct import struct_method
+
+class SftpClient(BaseComponent):
+    py_requires='public:Public'
+
+    @struct_method
+    def sftp_sftpClientLayout(self,pane,ftpname=None,
+                            datapath='.sftpclient',destdir=None,remotedir=None,**kwargs):
+        bc = pane.borderContainer(datapath=datapath,_anchor=True,**kwargs)
+        self.sftp_remoteTree(bc.roundedGroupFrame(region='left',title='!!Remote',
+                            datapath='.remote',width='50%',
+                            splitter=True),ftpname=ftpname,remotedir=remotedir)
+        self.sftp_localTree(bc.roundedGroupFrame(region='center',title='!!Local',
+                            datapath='.local'),ftpname=ftpname,destdir=destdir)
+
+    def sftp_remoteTree(self,frame,ftpname=None,remotedir=None):
+        resolver = self.getService('ftp',ftpname).sftpResolver()
+        frame.data('.tree',resolver())
+        self.sftp_fileTree(frame,nodeId='%s_src' %ftpname,topic='%s_upload' %ftpname)
+        frame.dataRpc(None,self.sftp_uploadFiles,ftp=ftpname,
+                    _onResult="""kwargs._dropnode.refresh(true);""",
+                    **{'subscribe_%s_upload' %ftpname:True})
+
+    def sftp_localTree(self,frame,ftpname=None,destdir=None):
+        resolver= DirectoryResolver(destdir or self.site.getStatic('site').path())
+        frame.data('.tree',resolver())
+        self.sftp_fileTree(frame,nodeId='%s_dest' %ftpname,
+                            topic='%s_download' %ftpname)
+        frame.dataRpc(None,self.sftp_downloadFiles,ftp=ftpname,
+                    _onResult="""kwargs._dropnode.refresh(true);""",
+                        **{'subscribe_%s_download' %ftpname:True})
+
+
+    def sftp_onDrag(self):
+        return """var children=treeItem.getValue('static')
+                  if(!children){
+                      dragValues['fsource']=[treeItem.attr.abs_path];
+                      return
+                  }
+                   result=[];
+                   children.forEach(function(n){
+                        if (n.attr.checked && !n._value){result.push(n.attr.abs_path);
+                    }},'static');
+                   dragValues['fsource']= result;
+               """
+
+    @public_method
+    def sftp_downloadFiles(self,sourcefiles=None,destfolder=None,ftp=None,**kwargs):
+        self.getService('ftp',ftp).downloadFilesIntoFolder(sourcefiles=sourcefiles,
+                                                destfolder=destfolder,**kwargs)
+
+    @public_method
+    def sftp_uploadFiles(self,sourcefiles=None,destfolder=None,ftp=None,**kwargs):
+        self.getService('ftp',ftp).uploadFilesIntoFolder(sourcefiles=sourcefiles,
+                                                destfolder=destfolder,**kwargs)
+
+
+    def sftp_fileTree(self,pane,topic=None,**kwargs):
+        tree = pane.treeGrid(storepath='.tree',hideValues=True,
+                      selectedLabelClass='selectedTreeNode',
+                      selected_abs_path='.abs_path',selected_file_ext='.file_ext',
+                      checked_abs_path='.checked_abs_path',
+                       autoCollapse=True,
+                      onDrag_fsource=self.sftp_onDrag(),
+                      headers=True,draggable=True,dragClass='draggedItem',
+                      onDrop_fsource="""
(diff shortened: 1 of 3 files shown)


dgpaci (commented):
Tested integrating this on erpylight (issue softwellsrl/erpylight#20 — SFTP polling batch for MAV ricevuti). With the original `pysftp/service.py` the same service config crashed at connection time with `SSHException: unpack requires a buffer of 4 bytes` (the bug from #845, paramiko trying DSS as fallback for an Ed25519 key). Switching the service `implementation` to `paramiko/service` fixes it cleanly:

- ✅ Connection with an Ed25519 private key
- ✅ `listdir()` on the configured `root`
- ✅ `isdir()` / `stat()` for filtering directories
- ✅ `get(remotepath, localpath)` download into a local temp dir
- ✅ `chdir(self.root)` correctly scopes operations to the configured folder

One gap on the wrapper surface area surfaced during the integration: the `_Connection` wrapper does not forward `remove`/`unlink`, so downstream callers that delete the remote file after a successful import lose drop-in compatibility with `pysftp.Connection`. Tracked separately as #848 — non-blocking for this PR.

cgabriel:
Closing, this must be an hotfix

PR #849, opened 2026-04-29
845 bug pysftp service fails with ed25519 and other modern private keys

### LINKED ISSUE #845: [BUG] pysftp service fails with Ed25519 (and other modern) private keys

cgabriel:
New implementation of the ftp service using sftp as a transport. This is using just paramiko, and not an external dependency like pysftp (which is abandonware, and doesn't support newer key types out of the box).

Pysftp implementation has been marked with deprecation warning, in order to have user switch to this new implementation.

Paramiko is already a dependency of the framework, so no new external dependencies are added.

close #845


dgpaci (changes requested):
The hotfix port from #847 is incomplete: `paramiko/service.py` is **byte-identical** to `paramiko/component.py`, so the actual paramiko `Service` implementation is missing from this PR.

## What I see

At HEAD `38ab0e598`:

```
paramiko/service.py    size=4371  sha=8bc7fa2acf…
paramiko/component.py  size=4371  sha=8bc7fa2acf…
```

Both files have the **same blob SHA** — they contain the same `SftpClient(BaseComponent)` UI code. The `Service(SftpService)` class with the `_Connection` wrapper, `__call__`, `downloadFilesIntoFolder` and `uploadFilesIntoFolder` — i.e. the ~139 lines that actually implement SFTP over paramiko and that close #845 — are not present anywhere in the PR.

In other words: this PR adds a UI component twice and no service. It does **not** fix #845 in its current state.

## Comparison with #847

PR #847 (closed, was on `develop`) contained three changes:

1. `paramiko/component.py` — new `SftpClient` UI component ✅ (present here)
2. `paramiko/service.py` — new `Service(SftpService)` paramiko implementation ❌ (missing here, replaced by a duplicate of component.py)
3. `pysftp/service.py` — deprecation warning + label fix ✅ (present here)

Only items 1 and 3 made it into #849. The most important file — the new service — is missing.

The pysftp deprecation block in this PR also has a couple of stray empty lines compared to #847; not blocking, but worth tidying up while you're at it.

## What I'd suggest

Easiest path: reset the `hotfix/845-…` branch and cherry-pick the original commits from `845-bug-pysftp-service-fails-with-ed25519-and-other-modern-private-keys`:

- `0d76ffc04` — pysftp deprecation warning
- `4c0d194cd` — new paramiko Service implementation
- `8a71af6dc` — flake8 cleanup

That branch was already reviewed and tested; reusing those commits avoids redoing the porting by hand and avoids exactly the kind of file-overwrite that happened here.

Once `paramiko/service.py` actually contains the `Service` class, I'm happy to re-review and repeat the positive opinion from #847.

dgpaci (approved):
The port is now complete and functionally equivalent to #847.

Verified at HEAD `b8b7b903`:

- `paramiko/component.py` — `SftpClient` UI component ✅
- `paramiko/service.py` — paramiko `Service(SftpService)` with `_Connection` wrapper, blob SHA `7688d0f40f…` matching the file from #847 ✅
- `pysftp/service.py` — deprecation warning + `'Private key' → 'Private key file path'` label fix ✅

Diff against #847 is reduced to two extra blank lines in `pysftp/service.py` — cosmetic, not blocking.

Repeating the positive opinion already given on #847: approach is sound (paramiko is already a framework dependency, no new external deps), pysftp deprecation is signalled cleanly, and the new service preserves the `pysftp`-compatible interface used by `SftpDirectoryResolver` and the upload/download helpers. Modern key types (Ed25519 etc.) now work via paramiko's `key_filename` connect option.

Approving on the same grounds as #847.

fporcari (dismissed):
Overall the change is in the right direction: dropping `pysftp` (abandonware) in favor of a direct `paramiko` implementation is the correct fix for #845, the `SftpService` interface is preserved, `preserve_mtime` is reimplemented correctly on both `get` and `put`, and no new dependencies are introduced. Nice work.

A few things should be addressed before merge:

### Must fix

1. **Deprecation warning placement** — `pysftp/service.py:14-18`
   The `warnings.warn(...)` is at module top level, so it fires once at import time, even when nobody is actually using the pysftp service (the resource loader may import the module during autodiscovery). Move it inside `Service.__init__` so it warns only when a pysftp-based service is actually instantiated, and on every instantiation.

2. **`thermo_wrapper(thermo_wrapper)` bug** — `paramiko/service.py:100` and `:121`
   ```python
   if thermo_wrapper:
       sourcefiles = thermo_wrapper(thermo_wrapper)
   ```
   This is a pre-existing bug copied from the pysftp implementation: it passes the wrapper to itself instead of wrapping `sourcefiles`. Should be `sourcefiles = thermo_wrapper(sourcefiles)`. Since the new file is being written from scratch, please fix it here (and ideally in the pysftp file too for consistency).

3. **`component.py` duplication**
   `paramiko/component.py` is a near-verbatim copy of `pysftp/component.py` (only a couple of removed comments and whitespace differ). 89 duplicated lines means a future fix has to be made twice. Please factor `SftpClient` into a shared module (e.g. `lib/services/ftp.py` or a common resource) and import it from both backends, or keep a single component since it is backend-agnostic.

### Should fix

4. **`AutoAddPolicy()`** — `paramiko/service.py:67`
   Silently accepts any host key (MITM open by default). It matches the previous pysftp behavior (`cnopts.hostkeys = None`), so functional parity is preserved, but at minimum add a comment explaining the choice; ideally expose a `known_hosts`/`verify_host_key` parameter as a follow-up.

5. **Redundant exception** — `paramiko/service.py:24`
   `except (FileNotFoundError, IOError)`: in Python 3 `IOError` is an alias of `OSError` and `FileNotFoundError` is a subclass of it, so `except IOError:` (or `except OSError:`) is enough.

### Nits

6. `pysftp/service.py:19-22`: three consecutive blank lines after the `warnings.warn` block (flake8 E303).
7. `paramiko/service.py:1`: `#!/usr/bin/env python` shebang on a library file is superfluous.
8. No tests for the new implementation. End-to-end testing requires an SFTP server, but the `_Connection` wrapper (e.g. `isdir` not raising on missing paths, `preserve_mtime` behavior) can be unit-tested by mocking `paramiko.SFTPClient` — here mocks are legitimate because we are isolating an external network dependency.

Happy to re-review once 1–3 are addressed.

---

genro:
Concordo con i 3 must-fix di @fporcari. In particolare 1 e 2 sono bloccanti:

- `thermo_wrapper(thermo_wrapper)` (service.py:97 e 115) e' funzionalmente rotto: il wrapper finisce per ricevere se stesso al posto della lista file. Bug ereditato dal pysftp, ma dato che il file paramiko viene scritto da zero va corretto qui.
- `warnings.warn` a livello modulo in `pysftp/service.py:15-19` scatta a import time anche durante l'autodiscovery delle resource: va dentro `Service.__init__`.
- Sul punto 3 (component.py duplicato): confermo via diff, le 89 righe sono near-verbatim — solo header rimosso e qualche spazio. Una futura modifica andrebbe fatta due volte.

Sui should-fix concordo: `AutoAddPolicy()` mantiene parita' funzionale ma andrebbe almeno commentato, e `except (FileNotFoundError, IOError)` puo' diventare `except OSError`.

genro (changes requested):
Verificato lo stato dei fix richiesti dopo i commit di stamattina.

**Must-fix indirizzati**:
- #1 deprecation warning spostata in `Service.__init__` (`pysftp/service.py:20-25`) ✅
- #3 `SftpClient` factorizzato in `lib/services/ftp.py`, i due `component.py` ora sono solo re-export ✅

**Must-fix #2 — fixato solo a metà**:

Il commit `61938aa` ha corretto `paramiko/service.py:100` (download) ma in `paramiko/service.py:121` (`uploadFilesIntoFolder`) c'è ancora:

```python
if thermo_wrapper:
    sourcefiles = thermo_wrapper(thermo_wrapper)
```

Stesso bug anche in `pysftp/service.py:91`, non toccato. Va corretto in entrambe le occorrenze (e idealmente anche nel pysftp originale, come scritto da fporcari).

**Should-fix indirizzati**: #4 commento `AutoAddPolicy` ✅, #5 `IOError` ✅, nit #6 E303 ✅, nit #7 shebang ✅.

**Aperto**: nit #8 test (non bloccante).

Sblocco al merge non appena il must-fix #2 è completo.

genro (dismissed):
Verificato lo stato dei fix richiesti dopo i commit di stamattina.

**Must-fix indirizzati**:
- #1 deprecation warning spostata in `Service.__init__` (`pysftp/service.py:20-25`) ✅
- #3 `SftpClient` factorizzato in `lib/services/ftp.py`, i due `component.py` ora sono solo re-export ✅

**Must-fix #2 — fixato solo a metà**:

Il commit `61938aa` ha corretto `paramiko/service.py:100` (download) ma in `paramiko/service.py:121` (`uploadFilesIntoFolder`) c'è ancora:

```python
if thermo_wrapper:
    sourcefiles = thermo_wrapper(thermo_wrapper)
```

Stesso bug anche in `pysftp/service.py:91`, non toccato. Va corretto in entrambe le occorrenze (e idealmente anche nel pysftp originale, come scritto da fporcari).

**Should-fix indirizzati**: #4 commento `AutoAddPolicy` ✅, #5 `IOError` ✅, nit #6 E303 ✅, nit #7 shebang ✅.

**Aperto**: nit #8 test (non bloccante).

Sblocco al merge non appena il must-fix #2 è completo.

fporcari (dismissed):
Most of the items are addressed (deprecation moved into `__init__`, `SftpClient` factored into `lib/services/ftp.py`, `AutoAddPolicy` annotated, redundant exception simplified, shebang removed). One residual issue:

### Still to fix

**`thermo_wrapper(thermo_wrapper)` in `uploadFilesIntoFolder`** — `paramiko/service.py:122`
```python
if thermo_wrapper:
    sourcefiles = thermo_wrapper(thermo_wrapper)
```
The fix was applied to `downloadFilesIntoFolder` (line 100 → `thermo_wrapper(sourcefiles)`) but the upload counterpart was missed. Same one-line change needed:
```python
sourcefiles = thermo_wrapper(sourcefiles)
```

### Nits (optional, won't block)

- A few trailing-whitespace lines around the fixed blocks (e.g. `paramiko/service.py:75`, `:101` and `pysftp/service.py:64`) — flake8 W291/W293.

Once the upload `thermo_wrapper` is fixed, this is good to go from my side.

---

fporcari (approved):
All previously requested changes are addressed:

- Deprecation warning moved into `Service.__init__` (`pysftp/service.py:20-24`), so it only fires when the service is actually instantiated.
- `thermo_wrapper(sourcefiles)` now correct in both `downloadFilesIntoFolder` and `uploadFilesIntoFolder`, in both pysftp and paramiko implementations.
- `SftpClient` extracted to `lib/services/ftp.py` and the two `component.py` files are now single-line re-exports — no more duplication.
- `AutoAddPolicy` annotated with a `FIXME` explaining the MITM trade-off and pointing at a possible follow-up (allowed host keys in service config).
- Redundant `(FileNotFoundError, IOError)` simplified to `IOError`.
- Stray shebang removed.

LGTM. Two minor nits, non-blocking:

- A few trailing-whitespace lines remain (`paramiko/service.py:75, 101`, `pysftp/service.py:69, 73, 95, 99`) — flake8 W293. Worth a quick `autopep8 -i` before merge if the project enforces flake8 on CI.
- The `AutoAddPolicy` FIXME has a typo: "possibile" → "possible".

Approving — feel free to clean up the nits before merge.

---

genro (approved):
Verificato il commit `e60bd13`: il must-fix #2 è ora completamente indirizzato in entrambe le occorrenze:

- `paramiko/service.py:121` (uploadFilesIntoFolder): `sourcefiles = thermo_wrapper(sourcefiles)` ✅
- `pysftp/service.py:91` (uploadFilesIntoFolder): `sourcefiles = thermo_wrapper(sourcefiles)` ✅

Tutti i must-fix e should-fix segnalati nella review iniziale di fporcari sono stati indirizzati. Resta aperto solo il nit #8 (test) che era esplicitamente non bloccante.

CI verde su build 3.11/3.12/3.13/3.14.

Approvo.

PR #852, opened 2026-04-29
Migrate gnrasync to asyncio/aiohttp; drop legacy GnrPdb debugger

genro:
## Summary

Migrazione completa del server WebSocket di GenroPy da **Tornado** ad **asyncio + aiohttp**, in 11 commit incrementali.

Il server `gnrasync` non era piu' in uso da tempo, quindi questa PR rappresenta un punto di partenza pulito per riprendere il lavoro sui WebSocket senza il vincolo di tornado.

## Update (2026-05-09): legacy GnrPdb debugger removed

In addition to the asyncio migration, this PR now includes the
**removal of the legacy PDB debugger** (`GnrPdb` + `DebugSession` +
gnride PDB-side UI + JS step/continue handlers). The system dates back
to 2015, was de-facto superseded by the existing `--debugpy` wiring
(debugpy + VS Code), and was no longer in active use by the team. The
removal is a single commit (`2bd3fd1f46`, 21 files, +26/-1217),
delete-only, with no application-level replacement in the framework:
the modern path is debugpy + VS Code. See breaking changes for the
operational details.

## Follow-up PR

A second PR — **#895 _Introduce neutral AsyncSession core_** — is
already open. It reintroduces, in a clean and modern form, the
async-session pattern that had grown silently inside the legacy
`DebugSession`. That PR depends on this one and is meant to be merged
right after.

**Please activate the review on #895 too**, since cgabriel will be
away from May 15 to June 6 and the goal is to land both PRs together
before that window closes.

## Cosa cambia

- `gnrasync.py` riscritto su `aiohttp.web` + `asyncio` (UnixSite + TCPSite, SSL preservato).
- `GnrWebSocketHandler` -> `GnrWebSocketSession` su `aiohttp.WebSocketResponse`.
- `SharedObject` / `SharedObjectsManager` portati a primitive asyncio (`asyncio.Lock`, `loop.run_in_executor`, `create_task`).
- POST `/wsproxy` (bridge per gunicorn workers) portato ad aiohttp.
- `DebugServer` portato a `asyncio.start_unix_server` con pipeline a 4 code preservata.
- `tornado_wsgi.py` eliminato; `tornado` rimosso da `pyproject.toml`.
- CLI dev: `--tornado` sostituito da `--async-port N` (spawna un processo `gnrasync` separato).
- JS lato client: aggiunto `websockets_port` per dev mode (gnrasync su porta separata) con fallback a `window.location.port` per produzione (nginx proxy). Buffer dei messaggi pre-OPEN. Guard `if(!this.socket)` su `send`/`call`.

## Breaking changes

1. **Socket name**: `sockets/async.tornado` -> `sockets/async.sock`. I config nginx/supervisor degli existing deploy vanno aggiornati prima di redeployare. Nessun fallback fornito.
2. **Modalita' Tornado-serves-WSGI rimossa**: `gnr web serve --tornado` e `GnrAsyncServer(web=True)` non esistono piu'. Lo sviluppo locale usa werkzeug; la produzione WSGI usa gunicorn (`gnr web stack`). `gnrasync` ora gestisce solo WebSocket, `/wsproxy` push, e il bridge TCP del pdb debugger.
3. **`GnrWsgiSite(tornado=...)`** kwarg rimosso (era accettato ma mai letto).
4. **Legacy PDB debugger removed**: `GnrPdb`, `GnrPdbClient`,
   `DebugSession`, `do_pdb_command`, `debug_queues`, `debugger.sock`,
   the entire gnride PDB-side UI (`do_pdb_out_*`, `sendCommand`,
   `do_step*`, `do_continue`, `isDebugging`, `onPdbAnswer_*`), the
   client JS (`onDebugstep`, `addToDebugged`, `removeFromDebugged`,
   `suspend_call`, `debugged_rpc`), and the `pdb_*` CSS classes are
   gone. Demos `test15/ws/gnrpdbtest.py` and `test15/ws/debugger.py`
   deleted. **No application-level replacement in the framework**:
   the modern debug path is `gnr web stack --debugpy` + VS Code (or
   VS Code Remote-SSH for prod-like sessions). The wiring is already
   in `serverwsgi.py:226+` and `genro_rpc.js:508-510` (which sets
   `timeout=0` automatically on RPC calls during a debug session).
   Note: point 2 above mentions "the pdb debugger TCP bridge" — that
   bridge and the `debugger.sock` are also removed by this commit.

## Test plan

- [x] flake8 zero issue sui file modificati
- [x] 6/6 nuovi end-to-end test in `gnrpy/tests/web/gnrasync_test.py` verdi (real aiohttp ClientSession, no transport mock)
- [x] suite completa: 1516/1516 passati (esclusi 2 pre-esistenti su develop in `test_compiler_coverage.py`)
- [x] nessuna nuova dipendenza (aiohttp gia' presente)
- [x] guardia contro reintroduzione di `import tornado` nel test suite
- [x] zero residual references to the legacy debugger (no match for
      `GnrPdb`, `debug_queues`, `debugger.sock`, `pdb_*`, etc.)
- [x] full test suite after the legacy debugger removal: 1700 passed,
      67 skipped, 22 failed (pre-existing on `develop`, unrelated)
- [ ] verificare manualmente che un deploy esistente con socket rinominato a `async.sock` continui a funzionare
- [ ] verificare le test page `test/websocket/collaborative` e `test/websocket/shared_objects` con `--async-port`

## Note

- Public surface preservata: tutti i simboli del legacy `gnrasync.py` sono esportati con la stessa firma. I 4 callsite tornado (`gnrasync.py`, `gnrwebsockethandler.py`, `gnrdeploy.py`, supervisor template) sono stati tutti migrati.
- Approccio incrementale: ogni commit ha messaggio dettagliato che spiega cosa e perche' (utile per future bisection).


fporcari (dismissed):
Approving — the migration is technically solid and the direction is right (Tornado dependency dropped, dead WSGI-via-Tornado mode removed, public API surface preserved). The 11-commit incremental structure is bisection-friendly and the test suite uses real `aiohttp.ClientSession` against an in-process server, which is the right call.

A few non-blocking observations worth addressing before/after merge:

**Shutdown vs autoSave race.** `SharedObject.onShutdown` schedules `save()` via `loop.create_task` and returns immediately. If the loop tears down before the task finishes, state is lost. The legacy code was fire-and-forget too, but the asyncio version makes the window more visible. Worth `await`ing the pending saves inside `_cleanup` (or gathering them with a bounded timeout) before closing the runner.

**`onSubscribePage(page)` signature mismatch.** In `SharedObject.subscribe()` we call `self.onSubscribePage(page)` (a page object) while subclasses (`SharedLogger`) declare `onSubscribePage(self, page_id)`. Inherited bug from the legacy code, but the migration was a good chance to align it.

**Cache `_get_async_port`.** `gnrwebpage.py:_get_async_port` opens the file on every `build_arg_dict` call (i.e. every page render). Cache it next to `_wsk_enabled`.

**Boot race for `--async-port`.** If the gnrasync child has not yet written `sockets/async_port` when the first client builds its arg_dict, `websockets_port` falls back to `window.location.port` (the werkzeug port, wrong). Quasi sempre va bene in pratica, ma non è garantito.

**E2E coverage gap.** The two manual checks in the test plan (existing deploy with renamed socket, `test/websocket/collaborative` and `test/websocket/shared_objects` with `--async-port`) are still open. All automated tests are Python-side: the JS path (dynamic port, pre-OPEN buffer, `if(!this.socket)` guards) has never been exercised in a real browser. Worth ticking those before declaring victory.

**`do_call` simplification.** The "double `getWsMethod` is not load-bearing" change has no test backing it (page is a MagicMock everywhere). Probably innocuous, but it is the kind of thing that surfaces in production.

**Subprocess monitoring.** `_spawn_async_server` does not watch the child; if it crashes the parent does not notice and the stale `async_port` file lingers on SIGKILL. Low priority but worth a TODO.

**Single source of truth for `AsyncWebSocketHandler`.** The legacy class still lives in `gnrwsgisite_proxy/gnrwebsockethandler.py` and is subclassed by the override in `gnrasync.py`. Consolidating later would help.

None of the above blocks merge — they are polish items that can land in a follow-up. Good work.

gnrpy/gnr/web/gnrwebpage.py
```diff
@@ -391,8 +391,15 @@ def wsk_enabled(self):
         if not hasattr(self, '_wsk_enabled'):
             self._wsk_enabled = self.wsk and self.getPreference('experimental.wsk_enabled',pkg='sys')
         return self._wsk_enabled
-    
-    @property 
+
+    def _get_async_port(self):
+        port_file = os.path.join(self.site.site_path, 'sockets', 'async_port')
```
cgabriel: The port configuration must be able to be defined also via an env var, which can contains an URL rather than a socket path, if the async service is deployed in a different container.

gnrpy/gnr/web/serverwsgi.py
```diff
@@ -330,98 +334,116 @@ def serve(self):
...
+            logger.info("Shutting down")
+
+    def _spawn_async_server(self, site_name, async_port):
+        """Spawn the gnrasync server as a child process for `--async-port`.
+
+        Registers an atexit cleanup so SIGTERM is sent on parent exit.
+        Skipped when running from the werkzeug reloader child to avoid
+        spawning two gnrasync processes.
+        """
+        gnrasync_bin = shutil.which('gnrasync')
```
cgabriel: Please use the correct command name. 'gnrasync' is deprecated, it's the old name, since it's an internal call we should use the new naming scheme, which is 'gnr app async' - btw, it should be moved to gnr.web, don't see a valid scope for async in the gnr.app package.

gnrpy/tests/web/gnrasync_test.py
```diff
@@ -1,9 +1,201 @@
...
+import json
+import os
 import sys
+import urllib.parse
+
+import pytest
+
+if sys.platform == 'win32':
+    pytest.skip(
+        'gnrasync requires AF_UNIX sockets which are not fully supported on Windows',
```
cgabriel: Not necessarily, deployment can also be made using AF_INET which are supported on Windows.

gnrpy/tests/web/gnrasync_test.py
```diff
@@ -1,9 +1,201 @@
...
+
+@pytest.mark.asyncio
+async def test_ping_pong(tmp_path):
+    server, fake_page = _build_server(tmp_path, PORT_BASE)
+    run_task = asyncio.create_task(server._run())
+    try:
+        await asyncio.sleep(0.4)
+        server.sharedStatus.registerPage(fake_page)
+        async with aiohttp.ClientSession() as s:
+            async with s.ws_connect(f'ws://127.0.0.1:{PORT_BASE}/websocket') as ws:
```
cgabriel, line 68: This test is made using AF_INET, so the initial check on windows Vs AF_UNIX is irrelevant.

cgabriel (changes requested):
Please review all the comments attached to implementation.

genro:
@cgabriel — addressed all 4 inline review comments in commit 3b5b95b.

**1. gnrwebpage.py:396 — env var + URL support.** Replaced `_get_async_port()` with the cached `async_endpoint` property. New resolution order:

1. `GNR_ASYNC_URL` env var (for deployments where the async service runs in a separate container; value can be a port number or a `ws://`/`wss://` URL)
2. `sockets/async_port` file (existing behavior)
3. `None`

The propagated kwarg is renamed `websockets_port` → `websockets_endpoint`. The JS bundles (d11 and d20) handle the URL form in `gnrwebsocket.js`: if the value starts with `ws://` or `wss://`, it's used as the base URL; otherwise it falls back to treating it as a port (existing behavior).

**2. serverwsgi.py:426 + module location.** Moved `gnrpy/gnr/app/cli/gnrasync.py` → `gnrpy/gnr/web/cli/gnrasync.py`. Updated the `pyproject.toml` entry point; kept the `gnrasync` script as a deprecated alias (pointing to the new module) so existing supervisord configs that invoke it keep working — `gnr web async` is now the documented invocation. `_spawn_async_server` invokes `gnr web async <site> -p <port>` instead of `shutil.which('gnrasync')`.

**3–4. gnrasync_test.py:18,74 — Windows skip.** Removed the module-level `pytest.skip(... allow_module_level=True)`. The only test that genuinely requires AF_UNIX is `test_unix_sockets_created` (asserts `async.sock` and `debugger.sock` are created), which now carries a targeted `@pytest.mark.skipif(sys.platform == 'win32')`. All other tests bind on AF_INET (TCP) and run on Windows.

I have not touched `gnrdeploy.py:572-573, 636-637`, which still reference the legacy `gnrasync` binary in supervisord program sections — these continue to work via the deprecated alias. Happy to update those in a follow-up if you'd like a single-pass cleanup across the codebase.

Verified locally: 6/6 tests pass in `gnrasync_test.py`, flake8 clean, `gnr.web.cli.gnrasync` imports correctly.

gnrpy/gnr/web/gnrwebpage.py
genro: Addressed in 3b5b95b: `GNR_ASYNC_URL` env var (port or `ws://`/`wss://` URL) on top of the existing `sockets/async_port` file.

gnrpy/gnr/web/serverwsgi.py
genro: Addressed in 3b5b95b: CLI moved to `gnr.web.cli`, spawned via `gnr web async`; legacy `gnrasync` kept as deprecated alias.

gnrpy/tests/web/gnrasync_test.py
genro: Addressed in 3b5b95b: removed the module-level Windows skip; tests on AF_INET now run on Windows too.

gnrpy/tests/web/gnrasync_test.py
genro: Addressed in 3b5b95b: removed the irrelevant module-level skip; only `test_unix_sockets_created` (the one that actually uses AF_UNIX) carries `@pytest.mark.skipif`.

cgabriel (changes requested):
gnr web serve's '-a' options doesn't allow to specify the async bind interface, only a port. When using AF_INET, the async service binds by default on all interfaces, without the option to choose one or more interfaces to bind on. This is a limitation that force the use of other methods to limit traffic access to the service.
 
I suggest to be allowed to define on 'gnr web async' also the interface when spawning the async server, being it along with the wsgi one or not. Surely, I'd introduce a safer default, like binding on loopback interface only.

I've also tried to use Test/Websocket/[Collaborative,Websocket,Shared objects] webpages, but none of them appears to be operationl. Wsk is enabled, and spawned via gnr web serve.

genro:
@cgabriel — quick proposal before you start digging.

On the points you raised, two things came up that we can simplify
across the board, beyond just fixing the test pages:

**1. Auto-spawn `gnr web async` from `gnr web serve` (dev only)**
If the siteconfig declares `<wsgi websockets="true"/>`, `gnr web serve`
now picks a free port and spawns the async subprocess automatically.
`-a PORT` keeps working as an explicit override. Scope is strictly
`gnr web serve` (werkzeug dev path); `gnr web stack`, `gnr web async`
standalone, supervisord/k8s templates are untouched — production still
runs as today, with two separately orchestrated services.

**2. Kill switch instead of opt-in for the DB preference**
The double gate today (siteconfig + `experimental.wsk_enabled = True`)
is collapsed to a single source of truth: siteconfig. The preference
becomes `experimental.wsk_disabled` (default false = enabled), so it
no longer needs to be set to activate WS, and it stays useful as an
emergency runtime kill switch in production (no rebuild/redeploy
needed if something goes wrong).

Result for the test pages procedure:

Before:
1. siteconfig: `<wsgi websockets="true"/>`
2. DB pref: `sys/experimental/wsk_enabled = True`
3. `gnr web serve sandboxpg -a 8086`
4. log check
5. test pages

After:
1. siteconfig: `<wsgi websockets="true"/>`
2. `gnr web serve sandboxpg`
3. test pages

Production stays exactly as today: gunicorn + `gnr web async` as
two separate orchestrated services, no auto-spawn involved.

I'm not pushing this — I don't want to walk over whatever you've
already started. Below is the diff (5 files, +65 / -13, verified
applies cleanly on `feature/asyncio-migration`):

<details>
<summary>Proposed diff (apply with <code>git apply</code> or use as reference)</summary>

```diff
diff --git a/gnrpy/gnr/web/gnrwebpage.py b/gnrpy/gnr/web/gnrwebpage.py
index 867081e461..8984473a32 100644
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -395,7 +395,7 @@ class GnrWebPage(GnrBaseWebPage):
     @property
     def wsk_enabled(self):
         if not hasattr(self, '_wsk_enabled'):
-            self._wsk_enabled = self.wsk and self.getPreference('experimental.wsk_enabled',pkg='sys')
+            self._wsk_enabled = self.wsk and not self.getPreference('experimental.wsk_disabled',pkg='sys')
         return self._wsk_enabled
 
     @property
diff --git a/gnrpy/gnr/web/serverwsgi.py b/gnrpy/gnr/web/serverwsgi.py
index 681ef87b83..ec48b06e0d 100644
--- a/gnrpy/gnr/web/serverwsgi.py
+++ b/gnrpy/gnr/web/serverwsgi.py
@@ -35,6 +35,7 @@ from werkzeug.wrappers import Response, Request
 from gnr.core.cli import GnrCliArgParse
 from gnr.core.gnrconfig import getGnrConfig, gnrConfigPath
 from gnr.core.gnrdict import dictExtract
+from gnr.core.gnrstring import boolean
 from gnr.app.pathresolver import PathResolver
 from gnr.web.gnrwsgisite import GnrWsgiSite
 from gnr.web import logger
@@ -162,8 +163,9 @@ class Server(object):
                             dest='async_port',
                             type=int,
                             default=None,
-                            help="Spawn the gnrasync server on this TCP port alongside the WSGI dev server "
-                                 "(WebSocket, wsproxy, debugger)")
+                            help="Force the gnrasync subprocess to listen on this TCP port. "
+                                 "When omitted, if siteconfig declares websockets=true the "
+                                 "subprocess is auto-spawned on a free port.")
         
         parser.add_argument('-o','--open',
                             dest='open_browser',
@@ -334,8 +336,20 @@ class Server(object):
         self.app_host = self.options.host
         site_name= f'{self.site_name}:{self.remote_db}' if self.remote_db else self.site_name
 
-        if self.options.async_port and not is_running_from_reloader():
-            self._spawn_async_server(site_name, self.options.async_port)
+        if not is_running_from_reloader():
+            async_port = self.options.async_port
+            websockets_enabled = boolean(self.siteconfig['wsgi?websockets'])
+            if async_port:
+                # explicit CLI request: spawn on the user-chosen port
+                if not self._async_server_already_running():
+                    self._spawn_async_server(site_name, async_port)
+            elif websockets_enabled:
+                # siteconfig declares websockets=true: auto-spawn on a free port
+                if not self._async_server_already_running():
+                    async_port = self._pick_free_port()
+                    self._spawn_async_server(
+                        site_name, async_port,
+                        reason='auto-spawned because siteconfig declares websockets=true')
 
         ssl_context = None
         if not self.debugpy and self.reloader and not is_running_from_reloader():
@@ -416,8 +430,8 @@ class Server(object):
         if not is_running_from_reloader():
             logger.info("Shutting down")
 
-    def _spawn_async_server(self, site_name, async_port):
-        """Spawn the gnrasync server as a child process for `--async-port`.
+    def _spawn_async_server(self, site_name, async_port, reason=None):
+        """Spawn the gnrasync server as a child process.
 
         Registers an atexit cleanup so SIGTERM is sent on parent exit.
         Skipped when running from the werkzeug reloader child to avoid
@@ -425,15 +439,19 @@ class Server(object):
         """
         gnr_bin = shutil.which('gnr')
         if not gnr_bin:
-            logger.error("'gnr' entrypoint not found on PATH; --async-port disabled")
+            logger.error("'gnr' entrypoint not found on PATH; async subprocess disabled")
             return
         try:
             child = subprocess.Popen([gnr_bin, 'web', 'async', site_name, '-p', str(async_port)])
         except Exception:
             logger.exception('failed to spawn gnr web async subprocess')
             return
-        logger.info('gnr web async subprocess started (pid=%s) for instance %s on port %s',
-                    child.pid, site_name, async_port)
+        if reason:
+            logger.info('gnr web async subprocess started on port %s (%s) — pid=%s, instance=%s',
+                        async_port, reason, child.pid, site_name)
+        else:
+            logger.info('gnr web async subprocess started (pid=%s) for instance %s on port %s',
+                        child.pid, site_name, async_port)
 
         def _cleanup():
             if child.poll() is None:
@@ -447,3 +465,37 @@ class Server(object):
                         pass
         atexit.register(_cleanup)
 
+    def _async_server_already_running(self):
+        """Return True if a gnrasync server is already listening on the
+        site's unix socket. A lingering stale socket file (no listener) is
+        treated as not-running: gnr web async itself replaces stale files
+        on startup.
+        """
+        socket_path = os.path.join(self.site_path, 'sockets', 'async.sock')
+        if not os.path.exists(socket_path):
+            return False
+        try:
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as s:
+                s.settimeout(0.2)
+                s.connect(socket_path)
+            return True
+        except (OSError, socket.error):
+            return False
+
+    def _pick_free_port(self, start=8086, end=8200):
+        """Pick a free TCP port for the auto-spawned gnrasync subprocess.
+
+        Tries the [start, end) range in order, falls back to kernel
+        ephemeral allocation (bind to 0) if nothing in the range is free.
+        """
+        for port in range(start, end):
+            with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+                try:
+                    s.bind(('127.0.0.1', port))
+                except OSError:
+                    continue
+                return port
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+            s.bind(('127.0.0.1', 0))
+            return s.getsockname()[1]
+
diff --git a/projects/gnrcore/packages/sys/resources/preference.py b/projects/gnrcore/packages/sys/resources/preference.py
index 785b6c36c6..fd0fb0c626 100644
--- a/projects/gnrcore/packages/sys/resources/preference.py
+++ b/projects/gnrcore/packages/sys/resources/preference.py
@@ -119,7 +119,7 @@ class AppPref(object):
     def developerPreferences(self, pane):
         fb = pane.formbuilder()
         fb.comboBox(value='^.experimental.remoteForm',lbl='!![en]Remote forms',values='onEnter,delayed')
-        fb.checkbox(value='^.experimental.wsk_enabled',lbl='!![en]WSK Enabled')
+        fb.checkbox(value='^.experimental.wsk_disabled',lbl='!![en]WSK Disabled (kill switch)')
         fb.checkbox(value='^.experimental.no_mako',lbl='!![en]No Mako rootPage')
 
     def tablesConfiguration(self, pane):
diff --git a/projects/gnrcore/packages/test/webpages/websocket/collaborative.py b/projects/gnrcore/packages/test/webpages/websocket/collaborative.py
index 9c16a52793..b1bbc93456 100644
--- a/projects/gnrcore/packages/test/webpages/websocket/collaborative.py
+++ b/projects/gnrcore/packages/test/webpages/websocket/collaborative.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-"Collaborative rooms with shared Object. Please enable experimental.wsk_enabled preference and use gnrwsgiserve --async-port PORT for testing"
+"Collaborative rooms with shared Object. Requires websockets=\"true\" in siteconfig; the dev server (gnr web serve) auto-spawns the gnrasync subprocess."
 
 class GnrCustomWebPage(object):
     
diff --git a/projects/gnrcore/packages/test/webpages/websocket/shared_objects.py b/projects/gnrcore/packages/test/webpages/websocket/shared_objects.py
index 4134ca5ba4..cb38b58644 100644
--- a/projects/gnrcore/packages/test/webpages/websocket/shared_objects.py
+++ b/projects/gnrcore/packages/test/webpages/websocket/shared_objects.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-"Shared Objects. Please enable experimental.wsk_enabled preference and use gnrwsgiserve --async-port PORT for testing"
+"Shared Objects. Requires websockets=\"true\" in siteconfig; the dev server (gnr web serve) auto-spawns the gnrasync subprocess."
 
 class GnrCustomWebPage(object):
     py_requires="gnrcomponents/testhandler:TestHandlerFull"
```

</details>

If you see issues with either simplification, or if you'd rather
extend the work in a different direction, feel free to push directly
on the branch — apply, tweak, replace, do whatever you think is best.
This proposal is a starting point, not a constraint.

On the bind interface point I leave that to you — agreed it's a
legitimate concern.

Cheers.

cgabriel (dismissed):
Ok for me now, I've been able to use also all tests. Feel free to apply the patch that removes the preference lock of the feature, after fixing the current conflicts the branch has.

genro:
@cgabriel @fporcari @mbertoldi — develop merged (resolved conflicts in `gnrwsgisite.py` and `serverwsgi.py` keeping the asyncio branch's structure: `tornado=None` kwarg dropped, legacy Tornado-serves-WSGI path removed). Kill-switch patch applied as agreed (`experimental.wsk_disabled` instead of `wsk_enabled`, auto-spawn `gnr web async` from `gnr web serve` when siteconfig declares `<wsgi websockets="true"/>`). 5 files, +65/-13. Ready for re-review — would like to land before May 15 if possible.

cgabriel (dismissed):
LGTM

genro:
Heads-up on commit `4cabcd5f02`: fixes a CI regression I introduced in `0c81f8a0b4`. While resolving the develop merge conflict on `gnrwsgisite.py` I used `--ours` and lost the #882 cleanup machinery (`_maybeRunCleanup`, `_runCleanup`, new `cleanup_threshold`/`interval_minutes` config). This commit takes develop's `gnrwsgisite.py` as the base and reapplies only the asyncio delta (`tornado=None` kwarg removal). Also drops the now-dead `--noclean` and `--counter` CLI flags on `serverwsgi.py` since #882 retired them.

Apologies for the premature ping — Giovanni had told me to wait for CI before re-requesting review, and I jumped the gun. No functional change to the asyncio migration itself; CI should go green now. Sorry for the noise.

cgabriel (approved):

fporcari (approved):

PR #854, opened 2026-04-30
fix(gnrhtml): allow GnrHtmlBuilder without page dimensions

### LINKED ISSUE #853: GnrHtmlBuilder raises for non-print HTML use cases (regression since 2022-07-26)
## Summary

`GnrHtmlBuilder` raises `GnrHtmlSrcError('Missing page dimensions')` when constructed without `page_width` / `page_height`. This silently broke `GnrHtmlPage` and `GnrHtmlDojoPage` for non-print use cases.

## Regression details

- **Commit**: `e234e6c36f58c899997bb5d1c04b93967fac8492` ("refactoring and weasyprint integration", 2022-07-26)
- **What changed**: the previous default fallback `page_height=280, page_width=200` (A4 in mm) was removed and replaced by an unconditional raise.

Before that commit, instantiating `GnrHtmlBuilder()` without dimensions silently fell back to A4 defaults — wrong for non-print but at least did not crash. After the commit, any non-print caller crashes at construction.

## Affected use cases

`GnrHtmlBuilder` is legitimately used outside of paged-print contexts:

- static rootPages built via struct (`GnrHtmlPage`, `GnrHtmlDojoPage`)
- HTML emails
- embedded HTML snippets / iframe content
- SPA shells

Concrete callers in external repos (silently broken since 2022-07-26):

- `genropynet/webpages/serverpages/home.py` — homepage of `genropynet` built with `GnrHtmlDojoPage`
- `genromed/webpages/report/ambulatorio_occupazione.py` — report page using `GnrHtmlDojoPage`

The breakage was not noticed in `genropy` itself because no in-repo caller exercises `GnrHtmlPage` without dimensions.

## Why the fix is safe

Inspection of all 28 `GnrHtmlBuilder` callsites (via Sourcerer) confirms:

- 27 of 28 are print callers and pass `page_width` / `page_height` explicitly (gnrtablescript, gnrbaghtml, RecordToHtmlPage, RecordToHtml, BagToHtml.newBuilder, Table.getHtmlBuilder, print examples, and similar across `genropy`, `pfrc`, `sandbox`).
- 1 caller (`GnrHtmlPage.onPreIniting`) intentionally builds without dimensions — this is the broken path.

Removing the raise restores the pre-2022 behaviour for the non-print path. Print callers are unaffected because their dimensions still flow through unchanged.

## Side effect to handle

`initializeSrc` always emitted an `@page` CSS block:

```css
@page {
    margin: 0;
    size: ${page_width}mm ${page_height}mm;
}
```

When dimensions are missing, this becomes `size: Nonemm Nonemm` — malformed CSS that breaks downstream PDF converters (weasyprint) and CSS validators. The fix emits `@page` only when both dimensions are set.

## Fix

PR #NNN — minimal change in `gnrpy/gnr/core/gnrhtml.py`: remove the unconditional raise, guard the `@page` emission with the dimensions check.

genro:
Closes #853.

## Summary

`GnrHtmlBuilder` used to raise `GnrHtmlSrcError('Missing page dimensions')` on construction without `page_width` / `page_height`. This broke `GnrHtmlPage` and `GnrHtmlDojoPage` for any non-print use case (static rootPages, HTML emails, embedded snippets, SPA shells).

The unconditional raise was introduced by commit `e234e6c36f` (2022-07-26, "refactoring and weasyprint integration") which also removed the previous A4 defaults.

## What changes

Two lines:

1. Remove the unconditional raise. Building HTML without page dimensions is a legitimate use case.
2. Emit the `@page` CSS rule only when both dimensions are set. Without dimensions, emitting `size: Nonemm Nonemm` produces malformed CSS that breaks PDF converters and CSS validators.

## Why the fix is safe

Audit of all 28 `GnrHtmlBuilder` callsites (Sourcerer search):

- 27 of 28 are print callers and pass `page_width` / `page_height` explicitly. They keep working unchanged because their dimensions still flow through.
- 1 caller (`GnrHtmlPage.onPreIniting`) is the non-print one — the path that has been broken since 2022-07-26.

External repos with affected callers (silently broken):
- `genropynet/webpages/serverpages/home.py`
- `genromed/webpages/report/ambulatorio_occupazione.py`

## Test plan

- [x] `python -m pytest gnrpy/tests/core/gnrhtml_test.py gnrpy/tests/core/gnrbaghtml_test.py gnrpy/tests/web/gnrhtmlpage_test.py` passes (9/9).
- [x] flake8 clean on `gnrpy/gnr/core/gnrhtml.py`.
- [x] Manual smoke test:
  ```python
  b = GnrHtmlBuilder()             # no dimensions -> works, no @page emitted
  b.initializeSrc()
  b.body.div(id='mainWindow')
  b.toHtml()

  b2 = GnrHtmlBuilder(page_width=200, page_height=280)  # print -> @page included
  ```
- [ ] Manual: load `genropynet/webpages/serverpages/home.py` and verify the homepage renders.

### DIFF
--- a/gnrpy/gnr/core/gnrhtml.py
+++ b/gnrpy/gnr/core/gnrhtml.py
@@ -290,8 +290,6 @@ class GnrHtmlBuilder(object):
         self.nextLetterhead = None
         self.page_height = page_height or top_layer['main.page.height']
         self.page_width = page_width or top_layer['main.page.width']
-        if not (self.page_height and self.page_width):
-            raise GnrHtmlSrcError('Missing page dimensions')
         self.page_margin_top = page_margin_top or top_layer['main.page.top'] or 0
         self.page_margin_left = page_margin_left or top_layer['main.page.left'] or 0
         self.page_margin_right = page_margin_right or top_layer['main.page.right'] or 0
@@ -373,12 +371,13 @@ class GnrHtmlBuilder(object):
                         }
                         """)
 
-        self.head.style(f"""
-            @page{{
-                margin:0;
-                size:{self.page_width}mm {self.page_height}mm;
-            }}
-        """)
+        if self.page_width and self.page_height:
+            self.head.style(f"""
+                @page{{
+                    margin:0;
+                    size:{self.page_width}mm {self.page_height}mm;
+                }}
+            """)
 
 
     def prepareTplLayout(self,letterhead_root):


fporcari (approved):
## Review

### Correctness ✅
The fix is correct and surgical. Verified by reading `gnrpy/gnr/core/gnrhtml.py`:

- The non-print path (`GnrHtmlPage.onPreIniting` → `GnrHtmlBuilder()` → `initializeSrc()` → `toHtml()`) never reaches the points where `page_width`/`page_height` are used arithmetically:
  - `newPage` (lines 443-481, formats `%smm` and subtracts margins)
  - `prepareTplLayout` (line 384+, idem)
  - line 753 (`width_mm = self.page_width - ...`)

  All of these are reachable only from print callers, which keep passing dimensions explicitly.
- Guarding the `@page` emission is necessary: without it, `size: Nonemm Nonemm` would produce malformed CSS that breaks weasyprint and CSS validators.

### Style and project conventions ✅
- Minimal scope (2 edit points). Aligned with the "don't add features beyond what the task requires" guideline.
- No unused imports, no backward-compatibility shims.
- The PR description documents the audit of all 28 callsites, which makes the blast radius easy to verify.

### Test coverage ⚠️ (main observation)
The PR claims "9/9 passes" but no regression test is added for the previously-broken path:

- `gnrpy/tests/web/gnrhtmlpage_test.py` is a pure smoke import (1 line, `# noqa: F401`).
- `gnrpy/tests/core/gnrhtml_test.py` only covers `calcRowsNumber` via a `FakeBuilder` stub — it never instantiates a real `GnrHtmlBuilder`.

No test demonstrates that `GnrHtmlBuilder()` + `initializeSrc()` + `toHtml()` works without dimensions and produces HTML free of `@page`.

Suggested addition (small, end-to-end, no mocks):

```python
def test_builder_without_dimensions_does_not_raise():
    b = GnrHtmlBuilder()
    b.initializeSrc()
    b.body.div(id='mainWindow')
    html = b.toHtml()
    assert '@page' not in html
    assert 'mainWindow' in html

def test_builder_with_dimensions_emits_page_rule():
    b = GnrHtmlBuilder(page_width=200, page_height=280)
    b.initializeSrc()
    html = b.toHtml()
    assert '@page' in html
    assert '200mm 280mm' in html
```

Without a regression test, a future refactor that reintroduces a use of `page_width` early in the construction path could silently re-break the same callers, exactly as happened in the 2022-07-26 commit. A real test (no mocks) would cement the contract.

### Minor side effect to flag
A caller that constructs `GnrHtmlBuilder()` without dimensions and then explicitly calls `newPage()` / `prepareTplLayout()` previously failed at `__init__`; now it would fail later with `TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'`. Unlikely in practice (print callers always pass dimensions), but worth noting.

### Verdict
**LGTM.** Approving. Adding the regression test before merge is recommended but not blocking.

---

genro:
Thanks for the approval and for the careful audit of the arithmetic callsites.

On the regression-test suggestion: I'd leave it out of this PR. A test targeted at "we don't break this specific regression" does not protect against what we should actually fear, which is future refactors that reintroduce different assumptions on the arithmetic points. If someone tomorrow moves a dimension-arithmetic call into `__init__` instead of `newPage`, the "no `@page` when dimensions are missing" test still passes and the bug comes back identical to 2022. To be effective the test would need to cover the full contract — "GnrHtmlBuilder used without dimensions does not raise and produces clean HTML" — which is exactly the contract that `GnrHtmlPage`/`GnrHtmlDojoPage` exercise in production (on external instances, e.g. `genropynet/webpages/serverpages/home.py`). That contract is better defended by an end-to-end test of the struct page, which will land naturally with the no-mako migration.

In that context the whole `gnrhtml.py` is an area to revisit and refactor: the class mixes a print engine and a generic DOM builder, and a good chunk of the code that currently lives under `if not (page_width and page_height)` will need to be rethought as a separate component. Anchoring a test now to the current API is debt we'd remove right after.

Proceeding to merge.

cgabriel (approved):
Agreed with @fporcari via claude tests' comment, the rest LGTM.

PR #861, opened 2026-04-30
feat(gnrdockerize): generate docker-compose via GnrYamlBuilder

genro:
Closes #859.

## Summary

Replaces the inline Mako template used to generate `<instance>-compose.yml` in `gnrdockerize.py` with a domain-aware `DockerComposeBuilder` (introduced in this PR under `gnr.utils.dockercompose`) layered on top of `GnrYamlBuilder`.

The builder models the five top-level sections of a compose document — `services`, `volumes`, `networks`, `configs`, `secrets` — as first-class citizens, each exposed via a lazy getter (`services()`, `volumes()`, ...) and an adder (`service()`, `volume()`, ...).

A new `--mako` CLI flag (default `False`) keeps the legacy template available as an opt-in fallback while the builder is validated on real instances.

## Why a builder over a template

The previous Mako template was a string that mixed three languages — YAML, Mako directives (`% if`, `% for`, `${expr}`), and shell-style escaping for the literal `${VAR}` interpolations destined for docker itself. Editing it required reasoning about all three at once, and the worst offender was the docker-env escaping:

    GNR_DB_HOST : ${r"${GNR_DB_HOST:-" + instanceName + "_db}"}

This is the only way to emit a literal `${VAR}` through Mako: a single misplaced quote breaks the output silently. With the builder the same line is a plain Python string:

    'GNR_DB_HOST': f'${{GNR_DB_HOST:-{name}_db}}'

The double brace is the standard f-string escape; no escape-of-escapes, no template engine involved.

Beyond readability, three concrete benefits drove the move:

1. **Type fidelity end-to-end.** Mako receives `extra_labels` as pre-formatted strings (`'key: 8888'`); the original `int 8888` is lost at the f-string boundary, and PyYAML guesses the type back at parse time. The builder takes native Python values and PyYAML serializes them with their actual types — `port: 8888` stays int by construction, not by coincidence.

2. **Tooling support.** A template is a string: the IDE cannot rename `instance_name` inside `${instanceName}`, type checkers see nothing, stack traces point at compiled-template line numbers. The builder is plain Python — refactoring, type checking, debugging all work.

3. **Testability.** `gnrdockerize.py` had no tests before this PR. A template only produces text, so testing meant parsing the YAML back and asserting on the result. The builder has a public API: this PR adds 30 tests (18 on `DockerComposeBuilder` itself, 12 on the compose generation in `gnrdockerize`) covering structure, type fidelity, `${VAR}` preservation, and the legacy Mako path.

[...]

PR #858, opened 2026-04-30
feat(gnrbag): add EnvResolver, UidResolver, TsResolver utility resolvers

genro:
Closes #857.

## Summary

Adds three small `BagResolver` subclasses to `gnrpy/gnr/core/gnrbag.py`, all expressing the same idiom — **"lazy state derived from process context"** — and all modulated by the existing `cacheTime` machinery (no new API surface):

- **`EnvResolver`** — exposes an environment variable as a lazy bag node
- **`UidResolver`** — generates a stable identifier on first access (default: 22-char `gnrlang.getUuid`)
- **`TsResolver`** — exposes the current timestamp with selectable freshness

Rationale and motivating examples: see #857.

## Scope

Self-contained, additive. One file modified, one new test file:

- `gnrpy/gnr/core/gnrbag.py` — three new classes inserted between `BagCbResolver` and `UrlResolver`; imports extended with `uuid`, `timezone`, `getUuid`
- `gnrpy/tests/core/gnrbag_resolvers_test.py` — 16 tests

No new dependencies. No changes to existing call sites.

## API

```python
# Environment variable, lazy
b['db_host'] = EnvResolver('GNR_DB_HOST', default='localhost')

# Stable id (22-char gnrlang.getUuid by default)
b['request_id']   = UidResolver()
b['standard_id']  = UidResolver(version='uuid4')

# Current timestamp
b['now']        = TsResolver()                         # live (re-read each access)
b['created_at'] = TsResolver(cacheTime=-1)             # snapshot, frozen forever
b['cached_ts']  = TsResolver(cacheTime=60)             # refresh every 60s
b['utc_now']    = TsResolver(tz='utc')                 # naive UTC
```

`cacheTime` semantics (existing `BagResolver` convention, verified in code at `gnrbag.py:2678-2701`):

| `cacheTime` | Behavior |
|---|---|
| `0` | re-read on every access (live) |
| `> 0` | cache for N seconds |
| `< 0` | never expires (snapshot stable) |

The defaults are chosen per resolver: `EnvResolver` defaults to `0` (live), `TsResolver` to `0` (live), `UidResolver` to `-1` (stable id is the common case).

## Error model

- `UidResolver(version='bogus')` → `ValueError` on access (not at construction, to stay consistent with the lazy nature of resolvers)
- `TsResolver(tz='bogus')` → `ValueError` on access

## Test plan

[...]

PR #850, opened 2026-04-29
staging colour styling via siteconfig / env vars

cgabriel:
retrieve staging_colour from site configuration via <gui> but can beoverriden from env vars (useful in deployment context)

If there's a value, a new inline style is injected into the root container.

It's the simplest implementation, without using CSS vars. I've added the env var feature which is really useful in deployment context (for example, can have have the env var injected in ALL demo/test cluster running containers, without touching the configuration of the single app)

PR #862, opened 2026-04-30
feat(no-mako): struct-based rootPage rendering, opt-in via preference

### LINKED ISSUE #853: GnrHtmlBuilder raises for non-print HTML use cases (regression since 2022-07-26)

genro:
Closes #853 (companion of merged #854).

## Summary

Replaces the Mako-rendered rootPage with a struct-based renderer driven by `BasePageTemplate` and `GnrHtmlBuilder`. The change is **opt-in** via the new preference `experimental.no_mako` (sys preferences, default off): when off the framework renders rootPage exactly as before; when on, it looks up a Python `<name>.py` template under `tpldirectories` (same override semantics as the legacy Mako lookup) and falls back to Mako if no struct template is found.

This is **PR #2** of the no-mako migration plan. PR #1 (`#854` — `GnrHtmlBuilder` regression fix) has been merged. Related follow-ups already open: `GnrYamlBuilder` → #856 (merged), utility resolvers → #858 (merged), dockerize → #861, scheduler dashboard SPA → future.

## Scope

Additive when the preference is off. The change is compartmentalized:

- Framework code: `GnrWebPage`, plugin Mako/RML, `GnrResourceLoader` — added a struct-template lookup branch, untouched legacy paths
- New: `BasePageTemplate` and `lookup_template_class` under `gnrpy/gnr/web/gnrwebpage_proxy/frontend/`
- Templates ported to Python under `resources/common/tpl/`: `standard.py`, `gnr_header.py`, `gnr_header_static.py`, `gnr_header_dojo.py`, `standard_print.py`, `standard_print_table.py`, `standard_export.py`
- Helpers extended: `GnrHtmlSrc.comment()`, `GnrHtmlBuilder._prettyfy()`, XHTML namespace on `PageBuilder`
- New URL parameter `_tpl=<name>` for ad-hoc template selection (use cases: viewport simulators, embeds, fullscreen demos)
- New env-var/siteconfig staging cue (`GNR_STAGING_STYLE`, `GNR_STAGING_COLOUR`) — wraps `#mainWindow` in `#stagingFrame` when set

## How it activates

- Default: nothing changes (preference off)
- Toggle: `experimental.no_mako = on` in sys preferences (now exposed via UI checkbox added to `gnrcore/packages/sys/resources/preference.py`)
- URL override: `?_tpl=<name>` switches the template per-request, security gated by the template's `check_access()` hook

## Test plan

- [x] 40 new functional-equivalence tests in `gnrpy/tests/web/test_no_mako_rootpage.py` covering: document structure (xml decl, doctype, xhtml ns, html/head/body, pretty-print), meta tags (charset, viewport, mobile-web-app, x-ua-compatible), body structure (mainwindow, pdb root, protection shield, body class), scripts (dojo, djConfig, module path, genro js imports, requires, client bootstrap), CSS (dojo, genro, requires, media print, localcss), optional features (PWA, sentry, favicon, google fonts, logo, dijit, custom headers), comments
- [x] Full `gnrpy/tests/web/` and `gnrpy/tests/core/` regression run: 394 pass / 57 skip / 20 pre-existing failures (gnrdate + gnrlocale, identical on develop — unrelated to this PR)
- [ ] Manual end-to-end on a real instance with preference toggled (sandbox / anaci / other) — to be done by reviewer or as a follow-up

## Reviewer checklist

- [ ] Default-off behavior verified (rendering with preference off == legacy Mako output)
- [ ] Per-package template override works (a `<pkg>/resources/tpl/<name>.py` wins over `resources/common/tpl/<name>.py`)
- [ ] `_tpl=` URL parameter respects `check_access()` deny path
- [ ] `experimental.no_mako` preference UI visible in sys preferences

## Out of scope

- Removing the preference / making struct rendering default — separate follow-up after field validation
- Conversion of `gnrtask_new.py` Mako template (deferred to a future scheduler dashboard SPA refactor)
- Conversion of `gnrdockerize.py` Mako template (separate PR — #861)


genro:
Marked as draft: while testing the struct-based rootPage on `test_invoice_pg` we noticed that the staging cue (`#stagingFrame` wrapper around `#mainWindow`, driven by `GNR_STAGING_STYLE` / `GNR_STAGING_COLOUR`) — which was present in the original commits of this branch — was removed in the review pass and is currently absent from `resources/common/tpl/standard.py`.

This is an improvement opportunity: bringing the staging wrapper back here would let this PR also subsume #850 (which is solving the same UX need at a different layer, in `frameindex.py`).

Will iterate on this branch.

genro:
Staging cue restored in commit d25e22e37c with an additional improvement over the original behaviour: the wrapper is now suppressed when the page is rendered inside another page's iframe.

Server-side detection uses the `_calling_page_id` request kwarg, which the client-side iframe builder (`setSrc_do` in genro_widgets) injects automatically with the parent's page id. `build_arg_dict` now sets `is_subframe=True` when that marker is present, and `_resolve_staging_style` returns `None` accordingly, so subpages embedded by frameindex.py (or any other htmliframe consumer) no longer get a 'frame-in-frame' cue.

Tested visually on `test_invoice_pg` with `GNR_STAGING_COLOUR='#ff8800'`: the orange wrapper appears only on the top-level browser window, never around tabbed pages or dialogs. 7 new tests in `TestStagingCue` cover both the emission and the subframe suppression paths.

gnrpy/gnr/core/gnrbagxml.py
```diff
@@ -348,14 +348,15 @@ def build(self, bag, filename=None, encoding='UTF-8', catalog=None,
...
+            xmlblock = self.buildTag('GenRoBag', self.bagToXmlBlock(bag,namespaces=[]), xmlMode=True, localize=False)
         if pretty:
-            from xml.dom.minidom import parseString
-            result = parseString(result)
-            result = result.toprettyxml()
-            result = result.replace('\t\n','').replace('\t\n','')
+            try:
+                xmlblock = self._prettyfy(xmlblock, indent=pretty if isinstance(pretty, str) else '\t')
+            except (ValueError, IndexError):
+                pass
```
cgabriel, line 358: Why silencing this exception?

cgabriel (approved):
LGTM, tested with the no mako preferences active, no issue. staging_colour working fine.

fporcari (approved):
LGTM. Cambio additivo, opt-in via `experimental.no_mako` con default off: il path Mako legacy resta invariato, il fallback a Mako quando il template struct non viene trovato evita rotture su template non ancora portati. CI verde su 3.11/3.12/3.13/3.14, già verificata manualmente con preference attiva (incluso lo staging cue). Suite di 40 test di equivalenza funzionale a copertura della struttura del documento, meta tag, body, script, CSS, feature opzionali e commenti. Approvo.

---

gnrpy/gnr/core/gnrbagxml.py
genro: This is intentional: pretty-print and XML validity are orthogonal concerns.

`pretty=True` is a cosmetic transformation `str -> str`. It doesn't validate, and shouldn't. If the caller wants validation, that's a separate explicit step (e.g. lxml parse, schema check) — done once, independent of pretty. The matrix is:

|              | valid       | invalid       |
|--------------|-------------|---------------|
| pretty       | nice+valid  | nice+invalid  |
| not pretty   | ugly+valid  | ugly+invalid  |

Conflating "indentation didn't work" with "the XML is broken" mixes the two dimensions and makes the broken-XML signal show up only when the caller happens to ask for indentation. That's the wrong channel.

The current behaviour:
- `pretty=True` on valid XML → indented XML (best case)
- `pretty=True` on invalid XML → unindented XML (graceful fallback, same string the caller would have got with `pretty=False`)
- `pretty=False` → unindented XML (independent of validity)

If validity is what we want to surface, the right fix is a separate validator that runs regardless of pretty — not a side-effect of the pretty path.

PR #864, opened 2026-05-01
Replace CodeMirror 5 with CodeMirror 6 in pane.codemirror() widget

### LINKED ISSUE #863: Replace CodeMirror 5 with CodeMirror 6 in pane.codemirror() widget
## Context

The vendored CodeMirror 5 (~1.1 MB under `resources/js_libs/codemirror/`)
is the upstream version that has been in stale maintenance since 2022. Its
successor CodeMirror 6 is a complete rewrite with first-class TypeScript,
state/effect primitives, modern packaging, and ongoing upstream releases.

`pane.codemirror(...)` is used across Genropy in:

- gnride (the in-browser IDE with pdb debugger integration)
- source_viewer (error highlight on save failures)
- docu (RST/HTML/CSS/JS snippets, base templates)
- dbadmin (SQL command runner)
- clientside_builder, ui_builder, print_tutorial, th_documentation
- third-party projects relying on the public widget API

## Goal

Replace the CM5 implementation with a CM6-based widget while keeping the
Python signature `pane.codemirror(value=..., config_mode=..., ...)`
backward-compatible, so call sites do not need to be updated.

## Scope

- Vendor a CM6 build under `resources/js_libs/codemirror6/` produced by
  an esbuild pipeline kept in `tools/codemirror6-build/` (entry.js +
  build.sh + package.json + README), so the bundle can be rebuilt
  reproducibly.
- Rewrite `gnr.widgets.codemirror` on top of `baseExternalWidget`
  in both `gnr_d11/js/genro_extra.js` and `gnr_d20/js/genro_extra.js`,
  exposing the lifecycle mixins (`gnr_value`, `gnr_markText`,
  `gnr_addLineClass`, `gnr_setGutterMarker`, `gnr_lineInfo`, ...) used
  by gnride and other consumers.
- Drop the legacy CM5 widget and the CM5 vendored bundle.
- Migrate gnride to the new widget: declarative `onGutterClick_<name>`
  handlers (replacing `cm.on('gutterClick', ...)`), pdb gutter attached
  only to Python sources, current-line highlight that survives editor
  rebuilds via `_cm6Snapshot.lineClasses`.
- Migrate `source_viewer.py` from `cm.doc.markText` to `gnr_markText` /
  `gnr_clearMark`.
- Provide CM5 compat hooks where reasonable: `mixin_setValue` alias for
  call sites reaching into `externalWidget.setValue()` directly; theme
  aliases `twilight -> oneDark`, `night -> dracula`; mode aliases
  `htmlmixed`, `htmlembedded`, `text/html`, `md`.

## Cache invalidation for vendored bundles

Lazy-loaded vendored bundles are not part of the Genropy `_jslib` hash
cache, so a static URL would be stuck on the first version the browser
saw. Introduce a generic mechanism: `gnrwebpage` exposes a
`gnr.vendoredMtime` dict (keyed by bundle short name), the widget reads
the value and appends `?mtime=<epoch>` to the URL. New vendored bundles
can be registered by adding entries to `_VENDORED_BUNDLES`.

## Out of scope (follow-up)

- `config_addon='lint'` (used in 2 builder pages) is silently ignored;
  CM6 lint requires a custom diagnostics source. Track separately.
- Mode `rst` / `stex` (used by docu): no native CM6 language pack;
  current behaviour falls back to plain text. Could be addressed later
  via `@codemirror/legacy-modes`.
- Other vendored bundles that would benefit from the same
  `gnr.vendoredMtime` cache pattern.

## Verification

- Demo pages under `projects/gnrcore/packages/test/webpages/tools/`:
  `codemirror6.py` (4 cards) and `codemirror6_playground.py` (live
  config tester).
- Manual smoke test of gnride: open `/sys/gnride`, set/clear breakpoints
  via gutter click, run a debug session, switch tabs, verify the
  current-line highlight is restored after editor rebuilds.
- Manual smoke test of `source_viewer` error highlight on a save with
  a syntax error.

genro:
Closes #863

## Summary

- Vendor a CodeMirror 6 build (~660 KB minified) under `resources/js_libs/codemirror6/` with an esbuild pipeline in `tools/codemirror6-build/`.
- Rewrite `gnr.widgets.codemirror` on top of `baseExternalWidget` in both `gnr_d11` and `gnr_d20`, drop the legacy CM5 widget and the CM5 vendored bundle (1.1 MB, 120 files). `pane.codemirror(...)` keeps the same Python signature.
- Migrate gnride and source_viewer to the new widget. Fix a longstanding bug in the gnride "Add ide" button where `ideId` was referenced as a JS variable but never exposed.
- Cache invalidation for lazy-loaded vendored bundles via a generic `gnr.vendoredMtime` map exposed by `gnrwebpage`.
- Add demo pages under `projects/gnrcore/packages/test/webpages/tools/`.

See #863 for the full design context, scope, out-of-scope items and verification plan.

## Notes for reviewers

- API is backward compatible: existing `pane.codemirror(...)` call sites work unchanged. The only required follow-up is removing dead `config_addon='lint'` parameters (now ignored) and replacing CM5-only theme names if desired (`twilight`/`night` are aliased anyway).
- CM5 method calls reaching into `externalWidget` directly were audited:
  - `th_documentation.py:270` uses `externalWidget.setValue(...)` — covered by the `mixin_setValue` alias.
  - `source_viewer.py` used `cm.doc.markText` / `tm.clear()` — migrated to `gnr_markText` / `gnr_clearMark`.
- Rebuild instructions for the CM6 bundle live in `tools/codemirror6-build/README.md`.

## Test plan

- [ ] Open `/sys/gnride`: set/clear breakpoints by clicking the gutter, run a debug session, step through code, switch between editor tabs, verify the current-line highlight is restored after editor rebuilds.
- [ ] Verify the pdb gutter is **not** shown on non-Python files.
- [ ] Click "Add ide" in the gnride toolbar: a new IDE tab is opened in the central stack.
- [ ] Open `/test/tools/codemirror6` and `/test/tools/codemirror6_playground` to exercise the four demo cards and the live config tester.
- [ ] Trigger a save error in `source_viewer` (e.g. introduce a syntax error and save) and verify the error line is highlighted and the message is shown.
- [ ] Smoke test other consumers: `dbadmin`, `clientside_builder`, `print_tutorial`, docu pages, `th_documentation`.
- [ ] After a `tools/codemirror6-build/build.sh` rerun, confirm the browser refetches the bundle thanks to the new `?mtime=<epoch>` cache-buster.


fporcari (dismissed):
Approving. The CM5 -> CM6 swap is well isolated, the bundle pipeline is clean, the demo pages are useful, and the snapshot pattern that lets selection / scroll / line-class survive editor rebuilds is a nice fit for the IDE use case. The `addIdeBtn` `ideId` fix is a real bugfix worth calling out.

A few priority fixes have already been pushed as a follow-up commit on branch `claude/review-pr-864-wHf2n` (1 commit on top of this PR head, +26 / -13). Feel free to cherry-pick / merge or take inspiration:

- `mixin_gnr_lineWrapping` was a no-op. Moved into a dedicated `Compartment` so the playground checkbox and any reactive `^.config.lineWrapping` binding actually toggle wrapping at runtime (same pattern as `readOnly` / `editable`).
- `mixin_gnr_value`: added a `doc.length === newValue.length` short-circuit before `doc.toString()` so unchanged large docs don't materialize the full rope on every external value update.
- Selection-restore `try/catch` in `initialize` now logs via `console.warn` instead of silently dropping the snapshot.
- Stray duplicate blank line in `gnrwebstruct.py` removed.

## Punch list (non-blocking, suggested follow-ups)

- [ ] **`isVisible` watcher leak (preexisting)**: every `mixin_gnr_value` call re-registers `sourceNode.watch('isVisible', ...)` without cleanup. Worth a separate fix.
- [ ] **Document the `evt_type` semantics shift in `onBreakpointGutterClick`**: CM5 used `info.gutterMarkers ? 'del' : 'ins'` (any gutter), CM6 now uses `info.gutterMarkers[gutter]` (specific gutter). Correct, but worth a line in the PR body so a "breakpoints don't toggle" report has context.
- [ ] **`_editorModeFor` fallback to `'python'`** for any unknown / dotless filename is surprising. Mitigated by the explicit `include` glob, but a `None` fallback would be safer.
- [ ] **`initialSpacer` IIFE in `buildExtensions`**: `return new (function(){...})()` is dead code, the inner body returns a separate `Object.create(CM.GutterMarker.prototype)`. Cosmetic cleanup possible.
- [ ] **`entry.js` re-exports unused symbols** on `window.CodeMirror6` (`rectangularSelection`, `crosshairCursor`, `highlightSpecialChars`, `WidgetType`, `HighlightStyle`, `openSearchPanel`, `closeSearchPanel`, `historyKeymap`). Either keep as a deliberate public API surface (and document) or trim.
- [ ] **`codemirror6.bundle.js.map` is committed** (~hundreds of KB). Useful for debugging but adds repo weight on every rebuild. Consider whether it belongs in the repo or should be a build-time artifact.
- [ ] **README "Updating CodeMirror" flow**: step 3 (`./build.sh`) calls `npm ci` against the just-rewritten `package-lock.json`. Works, but worth a one-liner clarifying the intent.
- [ ] **No automated tests** are added for the new widget. Reasonable for a DOM editor, but worth noting that the demo pages don't run in CI either, so the safety net is fully manual.
- [ ] **Behavioural changes to verify on the smoke-test targets in the PR body** (`dbadmin`, `clientside_builder`, `print_tutorial`, docu pages, `th_documentation`): the `editable=true` default for read-only editors is a deliberate departure from CM5 `'nocursor'` — anything depending on the old "no cursor at all" semantics needs an explicit `editable=False`.

None of the above is blocking. Approving so the merge can proceed.

---

fporcari (approved):
Re-approving on `3d067403`. The new commit applies exactly the four priority fixes from the previous review (verified: file contents at `3d067403` match the follow-up commit on `claude/review-pr-864-wHf2n` byte-for-byte across `gnrjs/gnr_d11/js/genro_extra.js`, `gnrjs/gnr_d20/js/genro_extra.js`, `gnrpy/gnr/web/gnrwebstruct.py`).

The previously listed punch-list items remain non-blocking suggested follow-ups; nothing else needs to land before merge.

---

cgabriel (approved):
LGTM, but since a whole distribution of an external package has been added to the repository, and using a difference license, please add a reference to LICENSE-3RD-PARTY file.

PR #867, opened 2026-05-04
feature: esm requirementstxt for packages ci bundling of modern esm libraries

### LINKED ISSUE #820: [FEATURE] esm_requirements.txt for packages — CI bundling of modern ESM libraries
## Summary

Introduce a convention where each GenroPy package can declare its ES module dependencies in a file `esm_requirements.txt` (sibling of `requirements.txt`). The CI pipeline introduced in #819 is extended to read these files, run a bundler (e.g. esbuild) on the declared packages, and ship the resulting self-contained bundles as static resources served by GenroPy's existing `/_rsrc/` mechanism.

This extends the CI bundling infrastructure of #819 from the gnrjs core to arbitrary modern JS libraries distributed as npm ES modules (Milkdown, TipTap, Lexical, Yjs, CodeMirror 6, etc.).

## Motivation

Modern JS libraries are increasingly distributed **only** as ES modules via npm. They are not single-file UMD/IIFE scripts: they import other npm packages internally and require a bundler (esbuild, Vite, webpack) to be usable in a browser.

The current GenroPy resource loading mechanism ([js_requires](https://github.com/genropy/genropy/blob/develop/gnrpy/gnr/web/gnrwebpage_proxy/gnrresourceloader.py), [loadJs](https://github.com/genropy/genropy/blob/develop/gnrjs/gnr_d20/js/genro_dom.js)) only supports classic `<script src>` loading of single-file libraries with globals. There is no path today for integrating modern npm-distributed libraries.

### Declaration

Each package may contain an optional `esm_requirements.txt` at its root (sibling of `requirements.txt`), listing the npm packages it needs:

```
@milkdown/crepes==7.3.0
yjs==13.6.0
```

Format mirrors `requirements.txt`: one package per line, pinned versions, comments with `#`.

### CI build (extends #819)

The CI workflow introduced in #819 is extended to:

1. Scan all installed packages for `esm_requirements.txt`
2. Aggregate the declarations into a unified set
3. Run `npm install` + esbuild to produce **self-contained bundles** (one per top-level package)
4. Write bundles to a conventional location (e.g. `resources/esm/<package>.bundle.js`)
5. Commit bundles and a manifest, same commit/PR pattern as #819

### Runtime usage

At runtime, widgets load the bundle via dynamic import of a well-known URL:

```js
const { Crepe } = await import('/_rsrc/esm/@milkdown/crepes.bundle.js')
```

Bundles are **self-contained** (esbuild resolves all bare imports internally), so no browser import map is required. Zero changes to server-side Python resource handling — the files are served as ordinary `/_rsrc/` static assets.

### Scope

**In scope**:
- Convention: `esm_requirements.txt` at package root
- CI extension (piggybacks on #819's workflow) that produces self-contained bundles
- Documentation of the widget-side pattern (`await import('/_rsrc/esm/...')`)

**Out of scope** (future issues):
- Actual widgets built on top of this infrastructure (Milkdown editor, etc.)
- Tooling to declare ESM requirements from Python code instead of a text file
- Local developer ergonomics (running the bundler outside CI)
- Import map support (not needed because bundles are self-contained)

## Relation to #819

This issue **depends on #819**. The CI workflow and the concept of committing pre-built JS artifacts to the repo are established there. This issue adds a second artifact category (ESM bundles) to the same pipeline.

## Files likely touched

- `.github/workflows/` — extend the workflow added in #819
- `tools/` — new script that scans packages for `esm_requirements.txt` and invokes the bundler
- Documentation: explain the `esm_requirements.txt` convention for package authors

## Benefits

- Opens GenroPy to the modern JS ecosystem (Milkdown, TipTap, Yjs, Lexical, etc.)
- Zero runtime dependency on Node or npm on client servers
- Consistent pattern with `requirements.txt` for Python dependencies
- No server-side Python changes: bundles are plain static files

cgabriel:
The PR includes:

* Allowing a package to defined a esm_requirements.txt file at top level, to list the ESM dependencies it needs
* Dependencies can be defined with 'alias=bundle@ver', allowing the definition of an easier name
* the 'esmbuilder' module is responsible to collect all the needs of all instance's packages, and bundle them via esbuild
* The builder generates single bundles and a global one called 'gnr_ext_bundle.js'
* All dependencies are cached, and checked via a manifest.json file.
* Bundles are placed into site/resources/esm, to they can be referenced via js_requires='esm/bundle_name'
* manifest is checked at every gnrapp startup, and dependencies automatically download. This is useful in docker context, will avoid distributing external code, and in a scaled container all container will have the needs without a sharing system.
* 'gnr app checkdep' will check also JS dependencies
* a dedicated 'gnr app esmbuilder' CLI tool is provided, to check and create the needed, with an options '-f' arg to invalidate the cache and rebuild everything from scratch.

close #820


genro (commented):
Reviewed end-to-end. The implementation is well-structured: clear naming,
solid threadpool pattern with sentinel for completion detection, cache
invalidation via manifest+specs+file existence, scoped package handling,
explicit per-platform esbuild pinning. The download/bundle pipeline is
internally consistent.

The one design question I'd like to settle before approving is **where**
and **when** the bundling should happen. The PR currently runs the
bundler at every `GnrApp` instantiation, which I think conflicts with
the pattern we've adopted for every other vendored library.

Note up front: the implementation in `esmbuilder.py` is essentially
**fully reusable** under either approach — the question is only about
where the entry point lives and when it fires, not about the bundling
logic itself. `EsmBuilder.bundle()`, `_download_all`, `download_package`,
`_resolve_version`, `parse_requirements`, manifest handling, esbuild
orchestration: all of it stays the same. What changes is whether the
caller is `GnrApp.__init__` (current) or a `tools/esm-deps-build/build.sh`
script invoked from CI (proposal). So this is an adaptation, not a
rewrite.

### Where the bundling should happen

We already have a consistent vendoring pattern for third-party JS, used
by CodeMirror 6, ProseMirror, and the Milkdown branch:

- `tools/<name>-build/` holds the npm dependencies (`package.json`,
  `package-lock.json`) and a `build.sh` that produces the bundle.
- The bundle is **committed** to `resources/js_libs/<name>/`.
- `LICENSE-3RD-PARTY` is updated alongside the vendored file.
- At runtime, the app loads the bundle from the vendored path. No
  network, no cache, no per-host divergence.
- Cache-bust on bundle update via `gnr.vendoredMtime` (see
  `_VENDORED_BUNDLES` in `gnrwebpage.py`).

What I'd propose for the ESM stack (originally my preference when we
first discussed this):

- `tools/esm-deps-build/` script that walks all packages'
  `esm_requirements.txt`, deduplicates, runs the existing `EsmBuilder`,
  writes `resources/js_libs/esm/<alias>.js` and `gnr_ext_bundle.js`.
- A CI workflow triggered on changes to any `**/esm_requirements.txt`
  rebuilds the bundles and either auto-commits or opens a follow-up PR.
- At runtime: nothing. The bundles are part of the repo, served as
  static assets like every other vendored library. Reproducible across
  dev / CI / prod / docker, offline-tolerant, and the `LICENSE-3RD-PARTY`
  update is part of the same automated step.

This avoids the problems with the current approach:

- **Auto-bundle at every `GnrApp` startup**: dev `--reload` cycles,
  every worker boot, and every script that imports an app instance pay
  the manifest + file-existence check, and the first-run path needs
  network. With no cache and no network, `urllib.request` raises and
  `GnrApp.__init__` fails — the whole app refuses to boot.
- **`~/.gnr/esm_cache` divergence**: each developer / CI runner /
  container has its own cache, so the actual bytes served differ unless
  you pin every transitive dependency, which the current resolver
  doesn't do.
- **No `LICENSE-3RD-PARTY` automation**: the bundle composition depends
  on whoever ran the bundler last; attribution would have to be
  updated manually.

The runtime hook in `check_package_dependencies` could either go away
entirely, or stay as a lightweight check that warns if the vendored
bundle is stale relative to the requirements files (no download).

### No tests

600+ lines of system code (semver resolution, threadpool download,
manifest cache, esbuild orchestration). At minimum unit tests on
`_parse_spec`, `_semver_tuple`, `_needs_resolution`, `_resolve_version`,
and `parse_requirements` would catch regressions cheaply. Not a hard
blocker for landing, but a follow-up issue would be appropriate.

### Minor

- `gnrapp.py`: three trailing blank lines after the new bundler call.
- `gnresmbuilder.py`: `sys.exit(0)` after `main()` is unreachable on
  exception and redundant on success.
- `package.py`: the `esm_requirements.txt` template comment ends with
  `# (alias=dist@ver) #` — the trailing `#` looks like a typo.
- `pyproject.toml`: nice trailing-tab cleanup on `gnrdeploybuilder`.

### Questions

- **Windows `os.symlink`** requires admin or developer mode. The Windows
  entry in `_ESBUILD_PACKAGES` suggests Windows is supported — has a
  symlink fallback been considered, or is the expected runtime
  Linux/Docker only?
- **`NPM_REGISTRY` is hardcoded** — useful to make configurable via env
  var for sites with internal mirrors / air-gapped environments?

Tagging COMMENTED so the review is on record without blocking. Happy
to APPROVE once we agree on the wiring. The bundler logic itself is
solid and reusable as-is — the discussion is purely about call-site
placement.

cgabriel:
@genro Thank you for the review. Let me address your consideration.

1/ There is not "consolidated pattern" on js bundles. The code mirrow/prose all came *after* the implementation of this PR. 

2/ The implementation proposed here follows discussion and consideration made during meetings: we don't want to have external vendor bundles to be committed in the project repository by the CI, introducing distribution issues, upgrading issues etc. Also, the unavailability of a new package added during local development, won't be added/bundles until it's committed, so it has to be pulled down. Very unconfortable.

3/ The check/download/build can happen manually ('gnr app esmbuilder'), useful for local development or during the dependency checks ('gnr app checkdep'), which will always need network in order to install possible python dependencies. The checkdep is also executed during docker image build, allowing all dependencies (python+javascript) to be bundled correctly into the application image. When an application starts (GnrApp instance), it just checkdep the dependencies, as it has always done with python deps, reporting the missing ones. This is indeed a consolidated pattern, checkdep has been around for more than 1 year, its usage is extensive, it just now check also for javascript deps among python ones.

All of the above addresses these requirements:

* Ability to express javascript versioned dependencies in packages
* Ability to install and bundles these dependencies
* Ability to use them locally during development/tests etc
* Easily verify if everything needed is correctly installed at the correct version.
* Won't introduce burden to have automatically committed files in branches, causing developers issues against their local work, and possibly distribution rights problem.
* Doesn't introduce new tools or nothing new during the development process, aside the 'esm_requirements.txt' file.

genro (commented):
Reviewed the bundler logic and the runtime wiring. The `EsmBuilder` implementation itself is solid — semver resolution, the threadpool-with-sentinel download, manifest-based cache invalidation and esbuild orchestration all hang together well. My questions are only about the runtime entry point, and I may be missing part of the intended design, so I'd rather ask than assume.

**1. Startup behaviour when offline (the point I'd like to clarify).**
`check_package_dependencies()` is called from `init()` (`gnrapp.py:990`), so it runs on every `GnrApp()` instantiation. Inside it, `bundler.run()` is called with no `try/except` around it. In the steady state this is cheap: if the manifest matches the specs and the output files exist, `_is_up_to_date()` returns early and nothing touches the network. But on the first run — or any time an `esm_requirements.txt` changes — `bundle()` calls `_download_all()`, which goes to the network. If the network isn't reachable there, the exception propagates all the way up through `init()` and the whole app fails to boot.

The Python-dependency path in the same method only *reports* problems (`logger.error`) and keeps going. The JS path instead downloads and builds, and can abort startup. In the thread you mention that at app start this should just `checkdep` the dependencies "as it has always done with python deps, reporting the missing ones" — but as written the JS branch doesn't report-and-continue, it builds and can raise. This is the part I'm not sure I'm reading correctly: is the intent that the build genuinely happens at every startup, or that startup should be report-only and the actual build stays in `gnr app esmbuilder` / `gnr app checkdep`? If it's the latter, wrapping the call so it logs and continues (or making the startup path report-only) would match the stated design and remove the offline-boot failure.

**2. Return value is unused (`gnrapp.py`).**
`output_dir, results = bundler.run()` — neither value is used afterwards, and there are three trailing blank lines. If the startup call becomes report-only, this can likely just become a check with a log line; otherwise the unused unpacking can go.

**3. `urllib.error` is referenced but not imported (`esmbuilder.py:71`).**
Only `urllib.parse` and `urllib.request` are imported. `except urllib.error.HTTPError` works today only because `urllib.request` pulls in `urllib.error` as a side effect — it's fragile. An explicit `import urllib.error` makes it robust.

**Minor (already noted by @genro, repeating for completeness):** unreachable `sys.exit(0)` after `main()` in `gnresmbuilder.py`; the `# (alias=dist@ver) #` trailing `#` in the `package.py` template reads like a typo; and 600+ lines of system code with no unit tests — `_parse_spec`, `_semver_tuple`, `_needs_resolution`, `_resolve_version`, `parse_requirements` would be cheap to cover, fine as a follow-up issue.

To be clear: I'm not pushing the commit-the-bundle approach — your reasoning in the thread about avoiding auto-committed vendored files is convincing. My only concern is the single failure mode where a network blip at startup takes the app down. Once that's addressed (or once I understand why it can't happen), this is good to go.

cgabriel:
This solution also would be improve SCA pipeline avoiding to manage sec issue internally.

fporcari (changes requested):
The builder itself reads well, but the runtime wiring reintroduces #1060 in the new code path: `collect_requirements` (gnrpy/gnr/app/esmbuilder.py:544) resolves every package with `self.app.pkg_name_to_path(package, project)` and discards the digest's attrs (`_pkgattrs`), so a package declared in the instanceconfig with an explicit `path` attribute raises `Error: package extpkg not found`. Since `bundler.run()` is wired into the startup dependency check (gnrapp.py:1100), an instance using path-declared packages fails at boot.

That is what CI is saying: all four builds are red on tests/app/test_package_path_attr.py::TestExternalPackagePath, which boots exactly such an instance.

The fix is the one #1060 already established for check_package_dependencies: keep the attrs from the digest and resolve with `self.app.pkg_path_from_attrs(package, pkgattrs, project)` instead of `pkg_name_to_path`.

cgabriel:
> The builder itself reads well, but the runtime wiring reintroduces #1060 in the new code path: `collect_requirements` (gnrpy/gnr/app/esmbuilder.py:544) resolves every package with `self.app.pkg_name_to_path(package, project)` and discards the digest's attrs (`_pkgattrs`), so a package declared in the instanceconfig with an explicit `path` attribute raises `Error: package extpkg not found`. Since `bundler.run()` is wired into the startup dependency check (gnrapp.py:1100), an instance using path-declared packages fails at boot.
> 
> That is what CI is saying: all four builds are red on tests/app/test_package_path_attr.py::TestExternalPackagePath, which boots exactly such an instance.
> 
> The fix is the one #1060 already established for check_package_dependencies: keep the attrs from the digest and resolve with `self.app.pkg_path_from_attrs(package, pkgattrs, project)` instead of `pkg_name_to_path`.

Issue addressed, the change in #1060 introduced the issue, this PR belong a long way back in time. All CI green now.

fporcari (approved):
Approving on `7352f2db` — my CHANGES_REQUESTED from 27/08 is addressed and no longer applies.

The #1060 regression is gone: `collect_requirements` now digests `#k,#a,#v` and resolves with `self.app.pkg_path_from_attrs(package, _pkgattrs, project)`, which is exactly the resolution `check_package_dependencies` uses. The four builds that were red on `tests/app/test_package_path_attr.py::TestExternalPackagePath` are green, CodeQL included.

What I am *not* deciding here is the design question @genro raised on 04/05 and left open in two COMMENTED reviews: bundling at runtime from `GnrApp.__init__` versus a `tools/esm-deps-build` step with the bundle committed. That call is his, and this PR still needs it — I am only clearing my stale block so it stops standing in front of that decision.

Not verified: I read the diff and the CI, I did not boot an instance against this branch.

genro (dismissed):
The feature is one we want, and the shape is right: a per-package `esm_requirements.txt`, an alias so call sites do not carry a bundle name and version, a cache with a manifest, and esbuild doing the bundling. I checked the collection path and the resource resolution and both work as described.

I cannot approve it as it stands, because of where and how the dependencies are fetched. These are all in `gnrpy/gnr/app/esmbuilder.py`.

**The esbuild binary is unpinned, unverified, and executed.** `get_esbuild` fetches `_npm_url(pkg_name, 'latest')` at `:106`, extracts the binary, `os.chmod(cached, 0o755)` at `:124`, and `subprocess.run` executes it at `:471`. Whatever that package publishes next runs on the machine doing the build. Pin an exact version.

**This runs at every app start, not only in CI.** `gnrapp.py:1102-1103` sits inside `check_package_dependencies`, which is called at `:996`, before the `checkdepcli` early return at `:997`. So every web worker, every CLI command and every batch process reaches the bundler. The call is also unguarded, so no network, an npm 5xx, an unsupported platform at `:102` or a non-zero esbuild exit raises out of `GnrApp.__init__` and kills the process. The title says CI bundling; the code does not.

**Nothing verifies what was downloaded.** `_download` at `:77-79` is a bare `urlretrieve`. The npm metadata you already fetch carries `dist.integrity` and `dist.shasum`, and `:107` and `:249` read `dist.tarball` from the same dict while ignoring both.

**The alias is package-controlled and unvalidated, and it is used two ways.** At `:487` it becomes a path, `os.path.join(output_dir, alias + '.js')`, so `../../../x` writes outside the output directory. At `:507` it is interpolated raw into JS, `export * as {alias} from '{name}';`, in a file esbuild then parses. The package name is interpolated into a single-quoted JS string at `:485` and `:507` with no escaping either.

**Tarball extraction is unfiltered on 3.11.** `_extractall` at `:81-85` passes `filter='data'` only on 3.12+, and `requires-python` is `>=3.11` with 3.11 in the CI matrix.

**There is no lockfile, and version matching is wrong rather than loose.** `_parse_spec` defaults a bare name to `latest` at `:165` and `:169`, `_resolve_version` returns `max(matching)` at `:227`, and `satisfies` at `:217-221` compares only the locked components — so `^18.2.0` matches `18.0.0`, below the stated floor, and `>=18` never matches 19. The tests at `test_esmbuilder.py:299-346` pass because they assert only the `max()` outcome.

Two more, smaller.

The manifest records only top-level aliases at `:382-394`, so a transitive dependency that publishes a new version is never invalidated on a warm cache.

Nothing in the repo emits `type="module"` — the emitter I read is `gnrhtmlpage.py:155`, a bare `script(src=...)` — while the bundles are `--format=esm` at `:490` and `:512`. A classic script tag on a file containing `export` is a SyntaxError. I did not trace the rootPage emitter, so tell me if I am looking at the wrong one.

On the tests: everything security-relevant is untested. `test_esmbuilder.py` covers URL construction, spec parsing and manifest round-tripping; nothing exercises `download_package`, `get_esbuild`, `_symlink_package` or `bundle`.

The branch is also 70 commits behind develop with conflicts in `gnrapp.py` and `pyproject.toml`, so the green CI ran against a stale base.

Finally, deployed instances will ship third-party code with no attribution. `LICENSE-3RD-PARTY` covers every bundle vendored so far; these are generated per instance, so there is nothing to commit, but the mechanism to produce the attribution does not exist either. Worth deciding now rather than after the first deploy.

cgabriel:
Thanks for the review — all findings addressed except one, which turned out to be moot.

- esbuild pinning: pinned to an exact version (0.28.2) instead of resolving latest at build time. 

- Bundling on every app start: moved the GnrInstanceEsmBundler call so it only runs under checkdepcli (i.e. the explicit dependency-check/deploy paths — gnrcheckdep, gnrdeploybuilder), never during ordinary GnrApp init. It's also now wrapped in try/except, so a network failure or unsupported-platform error can no longer take down a web worker, CLI command, or batch process.

- No download verification: every download (esbuild's binary and every npm package) is now checked against the npm registry's dist.integrity (sha512/384/256) or dist.shasum fallback before it's ever extracted.

- Alias/package-name injection: the alias is now validated as a plain identifier before it's used as a filename or JS binding name, which closes both the path-traversal (os.path.join) and the JS-injection vector you flagged. I also added validation for the package name itself — while reviewing this I found the raw name was used unsanitized in the node_modules symlink step too (a name containing / or quotes would reach os.symlink directly), so that's now closed as well. Both name and alias interpolation into generated JS now go through json.dumps rather than naive single-quoting, as defense in depth.

- Unfiltered extraction on 3.11: now uses tarfile.data_filter when the interpreter provides it, with a real manual fallback (rejects symlinks/hardlinks and any member that would land outside the destination) for interpreters that don't.

- Version matching: rewrote _resolve_version with correct ^/~/>=/>/<=/</= semantics. ^18.2.0 no longer matches 18.0.0 (floor is now enforced, not just major-equality), and >=18 now correctly matches 19, 20, etc. Added test cases for exactly the two failure modes you described.
  
 - Lockfile: added a per-package esm_requirements.lock, committed alongside esm_requirements.txt, pinning the exact resolved version and full transitive dependency tree. A build reuses the lock (skipping range re-resolution against the registry entirely) as long as the declared spec hasn't changed; it's regenerated automatically otherwise.
  
- Manifest only recording top-level aliases: bundle() now also writes dependencies.json with the full resolved tree (every transitive package, not just top-level aliases), so drift is auditable.

- type="module" / script emission: there's currently no code anywhere in the web layer (gnrhtmlpage.py, gnrwebpage.py, gnrbaseclasses.py) that emits a <script> tag for gnr_ext_bundle.js or any per-alias bundle. The bundles aren't wired into page rendering at all yet, so this isn't a live bug in the current branch — it's a gap in whatever follow-up wires bundle consumption into pages. Flagging so it's tracked, not silently dropped.

- Untested security paths: added tests for download_package, get_esbuild, _symlink_package, bundle, and the new lock/attribution mechanisms — all running against real tarballs, real filesystem operations, and a real (faked) esbuild subprocess; only the npm registry's JSON API is mocked.

- Stale base: the branch already merged develop and is 0 commits behind, so CI is running against current develop. This is normal when a review comes late and the destination branch proceeds
  
- License attribution: bundle() now writes THIRD-PARTY-LICENSES.txt into the output directory alongside the bundles, aggregating the license identifier and license text (when present in the package) for every package in the resolved dependency tree.
  
- All changes are covered by 98 tests in test_esmbuilder.py (flake8 clean, full suite green).

fporcari (changes requested):
Re-reviewed on d29e35e7. Most of @genro's 17/09 list is closed, and I checked it against the code rather than the reply: esbuild is pinned (`esmbuilder.py:25`, `:205`), every tarball is verified against `dist.integrity`/`dist.shasum` before extraction (`:134-161`), alias and package name are validated before they reach a path or a JS literal (`:40`, `:52`, `:261`, `:284`) and the entry files go through `json.dumps` (`:730`, `:752`), extraction uses `tarfile.data_filter` with a real fallback (`:163-184`), and `^18.2.0` / `>=18` now resolve correctly. Taking the bundler out of ordinary startup (`gnrapp.py:1202-1213`) removes the boot failure. `test_esmbuilder.py` passes here (98 tests) and flake8 is clean with the repo config.

Requesting changes on four points; the first two are defects in the current wiring.

**1. Package collection diverges from the Python dependency check again.** `check_package_dependencies` walks `self.package_closure()` and reads `entry['folder']` (`gnrapp.py:1191`), which covers packages reached through `required_packages()` and honours the `filename` attribute (`gnrapp.py:1137-1140`). `collect_requirements` walks only `config['packages']` and rebuilds the folder as `os.path.join(pkg_path_from_attrs(...), package)` (`esmbuilder.py:794-800`). So an `esm_requirements.txt` in a package loaded through `required_packages()` is silently ignored, and a package declared with `filename=` is looked up in the wrong folder. Same class as #1060; the fix is to iterate `self.app.package_closure().values()` and use `entry['pkgid']` / `entry['folder']`. #1383 is heading the same way for the image build: one closure walk, reading `required_packages()` with `ast`, whatever command ends up exposing it, that every per-package manifest is collected from. `esm_requirements.txt` belongs to that walk too; a third, private way of listing packages is what that issue is trying to retire.

**2. A failed bundle now passes checkdep.** Under `checkdepcli` the bundler is wrapped in `except Exception: logger.exception(...)` (`gnrapp.py:1216-1221`), and `gnrcheckdep` never looks at the outcome, so `gnr app checkdep` prints "All good!" and exits 0 (`gnrcheckdep.py:92`) with no bundle written. Ordinary startup no longer checks anything, so an image built this way boots without its bundles and nothing reports it. On 03/08 you argued that missing JS dependencies should stop things; this is the place for it: let checkdep fail with its own exit code, and keep ordinary startup to a network-free check that logs when the manifest does not match the declared specs. This is also where #1383 puts the guarantee for the whole image: the generated Dockerfile runs `checkdep` as a hard gate, and CI proves the built image. A bundle failure that still exits 0 goes straight through that gate, so whatever is decided on placement, the failure has to reach the exit code.

**3. The lock does not pin the transitive tree.** The reply says a build reuses the lock, skipping range re-resolution entirely. `load_lock` is read only for `aliases` (`esmbuilder.py:806`, `:818-821`); the `dependencies` map written at `:572` is never read back, and every transitive dependency is still resolved from its `package.json` range against the registry (`:463-475` -> `download_package` -> `_resolve_version`). Two cold caches built a week apart can bundle different bytes for the same lock. Either feed the locked versions into `_download_all`, or say that the lock pins top-level aliases only.

**4. Common npm range forms abort the whole build.** `_resolve_version` accepts a single comparator (`:318-324`); anything else raises `Unparseable version spec`, and `_download_all` re-raises the first error (`:510-512`). Calling it against a stubbed registry: `^16.8.0 || ^17.0.0 || ^18.0.0`, `*`, `1.x`, `>=17.x`, `>=1.0.0 <2.0.0` and `npm:other@1.0.0` all raise, and `1.0.0 - 1.5.0` is not recognised as a range at all (`:289`), so it is requested as a literal version. `||` ranges are common in `peerDependencies`, which are followed unless marked optional (`:466-469`), so a library that installs fine with npm can fail here, and with point 2 it fails silently.

Smaller:
- The cached esbuild is not keyed by version (`:191-195`): a machine that downloaded `latest` with the earlier code keeps running that binary, and a future bump of `_ESBUILD_VERSION` never reaches a warm cache. Putting the version in the cached filename fixes both.
- The description still says bundles can be referenced via `js_requires='esm/<name>'`. The root page emits `js_requires` as `<script type="text/javascript">` (`gnrjs/gnr_d11/tpl/gnr_header.tpl:69`), and an ESM bundle loaded that way is a SyntaxError. If page consumption is out of scope, drop that line from the description and open the follow-up issue.

Not verified: I did not boot an instance or run checkdep against the real registry. Points 1-3 come from reading the code at d29e35e7, point 4 from calling `_resolve_version` with a stubbed registry.

The placement question from @genro's first review stays his call; this review does not decide it.

cgabriel:
Thanks for checking. Everything below is addressed in 58c815e.

1. Package collection. collect_requirements now walks self.app.package_closure().values() and reads entry['pkgid'] / entry['folder']. That is the same walk check_package_dependencies uses, so there is no longer a separate way of listing packages. So packages reached only through required_packages() are now collected, and packages declared with filename= are looked up in the right folder. A new test on a real GnrApp in checkdep mode covers both cases. When #1383 lands a shared closure walk, esm_requirements.txt can move onto it with no other change. 
 
2. Failed bundles now fail checkdep. The GnrApp constructor no longer bundles at all.
  - Ordinary startup: runs check_esm_bundles(). It doesn't touch the network: it compares the manifest with the declared specs and checks that the output files exist. On a mismatch it logs a warning that points to gnr app checkdep <instance>.
  - gnr app checkdep: builds through app.build_esm_bundles(), which raises on any failure, and the CLI turns that into exit code 5. The existing codes are unchanged: 2 missing, 3 wrong, 4 --strict. A bundle failure now stops the checkdep gate that #1383 relies on.
  - gnrdeploybuilder: it used to bundle as a side effect of checkdepcli=True. Now it only runs the offline check.

3. The lock pins the transitive tree. The lock's dependencies map is now read back and passed into _download_all as pins. A pinned version replaces a range for any package, transitive ones included, whenever it satisfies that range. A lock is only used if its  aliases match the specs the package declares now, so a stale lock can't hold back a changed spec. When two packages' locks disagree on a version, the first package in closure order wins. A test publishes a newer transitive release after the lock is written and checks that a forced rebuild keeps the locked version without fetching that package's version list.
 
4. Range forms. _resolve_version now uses a range parser that follows node-semver's grammar:
  - || unions and space-separated comparator sets
  - hyphen ranges (1.0.0 - 1.5.0, which _needs_resolution now recognises as a range)
  - X-ranges (*, x, 1.x, >=17.x)
  - ~>, and whitespace after an operator
  
npm:other@range aliases are fetched as other and installed under the dependency's own name. Every form you listed now resolves, and each one has a test against a stubbed registry, including an || peer dependency followed end to end. Unparseable specs fail before any network call. git, URL and file: sources fail with an explicit "Unsupported dependency source" error. Pre-release versions are still never chosen.

Smaller points:
 - esbuild cache: the cached binary is now esbuild-<_ESBUILD_VERSION>. A binary left by the old unpinned download is ignored, and bumping the version reaches warm caches.
  - js_requires: I've dropped the js_requires='esm/<name>' line from the description. Loading the bundles in pages is out of scope here, and I'll open a follow-up issue for loading them as <script type="module">.
    
test_esmbuilder.py now has 145 tests and passes, and flake8 is clean on all the files I touched. 

Placement is still @genro's call.

fporcari (approved):
Re-reviewed on 58c815e: the four points of my 22/09 review are closed, and I checked them against the code rather than the reply.

1. `collect_requirements` walks `self.app.package_closure().values()` and reads `entry['pkgid']` / `entry['folder']` (esmbuilder.py:926-971), the same walk as `check_package_dependencies` (gnrapp.py:1191). `test_collects_required_and_filename_packages` covers both cases on a real `GnrApp`.
2. `gnr app checkdep` builds through `build_esm_bundles` and exits 5 on any failure (gnrcheckdep.py:94-97). Ordinary startup only runs `check_esm_bundles`, which is network-free and logs a warning (gnrapp.py:1210-1225).
3. The lock's `dependencies` are read back as pins (`_collect_pins`, :973) and replace a range in `_download_target` (:526) whenever they satisfy it, transitive packages included.
4. Against a stubbed registry, `^16.8.0 || ^17.0.0 || ^18.0.0`, `*`, `1.x`, `>=17.x`, `>=1.0.0 <2.0.0`, `1.0.0 - 1.5.0`, `1.2.3 - 2` and `npm:other@1.x` all resolve as node-semver would, and git, URL and `file:` sources fail with an explicit error.

`test_esmbuilder.py` and `test_package_path_attr.py` pass here (150 tests), and the CI flake8 selection is clean on the touched files. The three `tests/app` failures I get locally fail the same way on develop.

One note, not blocking: `_update_package_locks` (:1004-1016) writes the whole instance tree from `dependencies.json` into every package's `esm_requirements.lock`. A package used by two instances with different companions gets its committed lock rewritten differently by each, and it pins names it does not depend on. Restricting each lock to the subtree reached from that package's own aliases would keep it stable; a follow-up issue is fine.

As you say, placement (bundling from checkdep versus a build step with committed bundles) is still @genro's call, and this approval does not decide it.

Not verified: I did not boot an instance against this branch, and I did not run a build against the real npm registry and esbuild.

PR #868, opened 2026-05-04
map paramiko's remove() method to _Connection method with the same name

### LINKED ISSUE #848: [BUG] paramiko ftp service _Connection wrapper missing remove/unlink
## Problem Description

The new `paramiko/service.py` SFTP service introduced in #847 / PR #847 wraps `paramiko.SFTPClient` in a custom `_Connection` class. The wrapper exposes `listdir`, `isdir`, `stat`, `get`, `put`, `close` — but **not** `remove` / `unlink`.

`pysftp.Connection` (the class returned by the existing `pysftp/service.py`) exposes both `remove` and `unlink`. Any caller that downloads a file and then deletes it from the remote — a natural pattern for any polling/drain workflow — works against `pysftp/service` but breaks against `paramiko/service` with:

```
'_Connection' object has no attribute 'remove'
```

This makes `paramiko/service` non drop-in: instances that switch implementation to work around the Ed25519 issue (#845) lose functionality silently until they hit it at runtime.

### Root cause

[`projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py`](https://github.com/genropy/genropy/blob/845-bug-pysftp-service-fails-with-ed25519-and-other-modern-private-keys/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py) defines `_Connection` with a curated method set and does not forward `remove`/`unlink` to the underlying `paramiko.SFTPClient`, which does support them.

## Reproduction Steps

1. Configure a `sys.service` of type `ftp` with `implementation=paramiko/service` against any reachable SSH server (Ed25519 key OK).
2. From any code path that uses the service:
   ```python
   service = self.page.getService('ftp', service_name='<svc>')
   with service() as conn:
       files = conn.listdir()
       conn.remove(files[0])   # <- AttributeError
   ```
3. Observed: `'_Connection' object has no attribute 'remove'`.
   Expected: file removed from the remote, mirroring `pysftp.Connection.remove(...)`.

## Environment

- genropy: branch `845-bug-pysftp-service-fails-with-ed25519-and-other-modern-private-keys` (PR #847)
- Python: 3.12
- paramiko: 4.x
- OS: macOS / Linux (platform-independent)

## Proposed Fix

Add to `_Connection`:

```python
def remove(self, remotepath):
    self._sftp.remove(remotepath)

unlink = remove
```

Both names because `pysftp.Connection` exposes both — keeping the wrapper drop-in compatible.

A broader option, if the goal is full parity with `pysftp.Connection`, is to delegate any unknown attribute to `self._sftp` via `__getattr__`. The narrow fix above is enough to unblock the SFTP polling use case without expanding the surface area beyond what's needed.

cgabriel:
added remove() method as requested in #848 allowing file deletion, useful for batch that download files via sftp, removing them after the processing.

close #848

### DIFF
--- a/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py
+++ b/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py
@@ -31,6 +31,9 @@ class _Connection(object):
     def remove(self, path):
         self._sftp.remove(path)
 
+    def unlink(self, path):
+        self._sftp.remove(path)
+
     def get(self, remotepath, localpath, callback=None, preserve_mtime=False):
         self._sftp.get(remotepath, localpath, callback=callback)
         if preserve_mtime:
--- a/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py
+++ b/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py
@@ -28,11 +28,11 @@ class _Connection(object):
     def stat(self, path):
         return self._sftp.stat(path)
 
-    def remove(self, path):
-        self._sftp.remove(path)
+    def remove(self, remotepath):
+        self._sftp.remove(remotepath)
 
-    def unlink(self, path):
-        self._sftp.remove(path)
+    # alias, to maintain interface
+    unlink = remove
 
     def get(self, remotepath, localpath, callback=None, preserve_mtime=False):
         self._sftp.get(remotepath, localpath, callback=callback)
--- a/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py
+++ b/projects/gnrcore/packages/sys/resources/services/ftp/paramiko/service.py
@@ -28,6 +28,9 @@ class _Connection(object):
     def stat(self, path):
         return self._sftp.stat(path)
 
+    def remove(self, path):
+        self._sftp.remove(path)
+
     def get(self, remotepath, localpath, callback=None, preserve_mtime=False):
         self._sftp.get(remotepath, localpath, callback=callback)
         if preserve_mtime:


dgpaci (dismissed):
Thanks Christopher — the forward to `paramiko.SFTPClient.remove()` is correct. A couple of points before merge:

### Incomplete relative to #848
The issue explicitly asks for **both** `remove` and `unlink`:

> Both names because `pysftp.Connection` exposes both — keeping the wrapper drop-in compatible.

`pysftp.Connection.unlink` is part of the public surface, so any caller using `conn.unlink(...)` will still hit `AttributeError` against the paramiko implementation after this PR — same class of silent failure the bug is about. Suggest:

```python
def remove(self, remotepath):
    self._sftp.remove(remotepath)

unlink = remove
```

### Minor
- **Parameter name**: the rest of `_Connection` uses `remotepath` (`get`, `put`, `_sftp.utime(remotepath, ...)`). `path` here is inconsistent — please align to `remotepath` to match the surrounding style and the signature proposed in #848.
- **`RELEASE_NOTES.rst`**: not updated. A one-liner under Fixes referencing #848 would close the loop per the project rule.
- **Base branch**: PR targets `master`. Standard rule on this repo is `develop`; recent hotfixes (e.g. #849) did go to master, so this may be the accepted hotfix flow — leaving the call to the maintainers, just flagging.

### Tests
Not blocking, but the original failure was silent at runtime. If there's any harness for the paramiko service, a minimal upload → remove round-trip would have caught #848 and would catch the `unlink` regression too.

Once `unlink` and the release-note entry are added, happy to approve.

cgabriel:
> Thanks Christopher — the forward to `paramiko.SFTPClient.remove()` is correct. A couple of points before merge:
> 
> ### Incomplete relative to #848
> The issue explicitly asks for **both** `remove` and `unlink`:
> 
> > Both names because `pysftp.Connection` exposes both — keeping the wrapper drop-in compatible.
> 
> `pysftp.Connection.unlink` is part of the public surface, so any caller using `conn.unlink(...)` will still hit `AttributeError` against the paramiko implementation after this PR — same class of silent failure the bug is about. Suggest:
> 
> ```python
> def remove(self, remotepath):
>     self._sftp.remove(remotepath)
> 
> unlink = remove
> ```

Implemented unlink too as requested.
 
> ### Minor
> * **Parameter name**: the rest of `_Connection` uses `remotepath` (`get`, `put`, `_sftp.utime(remotepath, ...)`). `path` here is inconsistent — please align to `remotepath` to match the surrounding style and the signature proposed in #848.

Done

> * **`RELEASE_NOTES.rst`**: not updated. A one-liner under Fixes referencing #848 would close the loop per the project rule.

RELEASE_NOTES gets updated upon release, when all hotfix have been merged and a release is created.

> * **Base branch**: PR targets `master`. Standard rule on this repo is `develop`; recent hotfixes (e.g. #849) did go to master, so this may be the accepted hotfix flow — leaving the call to the maintainers, just flagging.

It's an hotfix.

dgpaci (commented):
Updates look good — both points from the previous round are addressed:

- ✅ `unlink = remove` added with a clear comment, restoring drop-in parity with `pysftp.Connection`
- ✅ parameter renamed to `remotepath` for consistency with `get`/`put`

One item still pending before merge:

- **`RELEASE_NOTES.rst`**: still not updated. Per the project rule, every change to `develop` (and hotfixes too, in practice) needs an entry. A one-liner under Fixes referencing #848 is enough — e.g. `paramiko SFTP service: add remove()/unlink() to _Connection wrapper (#848)`.

Add that and I'll approve. The base-branch question (master vs develop) I'll leave to the maintainers as before.

cgabriel:
> Updates look good — both points from the previous round are addressed:
> 
> * ✅ `unlink = remove` added with a clear comment, restoring drop-in parity with `pysftp.Connection`
> * ✅ parameter renamed to `remotepath` for consistency with `get`/`put`
> 
> One item still pending before merge:
> 
> * **`RELEASE_NOTES.rst`**: still not updated. Per the project rule, every change to `develop` (and hotfixes too, in practice) needs an entry. A one-liner under Fixes referencing #848 is enough — e.g. `paramiko SFTP service: add remove()/unlink() to _Connection wrapper (#848)`.
> 
> Add that and I'll approve. The base-branch question (master vs develop) I'll leave to the maintainers as before.

Please, please, read the previous comments regarding RELEASE_NOTES.rst.

genro (approved):

fporcari (approved):
LGTM. Fix minimo e corretto, esattamente come proposto in #848. Lascio il merge a te.

Issue #865, opened 2026-05-02
ProseMirror-based editor stack: introduce ProseMirror, Milkdown, TipTap


genro:
## Goal

Introduce a ProseMirror-based editor stack alongside the existing
CKEditor / TinyMCE / MDEditor widgets. The three editors will coexist
with the current ones — no replacement is planned at this stage.

### Phase 1 — ProseMirror bundle + test widget (infrastructure)

Vendor ProseMirror core under `resources/js_libs/prosemirror/` with an
esbuild pipeline in `tools/prosemirror-build/`. Expose
`window.ProseMirror.*` (state, view, model, transform, commands,
keymap, history, schema-basic, schema-list, inputrules). Add a
`loadProseMirror(cb)` helper in genro_extra.js with mtime-based cache
busting via `gnr.vendoredMtime.prosemirror`.

Add a thin Genropy widget `pane.proseMirror(value='^.html', ...)` on
top of `baseExternalWidget` so the bundle can be exercised end-to-end
in a real Genropy page (no toolbar, schema-basic + schema-list,
keyboard-driven editing). Demo page included.

**Output formats:** the widget serializes the editor state either as
HTML string (default, drop-in compatible with CKEditor/MDEditor
datapaths) or as ProseMirror JSON document (`format='json'`). When
operating server-side on the structured tree, JSON converts to/from
Bag with `Bag.fromJson(json_str)` / `bag.toJson()` — no custom schema
mapping in the widget.

### Phase 2 — Milkdown (markdown WYSIWYG widget)

Add `pane.milkdown(value='^.md', ...)` widget on top of the
ProseMirror bundle. Vendored Milkdown bundle with commonmark + gfm
presets. Bidirectional value sync, slash commands, inline editing
UX. MDEditor (Toast UI) stays in place — no replacement.

### Phase 3 — TipTap (HTML rich text widget with toolbar)

Add `pane.tiptap(value='^.html', toolbar='standard', ...)` widget.
Vendored TipTap bundle with starter-kit + table + link + image +
text-align + color + highlight + bubble-menu + floating-menu.
Custom Genropy toolbar (replicating CKEditor's toolbar_dict presets)
and dialogs (Link, Image, Table, Source). CKEditor stays in place
— no replacement.

## Out of scope

- Replacing CKEditor with TipTap or MDEditor with Milkdown.
  Coexistence only. Replacement, if ever, is a separate decision
  after the new widgets have been used in production for some time.

## Sub-PRs

[...]

PR #869, opened 2026-05-04
proseMirrorEditor: introduce ProseMirror bundle and widget

genro:
## Summary

Phase 1 of the editor stack consolidation (issue #865): introduce a vendored
ProseMirror bundle and a new `proseMirrorEditor` Genropy widget on top of it.

### What this PR lands

- **ProseMirror bundle** under `resources/js_libs/prosemirror/` produced by
  `tools/prosemirror-build/` (esbuild pipeline). Exposes `window.ProseMirror`
  with state, view, model, transform, commands, keymap, history, inputrules,
  schema-basic, schema-list, dropcursor, gapcursor, menu, example-setup,
  tables, trailing-node, collab, changeset.
- **`proseMirrorEditor` widget** (d11 + d20) on `gnr.widgets.baseExternalWidget`,
  with: schema/setup/menubar params, table dropdown, lifecycle hooks,
  markdown/html/bag formats, rebuild mixin.
- **Editor split**: large editor widgets (MDEditor, codemirror,
  proseMirrorEditor, CkEditor, TinyMCE) extracted from `genro_extra.js` into
  a dedicated `genro_editors.js` (d11 + d20). `genro_extra.js` shrinks to
  non-editor widgets only (GoogleLoader, fullcalendar, qrscanner, chartjs,
  dygraph).
- **Cache-bust**: PM bundle and CSS use the existing `gnr.vendoredMtime`
  pattern via `_VENDORED_BUNDLES` in gnrwebpage.
- **Demo pages** under `projects/gnrcore/packages/test/webpages/tools/`:
  `prosemirror.py` with HTML/JSON/Bag/shared-datapath samples, plus a
  `prosemirror_playground.py` exposing every config knob.
- **TestHandlerBase to TestHandlerFull migration** for ~50 test pages
  (separate commit, unrelated to PM but cleaner to ship together).
- **`LICENSE-3RD-PARTY` updated** with attributions for the new ProseMirror
  bundle and the previously-merged CodeMirror 6 bundle (closes deferred
  follow-up from PR #864 review by @cgabriel).

### Forward-port from PR #864

Commit `8d41f021c1` re-applies the four CM6 review fixes (lineWrapping
Compartment, doc.length short-circuit, console.warn on selection-restore
failure) into the codemirror block extracted into `genro_editors.js`.
Without this commit the codemirror widget in this branch would have
regressed to the pre-review-fix CM6 implementation.

## Items needing attention

- **Phase relationship**: this is Phase 1 of the editor stack consolidation
  decided in `temp/decision_editor_stack_2026_05_04.md`. Phase 2
  (`htmlEditor` with textColor mark replacing CKEditor 4 / TinyMCE) and
  Phase 3 (`mdEditor` on Milkdown replacing Toast UI) follow as separate
  PRs. The decision rationale is documented in the design doc.
- **Mark name canonicalization** (`proseMirrorEditor_PM_TO_BAG_MARKS`) is
  biased toward Tiptap StarterKit naming. Per the design doc the wrapper
  is currently not planned to swap, but the canonicalization is harmless
  and forward-compatible.
- **Bundle weight**: the bundle is "pumped" (~288 KB) because it includes
  menu, example-setup, tables, trailing-node, collab, changeset on top of
  the core. Phase 2 will trim what `htmlEditor` does not need; for now the
  fat bundle keeps every demo working end-to-end.
- **TestHandlerBase migration commit (`58f6014f43`)**: ~50 test pages
  migrated. Independently reviewable.

Closes-part-of: #865

## Test plan

- [ ] CI green on Python 3.11-3.14
- [ ] Visit `/test/tools/prosemirror`: HTML/JSON/Bag samples render and edit
- [ ] Visit `/test/tools/prosemirror_playground`: schema/menubar params toggle live
- [ ] Verify `pane.codemirror(...)` still works (CM6 fixes forwarded correctly)
- [ ] flake8 clean on touched Python files (verified locally)


cgabriel (changes requested):
General review is fine, but the editor when text is added doesn't respect the container size and overflows over element at the bottom of it, like in the attached screenshot:

Image

cgabriel (changes requested):
The ProseMirror playground still have problems with CSS:

Image

When more text is added, is flow under the container without a scroll bar, so it can't be read only by moving the cursor down. It need a fix in case this behaviour is present also where this editor is being used.

fporcari (changes requested):
## Review — proseMirrorEditor test pages

Overall the PR is well structured and the widget architecture is solid. However there are a few issues to fix before merging.

### 1. `[object Object]` when loading Bag samples (prosemirror_playground.py)

The "Simple bag" and "Formatted bag" buttons fire `dataRpc` to load a `Bag` into `.source`, but the editor format stays `html`. The editor calls `String(value)` on the Bag object and renders `[object Object]`.

Fix: switch format to `bag` before loading, e.g.:

```python
bag_row.button('Simple bag',
               action="SET .config.format = 'bag'; FIRE .loadBagSimple;",
               margin_right='4px')
bag_row.button('Formatted bag',
               action="SET .config.format = 'bag'; FIRE .loadBagFormatted;")
```

### 2. Keybinding for bold/italic/code not working

`Cmd-B`, `Cmd-I` and `` Cmd-` `` do nothing in any setup (minimal, example, full). The `_buildPlugins` function builds keymaps for undo/redo, lists and tables but never maps mark toggles. `PM.buildKeymap(schema)` is exported in the bundle but never called.

The playground sample text says "Try Cmd-B, Cmd-I" but they don't work.

Fix: add mark toggle keybindings in the example/full branch of `_buildPlugins`:

```javascript
var markKeys = {};
if(schema.marks.strong){ markKeys['Mod-b'] = PM.toggleMark(schema.marks.strong); }
if(schema.marks.em){ markKeys['Mod-i'] = PM.toggleMark(schema.marks.em); }
if(schema.marks.code){ markKeys['Mod-`'] = PM.toggleMark(schema.marks.code); }
plugins.push(PM.keymap(markKeys));
```

Or use `PM.buildKeymap(schema)` directly which handles all of this.

### 3. CSS overflow (still open after latest push)

As @cgabriel noted, the editor overflows its container when content grows — no scrollbar appears and text flows under elements below. The latest commit (38d2d1a) did not fully resolve this. The `.ProseMirror` div likely needs `overflow-y: auto` and a constrained height within its container.

genro:
@cgabriel @fporcari — addressed all three review points in commit 16f942c.

**1. CSS overflow with menubar** (cgabriel + fporcari)

Root cause: when the menubar is enabled, ProseMirror inserts a `.ProseMirror-menubar-wrapper` between `.gnr-prosemirror-host` and the actual `.ProseMirror` div. The existing rules targeted `.gnr-prosemirror-host > .ProseMirror`, which no longer matches when the wrapper is present — so the flex column chain broke and tall documents pushed past the host. Added rules to make the wrapper participate in the flex column and the inner editor surface scroll within the user-supplied height.

**2. Mark toggle keybindings** (fporcari)

Cmd-B / Cmd-I / Cmd-` now work in the example/full setups. Added a `markKeys` keymap (Mod-b/strong, Mod-i/em, Mod-`/code) in `_buildPlugins`, applied to both d11 and d20 bundles. Went with the chirurgical 3-binding approach rather than `PM.buildKeymap(schema)` to avoid clashing with the undo/redo/list keymaps already pushed.

**3. Bag samples → `[object Object]`** (fporcari)

Buttons now `SET .config.format = 'bag'` before firing the loader, exactly as proposed in the review.

Caveats: fixes 1 and 2 require browser verification — they are visual/interactive and I have not been able to test them in a real browser. The playground at `test/tools/prosemirror_playground` is the natural place to check both. Fix 3 is testable by clicking the "Simple bag" / "Formatted bag" buttons.

Branch rebased on current `develop`; CI should re-run on the new commit.

cgabriel (approved):
With the latest fixes, LGTM.

fporcari (dismissed):
All three points from my May review are resolved. I verified them in a real browser this time rather than by reading the diff, since two of the three are behavioral and the overflow one had already survived one fix attempt. Approving.

Setup: `sandbox` instance on this branch, `test/tools/prosemirror_playground`, schema `basic + lists`, menubar on.

### 1. Overflow with the menubar — fixed

The diagnosis in the commit message is right: the menubar wrapper was breaking the flex chain, not the host rules. Measured with the document grown to 866 px against a 535 px editing surface:

- `.ProseMirror` → `scrollHeight 866`, `clientHeight 535`, scrolls internally (`scrollTop` reaches 331, i.e. exactly the overflow)
- `.gnr-prosemirror-host` → `clientHeight 572`, `scrollHeight 572`, does **not** grow past its parent
- the document gains no vertical scrollbar, and the toolbar stays pinned at the top

That is the behavior @cgabriel asked for, with the menubar enabled, which is the case that was broken.

### 2. Mark keybindings — fixed in `example` and `full`

Whole document selected, then:

- `Cmd-B` → `<strong>` 1 → 0 (toggle off across the selection), again → 48 (toggle on)
- `Cmd-I` → the sample's single `<em>` removed
- `` Cmd-` `` → all 9 `<code>` removed

Control test, same selection in setup `minimal`: `Cmd-B` leaves the 46 `<strong>` untouched. Worth stating because it rules out the browser's native contentEditable bold as an alternative explanation for the positives above — the marks really are coming from the new `markKeys` keymap. `toggleMark` is genuinely exported by the bundle (`tools/prosemirror-build/entry.js:45`), so nothing is silently a no-op.

`minimal` leaving marks unbound matches the preset's documented contract ("just baseKeymap+history+cursors"), so I read that as intended — but the sample text says "Try `Cmd-B`, `Cmd-I`" in every setup, so in `minimal` the page tells the user to press keys that do nothing. Trivial, and only in the playground.

### 3. Bag format switch — fixed, and it uncovered the mirror image

"Simple bag" now flips Format to `gnr.Bag tree` and renders the sample correctly, no more `[object Object]`.

The symmetric case is still open, though, and I reproduced it: with the format left at `bag` after using a Bag button, picking any HTML sample from the **Sample content** selector leaves the editor with a single empty `<p>` and no error at all. The selector only sets the source:

```python
# projects/gnrcore/packages/test/webpages/tools/prosemirror_playground.py:188
controls.dataController("SET .source = SAMPLES_HTML[sample];",
                        sample='^.config.sample', ...)
```

so the HTML string goes through the `bag` branch of `parseInitialDoc`: `gnr.GnrBag().fromXml(html)` yields cells with no `tag` attribute, `_bagChildrenToPmFragment` skips every one of them, and the result is an empty document. Same one-line cure as the Bag buttons got:

```python
controls.dataController("SET .config.format = 'html'; SET .source = SAMPLES_HTML[sample];",
                        sample='^.config.sample', ...)
```

Not blocking — it is a test page, and the PR is otherwise in better shape than when I last looked at it. But it is the exact defect I reported in May with the arrow reversed, so it is worth closing while the context is fresh.

### Session hygiene

Zero application JS errors and zero failed requests during the run. The four `not existing nodeId:test_0_playground` console errors in my session were mine — failed `genro.getData` probes from the console, not the page.

fporcari (approved):
Merged develop into the branch (`79cf76eda`) so this stops rotting — it was `DIRTY` and none of the conflicts were yours to guess at, they are all consequences of things that landed on develop over the last two months. Re-approving your work, but see the last paragraph: the resolution itself is mine and wants a second pair of eyes.

### What the conflicts actually were

**`gnrjs/gnr_d20` no longer exists.** develop dropped the whole tree in `d7af14807` ("drop gnr_d20 JS duplicate"). I took the deletion, so the d20 half of this PR is gone — including the new `gnr_d20/js/genro_editors.js` — and `dojo_20.py` keeps develop's js list, so the branch no longer references a file it does not ship. Worth flagging for anyone doing a similar merge: **git reports no conflict at all for a file added inside a tree the other side deleted**, so that new d20 editors file would have quietly resurrected the tree. It had to be removed by hand.

**`gnrwebstruct.py` became a package** (`890ce7136`). The one line this branch added there — `'proseMirrorEditor'` in the dojo11 `gnrNS` list — is now in `gnrpy/gnr/web/gnrwebstruct/dojo11.py`. develop also freezes the cardinality of that namespace in `gnrpy/tests/web/gnrwebstruct_test.py` ("drift in either direction must be intentional"), so the snapshot moves 256 → 257: the drift is intentional and it is this widget.

**The `genro_extra.js` → `genro_editors.js` move would have reverted a landed fix.** This branch moves the editor widgets out of `genro_extra.js`; meanwhile `5f461bbf6` (issue #962, June 30) fixed `loadCodeMirror6` in the version of that file still on develop, so the moved copy here carried the pre-fix code. Taking either side wholesale loses something. I kept your move and ported the dedupe verbatim into `genro_editors.js` — diffed byte-for-byte against develop's version.

**`bordercontainer.py` / `framepane.py`**: took develop's content, including its new `test_5_top_autoheight_hidden_toggle` and its removal of the `dashboard_component` require, keeping your `TestHandlerBase` → `TestHandlerFull` migration (develop is at 88 Full / 36 Base, so that direction is settled).

### Verified on the merged tree, not just on the pre-merge branch

Re-ran the checks in a real browser after the merge, since 177 develop commits is enough to break a widget quietly:

- the playground builds, the flex chain is intact (`.ProseMirror` `overflow-y: auto`, `flex: 1 1 auto`, wrapper in the column), and a tall document scrolls inside the host (`scrollHeight 779` vs `clientHeight 518`, max `scrollTop` 261) without the host growing or the page scrolling
- `Cmd-B` still toggles across the selection
- "Simple bag" still flips the format and renders, no `[object Object]`
- **the ported #962 fix works where it matters**: `test/tools/codemirror6` builds 5 CodeMirror 6 editors in one tick and appends exactly **one** bundle `<script>`, with no "multiple instances of @codemirror/state". Without the port those 5 editors would have raced.

flake8 clean on the touched Python files. Full `gnrpy` suite: 1893 passed, 70 skipped, 1 failure — `core/gnrlist_test.py::TestCsvReader::test_auto_dialect`, which I verified fails identically on clean develop.

### Two things for you

- The `SET .config.format = 'html'` one-liner from my previous review is still open — the **Sample content** selector leaves the editor empty when the format is still `bag`.
- The push dismissed both approvals (`dismiss_stale_reviews` is on for `develop`). My approval here covers your changes, which I have now verified twice, but **the merge resolution is my own work and I should not be its only reviewer** — @cgabriel, when you re-confirm, the four resolutions above are the part worth reading, especially the `loadCodeMirror6` port.

Issue #872, opened 2026-05-04
[BUG] watermarkedPDF fails with 'source object number out of range' on PDFs saved incrementally


dgpaci:
## Problem Description

`pdfService.watermarkedPDF()` (in `resources/common/services/pdf/pypdf.py`) raises `RuntimeError: code=4: source object number out of range` when applied to PDFs that were saved with multiple incremental xref generations (e.g. PDFs that have been opened and re-saved by macOS Preview/Anteprima with annotations or signatures, PAdES-signed PDFs, PDFs round-tripped through Adobe Reader with markups, etc.).

The error is raised by MuPDF (the underlying C library used by PyMuPDF) inside `Document.insert_pdf()`, but the trigger is the prior `Document.save()` call done without garbage collection.

Expected behavior: `watermarkedPDF` should produce a watermarked output for any valid input PDF, regardless of how the source PDF was saved.

## Reproduction Steps

1. Take any PDF that has been re-saved incrementally (a quick way: open a PDF with macOS Preview, add a text annotation, save). The file will contain multiple `%%EOF` / `startxref` markers — one per generation.
2. Call `pdfService.watermarkedPDF(filepath, watermark='In attesa validazione', watermark_color='red')` (or any other watermark args) on that file.

Observed result:

```
File ".../resources/common/services/pdf/pypdf.py", line 131, in watermarkedPDF
    for page in self.multipartPDF(input_pdf, output=output_pdf_bytes):
File ".../resources/common/services/pdf/pypdf.py", line 103, in multipartPDF
    resultdoc.insert_pdf(fitz.open('pdf', tmpdoc.read()))
RuntimeError: code=4: source object number out of range
```

PDFs that have a single xref generation (most of those produced internally by reportlab/wkhtmltopdf) are not affected.

## Root Cause

In `multipartPDF` (current code):

```python
srcdoc = fitz.open('pdf', f.read())
tmpdoc = BytesIO()
for page in srcdoc:
    yield page                 # caller does clean_contents() + insert_textbox
srcdoc.save(tmpdoc)            # default save: no garbage collection
tmpdoc.seek(0)
resultdoc.insert_pdf(fitz.open('pdf', tmpdoc.read()))   # crashes here
```

When `srcdoc` has multiple xref generations and pages have been mutated, `srcdoc.save()` collapses the xref but leaves cross-generation references that no longer resolve. The next `insert_pdf` then trips on those orphan references.

[...]

PR #873, opened 2026-05-04
fix(#872): normalize PDF xref before watermark to avoid MuPDF crash

dgpaci:
## Summary

- Fixes ``RuntimeError: code=4: source object number out of range`` raised by ``Service.watermarkedPDF`` when the source PDF has multiple incremental xref generations (e.g. macOS Preview annotations, PAdES signatures, Adobe Reader markups).
- Rewrites the source document with ``garbage=4, clean=True, deflate=True`` at read time so the rest of the watermark pipeline always sees a single, clean xref.
- Adds a release note under a new ``Unreleased`` section in ``gnrpy/RELEASE_NOTES.rst``.

## Root cause

Inside ``multipartPDF``:

```python
srcdoc = fitz.open('pdf', f.read())
for page in srcdoc: yield page                # caller mutates pages (clean_contents + insert_textbox)
srcdoc.save(tmpdoc)                           # default save: no garbage collection
resultdoc.insert_pdf(fitz.open('pdf', tmpdoc.read()))  # crashes on incremental-saved PDFs
```

When ``srcdoc`` carries multiple xref generations, ``save()`` collapses them but leaves cross-generation references that no longer resolve, and ``insert_pdf`` trips on those orphans. This is a long-standing PyMuPDF behavior — see [pymupdf/PyMuPDF#856](https://github.com/pymupdf/PyMuPDF/issues/856) and [discussion #3640](https://github.com/pymupdf/PyMuPDF/discussions/3640). The recommended workaround in the [Using-Incremental-Saves wiki](https://github.com/pymupdf/PyMuPDF/wiki/Using-Incremental-Saves) is to rewrite the document with ``garbage=4, clean=True, deflate=True``.

## Fix

One-liner inside ``multipartPDF`` that normalizes the source document before the page iteration:

```python
srcdoc = fitz.open('pdf', f.read())
srcdoc = fitz.open('pdf', srcdoc.tobytes(garbage=4, clean=True, deflate=True))
```

Applied at read time (rather than at save time) so the entire downstream pipeline stays unchanged and any future code added to ``multipartPDF`` is automatically protected.

## Test plan

- [x] Reproduced the original crash on a real-world Preview-saved PDF (1 page, 407 KB, ``Producer: macOS ... Quartz PDFContext, AppendMode 1.1``, 3 ``%%EOF`` markers).
- [x] Imported the patched ``pypdf.Service`` directly and re-ran ``watermarkedPDF()`` on the same file: now succeeds, output 400 KB.
- [x] Re-ran on a clean control PDF (single xref): no regression, output equivalent.
- [x] ``flake8`` clean on ``resources/common/services/pdf/pypdf.py`` (``--max-line-length=120``).

## Backward compatibility

None affected. PDFs that already had a single xref generation pass through ``tobytes(garbage=4, ...)`` essentially unchanged (a few bytes of difference from the rewrite, no semantic change).

## Downstream

Tracking ticket on the field deployment that surfaced the bug: softwellsrl/anaci#337.

### DIFF
--- a/gnrpy/RELEASE_NOTES.rst
+++ b/gnrpy/RELEASE_NOTES.rst
@@ -1,3 +1,18 @@
+Unreleased
+==========
+
+Bugfixes
+--------
+
+* PDF watermark service: fix ``code=4: source object number out of
+  range`` when ``watermarkedPDF`` is applied to PDFs saved with
+  multiple incremental xref generations (e.g. macOS Preview
+  annotations, PAdES signatures). The source PDF is now normalized
+  with ``garbage=4, clean=True, deflate=True`` before the watermark
+  is composed, following the official PyMuPDF workaround. No
+  backward-compatibility impact: PDFs that already had a single xref
+  generation are unaffected. (#872)
+
 Release 26.04.30
 ================
 
--- a/resources/common/services/pdf/pypdf.py
+++ b/resources/common/services/pdf/pypdf.py
@@ -95,6 +95,10 @@ class Service(PdfService):
         for docpath in documents:
             with self.parent.storageNode(docpath).open('rb') as f:
                 srcdoc = fitz.open('pdf',f.read())
+                # rewrite xref to one clean generation: incremental-saved
+                # PDFs (macOS Preview, PAdES signatures, ...) otherwise make
+                # the later insert_pdf raise 'source object number out of range'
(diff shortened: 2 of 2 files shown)


cgabriel (approved):

genro (approved):
LGTM. CI green on 3.11/3.12/3.13/3.14. Targeted fix that normalizes the PDF xref before applying the watermark, avoiding the MuPDF crash on files with an out-of-range object. Already approved by cgabriel.

PR #876, opened 2026-05-04
feat(#874): consolidate connection folders cleanup into single on-event mechanism

### LINKED ISSUE #874: Connection folders cleanup: consolidate into a single on-event mechanism
## Background

The `data/_connections/<conn_id>/<page_id>/` filesystem hierarchy is used by the framework to store transient per-page artifacts (frozen selections, page-local documents created via `pageLocalDocument`, `freezeSelection`, `temporaryDocument`). Over the years the cleanup strategy has gone through several incarnations and the current state is fragmented and broken in subtle ways.

### Historical timeline

**2009-09-13** (`ed5218d71e`): introduced `if counter == 0: onInited(clean=True)` to discriminate the first server boot from restarts. The counter was incremented externally by a self-restart wrapper (`serverwsgi_old.py`), which at every reload spawned the subprocess with `--counter N+1`. At first boot the cleanup ran; on subsequent restarts it didn't.

**2010-05-20** (`5a93134024`): added a runtime amortized cleanup, `connFolderRemove(connection_id, rnd=True)`, called on connection close. It removed the connection's own folder, and with `random.random() > 0.9` (10% probability) it also scanned for orphan folders not present in the live register. Quiet, distributed cleanup.

**Some later refactor** (untracked): the `connFolderRemove` definition was commented out. From this point on, runtime filesystem cleanup stopped happening at all. Folders accumulated silently, but nobody noticed because there was no log.

**2019-04-11** (`91158db872`): removed `serverwsgi_old.py` together with seven other legacy entry points. The self-restart wrapper that incremented `--counter` was deleted. From this moment on, `counter` is always `0` at every process boot, including subprocesses of werkzeug `run_with_reloader`. The check `if counter == 0 and self.debug:` therefore always passes in dev mode with `--reload`. No one noticed because:

- production runs with `debug=False`, so the check doesn't fire there;
- in dev, the cleanup happened silently (no log) until 2025.

**2025-04-15 (PR #159)**: re-introduced `connFolderRemove` (without the lottery), added `logger.info("Purging connection folders")` to `dropConnectionFolder` and `connFolderRemove`. Behavior is functionally identical to before, but now visible in the logs. Suddenly, dev users on `--reload` started seeing "Purging connection folders" at every reload of a `.py` file, and asking why their live sessions were being wiped.

### Current state (verified empirically)

On a production server (gunicorn + multiple workers, `debug=False`):

- The startup purge never fires (`debug=False`).
- `connFolderRemove`, `pageFolderRemove` exist as functions but are never called from runtime code (verified by `grep` over the entire repo).
- `register.cleanup()` drops connections from the in-memory register at expire (default 600s) but does **not** touch the filesystem.
- `rpc_logout` and `rpc_onClosePage` drop from register but do **not** touch the filesystem.

Test on a live deploy: opened a page with a `freezeSelection` query → `_connections/<conn>/<page>/V_*.pik` files created. Closed the page → folder remains. Logged out → folder remains. **Folders accumulate indefinitely**.

In development with werkzeug `--reload`:

- The startup purge fires at every reload.
- Live sessions (in the register) survive but their on-disk freeze files are wiped.
- On a Windows dev machine the symptom is amplified by `shutil.rmtree` interacting with Windows Defender locks.

## Problem statement

We need to put an end to four conflicting partial mechanisms:

1. Startup purge gated by a check (`counter == 0 and self.debug`) that is always true on every werkzeug-reload subprocess.
2. `--noclean` CLI flag whose default makes the purge fire.
3. `connFolderRemove` / `pageFolderRemove` functions defined but never called.
4. No connection of any kind between in-memory register lifecycle (drop on expire/logout) and filesystem cleanup.

The result, depending on environment:

- **Production**: filesystem grows without bound. No automatic cleanup.
- **Development with werkzeug `--reload`**: aggressive purge at every reload, surprising and sometimes destructive.
- **Development with `--noclean`**: same as production (no cleanup at all).

### Remove all startup mechanisms

- Remove the `counter == 0 and self.debug` check in `gnrwsgisite.py:516` and the call to `onInited(clean=...)` it gates.
- Remove the `--noclean` CLI flag in `serverwsgi.py:200-203`.
- Remove the `noclean` parameter and the `counter` parameter from `GnrWsgiSite.__init__`.
- Keep `dropConnectionFolder()` only as the underlying primitive used by the existing `gnrcleanup` CLI for explicit on-demand wipes.

### Single on-event cleanup driven by `rpc_onClosePage`

Coordination is delegated to the register daemon via a new atomic method `claim_cleanup(min_gap_seconds)`. The daemon is the single source of truth for `last_cleanup` and is shared by all workers (and by all pods in Kubernetes deployments where the register is centralized).

When `rpc_onClosePage` arrives:

1. **Local lottery** in the worker: `random()*100 < threshold` (default 5%). If it loses, return immediately. No roundtrip, no I/O.
2. **If it wins**, call `register.claim_cleanup(interval_seconds)` on the daemon. The daemon, atomically:
   - reads `last_cleanup`,
   - if `time.time() - last_cleanup < interval` → returns `False`;
   - otherwise sets `last_cleanup = time.time()` and returns `True`.
3. **If `False`**, return. Another worker (or another invocation) already claimed the slot, or not enough time has passed.
4. **If `True`**, spawn a daemon thread inside the current worker. The thread:
   - lists immediate children of `_connections/`;
   - for each subdir with `mtime` older than `min_age` (default 120 min): if `connection_id` is **not** in `register.connections()` → `shutil.rmtree(..., ignore_errors=True)`;
   - returns. No further coordination needed: the daemon already updated `last_cleanup`.

Because `claim_cleanup` is atomic (Pyro4 serializes calls on the singleton register), there is no need for filesystem `flock` or process-level `threading.Lock`. The daemon is the arbiter.

### Resilience to crashes

If the worker that won the claim crashes during the cleanup (worker recycled by gunicorn `max_requests`, OOM kill, network error, etc.), `last_cleanup` has already been advanced on the daemon. Consequence:

- The cleanup that was in progress is **truncated**: some orphan folders that should have been removed remain on disk.
- For the next `interval` minutes (4h default), no other worker will start a new cleanup, because `claim_cleanup` will return `False` for them.
- After `interval` minutes the next `rpc_onClosePage` that wins the lottery and calls `claim_cleanup` will succeed → a fresh cleanup starts → the folders missed by the crashed cleanup will be picked up at this point.

So the system **self-heals** within at most one `interval` window. No supervisor or watchdog is needed. Worst case: leftover orphans live a few extra hours before being collected.

### Configuration: `<cleanup>` in siteconfig.xml

Three new attributes, all optional with sensible defaults:

\`\`\`xml
<cleanup
    folder_purge_threshold="5"
    folder_purge_min_age_minutes="120"
    folder_purge_interval_minutes="240"
/>
\`\`\`

| Attribute | Default | Meaning |
|---|---|---|
| `folder_purge_threshold` | `5` | Percent probability the lottery fires on each `rpc_onClosePage` |
| `folder_purge_min_age_minutes` | `120` | Minimum folder age (minutes) to be a removal candidate |
| `folder_purge_interval_minutes` | `240` | Minimum gap (minutes) between two cleanup cycles |

The existing `<cleanup interval="..." page_max_age="..." connection_max_age="..." guest_connection_max_age="..."/>` attributes are preserved as-is.

### What stays the same

- `gnrcleanup <site>` CLI: explicit on-demand full wipe. Unchanged.
- Register expire logic (in-memory drop after `connection_max_age`): unchanged.
- `connectionLog('open'/'close')` and `adm.connection` table integration: unchanged.

### Why this approach

- **Zero overhead on hot paths**: regular HTTP requests (load, query, save) pay nothing. Cleanup logic only executes on the relatively rare `rpc_onClosePage` event, with a 5% chance, and with an additional 4-hour gate enforced by the daemon.
- **Self-balancing across deployments**: high-traffic sites have many close-page events → the lottery succeeds often → `last_cleanup` advances regularly. Low-traffic sites accumulate slowly and clean rarely. Both cases scale naturally.
- **One coordinator, no distributed locks**: the daemon is already a singleton across all workers and (when configured as such) across pods in Kubernetes. Atomic check-and-set on `last_cleanup` is sufficient. No filesystem `flock`, no per-process `threading.Lock`.
- **Crash-safe**: a crashed cleanup leaves leftovers but does not corrupt state. Self-heals within one `interval` window.
- **Werkzeug reload safe**: no purge at boot means subprocesses spawned by `run_with_reloader` no longer destroy live sessions on dev machines.
- **Conservative**: a folder is only removed if no live connection claims it AND it is at least `min_age_minutes` old. New connections are safe by construction.

## Files affected

- `gnrpy/gnr/web/gnrwsgisite.py`
- `gnrpy/gnr/web/serverwsgi.py`
- `gnrpy/gnr/web/gnrwebpage.py` (the `rpc_onClosePage` path)
- `gnrpy/gnr/web/daemon/siteregister.py` (add `claim_cleanup` method)
- `gnrpy/gnr/web/daemon/siteregister_client.py` (proxy method)

## Out of scope (separate tickets)

- Tornado branch reloader behavior.
- Resurrecting full orphan scans on demand. The new mechanism subsumes that case.
- Documenting `gnrcleanup` more visibly in the deploy templates.
- Investigating Windows-specific filesystem lock issues (Defender) — likely irrelevant once the on-event cleanup replaces mass-delete-at-startup.

## Acceptance criteria

- No `Purging connection folders` log lines during normal operation.
- Werkzeug `--reload` reloads do not destroy live `_connections/<conn>/<page>/` content.
- Production deployments (single host or Kubernetes) do not accumulate orphan folders indefinitely (verified after 24h of normal traffic).
- A simulated worker crash during cleanup does not leave the system in an inconsistent state — the next cleanup cycle picks up where the crashed one left off.
- `gnrcleanup <site>` CLI continues to work as a full manual wipe.
- All three new `<cleanup>` defaults can be overridden via siteconfig.

genro:
## Summary

- Removes startup connection-folder purge (counter, noclean, --noclean, --counter)
- Adds on-event cleanup hooked to `rpc_onClosePage` with lottery + atomic `claim_cleanup`
- Three new optional `<cleanup folder_purge_*>` attributes, defaults 5%, 120 min, 240 min
- Self-healing within one interval window if a worker crashes during cleanup
- Preserves `gnrcleanup` CLI for explicit on-demand wipes (now calls `dropConnectionFolder()` directly)
- Unit tests cover `claim_cleanup` atomicity, lottery branches, and orphan selection logic

Closes #874

## Test plan

- [x] No runtime imports
- [x] Unit tests pass (8/8 in 0.4s)
- [x] CI green on Python 3.11, 3.12, 3.13, 3.14
- [x] Sourcerer cross-repo check: no downstream callers of removed APIs
- [x] Manual T1 verified: werkzeug --reload preserves session `.pik` files (no `Purging` log line)
- [x] Manual T4 verified: `gnrcleanup` CLI still wipes `data/_connections/`

E2E tests with real wall-clock timing are deliberately not added to CI: see the
docstring of `gnrpy/tests/web/gnrwsgisite_folder_cleanup_test.py` for the rationale.

### DIFF
--- a/gnrpy/gnr/web/cli/gnrcleanup.py
+++ b/gnrpy/gnr/web/cli/gnrcleanup.py
@@ -11,7 +11,8 @@ def main():
     parser.add_argument('site_name')
 
     options = parser.parse_args()
-    site = GnrWsgiSite(options.site_name, noclean=False)
+    site = GnrWsgiSite(options.site_name)
+    site.dropConnectionFolder()
 
 if __name__ == "__main__":
     main()
--- a/gnrpy/gnr/web/daemon/processes.py
+++ b/gnrpy/gnr/web/daemon/processes.py
@@ -150,7 +150,7 @@ class GnrRemoteProcess(object):
 
     def _makeSite(self):
         from gnr.web.gnrwsgisite import GnrWsgiSite
-        self._site =  GnrWsgiSite(self.sitename,noclean=True)
+        self._site = GnrWsgiSite(self.sitename)
         self._site_ts = datetime.now()
         self.logger.debug('Created site for PID {}'.format(os.getpid()))
     
@@ -181,7 +181,7 @@ class GnrDaemonServiceManager(object):
     def site(self):
         if not hasattr(self, '_site'):
             from gnr.web.gnrwsgisite import GnrWsgiSite
-            self._site = GnrWsgiSite(self.sitename, noclean=True)
+            self._site = GnrWsgiSite(self.sitename)
         return self._site
 
     def terminate(self):
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -710,6 +710,13 @@ class SiteRegister(BaseRemoteObject):
         self.last_cleanup = time.time()
         return dropped_connections
 
+    def claim_cleanup(self, min_gap_seconds):
+        now = time.time()
+        if now - self.last_cleanup < min_gap_seconds:
+            return False
+        self.last_cleanup = now
+        return True
+
     def get_register(self, register_name):
         return getattr(self, '%s_register' % register_name)
 
--- a/gnrpy/gnr/web/daemon/siteregister_client.py
+++ b/gnrpy/gnr/web/daemon/siteregister_client.py
@@ -324,6 +324,9 @@ class SiteRegisterClient(object):
         else:
             logger.info('UNABLE TO LOAD REGISTER %s' % self.site.site_name)
 
+    def claim_cleanup(self, min_gap_seconds):
+        return self.siteregister.claim_cleanup(min_gap_seconds)
+
     def __getattr__(self, name):
         h = getattr(self.siteregister, name)
         if not callable(h):
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -525,6 +525,9 @@ class GnrWsgiSite(object):
         self.cleanup_interval = int(cleanup.get('interval') or 120)
         self.page_max_age = int(cleanup.get('page_max_age') or DEFAULT_PAGE_MAX_AGE)
         self.connection_max_age = int(cleanup.get('connection_max_age')or 600)
+        self.folder_purge_threshold = float(cleanup.get('folder_purge_threshold') or 5)
+        self.folder_purge_min_age_minutes = int(cleanup.get('folder_purge_min_age_minutes') or 120)
+        self.folder_purge_interval_minutes = int(cleanup.get('folder_purge_interval_minutes') or 240)
 
         self.db.closeConnection()
 
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -1,4 +1,5 @@
 import os
+import random
 import re
 import io
 import shutil
@@ -1739,6 +1740,42 @@ class GnrWsgiSite(object):
         self.pageLog('close', page_id=page_id)
         self.clearRecordLocks(page_id=page_id)
         page._closed = True
+        self._maybeRunFolderCleanup()
+
+    def _maybeRunFolderCleanup(self):
+        """Lottery + atomic claim + spawn cleanup thread.
+
+        Runs only when the local lottery (folder_purge_threshold) wins AND
+        the daemon's claim_cleanup gates the call by interval.
+        """
+        if random.random() * 100 >= self.folder_purge_threshold:
+            return
+        interval_seconds = self.folder_purge_interval_minutes * 60
+        if not self.register.claim_cleanup(interval_seconds):
+            return
+        Thread(target=self._runFolderCleanup, daemon=True).start()
+
+    def _runFolderCleanup(self):
+        """Worker thread: removes orphan connection folders older than min_age."""
+        min_age_seconds = self.folder_purge_min_age_minutes * 60
+        now = time()
+        try:
+            live_connections = set(self.register.connections().keys())
+            for entry in os.listdir(self.allConnectionsFolder):
+                path = os.path.join(self.allConnectionsFolder, entry)
+                if not os.path.isdir(path):
+                    continue
+                if entry in live_connections:
+                    continue
+                try:
+                    mtime = os.stat(path).st_mtime
+                except OSError:
+                    continue
+                if now - mtime < min_age_seconds:
+                    continue
+                shutil.rmtree(path, ignore_errors=True)
+        except Exception:
+            logger.exception("Folder cleanup thread failed")
 
     def sqlDebugger(self,**kwargs):
         page = self.currentPage
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -14,6 +14,7 @@ from threading import RLock, Thread
 import warnings
 
 import requests
+from werkzeug.serving import is_running_from_reloader
 from werkzeug.wrappers import Request, Response
 from werkzeug.utils import redirect
 from werkzeug.exceptions import (HTTPException, InternalServerError,
@@ -407,13 +408,12 @@ class GnrWsgiSite(object):
     """TODO"""
 
     def __init__(self, script_path, site_name=None, _config=None,
-                 _gnrconfig=None, counter=None, noclean=None,
+                 _gnrconfig=None,
                  options=None, tornado=None, websockets=None,
                  debugpy=False):
-        
+
         global GNRSITE
         GNRSITE = self
-        counter = int(counter or '0')
         self.storageTypes = STORAGE_TYPES + STATIC_HANDLER_TYPES
         self.pathfile_cache = {}
         self._currentAuxInstanceNames = ThreadedDict()
@@ -513,11 +513,11 @@ class GnrWsgiSite(object):
         self.register
         
         self.datacollector = DataCollector(self.register.siteregister)
-        
-        if counter == 0 and self.debug:
-            self.onInited(clean=not noclean)
-            
-        if counter == 0 and options and options.source_instance:
+
+        if not is_running_from_reloader():
+            self.onInited()
+
+        if not is_running_from_reloader() and options and options.source_instance:
             self.gnrapp.importFromSourceInstance(options.source_instance)
             self.db.commit()
             logger.info('End of import')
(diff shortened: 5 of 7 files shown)


fporcari:
## PR review

Two blocking issues, plus a behavior change worth discussing and a few minor notes.

### Blocking

**1. `register.connections().keys()` will raise `AttributeError` at runtime.**

In `_runFolderCleanup`:

```python
live_connections = set(self.register.connections().keys())
```

But `SiteRegister.connections()` (`siteregister.py:658`) → `connection_register.connections()` → `self.values()` → `list(self.registerItems.values())`. It returns a **list of dicts**, not a dict. `.keys()` on a list raises.

The unit tests don't catch this because `_FakeRegister.connections()` returns a real `dict` (`{'live1': {}}`). The fake doesn't match the production return shape, so the test passes while production crashes on the first cleanup that wins both gates.

Suggested fix:

```python
live_connections = {c['register_item_id'] for c in self.register.connections()}
```

And update the fake to return a list of dicts so the test exercises the production shape.

**2. `claim_cleanup` shares `last_cleanup` with `SiteRegister.cleanup()` — folder cleanup will essentially never fire.**

`SiteRegister.cleanup()` (`siteregister.py:693-710`) is called on every request via `gnrwsgisite.py:1265`. Every `cleanup_interval` seconds (default 120s) it runs and writes `self.last_cleanup = time.time()`.

The new `claim_cleanup(min_gap_seconds)` reads/writes the **same field**. With `folder_purge_interval_minutes=240` (= 14400s):

- The in-memory expire path resets `last_cleanup` every ~120s under any normal traffic.
- `claim_cleanup(14400)` checks `now - last_cleanup < 14400`, which is essentially always true after a recent expire pass.
- Net effect: `claim_cleanup` always returns `False`. The folder cleanup never runs.

The interference also goes the other way: when `claim_cleanup` does succeed, it advances `last_cleanup` to now, so the next `register.cleanup()` call skips its in-memory expire because the gap got reset.

Suggested fix: introduce a separate field on `SiteRegister`, e.g. `self.last_folder_cleanup` initialized to `0`, used only by `claim_cleanup`. The two cleanup mechanisms have different cadences and shouldn't share state.

### Behavior change worth discussing

**3. `is_running_from_reloader()` makes `initializePackages()` run in production for the first time.**

Before:

```python
if counter == 0 and self.debug:    # False in prod (debug=False)
    self.onInited(clean=not noclean)   # → initializePackages() not called in prod
```

After:

```python
if not is_running_from_reloader():   # True in prod
    self.onInited()                  # → initializePackages() now called in prod
```

`initializePackages()` calls `pkgBroadcast('onSiteInited')`. Before the patch, in production this hook never fired (the gate required `self.debug=True`, which is false in prod). After, it fires on every gunicorn worker boot.

This may well be the *correct* behavior — the original `counter==0` design from 2009 implied `onSiteInited` should fire once at startup, and since 2019 the gate was effectively broken. But it's a real behavior change beyond the stated scope of #874. Worth:

- Documenting in the commit message and PR description.
- Auditing `onSiteInited` handlers across packages for re-entrancy and multi-worker safety (each gunicorn worker will fire it now).

Same concern for the `importFromSourceInstance` branch: with reloader active in dev, it will silently not run (since `is_running_from_reloader()=True` in the werkzeug child where `GnrWsgiSite` is actually instantiated). If someone passes `--source_instance` in a dev-with-reload setup, the import will be skipped without warning.

### Per-page accumulation still unaddressed

I raised this on the issue: the lottery cleanup scans only at the connection level. A long-lived connection (user with the app open all day) accumulates `_connections/<conn>/<page>/` subfolders for every closed page; the connection-level scan correctly skips them as live, so per-page orphans never get collected.

Calling `page.pageFolderRemove()` deterministically inside `rpc_onClosePage` (before `_maybeRunFolderCleanup`) closes that gap cheaply — we already know the exact path to remove.

### Minor

- **No log on successful cleanup.** `_runFolderCleanup` only logs on exception. Add an INFO line with the count of removed folders so operators can verify it runs in production.
- **Empty `_connections/` directory.** Fresh install or freshly-purged sites: `os.listdir(self.allConnectionsFolder)` raises `FileNotFoundError`, which gets swallowed by `except Exception` and logged at exception level on every triggering cleanup. Cheap fix: `if not os.path.isdir(self.allConnectionsFolder): return` at the top of `_runFolderCleanup`.
- **`test_runfoldercleanup_swallows_register_errors`** docstring claims *the error is logged, not propagated*, but the test only checks no exception escapes — the *is logged* half is not asserted. Use `caplog` or rephrase the docstring.
- **`--noclean` removed without a deprecation step.** If any deploy script still passes `--noclean`, argparse will error out. Optional: keep the flag for one release as a no-op + DeprecationWarning.
- **Tornado branch follow-up.** A `# TODO #874-followup` marker in the tornado path would prevent the out-of-scope item from getting lost.

genro:
Superseded by #882.

The original design here had two parallel cleanup mechanisms (legacy `SiteRegister.cleanup()` on every request + new on-event folder purge), which is what triggered the two blocking issues in fporcari's review (`connections().keys()` shape mismatch + `last_cleanup` shared field).

#882 starts from `develop` after #877 (which removed the orphan `onSiteInited` listener) and consolidates everything into a single on-event mechanism: lottery + atomic `claim_cleanup` + thread that expires register state and removes filesystem folders in one pass. The history is clean (6 commits, no force-push), the review feedback from this thread is addressed there.

PR #880, opened 2026-05-05
use  GNR_DB_DSN env var which can override every db configuration

cgabriel:
using a DSN, allowing to defined a different connection if needed, leaving the other genropy's parameters (dbbranch, remotedb etc) intact.

Use cases: 1/ temporary connection to a different database for the same local application. 2/ simplify secret configuration for docker (1 variable instead of 5 variables)


genro (dismissed):
Useful feature, but a few things to address before merge — mainly around making the DSN parsing actually generic instead of postgres-specific.

## Blocking

**1. `implementation` is hardcoded to `'postgres'`**

```python
return {
    'implementation': 'postgres',
    ...
}
```

A DSN URL carries its scheme (`postgres://`, `mysql://`, `sqlite:///`, ...). Hardcoding `postgres` means `GNR_DB_DSN=mysql://...` is silently treated as postgres and will fail at first connect. Suggest:

```python
'implementation': parsed.scheme.split('+')[0]  # postgres+psycopg2 → postgres
```

**2. Query parameters are silently dropped**

A typical postgres DSN: `postgres://user:pass@host:5432/dbname?sslmode=require&connect_timeout=10`. The current parser ignores `parsed.query` entirely. `sslmode`, `connect_timeout`, `application_name`, etc. are critical for production deploys (especially the docker/k8s use case in the PR description). They should be parsed and merged into `dbattrs` (or passed as adapter kwargs).

**3. Username/password not URL-decoded**

`urlparse` doesn't decode `%`-encoded characters. A password containing `@`, `:`, `/`, or `%` must be URL-encoded in the DSN, and the code needs to decode it back:

```python
from urllib.parse import urlparse, unquote
...
'user': unquote(parsed.username) if parsed.username else None,
'password': unquote(parsed.password) if parsed.password else None,
```

Otherwise the docker secret use case breaks for non-trivial passwords.

## Non-blocking

**4. Default port `'5432'` is postgres-specific**

If we deduce `implementation` from the scheme (point 1), the default port should also be scheme-aware (mysql 3306, mssql 1433, oracle 1521). Or just leave `port` unset when not in the DSN and let the adapter pick its own default.

**5. No tests**

The `dsn_to_config()` parsing has several edge cases (missing port, scheme variants, special chars in password, query params). One unit test would document the contract clearly.

**6. No release notes / docs**

This is a user-facing env var that changes behavior. Worth a line in `RELEASE_NOTES.rst` and a mention in the deployment docs.

**7. PR title/description imprecision**

"override every db configuration" is a bit broad — it overrides the connection *target* (host, port, user, password, dbname, implementation) but leaves orthogonal attributes intact (`dbbranch`, `external_host`, pool/adapter kwargs). That's the right behavior, but the description could clarify it to avoid misreading.

**8. Cosmetic**

Trailing whitespace on the line after `dbattrs['application'] = self`, and the new blank line after `dbattrs['implementation'] = ...` doesn't add value.

## Summary

The intent is good — DSN-as-env is a common 12-factor pattern. The implementation just needs to be honest about being scheme-aware (1, 2, 4), URL-safe (3), and tested (5). Once those are addressed I'd be happy to approve.

cgabriel:
> Useful feature, but a few things to address before merge — mainly around making the DSN parsing actually generic instead of postgres-specific.
> 
> ## Blocking
> **1. `implementation` is hardcoded to `'postgres'`**
> 
> ```python
> return {
>     'implementation': 'postgres',
>     ...
> }
> ```

Fixed. The implementation now is coming from DSN scheme, which map the genropy adapter of choice.
 
> **2. Query parameters are silently dropped**
> 
> A typical postgres DSN: `postgres://user:pass@host:5432/dbname?sslmode=require&connect_timeout=10`. The current parser ignores `parsed.query` entirely. `sslmode`, `connect_timeout`, `application_name`, etc. are critical for production deploys (especially the docker/k8s use case in the PR description). They should be parsed and merged into `dbattrs` (or passed as adapter kwargs).

These are not defined via genropy db config and not used, so why bother?
 
> **3. Username/password not URL-decoded**
> 
> `urlparse` doesn't decode `%`-encoded characters. A password containing `@`, `:`, `/`, or `%` must be URL-encoded in the DSN, and the code needs to decode it back:
> 
> ```python
> from urllib.parse import urlparse, unquote
> ...
> 'user': unquote(parsed.username) if parsed.username else None,
> 'password': unquote(parsed.password) if parsed.password else None,
> ```
> 
> Otherwise the docker secret use case breaks for non-trivial passwords.

You're right - fixed.


> ## Non-blocking
> **4. Default port `'5432'` is postgres-specific**
> 
> If we deduce `implementation` from the scheme (point 1), the default port should also be scheme-aware (mysql 3306, mssql 1433, oracle 1521). Or just leave `port` unset when not in the DSN and let the adapter pick its own default.

Handled, if port is provided use it otherwise default to None
  
> **6. No release notes / docs**
> 
> This is a user-facing env var that changes behavior. Worth a line in `RELEASE_NOTES.rst` and a mention in the deployment docs.

Please remove this check from your agent configuration/skill. RELEASE_NOTES.rst are handled at release time, pushing this into develop (like this PR) will cause conflicts that needs to be handled manually when doing a release.

> **7. PR title/description imprecision**
> 
> "override every db configuration" is a bit broad — it overrides the connection _target_ (host, port, user, password, dbname, implementation) but leaves orthogonal attributes intact (`dbbranch`, `external_host`, pool/adapter kwargs). That's the right behavior, but the description could clarify it to avoid misreading.

Your call about this, feel free to change the description of the PR. Important thing is the commit log text.
 
> **8. Cosmetic**
> 
> Trailing whitespace on the line after `dbattrs['application'] = self`, and the new blank line after `dbattrs['implementation'] = ...` doesn't add value.

I like it that way

mbertoldi (approved):

PR #884, opened 2026-05-06
fix(#883): honor dbSelect notnull parameter on the client

### LINKED ISSUE #883: [BUG] dbSelect 'notnull' parameter is ignored client-side, forcing coupling with 'validate_notnull'
## Problem Description

The `dbSelect` server-side handler (`gnrpy/gnr/web/gnrwebpage_proxy/apphandler/db_select.py`) declares a `notnull` parameter that, when `True`, suppresses the empty/null option row in the dropdown:

```python
notnull: When ``True`` no empty-label row is added.
```

The intended semantics: pass `notnull=True` to remove the blank option *without* making the field mandatory. This is distinct from `validate_notnull=True`, which marks the field as required.

In practice the parameter never reaches the server when set on the widget, because the client-side `DynamicBaseCombo.creating()` builds the resolver attributes like this (`gnrjs/gnr_d20/js/genro_widgets.js:4761-4762`, mirrored in `gnr_d11`):

```javascript
var resolverAttrs = objectExtract(attributes, 'method,columns,limit,auxColumns,hiddenColumns,alternatePkey,rowcaption,order_by,preferred,invalidItemCondition,subtable');
resolverAttrs['notnull'] = attributes['validate_notnull'];
```

Two issues:
1. The `objectExtract` allow-list does **not** include `notnull`, so a `notnull` set on the widget is dropped before the resolver is built.
2. The next line **unconditionally overwrites** `resolverAttrs['notnull']` with `attributes['validate_notnull']`.

Result: `notnull` and `validate_notnull` are de-facto coupled. The only way to suppress the empty row is to set `validate_notnull=True`, which also makes the field mandatory — defeating the purpose of having a separate `notnull` flag.

## Reproduction Steps

1. Define a `dbSelect` (or `dbselect`-based field) bound to a table, with `notnull=True` and **without** `validate_notnull`.
2. Open the dropdown.
3. Observe the empty/blank row at the top of the result list.

Expected: no empty row, but the field can still be left blank (no mandatory validation).
Actual: empty row is present; the only workaround (`validate_notnull=True`) also makes the field required.

## Environment

- Affected files: `gnrjs/gnr_d20/js/genro_widgets.js` and `gnrjs/gnr_d11/js/genro_widgets.js`
- Affected branches: `develop` (and `master`, since the issue predates recent refactors — first introduced in commit `7683c9123`)
- Server-side logic in `gnrpy/gnr/web/gnrwebpage_proxy/apphandler/db_select.py:195` is correct; only the client-side wiring needs to change.

## Proposed Fix

Include `notnull` in the `objectExtract` allow-list and only fall back to `validate_notnull` when an explicit `notnull` was not provided:

```javascript
var resolverAttrs = objectExtract(attributes, 'method,columns,limit,auxColumns,hiddenColumns,alternatePkey,rowcaption,order_by,preferred,invalidItemCondition,subtable,notnull');
if (resolverAttrs['notnull'] === undefined) {
    resolverAttrs['notnull'] = attributes['validate_notnull'];
}
```

This preserves the existing behavior when only `validate_notnull` is set, and lets `notnull=True` work standalone as documented.

dgpaci:
## Summary

- Fixes #883: the `notnull` parameter on `dbSelect` was being silently overwritten by `validate_notnull` in `DynamicBaseCombo.creating()`, so the documented behavior (suppress the empty option row without making the field mandatory) was unreachable.
- Includes `notnull` in the `objectExtract` allow-list and only falls back to `validate_notnull` when no explicit `notnull` was provided.
- Applied symmetrically to `gnr_d11` and `gnr_d20`.

## Background

The server-side handler in `gnrpy/gnr/web/gnrwebpage_proxy/apphandler/db_select.py` has always documented `notnull` as a flag that suppresses the empty-label row, distinct from `validate_notnull` (which makes the field mandatory). The server logic at `db_select.py:195` (`if not notnull and not _id:`) is correct — but the client never delivered the flag.

In `DynamicBaseCombo.creating()` the resolver attributes were built like this:

```javascript
var resolverAttrs = objectExtract(attributes, 'method,columns,limit,...,subtable');
resolverAttrs['notnull'] = attributes['validate_notnull'];
```

`notnull` was not in the allow-list, and the next line unconditionally overwrote it. The two flags were de-facto coupled: the only way to remove the empty row was to also force the field to be mandatory.

## Test plan

- [x] Manually verified on a `dbSelect` with `notnull=True` and **no** `validate_notnull`: the empty row no longer appears and the field is **not** mandatory.
- [ ] Verify a `dbSelect` with only `validate_notnull=True` (no `notnull`): empty row removed **and** field mandatory — existing behavior preserved.
- [ ] Verify a `dbSelect` with neither flag: empty row still present (default behavior preserved).
- [ ] Hard reload the browser to bust the JS cache before testing.

## Notes

- No backward-compatibility impact: callers that only used `validate_notnull` see no change.
- Release notes updated under `Unreleased / Bugfixes`.

### DIFF
--- a/gnrjs/gnr_d11/js/genro_widgets.js
+++ b/gnrjs/gnr_d11/js/genro_widgets.js
@@ -4772,8 +4772,10 @@ dojo.declare("gnr.widgets.DynamicBaseCombo", gnr.widgets.BaseCombo, {
             attributes.hasDownArrow = false;
             attributes['tabindex'] = -1;
         }
-        var resolverAttrs = objectExtract(attributes, 'method,columns,limit,auxColumns,hiddenColumns,alternatePkey,rowcaption,order_by,preferred,invalidItemCondition,subtable');
-        resolverAttrs['notnull'] = attributes['validate_notnull'];
+        var resolverAttrs = objectExtract(attributes, 'method,columns,limit,auxColumns,hiddenColumns,alternatePkey,rowcaption,order_by,preferred,invalidItemCondition,subtable,notnull');
+        if (resolverAttrs['notnull'] === undefined) {
+            resolverAttrs['notnull'] = attributes['validate_notnull'];
+        }
         savedAttrs['auxColumns'] = resolverAttrs['auxColumns'];
         var selectedColumns = objectExtract(attributes, 'selected_*');
         var selectedCb = objectPop(attributes,'selectedCb');
--- a/gnrjs/gnr_d20/js/genro_widgets.js
+++ b/gnrjs/gnr_d20/js/genro_widgets.js
@@ -4758,8 +4758,10 @@ dojo.declare("gnr.widgets.DynamicBaseCombo", gnr.widgets.BaseCombo, {
             attributes.hasDownArrow = false;
             attributes['tabindex'] = -1;
         }
-        var resolverAttrs = objectExtract(attributes, 'method,columns,limit,auxColumns,hiddenColumns,alternatePkey,rowcaption,order_by,preferred,invalidItemCondition,subtable');
-        resolverAttrs['notnull'] = attributes['validate_notnull'];
+        var resolverAttrs = objectExtract(attributes, 'method,columns,limit,auxColumns,hiddenColumns,alternatePkey,rowcaption,order_by,preferred,invalidItemCondition,subtable,notnull');
+        if (resolverAttrs['notnull'] === undefined) {
+            resolverAttrs['notnull'] = attributes['validate_notnull'];
+        }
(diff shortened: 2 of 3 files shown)


fporcari (approved):
Fix correttto e minimale: aggiunge `notnull` all'allow-list di `objectExtract` e mantiene il fallback su `validate_notnull` solo quando non è esplicitamente impostato. Backward-compatibile, simmetrico tra gnr_d11 e gnr_d20, server-side già corretto. LGTM.

PR #886, opened 2026-05-06
feat(sourcerer): rpc_query endpoint (alpha, opt-in)

genro:
## Summary

Adds an `rpc_query` method on `/sys/ep_sourcerer` that exposes `getSelection`
to authenticated Sourcerer callers, plus a new `query_enabled` flag on the
`sourcerer` Service to gate it.

- Pass-through of `getSelection` kwargs (table, columns, where, condition, order_by, limit, offset, group_by, having, distinct, sqlparams, pkeys, subtable, countOnly + kwargs forwarded to `query()`)
- `recordResolver=False` forced server-side (lazy resolvers don't make sense outside the SPA)
- Silent cap at 1000 rows with `info.more=true` flag; `countOnly` bypasses the cap
- Postgres-only `SET LOCAL statement_timeout = 3000` (no-op on other adapters)
- Uniform JSON envelope `{ok, ticket_code, info, data}` for both success and error
- `query_enabled` flag default OFF, exposed as a checkbox in the service parameters form. When OFF the endpoint replies with `ok=false, info.error="endpoint_disabled"`.

Re-uses the existing `SourcererBearerVerifier` for auth.

## Why this is alpha and not a regression

The whole Sourcerer connection is alpha and only meaningful on test sites with test customers. **Default-off** means every existing site behaves exactly as before — `query_enabled` has to be explicitly turned on in the service parameters panel for the endpoint to do anything. Sites that never configure or enable the Sourcerer service are not affected at all.

Once we have field experience we can layer on rate limiting, audit log, table whitelisting and similar hardening.

## Files changed

- [projects/gnrcore/packages/sys/resources/services/sourcerer/sourcerer.py](../blob/feature/sourcerer-rpc-query/projects/gnrcore/packages/sys/resources/services/sourcerer/sourcerer.py) — `query_enabled` field + `is_query_enabled()` + checkbox in service form
- [projects/gnrcore/packages/sys/webpages/ep_sourcerer.py](../blob/feature/sourcerer-rpc-query/projects/gnrcore/packages/sys/webpages/ep_sourcerer.py) — new `rpc_query` next to existing `rpc_health`

## Response shape

```json
{
  "ok": true,
  "ticket_code": "<echoed from input>",
  "info": {
    "endpoint": "rpc_query",
    "error": null,
    "limit_applied": 1000,
    "totalrows": 1000,
    "more": true,
    "servertime_ms": 5
  },
  "data": [{ "id": "...", "_pkey": "...", "...": "..." }]
}
```

On error (`endpoint_disabled`, SQL error, getSelection exception): `ok=false`, `data=[]`, `info.error` set.

## Test plan

Smoke tests done locally against `test_invoice` (SQLite, 3200 rows in `invc.customer`):

- [x] flag OFF → `endpoint_disabled`
- [x] flag ON, basic query → 5 flat rows, `_pkey` matches direct `query().fetch()`
- [x] cap silently applied (`limit=5000` → `data` 1000 rows, `info.more=true`, `info.limit_applied=1000`)
- [x] `countOnly=true` → `data=[]`, `info.totalrows=3200` (cap skipped)
- [x] SQL error → `ok=false`, `info.error` carries the message
- [x] flake8 clean on both touched files
- [ ] HTTP-level Bearer auth (verifier) — needs running server, deferred to field test
- [ ] Postgres `statement_timeout` — needs PG instance, deferred to field test (SQLite skips silently as designed)
- [ ] JSON serialization for non-trivial column types (Decimal, datetime) — `invc.customer` has none; to be checked on richer tables

## Known limitations (intentional, deferred)

- No rate limiting
- No audit log on `sys.externalcall_log`
- No table whitelisting
- Timeout effective only on Postgres

### DIFF
--- a/projects/gnrcore/packages/sys/webpages/ep_sourcerer.py
+++ b/projects/gnrcore/packages/sys/webpages/ep_sourcerer.py
@@ -5,6 +5,9 @@ from gnr.core.gnrbag import Bag
 from gnr.web.verifiers import AuthorizationBearerVerifier
 
 
+MAX_ROWS = 1000
+
+
 class SourcererBearerVerifier(AuthorizationBearerVerifier):
     def get_token_to_verify(self):
         service = self.page.getService('sourcerer')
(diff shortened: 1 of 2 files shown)


genro:
**Test suite status note**

Running the full test suite locally returns `1695 passed, 22 failed, 67 skipped, 4 errors`. The 22 failures + 4 errors are **pre-existing** and unrelated to this change:

- `gnrpy/tests/core/gnrdate_test.py` — multiple `KeyError: 'february'` / `KeyError: 'from'`, locale-dependent month/period parsing
- `gnrpy/tests/core/gnrlocale_test.py` — number/currency formatting tests sensitive to system locale
- `gnrpy/tests/sql/test_compiler_coverage.py` — two `TestExpandPeriod` cases
- `gnrpy/test_xmltrasform.py` — collection errors

All failing modules are completely unrelated to `gnrcore/sys/{webpages,resources/services}` (the only paths this PR touches). `git diff origin/develop HEAD --name-only` confirms only the two Sourcerer files are modified.

The five smoke checks in the test plan all pass against `test_invoice` (real SQLite DB, 3200 rows in `invc.customer`).

fporcari (commented):
approved

projects/gnrcore/packages/sys/webpages/ep_sourcerer.py
```diff
@@ -18,3 +21,54 @@ class GnrCustomWebPage(object):
...
+            info['limit_applied'] = kwargs['limit']
+
+        kwargs.pop('recordResolver', None)
+
+        try:
+            adapter_name = type(self.db.adapter).__name__.lower()
+            if 'postgres' in adapter_name or 'pg' in adapter_name:
+                self.db.execute("SET LOCAL statement_timeout = 3000")
+        except Exception:
+            pass
```
cgabriel, line 53: This exception shouldn't be silenced, if the execute fails at least a warning log should be issued.

cgabriel (approved):
Left minor comment.

fporcari (approved):
approved

genro:
@cgabriel grazie del rilievo, hai ragione: `except Exception: pass` non va bene così com'è — almeno un warning log. Lo affronto in PR di follow-up per non invalidare le review già concesse e tenere mergeable la #886. Ho marcato il thread come resolved.

PR #888, opened 2026-05-06
feat(#881): default sans-serif font-family for HTML/PDF prints

### LINKED ISSUE #881: [FEATURE] Default sans-serif font-family fallback in print resources
## Summary

Apply a baseline `font-family: sans-serif` fallback in HTML/print resources (`TableScriptToHtml`, `RecordToHtml`) so that prints always render with a sensible default font when no `font_family` is explicitly set.

## Problem or Need

Today in `gnrpy/gnr/web/gnrtablescript.py` the class attribute `font_family = None` is the default, and the body-level `font-family` CSS rule is emitted only when the attribute is set:

```python
if self.font_family:
    self.builder.font_family = self.font_family
    self.body.style('body {{ font-family: {f}; }}'.format(f=self.font_family))
```

When a project does not set `font_family` and does not declare an explicit `font-family` in its own styles, the rendered HTML/PDF inherits whatever default the rendering engine picks — typically a serif font (Times). This forces every print resource to either:

- hardcode a font stack (e.g. `"Lucida Grande", Lucida, Verdana, sans-serif`) just to avoid Times, or
- live with a serif default that nobody asked for.

We just hit this in `genromed` (softwellsrl/genromed#34), where removing hardcoded `"Lucida Grande"` font stacks from several print resources caused the prints to fall back to Times. The right fix is at the framework level, not in every project.

## Proposed Solution

Have the print base class always emit a baseline `body { font-family: sans-serif; }` rule (or another generic family agreed upon, e.g. `Helvetica, Arial, sans-serif`) regardless of whether `font_family` is set. When `font_family` is set, it overrides the baseline as it does today.

Concretely, in `TableScriptToHtml.mainLayout` (and the equivalent path in `RecordToHtml`), unconditionally include a default `font-family` in the body style block, then let the existing `if self.font_family:` branch override it.

This way:
- prints with no explicit font get a clean, neutral sans-serif rendering by default;
- projects that set `font_family` keep working unchanged;
- project-level resources can drop hardcoded font stacks without regressions.

## Considered Alternatives

- Setting `font_family` to a default value like `'sans-serif'` instead of `None`. Rejected because `font_family` is also used by `self.builder.font_family` for exact row-height calculation, which expects a *mapped* font name (per the comment at line 127). A pure CSS fallback in the body style is decoupled from that machinery.
- Asking each project to set `font_family` explicitly. Rejected: it pushes a framework-level concern onto every consumer and is exactly what causes the kind of drift fixed in genromed#34.

## Impact and Scope

- **Affected files**: `gnrpy/gnr/web/gnrtablescript.py` (`TableScriptToHtml`, `RecordToHtml`).
- **Beneficiaries**: any project using `TableScriptToHtml` / `RecordToHtml` for HTML/PDF prints.
- **Backward compatibility**: prints that already declare `font_family` or that override the body `font-family` in their own styles are unaffected (CSS specificity / later rules win). Prints that today rely on the implicit serif default would visibly change to sans-serif — this is the intended improvement and should be called out in `RELEASE_NOTES.rst`.

## Additional Information

Reference: softwellsrl/genromed#34 — removed `"Lucida Grande", Lucida, Verdana, sans-serif` font stacks from 6 print resources; this issue addresses the framework-level default that should make those local declarations unnecessary.

dgpaci:
## Summary

- Emit a baseline `body { font-family: sans-serif; }` rule from `BagToHtml.defineStandardStyles` (covering `TableScriptToHtml`, `TableTemplateToHtml` and any other `BagToHtml` subclass) and from the `RecordToHtmlPage` / `RecordToHtml` hierarchy in `gnrtablescript.py`.
- Reorder `TableScriptToHtml.defineStandardStyles` to call `super()` first and emit the explicit `font_family` override last, so the override still wins by CSS source order over the new baseline.
- Document the change under "Unreleased / Enhancements" in `RELEASE_NOTES.rst`, including the visible behavior change (Times → sans-serif fallback).

Fixes #881.

## Rationale

Previously, prints with no explicit `font_family` and no project-level body `font-family` rule fell back to the rendering engine's default (typically Times). Each project had to either hardcode a font stack in every print resource or live with the serif default. The fix moves the baseline to the framework so consumers can drop their hardcoded stacks without regression.

## Backward compatibility

- Prints that already declare `font_family` on the print class are unaffected — the override is now emitted *after* the baseline and wins by CSS source order.
- Prints that override the body `font-family` in their own styles are unaffected.
- Prints that previously relied on the implicit serif fallback will now render in sans-serif. Projects that want the old serif rendering can declare it explicitly in their own styles.

## Test plan

- [ ] Generate a print from a `TableScriptToHtml` subclass with no `font_family` and no project-level body font rule — verify it renders in sans-serif (was Times).
- [ ] Generate a print from a `TableScriptToHtml` subclass with `font_family = 'Helvetica'` — verify the explicit font is applied (override wins).
- [ ] Generate a print from a project resource that already declares `body { font-family: ... }` in its own styles — verify project rule wins.
- [ ] Generate a print from a `RecordToHtmlPage` / `RecordToHtml` resource — verify same baseline behavior.


cgabriel (dismissed):
Looking good to me, need to test it out in a container context with limited fonts to verify if the override is fine.

genro (dismissed):
Approving. Diagnosis is correct and the fix lives in the right place: the baseline on BagToHtml covers the whole print hierarchy without touching font_family (which stays reserved for the builder's row-height calculation). Calling super() first so the explicit override wins by CSS source order is the cleanest choice.

One non-blocking note: using the bare sans-serif generic family maximizes portability, but in containers with a limited font set the rendering depends on whichever sans-serif the engine picks. Worth a smoke test in a real CI/print container, as already noted in the prior review.

genro:
@dgpaci can you bring develop into this branch and resolve the conflict? `mergeStateStatus` is now `DIRTY` — likely a collision on `RELEASE_NOTES.rst` after other PRs landed since 2026-05-08.

Per team convention, please use `git merge origin/develop` (not rebase, since the PR is already under review). After the merge commit lands and CI is green again, I'll merge.

Two approvals already in place (cgabriel + genro), CI was green on the last push, no other blockers.

dgpaci:
Per internal policy, merging `develop` into a feature branch is normally a maintainer's responsibility. Since I performed the merge myself to resolve the conflict in `RELEASE_NOTES.rst` and unblock the PR, the previous approvals were auto-dismissed by GitHub. A fresh review is required before merge.

Conflict resolution: kept both `Unreleased` sections — `New features` (#874, from develop) followed by `Enhancements` (#881, this PR).

cgabriel (approved):
LGTM, as per previous review.

PR #892, opened 2026-05-08
feat(no-mako): convert scheduler dashboard + warn on Mako fallback

genro:
## Summary

Two related steps in the no-mako roadmap.

## 1. Scheduler dashboard → PageBuilder

`GnrTaskScheduler.dashboard` rendered a ~100-line inline Mako template. Replace with `PageBuilder` (introduced in #862). Output is functionally equivalent — same Bootstrap stylesheet, same auto-reload script, same sections and tables. The only visible difference is XHTML strict output (consistent with the templates migrated in #862).

This conversion is unconditional (no preference branch): the dashboard endpoint is standalone, simple, and there is no reason to keep a Mako fallback for it alone.

## 2. Warn on Mako fallback when `no_mako=True`

When `experimental.no_mako` is on but a struct `<name>.py` template is missing, the rootPage flow (`gnrwebpage.py`) and the Mako plugin (`gnrwebpage_plugin/mako.py`) silently fall back to the legacy Mako `.tpl`. This hides incomplete migrations.

Emit a `logger.warning` at each fallback point naming the missing struct template and the Mako template that took over. An instance running with `no_mako=True` and zero such warnings in its logs is a safe candidate for dropping the Mako fallback entirely.

## Roadmap context

This is the first step toward removing Mako:

1. **Now** (this PR): every path can run with or without Mako depending on `experimental.no_mako` (default off). Warning surfaces any gap.
2. **In ~2 weeks**: flip `experimental.no_mako` default to `True`, fallback still present.
3. **In ~4 weeks**: drop the switch, the Mako plugin, all `.tpl` files, and remove `mako` from `pyproject.toml`.

## Test plan

- [x] `pytest gnrpy/tests/web/gnrtask_test.py` — 28/28 pass
- [x] Smoke test of `_render_dashboard()` with mock payload: produces valid XHTML with all expected sections
- [ ] Manual end-to-end: open the scheduler dashboard URL in a browser, verify visual parity with current Mako output
- [ ] Manual end-to-end: run an instance with `experimental.no_mako=True`, confirm no warnings in logs (= all standard templates have struct equivalents)

### DIFF
--- a/gnrpy/gnr/web/gnrtask_new.py
+++ b/gnrpy/gnr/web/gnrtask_new.py
@@ -33,7 +33,7 @@ from datetime import datetime, timezone
 from collections import defaultdict
 from typing import Any
 
-from mako.template import Template
+from gnr.web.gnrwebpage_proxy.frontend.basepagetemplate import PageBuilder
 import aiohttp
 import asyncio
 from aiohttp import web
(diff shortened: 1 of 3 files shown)


gnrpy/gnr/web/gnrtask_new.py
```diff
@@ -451,119 +451,103 @@ async def schedule_single_execution(self, task_def):
...
         )
 
+    def _render_dashboard(self, payload):
+        builder = PageBuilder()
+        head = builder.head
+        body = builder.body
+
+        head.child('meta', charset='utf-8')
+        head.child('meta', name='viewport',
+                   content='width=device-width, initial-scale=1')
```
cgabriel: This is wrongly rendered as

```
		<meta name="viewport">width=device-width, initial-scale=1</meta>
```

cgabriel (changes requested):
See code annotation, the template is rendered wrongly.

gnrpy/gnr/web/gnrtask_new.py
genro: Fixed in 58c667d. The bug was `content=` (text child) vs `_content=` (HTML attribute) in `PageBuilder`: the viewport line was passing the string as text, producing `<meta name="viewport">width=...</meta>`. Now uses `_content=` matching the convention in `standard.py:19+`. Verified locally: full dashboard renders as well-formed XHTML, all `<meta>` self-closed with proper `content="..."` attributes.

cgabriel (approved):

PR #894, opened 2026-05-09
Introduce neutral AsyncSession core (depends on #852)

genro:
## Summary

Reintroduces, in **modern and neutral form**, the full-duplex async-session
pattern that lived inside the legacy `DebugSession` class (removed in #852
via commit `2bd3fd1f46`).

The pattern was good infrastructure but tailored to the PDB use case.
Removing the legacy debugger without preserving the pattern would have
meant rewriting it identically at the first future use. Reintroducing it
here, neutral and well-tested, locks down the API ahead of real consumers.
The first expected one is `gnrorchestrator` (a future conversational
orchestration component).

## What it adds

A single module `gnrpy/gnr/web/gnrasync_session.py` exposing two public
classes:

- **`AsyncSession[T, U]`** — a full-duplex session bridging a server-side
  peer with a browser-side WebSocket client. Three semantic operations:
    - `notify(payload)` — fire-and-forget event peer → client
    - `request(payload, timeout=...)` — peer asks client and awaits the
      correlated response (`_request_id` under the hood)
    - `events()` — async iterator of client → peer messages that are not
      responses to a pending request
- **`AsyncSessionManager`** — registry of sessions with lifecycle managed
  by an async context manager (`async with manager.session(key, ...) as session:`).
  Per-page lookup and bulk close (`close_for_page`) are first-class.

## Design choices

- **Python 3.11+ idioms**: `asyncio.TaskGroup` for structured cancellation,
  `asyncio.timeout()`, union types `X | Y`, `@dataclass`,
  `asynccontextmanager`.
- **JSON framing by default**, with pluggable `encode_to_client` /
  `decode_from_client` hooks for consumers that need a different wire
  format (binary, msgpack, …).
- **Correlation hidden from the consumer**: the payload type `T` is
  opaque; `request()` injects `_request_id` in the wrapper and the client
  echoes `_response_to`. The consumer only sees the semantic API.
- **No global registry, no leaks**: a session exists only within the
  async-context-manager scope; the manager unregisters on exit.
  Structurally fixes the `debug_queues` leak the legacy `DebugSession`
  had.
- **Immediate wakeup of `events()` on close** via an internal sentinel —
  no busy polling.

## What it does not add (on purpose)

- No consumer. No listener (Unix socket, TCP, …). No changes to any
  existing framework file.
- The scaffold stands on its own, is covered by pure unit tests
  (no I/O, no framework mocks), and waits for its first user.

## Test plan

- [x] flake8: zero issues on added files
- [x] 11/11 unit tests in `gnrpy/tests/web/gnrasync_session_test.py` green:
      lifecycle via context manager, notify, request/response correlation,
      request timeout, events iterator, event/response separation,
      multi-session isolation, close_for_page, custom framing, error
      propagation through TaskGroup, immediate wakeup of `events()` on
      close
- [x] asyncio regression suite (`gnrasync_test.py`): 6/6 green
- [x] full `gnrpy/` test suite: 1711 passed (22 pre-existing failures on
      `feature/asyncio-migration`, in unrelated modules: `gnrlocale`,
      `gnrdate`, `test_compiler_coverage`)
- [x] no dirty imports, no logic in CLI modules

## Dependency

This PR builds on commit `2bd3fd1f46` of #852 (legacy debugger removal).
**Merge #852 first, then this one.** Back to the concept:

> #852: removes the legacy PDB debugger and completes the asyncio migration
> this PR: reintroduces the good part (async session pattern) in modern
>          form, ready for gnrorchestrator and future use cases

## Review window

cgabriel will be away from May 15 to June 6. **Please start the review
before May 15** so that both PRs (#852 and this one) can land together
without a month-long gap.

### DIFF
--- a/projects/gnrcore/packages/sys/lib/sourcerer_endpoint.py
+++ b/projects/gnrcore/packages/sys/lib/sourcerer_endpoint.py
@@ -126,104 +126,58 @@ def _apply_statement_timeout(db, timeout_ms):
         pass
 
 
-def run_query(db, table=None, count_only=False,
+def run_query(site, table=None,
               max_rows=DEFAULT_MAX_ROWS,
               statement_timeout_ms=DEFAULT_STATEMENT_TIMEOUT_MS,
-              app=None, **kwargs):
-    """Execute a query against a GenroPy DB, returning the same shape that
-    rpc_query produces over HTTP.
-
-    Two execution paths:
-        - If ``app`` is provided, uses ``app.getSelection`` (web context: full
-          virtual_column resolution, alias expansion, etc.).
-        - Otherwise falls back to ``db.table(table).query(**kwargs).fetch()``
-          for in-process callers without a web app.
+              **kwargs):
+    """Execute a getSelection against a GenroPy site.
 
     Args:
-        db: GenroPy db connection.
-        table: Fully qualified table name (e.g. ``'fatt.fattura'``).
-        count_only: If True, return only the row count.
-        max_rows: Cap applied when ``limit`` is missing or higher.
-        statement_timeout_ms: PostgreSQL statement_timeout cap (best effort).
-        app: Optional GnrApp / web app proxy with ``getSelection``.
-        **kwargs: Forwarded to getSelection / tbl.query (columns, where,
-            distinct, group_by, having, order_by, limit, etc.).
+        site: GnrWsgiSite (provides ``site.db`` and ``site.dummyPage.app``).
+        table: fully qualified table name (e.g. ``'fatt.fattura'``).
+        max_rows: cap applied when ``limit`` is missing or higher.
+        statement_timeout_ms: PostgreSQL statement_timeout cap.
+        **kwargs: forwarded to ``app.getSelection`` (columns, where, condition,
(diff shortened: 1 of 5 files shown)


projects/gnrcore/packages/sys/lib/sourcerer_endpoint.py
cgabriel, line 1: HOw is sourcerer endpoint related to asyncsession core?

projects/gnrcore/packages/sys/webpages/ep_sourcerer.py
cgabriel, line 1: Also here, how is all this related to the PR scope?

cgabriel (changes requested):
I think you got some problems with the PR: any of the files described in the PR are actually in the diff, looks like the PR is made starting from a branch, but then a different branch is associated to it. Please check and advise.

genro:
Closing this PR — operator error on my side: the head branch was set to 'feature/sourecer-enancing' instead of the intended 'feature/async-session-core'. The actual AsyncSession commit is on origin at 'feature/async-session-core' (commit f190d5c83e). Reopening a fresh PR with the correct head. Apologies for the noise, @cgabriel.

PR #904, opened 2026-05-15
feat(gnrwebstruct): registry-driven widget namespace + W3C HTML5 catalog

genro:
**Draft — depends on #903 (split) which depends on #902 (tests).**
After both upstream PRs merge, this branch will be rebased on `develop`
and marked ready for review.

## Summary

Third of three PRs from #901. Two intertwined changes:

1. **Registry-driven dispatch**: the four hardcoded namespace lists
   inside `GnrDomSrc_dojo_11` (`htmlNS`, `dijitNS`, `dojoxNS`, `gnrNS`)
   are replaced by introspection of `AllWidgets._widget_names`, a
   composed class that gathers declarations from four dialect mixins
   in `gnr/web/widgets/`. Adding/removing widgets now happens by
   editing those mixins, not by mutating list literals buried in a
   3000-line module.
2. **W3C HTML5 catalog**: `widgets/html.py` is ported from
   `genro-builders/contrib/html/html5_elements.py` (112 elements,
   schema-derived `sub_tags` metadata). 25 modern semantic tags become
   available framework-wide: `article`, `section`, `nav`, `header`,
   `footer`, `main`, `aside`, `dialog`, `details`, `summary`,
   `picture`, `source`, `track`, `figure`, `figcaption`, `mark`,
   `time`, `data`, `bdi`, `s`, `u`, `wbr`, `ruby`, `rp`, `rt`, `slot`,
   `template`, `form`, `output`, `datalist`, `meter`, `progress`,
   `search`, `hgroup`.

## Commit-by-commit

| Commit | Purpose |
|---|---|
| 1 — infrastructure | `@element` decorator, `WidgetMixinBase`, `AllWidgets`, 24 class-level tests |
| 2 — wiring | `genroNameSpace = AllWidgets._widget_names` |
| 3 — restore | 7 layout containers + `TracebackViewer` + `del_` keyword alias |
| 4 — HTML5 port | replace `widgets/html.py` with the W3C catalog, move GenroPy layout primitives to `genro.py` |

## Cardinality

| Mixin | Pre | Post | Delta |
|---|---:|---:|---:|
| HtmlWidgets | 87 (curated) | 112 (W3C HTML5) | +25 |
| DijitWidgets | 35 | 42 | +7 |
| DojoxWidgets | 31 | 31 | 0 |
| GenroWidgets | 97 | 102 | +5 |
| **AllWidgets** | 256 | **282** | **+26** |

## Cross-dialect collisions

Five widget names appear in two dialects each. The MRO order
[...]

PR #903, opened 2026-05-14
refactor(gnrwebstruct): split monolithic module into package

genro:
**Draft — depends on #902 merging first.** After that, this branch will
be rebased on `develop` and marked ready for review.

## Summary

Second of three PRs from #901. Promotes
`gnrpy/gnr/web/gnrwebstruct.py` (3186 lines) into a package
`gnrpy/gnr/web/gnrwebstruct/`, mirroring the prior `gnr.sql.gnrsql`
split.

Pure refactor — zero behavioural change.

## Package layout

| Module           | Lines | Content |
|------------------|------:|---------|
| `__init__.py`    |    50 | public re-exports only |
| `_helpers.py`    |   170 | `cellFromField`, `_selected_defaultFrom`, `externalStore` |
| `base.py`        |  1217 | `GnrDomSrc` + `StructMethodError` + `struct_method` + `GnrDomSrcError` + `GnrDomElem` |
| `dojo11.py`      |  1112 | `GnrDomSrc_dojo_11` |
| `dojo20.py`      |    13 | `GnrDomSrc_dojo_20` stub (kept separate on purpose — marks the d11/d20 alignment boundary) |
| `formbuilder.py` |   404 | `GnrFormBuilder` |
| `gridstruct.py`  |   380 | `GnrGridStruct` |

`base.py` is intentionally kept at ~1100 lines: PR #3 will trim it by
replacing the trivial HTML widget methods (`h1`-`h6`, `div`, `span`,
...) with namespace-driven dispatch via the new `@widget` decorator.

## Compatibility

`pyproject.toml` updated under `[tool.setuptools].packages` to register
`gnr.web.gnrwebstruct` (parallel to the existing `gnr.sql.gnrsql`
entry).

Every previously exported symbol stays importable as
`from gnr.web.gnrwebstruct import X`. The 17 characterization tests
added in #902 pin this contract — they pass identically before and
after the split.

## Dropped

The trailing `if __name__ == '__main__':` demo block (lines 3170-3186
of the old file). It referenced a hard-coded `/usr/local/genro/...`
path that has not existed for years, and a package directory cannot be
executed as a script anyway.

## Test plan

- [x] `pytest gnrpy/tests/web/gnrwebstruct_test.py` → 17/17 pass
- [x] `pytest gnrpy/tests/` → 1742 pass / 22 fail (pre-existing, identical to pre-split)
- [x] `flake8 gnrpy/gnr/web/gnrwebstruct/` → zero errors
[...]

Issue #901, opened 2026-05-14
refactor(gnrwebstruct): three-PR roadmap to modularize gnrwebstruct and align widgets to W3C HTML5


genro:
## Background

`gnrpy/gnr/web/gnrwebstruct.py` is a 3186-line monolith hosting:

- `GnrDomSrc` / `GnrDomSrc_dojo_11` / `GnrDomSrc_dojo_20` — DOM source builders
- `GnrFormBuilder` — layout-tabular form builder
- `GnrGridStruct` — grid struct builder
- `@struct_method` registry — extension hook
- 4 hardcoded namespace lists (`htmlNS`, `dijitNS`, `dojoxNS`, `gnrNS`) consumed
  by `__getattr__` to dispatch widget calls to `child(tag)`

The 4 lists drive widget resolution but live deep inside a 3000-line file.
The companion package `gnrpy/gnr/web/widgets/` already declares mixins for
the same widgets (intended as IDE/typing documentation) but is **not
imported anywhere at runtime** — divergence is structural.

## Goal

Three sequential PRs:

### PR 1 — test characterization (this issue starts here)
Branch: `chore/gnrwebstruct_tests`

Add focused tests that pin current behaviour of `gnrwebstruct`:

- `__getattr__` six-fallback ladder
- `genroNameSpace` content snapshot
- `child()` primitive
- `@struct_method` extension hook (already covered, keep as-is)
- smoke instantiation of `GnrDomSrc_dojo_11`

Acceptance: snapshot of `genroNameSpace` (count + sample keys) frozen, so
later refactors can verify bit-perfect equivalence.

### PR 2 — promote module to package
Branch: `feat/gnrwebstruct_split`

Pure move: `gnrwebstruct.py` → `gnrwebstruct/` package, modelled on
`gnrpy/gnr/sql/gnrsql/`. Split into focused modules (`base`, `dojo11`,
`dojo20`, `formbuilder`, `gridstruct`, `registry`, `exceptions`).
`__init__.py` re-exports all public symbols. Zero behavioural change —
validated by PR 1 tests.

### PR 3 — lightweight `@widget` decorator + registry-driven namespace
Branch: `feat/gnrwebstruct_widgets`

- Introduce a minimal `@widget` decorator in `widgets/__init__.py`,
  signature-compatible with `genro-builders`' `@element`
- Mixins (`widgets/dijit.py`, `widgets/dojox.py`, `widgets/genro.py`,
[...]

PR #906, opened 2026-05-15
feat(gnrwebstruct): registry-driven widget namespace + W3C HTML5 catalog

genro:
Third of three PRs from #901, replacement of the now-closed #904.
Cherry-picked onto develop after #903 was merged, so the diff is
exactly the four widget commits with no leftover noise from the
split that #903 already delivered.

## Summary

Two intertwined changes:

1. **Registry-driven dispatch**: the four hardcoded namespace lists
   in `GnrDomSrc_dojo_11` (`htmlNS`, `dijitNS`, `dojoxNS`, `gnrNS`)
   are replaced by introspection of `AllWidgets._widget_names`, a
   composed class that gathers declarations from four dialect mixins
   in `gnr/web/widgets/`. Adding/removing widgets now happens by
   editing those mixins.
2. **W3C HTML5 catalog**: `widgets/html.py` is ported from
   `genro-builders/contrib/html/html5_elements.py` (112 elements,
   schema-derived `sub_tags` metadata). 25 modern semantic tags
   become available framework-wide.

## Commit-by-commit

| Commit | Purpose |
|---|---|
| 1 — infrastructure | `@element` decorator, `WidgetMixinBase`, `AllWidgets`, 24 class-level tests |
| 2 — wiring | `genroNameSpace = AllWidgets._widget_names` |
| 3 — restore | 7 layout containers + `TracebackViewer` + `del_` keyword alias |
| 4 — HTML5 port | replace `widgets/html.py` with the W3C catalog, move GenroPy layout primitives to `genro.py` |

## Cardinality

| Mixin | Pre #903 | Post | Delta |
|---|---:|---:|---:|
| HtmlWidgets | 87 (curated) | 112 (W3C HTML5) | +25 |
| DijitWidgets | 35 | 42 | +7 |
| DojoxWidgets | 31 | 31 | 0 |
| GenroWidgets | 97 | 102 | +5 |
| **AllWidgets** | 256 | **282** | **+26** |

## Cross-dialect collisions

Five widget names appear in two dialects each. The MRO
`(Genro, Dojox, Dijit, Html)` resolves them, preserving historical
behaviour:

- `button`, `dialog`, `menu`, `textarea` -> Dijit wins
- `script` -> Genro wins

## Behavioural notes

- Dispatch contract unchanged: `__getattr__` still looks up the
  lowercased name in `genroNameSpace` and builds a `GnrDomElem`.
- Wire-format: some tag values now lowercase (`borderContainer`
  instead of `BorderContainer`, etc.) because mixin method names
  follow PEP 8. The client-side handler registry lowercases tags
  before lookup (`genro_wdg.js:469`), so no behavioural impact.
- New: `pane.del_(...)` produces `<del>`. Previously the element
  was reachable only via `pane.child('del')`.
- HTML4 deprecated tags removed by the W3C port (zero usage):
  `acronym`, `big`, `tt`, `frame`, `frameset`, `noframes`.

## Addressing #903 review notes (non-blocking)

@fporcari left two follow-up notes on #903:

1. The 17 characterization tests pin dispatch/registration but
   never call `.toXml()` / `.source()` / build a real form or grid.
   "PR #3 will introduce real behavioural changes and will need
   rendering-level coverage to be safe."
   -> Not yet addressed in this PR. Should be added before merge.
2. Rename `externalStore` -> `_externalStore` for consistency with
   `_selected_defaultFrom`.
   -> Out of scope for this PR; deferred.

## Out of scope

- Trimming the 16 trivial HTML wrapper methods in `base.py`
  (`h1`-`h6`, `div`, `span`, `td`, `th`, `li`, `pre`, `style`,
  `details`, `summary`, `a`, `dt`, `option`, `caption`) — separate
  surgical commit.
- Direct dependency on `genro-builders.@element`. Locally
  reimplemented with isomorphic signature.
- Runtime validation of `sub_tags`. Data is stored on every HTML5
  method but no check consumes it yet.

## Test plan

- [x] `pytest gnrpy/tests/web/widgets_test.py` -> 25/25 pass
- [x] `pytest gnrpy/tests/web/gnrwebstruct_test.py` -> 17/17 pass
- [x] `pytest gnrpy/tests/` -> 1767 pass / 22 fail (pre-existing,
      identical to develop)
- [x] `flake8 gnr/web/widgets/ gnr/web/gnrwebstruct/ tests/web/` ->
      zero
- [x] `GnrDomSrc_dojo_11.genroNameSpace` has 282 entries
- [ ] **TODO before merge**: rendering-level smoke test on a real
      instance (per @fporcari's review note)

Refs #901, closes #904 (which was the same work on an out-of-date
base)


genro:
@fporcari pushed the nested `_widgets` refactor (`abe940b11`) on top of the four widget commits — the dispatch package now lives next to its dispatcher.

Could you run a rendering-level smoke test on a real instance when you have a slot? Per your review note on #903, the 42 unit tests pin dispatch and registration but never call `.toXml()` / `.source()` nor build a real form/grid. Decided with @genro to do that coverage live rather than in characterization tests — they have more signal in the browser than in pytest.

Suggested smoke surfaces:
- HTML5 semantic tags introduced by the W3C port (\`article\`, \`section\`, \`nav\`, \`mark\`, \`figure\`, etc.)
- Restored Dojo layout containers (\`tabContainer\`, \`accordionContainer\`, \`borderContainer\`)
- Cross-dialect collisions (\`button\`, \`dialog\`, \`menu\`, \`textarea\` → Dijit; \`script\` → Genro)
- Lowercase wire format for tags that previously rendered camelCase (\`borderContainer\` was already camelCase, so no change expected — worth confirming)

PR stays draft until smoke is green.

fporcari (commented):
Reviewed this in full. The design idea is right — four hardcoded namespace lists in `dojo11.py` are exactly the kind of thing that should be derived from the widget declarations rather than maintained by hand, and the mixin split makes "add a widget" a one-place edit. The cardinality claims check out precisely: 112 html / 42 dijit / 31 dojox / 102 genro, 282 total, and the five cross-dialect collisions (`button`, `dialog`, `menu`, `textarea` → dijit; `script` → genro) are the ones the body lists.

The problem is that the branch has aged past the point where it can be evaluated as-is. It is based on a commit predating #869, which is already on develop, and it is `CONFLICTING`. Three findings, and the first two would ship real breakage:

1. **`proseMirrorEditor` disappears from the namespace.** It is declared on develop (`gnrpy/gnr/web/gnrwebstruct/dojo11.py:67`) and there is no equivalent in the branch's `_widgets/genro.py`. `webpages/tools/prosemirror.py:61,70,84,96,99` and `prosemirror_playground.py:197` would fail with `AttributeError` at page build. Green CI does not see it because the test pins cardinality and lowercasing, not "no previously registered name disappeared".

2. **The 25 new HTML5 tags have no client-side handler.** `gnrjs/gnr_d11/js/genro_wdg.js:120-124` (`htmlspace`) does not list `article, aside, section, nav, header, footer, main, figure, mark, time, data, output, meter, template, slot, source, track, picture, …`. `getHandler` returns `undefined`, `genro.assert` only logs, and `create` then dereferences `handler._beforeCreation` → TypeError. So `pane.article(...)` builds server-side and dies in the browser. `gnr/core/gnrhtml.py:52` and `gnr/web/gnrhtmlpage.py:33-42` also keep their own pre-HTML5 lists.

3. **The CamelCase → lowerCamel rename of the 72 dijit tags is not behaviour-neutral server-side.** Client dispatch is case-insensitive, so that half of the claim holds, but `resources/common/th/th.py:99` and `:949` compare `tag == 'ContentPane'` to force `overflow='hidden'`. Built through `pane.contentPane(...)` the guard silently stops firing — a layout regression with no error. Aggravating: `th.py:272,290` still pass the literal `tag='ContentPane'`, so both spellings would coexist in one tree.

Also worth noting, in the "good news" column: the priority inversion the new names introduce in `base.py:170-205` (namespace lookup now precedes `_external_methods` / autoslots / subtags) has **zero** collisions inside this repo — 278 registered struct-method names, no intersection with the 34 added tags — so the blast radius is far smaller than it looks. Application repos were not checked.

My read is that this and #842 were both proofs of concept that made their point and then sat for months while the ground moved. Rather than carry them as open PRs that need a rebase nobody is going to do speculatively, I would **close both and reopen from a fresh branch when the idea is actually wanted** — the registry-driven namespace is worth doing, just not on a base that predates #869. If you would rather keep this one alive, the minimum to make it reviewable again is: rebase on develop, restore `proseMirrorEditor`, add the new tags to `htmlspace`, and add a subset assertion freezing the pre-existing widget names so a future removal cannot slip past CI.

Your call as author — say the word and I close it, or leave it open and I re-review after the rebase.

Issue #910, opened 2026-05-18
RFC: should grouplet accept a client-side handler (handlerJs=)?


genro:
## Background

Every grouplet today resolves its content through `gr_loadGrouplet` server-side:

- `handler=` is a Python `@public_method` that builds a `GnrDomSrc` tree
- `resource=` is a Python file under `resources/grouplets/` whose `grouplet_main(pane, **kwargs)` builds the same tree
- In both cases the server serializes the resulting struct to XML, the client receives it and mounts the dijits

For `groupletGrid` this is fine: one RPC at boot, then zero RPC because the template lives in client memory as a `GnrDomSource` that the controller deep-clones per row.

For the rest of the family, *every* mount is a server round-trip — even when the grouplet contents are completely static and could equally well be expressed client-side.

## The observation

The fluent builder API used by the Python side (`pane.contentPane(...)._formbuilder(cols=2)._textbox(value='^.x')`) **already has a JS twin** in `gnr.GnrDomSource._()`:

```js
sourceNode._('contentPane', {datapath: '.myform'})
          ._('formbuilder', {cols: 2})
          ._('textbox', {value: '^.email', lbl: 'Email'});
```

defined at [gnrjs/gnr_d20/js/gnrdomsource.js:2050](gnrjs/gnr_d20/js/gnrdomsource.js#L2050). Every existing widget uses it internally (`sourceNode._('div', ...)`, `sourceNode._('button', kw)`, ...). So the machinery to build a struct on the client without a Python round-trip is already there.

## The proposal (to discuss, not necessarily to ship)

Add a third resolution path on `pane.grouplet(...)`:

```python
pane.grouplet(
    value='^.contact',
    handlerJs=\"\"\"
        const fb = pane._('formbuilder', {cols: 2});
        fb._('textbox', {value: '^.email', lbl: 'Email'});
        fb._('textbox', {value: '^.phone', lbl: 'Phone'});
    \"\"\")
```

The string is `funcCreate`-d on the client into a function `(pane, kw) => { ... }` and invoked at mount time, **skipping the `kw.remote = 'gr_loadGrouplet'` round-trip**. The widget would be re-built from scratch every time the value path changes — same as the server-side equivalent — only the resolution lives client-side.

[...]

Issue #909, opened 2026-05-18
Grouplet family: surface silent failures and identifier collisions


genro:
While drafting a "For Dummies" tour of the grouplet family (an internal doc exercise across the 7 widgets — `grouplet`, `groupletChunk`, `groupletForm`, `groupletPanel`, `groupletWizard`, `groupletGrid`, Topic Grid + the three `genro.dlg.*DataEditor` variants), several **silent failure modes** kept surfacing while reading the source side by side with the existing tests. None of them produce a traceback, a log line, or a visible error in the browser — they just produce surprising behavior that costs the developer 15-30 minutes of head-scratching to diagnose.

Filing them as a single umbrella because the fix shape is the same: convert "silent surprise" into "loud signal". Each item below is independent and could be split into its own PR.

## A) Handler missing `@public_method` decorator

When `pane.grouplet(handler=self.my_method)` resolves a handler that isn't decorated with `@public_method`, the RPC roundtrip returns nothing. The pane stays empty, browser shows no error, server log shows nothing.

**Fix shape**: server-side, on `gr_loadGrouplet`, check `getattr(handler, 'is_rpc', False)` and `raise` a `GnrException` like `"grouplet handler '<name>' is not decorated with @public_method"` if not.

## B) `handler=` and `resource=` both passed — `resource` wins silently

`pane.grouplet(value='^.x', handler=self.x_handler, resource='x_template')` silently drops the handler. Today's behavior is to prefer `resource`, with no warning.

**Fix shape**: in the builder, raise `GnrException` if both are non-`None`. (Or accept both and warn loudly; the strict variant is safer.)

## C) Duplicate `name=` across `groupletChunk` / dialog editors

Two chunks with `name='edit'` collide on the page-wide `genro.dlg.memoryDataEditor` registry. The second click reuses the first dialog's `handler` and `title`, inheriting the wrong state. No error.

**Fix shape**: in `genro.dlg._groupletForm` (in `gnrjs/gnr_d20/js/genro_dlg.js`), detect a duplicate `dlgId` already in the registry and either (a) error explicitly, or (b) `console.error` with the conflicting names so the bug is visible in devtools.

## D) Duplicate `formId=` across `groupletForm`

Same shape as C, on a different registry: `genro.formById('gf_main')` returns the first registration; subsequent `groupletForm` instances with the same id silently shadow it and lose all external `load()` / `save()` traffic.

**Fix shape**: at form registration time, warn or error on collision.

[...]

PR #905, opened 2026-05-15
feat(groupletGrid): new widget for Bag-driven repeatable mini-forms

fporcari:
## Summary

Introduce **`groupletGrid`**, a genropy widget that renders a Bag as N tiles where each tile is the same reusable "grouplet" template (a mini-form). One RPC for the row template at boot, then **zero RPC** for every add / remove / edit / save — everything flows through Bag triggers.

It targets use cases that don't fit a flat `gnr.Grid`: invoice line editors, kanban boards, framed card galleries, spreadsheet-style data entry (`struct=` mode), multi-pane tab editors, and recursive master-detail forms.

## What's in

- **Python helper** `gr_groupletGrid(...)` (`resources/common/gnrcomponents/grouplet/grouplet.py`) — emits a CSS-grid container with autoslots (top/bottom/left/right), a sibling `BagStore` for change/lock/delete integration, and a `_onBuilt` dataController that boots the JS controller.

- **JS controller** (`resources/common/gnrcomponents/grouplet/grouplet_grid.js`) — five `gnr.*` classes:
  - `GroupletGridStructAdapter` — maps a Grid-style `struct` to a spreadsheet-like tile.
  - `GroupletGridDnD` — DOM-pure HTML5 drag-and-drop dispatcher shared by tile and chip paths.
  - `GroupletGridController` — lifecycle, action bus, layout switching, template loading (with `resourceField=` multi-template mode), `GridChangeManager` bridge for formulas / totalizers.
  - `GroupletGridTile` — per-pkey view with chrome (×, kebab, drag handle ⠿) and body grafted from a deep-copied template.
  - `GroupletDataStore` — thin view over the sibling `BagStore`; forwards getData / addRow / removeRow / moveRow / updateRow / deleteRowAsk / hasChanges / hasErrors / setLocked.

- **CSS** (`resources/common/gnrcomponents/grouplet/grouplet.css`) — theme tokens, layout primitives for cards / tabs / vtabs / struct modes, DnD state classes aligned with framework conventions (`draggedItem`, `canBeDropped`, `cannotBeDropped`, `body.drag_started`).

## Demo gallery

All demos live under `projects/gnrcore/packages/test/webpages/gnrwdg/test_grouplet_grid/`:

| Page | Scenario |
|---|---|
| `01_gallery.py::test_01_invoice_baseline` | invoice baseline (cards single-grid) |
| `test_02_todolist_handler` | todolist via handler callable |
| `test_03_invoice_responsive` | responsive cols + min_width |
[...]

PR #912, opened 2026-05-18
docs(grouplet): For Dummies tour of the grouplet family

genro:
## Summary

Adds a multi-tab self-contained HTML documentation site that walks through the 7 widgets in the grouplet family — `grouplet`, `groupletChunk`, `groupletForm`, `groupletPanel`, `groupletWizard`, `groupletGrid`, Topic Grid — plus the three `genro.dlg.*DataEditor` variants and a final cheat sheet.

## Layout

```
docs/grouplet-family-for-dummies/
├── index.html       # shell: dark theme, tab navigation, copy buttons
├── serve.py         # zero-dep launcher (stdlib http.server)
└── tabs/
    ├── why.html         # overview + reactive flow SVG
    ├── quickstart.html  # minimal handler + minimal resource
    ├── grouplet.html    # the base widget
    ├── chunk.html       # groupletChunk + template syntax
    ├── form.html        # groupletForm + memory/record stores
    ├── panel.html       # tree / multiButton navigation
    ├── wizard.html      # step-by-step (with stepper SVG)
    ├── grid.html        # groupletGrid (architecture SVG + 7 modes)
    ├── topic.html       # CSS grid of folder of grouplets
    ├── dialogs.html     # memory/document/recordDataEditor
    └── cheatsheet.html  # 43-row "I want to..." reference + anti-patterns
```

Total: 13 files, ~2385 added lines.

## How to view

```bash
python3 docs/grouplet-family-for-dummies/serve.py
# opens http://127.0.0.1:8000/ in the browser
```

Direct `file://` doesn't work because `index.html` uses `fetch()` to load the tab fragments, which the browser blocks for CORS on local files. The launcher script picks a free port automatically (8000+), serves the directory, and opens the browser.

Flags: `--port N` for explicit port, `--no-browser` to skip auto-open.

## Why fragments instead of one monolithic HTML

Multi-file is friendlier for refactoring — each tab can be edited independently without touching the shell, and the shell's CSS/JS is shared (no duplication). Tradeoff is that `file://` doesn't work; `serve.py` is the answer.

## Tone and content

Each tab follows a consistent shape:

1. One-sentence elevator pitch
2. "When to use" / "Why" (problem framing, not API listing)
3. Quick example, code is copy-pasteable
4. One or more advanced patterns
5. Full parameters table
6. Gotchas (as warning call-outs)
7. Reference files (source paths under `resources/common/gnrcomponents/grouplet/` and the test webpages)

Code samples are grounded on the actual test files under `projects/gnrcore/packages/test/webpages/gnrwdg/test_grouplets/` (legacy family) and `test_grouplet_grid/` (new from #905). The reader is assumed to know GenroPy basics (Bag, datapath, formbuilder) — no remedial explanations.

## Related issues

Several silent-failure modes surfaced during the read-through and are filed in companion issues:

- #909 — umbrella bug: silent collisions on `name=` / `formId=` / `frameCode=` registries, missing `@public_method` not raising, `handler+resource` precedence undefined, etc.
- #910 — RFC discussion: should `grouplet` accept a client-side `handlerJs=` for static templates that don't need a server round-trip? Includes a measurement plan and three decisional alternatives.

## Test plan

Documentation-only PR — no production code changes.

- [x] All 11 tab fragments parse as valid HTML (no `<html>`/`<head>`/`<body>` wrappers, no `<style>`/`<script>` blocks)
- [x] All fragments served via `python -m http.server` return HTTP 200 (smoke-tested locally)
- [x] No AI/LLM references in any of the 13 files (greps clean)
- [x] `flake8` clean on `serve.py` (zero errors, strict mode)
- [x] `python3 -m py_compile serve.py` passes
- [x] Server launcher tested with auto-port detection and `--port`/`--no-browser` flags
- [ ] Reviewer to verify the doc renders correctly on Chrome / Firefox / Safari

## Suggested follow-ups (not in this PR)

- Hook the doc into a wider GenroPy docs index when one exists
- Consider auto-publishing to GitHub Pages on tag releases
- Iterate on tab content as the family evolves (e.g. when #910 lands, add a "JS handler" subsection to the relevant tabs)


fporcari (commented):
This fills a real gap. `grouplet` is not mentioned anywhere under `docs/` on develop, so this is currently the only written account of the family, and the shape you chose for each tab — problem framing first, parameters table late, gotchas as call-outs — is the right order for a reader who is deciding *whether* to reach for the widget rather than looking up an argument they already know. The vocabulary section in `tabs/grid.html` (item / pkey / tile / chrome / body) is worth keeping whatever happens to the rest: those five words are used throughout the source with no definition anywhere.

Two things stand between it and a merge.

## The groupletGrid chapter predates the refactor

`a885a0680 refactor(groupletGrid): client-side store, controllerPath, cleanup (#916)` landed after this branch was cut, and `tabs/grid.html` still describes the architecture it replaced.

The concrete marker: `controllerPath` appears 16 times in the current code — `resources/common/gnrcomponents/grouplet/grouplet.py:735` in the signature, with the fallback semantics spelled out at `grouplet.py:801-805`, and `resources/common/gnrcomponents/grouplet/grouplet_grid.js:11-27` where the controller takes it as a constructor argument — and zero times across all 13 files of this PR. The "How it works (architecture)" section, its SVG, and the "The sibling BagStore is mandatory by design" call-out all describe the pre-#916 wiring where `storepath` is the only path involved.

The rest of the chapter has aged better than that suggests — `gr_getGroupletGridTemplate` is still the boot-time template fetch, and the tile lifecycle is unchanged — so this is a section rewrite, not a rewrite of the tab.

Also missing, all post-branch: `lazyTabs`, `fillParent`, `groupletChunk` record mode via `record_id`, and drop-on-`+`.

## The docs index already exists

The follow-up list proposes hooking this into a wider docs index "when one exists". `docs/genro/` is a complete Sphinx tree — `conf.py`, `Makefile`, `index.rst`, and a `components/` section with `writing.rst` and `components_list.rst` — and it predates this branch. As it stands the PR installs a second documentation toolchain beside it, with its own shell, its own CSS and its own `serve.py`.

That is a real cost but not obviously the wrong trade: converting ~2385 lines of hand-built HTML into reStructuredText is weeks, and the content is useful now.

## Suggestion

Update the `grid.html` architecture section to the post-#916 wiring and merge as-is, with a follow-up issue for the migration into `docs/genro/components/`. Documentation that describes a superseded architecture is worse than none, because a reader has no way to tell which half is stale — but that is one section, and it should not hold the other twelve files hostage.

One note on why this sat untouched since May: `docs/` matches no rule in `.github/CODEOWNERS` on develop, so this PR opened with an empty `reviewRequests` and nobody was ever notified. #1093 adds a catch-all that closes that.

Marking as a comment rather than a block — the structural question is yours to settle, and the `grid.html` fix is the only thing I would insist on before merge.

fporcari:
Re-checked at `747aff66`: the push on 19 August is the `develop` merge only, so the one thing I asked for is still open and this PR is waiting on you rather than on a reviewer. Nothing else stands between it and a merge — checks are green on all four Python versions and there is no other objection from me.

The three markers, re-run just now against the current head:

- `controllerPath` — 0 occurrences across the 13 files of this PR, 16 in `resources/common/gnrcomponents/grouplet/grouplet.py` and `grouplet_grid.js` on `develop`
- the "the sibling BagStore is mandatory by design" call-out is still in `tabs/grid.html`
- `lazyTabs`, `fillParent` and `groupletChunk` record mode via `record_id` are still unmentioned

So it is still the "How it works (architecture)" section of `tabs/grid.html` and its SVG, not the tab and not the other twelve files. Everything else in the chapter has aged fine.

One practical note on why nobody picked this up in three months: `reviewRequests` is empty and there is no assignee, so this PR notifies nobody. `docs/` matched no rule in `.github/CODEOWNERS` when it opened. Assigning it to yourself and requesting a reviewer explicitly would put it back in somebody's queue — mine included, if you want it there.

PR #907, opened 2026-05-15
fix(d11/d20): make setHidden idempotent to preserve native double-click

fporcari:
## Summary

- Make `GnrDomSourceNode.setHidden` idempotent: skip `dojo.style` write and the trailing `genro.fakeResize()` when the requested visibility already matches the current one.
- Fix a latent bug in the same method: `statusChanged` was overwritten on every iteration of the `forEach` over `_hiddenTargets`, so only the last target's transition determined whether `fakeResize` ran. It now stays `true` once any target actually changed.
- Apply the same fix symmetrically to `gnrjs/gnr_d11/js/gnrdomsource.js` and `gnrjs/gnr_d20/js/gnrdomsource.js` to keep the two dialects aligned.

## Motivation

When a hidden binding like `hidden=^.grid.selectedId?=!#v` is wired to a toolbar element above a `Grid` row, every click on the grid re-fires the binding with the same value. Pre-fix, `setHidden` always wrote `display` inline and triggered `fakeResize()`, which reflowed the layout in the middle of the click sequence and caused the browser to lose track of the native double-click on the underlying row. With the no-op short-circuit, the second click on an already-selected row no longer fires a spurious reflow and the native double-click is preserved.

## Test plan

- [x] Manual: d11 page with `Grid` + toolbar bound via `hidden=^...selectedId?=!#v` — double-click on a selected row now fires reliably.
- [x] Manual: single click toggling the toolbar visibility (selected vs. no selection) still works as before — `dojo.style` and `fakeResize` are only invoked when the state actually changes.
- [ ] Smoke check on other consumers of `setHidden` / `setHiddenChild` across the app.

PR #913, opened 2026-05-18
fix(setHidden): trigger fakeResize only on real layout change

fporcari:
## Summary
- PR #907 made `setHidden` idempotent by short-circuiting the `dojo.style` write when `genro.dom.isVisible(node)` matched the requested state. That short-circuit also skipped the write that the `labelWrapper` / formlet hide-propagation depended on, leaving labels visible while their widgets were hidden (visible e.g. on the login form's optional `dbSelect` / `dateTextBox` fields).
- The `display` style is now always written, so wrapper propagation works again. `fakeResize` is fired only when both the hidden state and the node's `offsetHeight` actually change — preserving #907's intent (no spurious `fakeResize` on already-selected grid row clicks; no interference with the browser's native double-click on previously-unselected rows).
- Mirrored on `gnr_d11` and `gnr_d20` to keep them in sync.

## Drive-by cleanup
- Adds `test_5_top_autoheight_hidden_toggle` in `bordercontainer.py` to verify that a region sized by its content correctly relayouts when a child is shown/hidden.
- Removes a dead `dashboard_component/dashboard_component:DashboardItem` `py_requires` from `bordercontainer.py` and `framepane.py`. The component lives in the `biz` package and is not resolvable from the `test` package, which made both pages fail to load. The symbol was never used in either file.

## Test plan
- [x] Login form: hidden optional fields no longer leak their label (regression introduced by #907 fixed)
- [x] Grid with `hidden=.grid.selectedId?=!#v` on toolbar items: native double-click on a previously-unselected row works at first try
- [x] `test/layout/bordercontainer` test_5: toggling the checkbox grows / shrinks the top region and the center region adapts accordingly
- [x] Backend test suite passes (1553 passed, 260 skipped — pre-existing skips)
- [x] flake8 clean on changed Python files

### DIFF
--- a/gnrjs/gnr_d11/js/gnrdomsource.js
+++ b/gnrjs/gnr_d11/js/gnrdomsource.js
@@ -1593,10 +1593,11 @@ dojo.declare("gnr.GnrDomSourceNode", gnr.GnrBagNode, {
         var targets = this._hiddenTargets || [this.domNode || this.widget.domNode];
         var statusChanged = false;
         targets.forEach(function(domNode){
-            let currenHidden = !genro.dom.isVisible(domNode);
-            if(currenHidden != hidden){
+            let wasHidden = domNode.style.display === 'none';
+            let prevHeight = domNode.offsetHeight;
+            domNode.style.display = hidden ? 'none' : '';
+            if(wasHidden !== hidden && domNode.offsetHeight !== prevHeight){
                 statusChanged = true;
-                dojo.style(domNode, 'display', (hidden ? 'none' : ''));
             }
         });
         if(statusChanged){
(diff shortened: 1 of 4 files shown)


genro (approved):
LGTM. The fix correctly resolves the regression introduced by #907.

## Key points verified

- `genro.dom.isVisible()` (genro_dom.js:1728) also checks `clientHeight==0`, `clientWidth==0` and `isWindowVisible()` — so on a widget with a label, the post-#907 check could return "already hidden" for reasons unrelated to the inline `display`, skipping the write and leaving the label visible. The unconditional write of `style.display` removes this dependency.

- The `wasHidden !== hidden && offsetHeight !== prevHeight` gate preserves #907's intent (no spurious fakeResize on Grid double-click) using a real layout-shift signal instead of computed style.

- The `statusChanged` OR-accumulation latent-fix from #907 is preserved for multi-target `_hiddenTargets` (e.g. checkbox wrapper + checkbox itself).

- d11/d20 kept in sync — diff is identical between the two dialects.

## Drive-by cleanup

The removal of the dead `dashboard_component:DashboardItem` `py_requires` from `bordercontainer.py` and `framepane.py` is correct: the component lives in `biz` and isn't resolvable from `test`, breaking both pages. The added `test_5_top_autoheight_hidden_toggle` is a useful visual regression check for the auto-height case.

## Minor (non-blocking)

A one-line comment above the `offsetHeight !== prevHeight` condition explaining *why* the fakeResize is gated on height delta would help future readers — easy to misread as redundant.

PR #914, opened 2026-05-18
fix(groupletGrid): serialize template loader RPC for dynamic mixin support

fporcari:
## Summary

- **Bug fix**: the JS controller invoked `gr_getGroupletGridTemplate` / `gr_getGroupletGridTemplateMap` by bare name via `genro.serverCall`, which only resolved when the methods were directly on the current `WebPage`. In dynamic-mixin contexts the name lookup missed the right implementation. The Python widget now passes the two bound `public_method`s as kwargs of the `dataController`; the framework's xml serializer turns them into full RPC path strings, which the JS controller stores on `this` and uses in `serverCall`.
- **Struct maintable propagation**: when `gr_groupletGrid(..., table=..., struct=callable)` is used, `newGridStruct(maintable=table)` is now called so column shortcuts (field name → table column metadata) resolve. Grouplets can be table-scoped — this case was missed.
- **Cleanup**: removed the `_UNSET` sentinel pattern. `delitem=True` / `editmenu=False` as direct kwarg defaults are semantically identical (no caller uses `delitem=None`) and follow the codebase convention.

## Linked Issue

None — micro-fixes on `gr_groupletGrid` (merged in #905) discovered during follow-up review.

## Test plan

- [x] `flake8` clean on modified files
- [x] Full pytest suite passes (1553 passed, 260 skipped)
- [x] Browser smoke: open a page using `gr_groupletGrid(...)` in single-template mode → verify the POST hits the **serialized full RPC path** (not the bare `gr_getGroupletGridTemplate` name)
- [x] Browser smoke: open a page using `resourceField=` (multi-grouplet mode) → idem for `gr_getGroupletGridTemplateMap`
- [x] Add/remove rows and drag&drop still work
- [x] Dynamic-mixin case (the original bug): the template loads correctly when `GroupletGridHandler` is reached via mixin, not direct `py_requires`
- [x] `gr_groupletGrid(table='foo.bar', struct=lambda s: s.col('field_name'))` → struct resolves `field_name` against `foo.bar` columns

### DIFF
--- a/resources/common/gnrcomponents/grouplet/grouplet.py
+++ b/resources/common/gnrcomponents/grouplet/grouplet.py
@@ -6,11 +6,6 @@ from gnr.core.gnrlang import gnrImport
 from gnr.web.gnrbaseclasses import BaseComponent
 from gnr.web.gnrwebstruct import struct_method
 
-# Sentinel for kwargs whose default may need to be resolved later. Lets
-# us tell "not passed" apart from "explicitly False/{}".
-_UNSET = object()
-
-
 class GroupletHandler(BaseComponent):
     css_requires = 'gnrcomponents/grouplet/grouplet'
     js_requires = 'gnrcomponents/grouplet/grouplet'
@@ -734,7 +729,7 @@ class GroupletGridHandler(BaseComponent):
                         table=None, grouplets_root=None,
                         cols=1, min_width=None, gap='12px',
                         height=None, max_height=None,
-                        additem=True, delitem=_UNSET, editmenu=_UNSET,
+                        additem=True, delitem=True, editmenu=False,
                         layout='cards',
                         titleField=None,
                         emptyTitle='!!Untitled',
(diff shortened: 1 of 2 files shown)


genro (approved):
LGTM. Three small, well-scoped fixes on `gr_groupletGrid` (merged in #905).

## Verified

1. **RPC serialization for dynamic mixins** — architecturally sound:
   - `gnrbagxml.py:465-466` preserves attrs with `is_rpc` during XML serialization
   - `gnrdecorator.py:62-77` shows `@public_method` sets `func.is_rpc = True`
   - `gnrlang.py:476-498` (`serializedFuncName`) turns the bound method into the full RPC path, including `__mixin_pkg` / `__mixin_path` markers for dynamic mixins — exactly the case the bug surfaced
   - `genro.js:2191` → `genro_rpc.js:448-450` show `genro.serverCall(method, params, cb)` accepts the method as an arbitrary string, so the serialized path passes through unchanged

2. **`newGridStruct(maintable=table)`** — `newGridStruct` defaults `maintable=None` (gnrwebpage.py:1597), so the change is backward-compatible. No in-repo caller currently uses `table=` with a callable `struct=`, so this is preventive and matches the updated docstring.

3. **`_UNSET` sentinel cleanup** — `delitem=True` / `editmenu=False` as direct defaults are semantically identical (no caller passes `None`), consistent with the codebase convention.

## Minor (non-blocking)

- Three logically distinct fixes (RPC, maintable, cleanup) bundled in one commit — bisecting will be less precise if a regression surfaces. Not worth splitting at this size, just noting.

- The test plan still has 4 browser-smoke items unchecked, including the one that matters most: the dynamic-mixin case (the original bug). Worth confirming that one was actually exercised in a browser before merge.

PR #917, opened 2026-05-19
fix(gnrwebstruct): attached sub-nodes win over widget namespace in __getattr__

fporcari:
## Summary

Fix a regression introduced in #906 by porting the full W3C HTML5 catalog into `AllWidgets._widget_names`. The new catalog registers tags such as `form`, `header`, `footer`, `section`, `nav`, `main`, `dialog`, `menu`, `select`, `button`, `a`, ... — names that are also commonly used as **attached sub-node names** inside genro structs (the tableHandler attaches a `form` sub-node to `th`, for example).

Because `GnrDomSrc.__getattr__` consulted the widget namespace **before** `getNode`, expressions like `th.form` started returning a fresh `GnrDomElem('form')` instead of the attached form sub-node. That broke every consumer that reads attributes on the attached node — concretely:

```python
File ".../resources/common/public.py", line 700, in _th_parentFrameMessageSubscription
    fid = form.attributes['formId']
AttributeError: 'GnrDomElem' object has no attribute 'attributes'
```

Any tablehandler page in any project (reproduced on the sandbox `th_prodotto` of `fatt.prodotto`) crashes on render and redirects to `/console?error=...`.

## Fix

Reorder the lookup ladder in `GnrDomSrc.__getattr__` so that `getNode(fname)` runs **before** the `genroNameSpace` lookup. An explicit named child is stateful (attributes, children, nodeId, …) and must shadow the generic constructor offered by the widget catalog. The previous order was effectively silent data loss: the attached node was orphaned and unreachable via attribute access whenever its name collided with a registered tag.

External methods (`_external_methods`) keep their historical position after the widget namespace.

The HtmlWidgets catalog and the `@element` / `AllWidgets` registry introduced by this PR are untouched — this is a dispatch-order fix, not a catalog rollback.

## Why not other approaches

- **Restricting the HTML5 catalog** (dropping `form` and friends): masks the bug but leaves the same trap for any future catalog addition. The structural fix is in the dispatcher.
- **Renaming the colliding sub-nodes** (`th.form` → `th._form`): cascades through countless callers in genropy and downstream projects.

## Regression test

Added `test_attached_subnode_wins_over_widget_namespace` in `gnrpy/tests/web/gnrwebstruct_test.py`: attaches a sub-node named `form`, asserts that `root.form` returns the attached node (and is not a `GnrDomElem`), and that `'form'` is indeed in `genroNameSpace` (so the test would have caught the pre-fix behaviour).

## Test plan

- [x] `pytest gnrpy/tests/web/gnrwebstruct_test.py` (18/18 pass, including the new regression test)
- [x] `pytest gnrpy/tests/web/widgets_test.py` (25/25)
- [x] Full pre-commit test run via hook: 1579 passed, 260 skipped
- [x] `flake8` clean on modified files
- [x] Manual repro on sandbox `th_prodotto` of `fatt.prodotto` — page renders without redirecting to `/console?error=`

### DIFF
--- a/gnrpy/gnr/web/gnrwebstruct/base.py
+++ b/gnrpy/gnr/web/gnrwebstruct/base.py
@@ -167,10 +167,16 @@ class GnrDomSrc(GnrStructData):
             return self.parent.parentfb
     parentfb = property(_get_parentfb)
             
-    def __getattr__(self, fname): 
+    def __getattr__(self, fname):
         fnamelower = fname.lower()
         if (fname != fnamelower) and hasattr(self, fnamelower):
             return getattr(self, fnamelower)
+        # Attached sub-nodes must win over the widget catalog: an explicit
+        # named child is stateful (attributes, children), while the widget
+        # entry would build a brand-new element and orphan the existing node.
+        attachnode = self.getNode(fname)
+        if attachnode:
+            return attachnode._value
         if fnamelower in self.genroNameSpace:
             return GnrDomElem(self, '%s' % (self.genroNameSpace[fnamelower]))
         if fname in self._external_methods:
@@ -182,9 +188,6 @@ class GnrDomSrc(GnrStructData):
                     "Struct method '%s' not found in page '%s'"
                     " — check py_requires" % (method_name, page_name))
             return lambda *args, **kwargs: handler(self, *args,**kwargs)
-        attachnode = self.getNode(fname)
-        if attachnode:
-            return attachnode._value
         autoslots = self._parentNode.attr.get('autoslots')
         if autoslots:
             autoslots = autoslots.split(',')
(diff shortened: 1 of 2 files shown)


genro (approved):
LGTM. Targeted fix for the regression introduced by the W3C HTML5 catalog landing in #906.

Notes for follow-up (not blockers):

1. `getNode(fname)` now runs on every `__getattr__` miss, including widget-only paths
   like `th.span` / `pane.button`. Lookup is O(n) on the node's children list
   (`Bag._index` is a linear scan over `_nodes`). On typical struct sizes this is
   negligible; worth keeping an eye on if heavier pages start feeling slow.
2. The reorder also moves attached sub-nodes ahead of `_external_methods`: an
   explicit `@struct_method` named `form` would now lose to an attached sub-node
   called `form`. I don't think this collision exists in practice, but a one-line
   regression test pinning the new precedence (sub-node > namespace > external
   methods) would document the choice.
3. Chained on top of #906 — should be merged together with or right after the
   parent.

Issue #919, opened 2026-05-20
Add per-query timeout (statement_timeout) support to db.execute / SqlTable.query


genro:
## Motivation

`gnrsql` currently has no way to bound the wall-clock duration of a
single query. Adapters expose `timeout` only for `listen()` (async
pub-sub) — not for `SELECT` / `UPDATE` / `INSERT`.

In contexts where queries are driven by external or untrusted input
(HTTP/RPC endpoints, integrations consuming user parameters, batch
actions), a stuck or pathological query can block a worker or a
connection indefinitely. A per-query timeout would let callers cap
the exposure at the framework level.

## Proposal

Add per-query timeout support to `gnrsql`. Two design directions
(maintainers to choose, or combine):

### A. Via `currentEnv` / `tempEnv`

The caller scopes a timeout via the existing environment mechanism:

```python
with db.tempEnv(statement_timeout_ms=200):
    rows = db.table('invc.invoice').query(where='...').selection()
```

The framework reads `currentEnv['statement_timeout_ms']` inside
`db.execute()` and applies it on the actual cursor that runs the
statement, then restores the previous value. This composes naturally
with other `tempEnv`-scoped settings (locale, storename, workdate)
and keeps the call site of `.query()` unchanged.

### B. Via explicit parameter

A keyword `max_time_ms` (or similar) added to `SqlTable.query()`
and/or `GnrSqlDb.execute()`:

```python
db.table('invc.invoice').query(where='...', max_time_ms=200).selection()
```

More explicit but requires plumbing the parameter through the query
compiler.

### Shared semantics

Regardless of the surface chosen:

- on PostgreSQL, applies the appropriate session/local
  `statement_timeout` and resets it afterwards
- on other adapters, best-effort or no-op, with documented behavior
- raises a dedicated exception on expiry (e.g.
  `GnrSqlTimeoutException`, derived from `GnrSqlException`) so
  callers can distinguish a timeout from a generic SQL failure

## Why this belongs in the framework

A user-side `SET SESSION statement_timeout = ...` issued through the
public `db.execute()` API does not reliably propagate to the cursor
that subsequently runs `.query().selection()`: the gnrsql layer wraps
each call in its own transactional scope, so the SET does not persist
across the boundary. The correct fix lives where the connection
lifecycle is known.

## Scope

- core support in `gnrsql`
- Postgres adapter implementation
- exception type and documented behavior on adapters that do not
  natively support per-statement timeouts

PR #920, opened 2026-05-20
feat(app): introduce gnr.app.api_engine.ApiEngine

genro:
## Summary

Introduces `gnr.app.api_engine.ApiEngine`, an application-level facade
for database introspection and JSON-safe read access — designed for
consumption by HTTP/RPC endpoints, CLI tools and external integrations.

```python
from gnr.app.api_engine import ApiEngine

engine = ApiEngine('my_instance')          # or ApiEngine(my_app)
engine.package_names()                      # ['adm', 'invc', 'sys']
engine.table_names()                        # full list
engine.table_names('invc')                  # filtered by package
engine.table_schema('invc.invoice')         # relation tree
engine.table_columns('invc.invoice')        # flat per-column typing
engine.openapi_schema('invc.invoice')       # OpenAPI schema (read+write)
engine.run_query('invc.invoice',
                 columns='$id,$total',
                 where='$total > :t',
                 sqlparams={'t': 100},
                 limit=10)                  # JSON-safe envelope
```

The engine bundles five surfaces under one class:

- **Enumeration**: `package_names`, `table_names(target=None)`
- **Relation-tree schema**: `table_schema(target, depth, relations_only)`
- **Per-column typing**: `table_columns(target)` — `dtype`,
  `python_type`, `openapi_type`, `openapi_format`, `nullable`,
  `pkey`, `fkey`, `size`, etc.
- **OpenAPI schema synthesis**: `openapi_schema(target, readonly)` —
  opt-in via `openapi=True` on table/package, opt-out per column via
  `openapi=False`, read-only per column via `openapi='R'`. System
  columns (`__` prefix) are never writable; `pycolumn` never appears.
- **Query execution**: `run_query(table, ...)` — JSON-safe envelope
  `{rows, rowcount, truncated, elapsed_ms, error}`, configurable
  safety cap `max_rows`.

Targets across the introspection methods are uniformly polymorphic:
`None` (entire db), package name, table fullname, or list of either.
Output is always wrapped as `{table_fullname: ...}`.

## Robustness against imperfect models

Two real-world model imperfections that previously crashed ApiEngine
are now handled explicitly:

- **Non-string dtype** (e.g. `dtype=True` from a typo): raises
  `ApiEngineError` with a diagnostic message naming the offending
  table, column and value.
- **Unknown dtype string** (e.g. `dtype='Z'`): raises
  `ApiEngineError` listing the accepted codes.
- **FK whose target table belongs to an unloaded package**: silently
  omits the `fkey` entry. `model.column()` raises `AttributeError`
  in this case because `model.table()` returns `None`; the engine
  catches it and still describes the column.

The principle: ApiEngine describes what the model exposes — it
surfaces structural bugs but tolerates partial models.

## Out of scope

- **Per-query wall-clock timeout**: gnrsql has no native primitive
  and a user-side `SET SESSION statement_timeout` does not persist
  across gnrsql's transactional boundary. Tracked by #919. Callers
  needing to bound query duration use external means (asyncio,
  signal alarm) until #919 lands.
- **User-bound filtering** (allowed tables/columns from
  `app.checkResourcePermission`, `_getUserConfiguration`): the
  engine is "all or nothing" today; a follow-up PR will layer
  user-level filters on top.
- **DML primitives** (`insert`/`update`/`delete`): a future module
  in the same package will add write support.
- **Sourcerer endpoint refactor**: `projects/.../ep_sourcerer.py`
  still duplicates introspection helpers. A follow-up PR will
  rewrite it on top of ApiEngine.

## Test plan

- [x] 46 unit tests in `gnrpy/tests/app/test_api_engine.py` run on
  a real SQLite database populated from
  `projects/test_invoice/data/export/`. Coverage: construction,
  enumeration, table_columns (typing, fkey detection, system flag,
  pycolumn exclusion, package/list targets), table_schema
  (relations-only, depth, package target), openapi_schema (opt-in,
  veto, readonly, column-level override), run_query (envelope,
  max_rows clamp engine and per-call, error handling, opt_kwargs
  validation, JSON-safe), partition_kwargs (single value, list,
  currentEnv restoration), dtype validation regressions
  (non-string, unknown), FK orphan tolerance (both
  `related.table is None` and `relatedColumn` raising
  `AttributeError`).
- [x] flake8 zero output on package and tests.
- [x] All 46 tests pass; pre-existing failures elsewhere in the
  gnrpy suite are unchanged.

## Cleanup note for a follow-up

`gnrpy/tests/app/conftest.py` is a deliberate SQLite-only copy of
`gnrpy/tests/sql/conftest.py`. A future cleanup should factor the
shared fixture into a single reusable module — the duplication was
chosen here to keep the PR strictly additive (no edits to existing
infrastructure).


fporcari (dismissed):
LGTM — additive, well-tested against a real SQLite DB (no mocks), commit messages are great, and the `statement_timeout_ms` story (drafted → removed → tracked by #919 → pinned by regression test) is exemplary. The two FK-orphan fixes with separate regression tests for each branch are also nicely done.

A few non-blocking suggestions to consider in a follow-up (or fold in now if convenient):

**1. "JSON-safe envelope" claim in `run_query`**

`selection.output(mode='dictlist')` materializes raw `Decimal` / `datetime` / `date` / `Bag` values, which stdlib `json.dumps` cannot serialize without a custom encoder. The `test_json_safe_envelope` test passes thanks to `default=str` — i.e. the envelope is *coercible* to JSON, not strictly *JSON-safe*.

Two options:
- coerce inside the engine (Decimal → str, date/datetime/time → ISO, Bag → dict/None) so callers can call plain `json.dumps`;
- soften the docstring/PR claim and document that callers must supply an encoder.

Option 1 fits the "feed HTTP/RPC consumers" use case better, but it can absolutely land in a follow-up.

**2. `truncated` false positive at exact cap**

```python
truncated = (
    cap is not None
    and requested_limit is not None
    and requested_limit > int(cap)
) or (cap is not None and rowcount >= int(cap))
```

If the table has exactly `cap` rows (or the caller passed `limit=cap` and the result fills it exactly), `rowcount >= cap` flags truncated even though nothing was actually cut. Existing tests don't catch it because the table has > cap rows. Suggested shape:

```python
truncated = (
    cap is not None
    and rowcount == int(cap)
    and (requested_limit is None or requested_limit > int(cap))
)
```

plus a regression test where table size == cap and `limit` is not the limiting factor.

**3. Dead `addPkeyColumn=False`**

```python
query_kwargs = dict(..., addPkeyColumn=False)
...
query_kwargs = {k: v for k, v in query_kwargs.items() if v is not None}
query_kwargs.setdefault('addPkeyColumn', False)  # dead — False survives the None-filter
```

The `setdefault` line is unreachable in practice. Pick one of the two and drop the other.

**4. Track the `ep_sourcerer` refactor**

The PR body already commits to rewriting `projects/gnrcore/packages/sys/webpages/ep_sourcerer.py` on top of `ApiEngine`. Worth opening a tracking issue now so the `_classify_column` / `_attr_to_json` / `_bag_to_json` duplication doesn't drift — every change to those helpers currently has to be mirrored in two places.

Same goes for the `tests/app/conftest.py` ↔ `tests/sql/conftest.py` duplication you already flagged.

None of the above block the merge. Approving.

genro:
Thanks for the careful review. All three required changes have been
addressed in the three commits just pushed:

**R1 — JSON-safe envelope (97043650)**
\`run_query\` now coerces row cells before returning: Decimal → str
(precision preserved), date/datetime/time → ISO 8601 strings,
timedelta → \`str(td)\`, bytes → UTF-8, objects with \`toJson\` → that
method, dicts/lists recursed. The docstring documents the
guarantee. The \`json.dumps(default=str)\` crutch is gone; a new
test selects \`\$total\` (Decimal) and \`\$date\` (date) and verifies a
plain \`json.dumps\` works without any custom encoder.

**R2 — truncated false-positive (ccc95bba)**
Switched to the probe-limit-plus-one pattern: when a cap is active,
the engine internally fetches \`cap + 1\` rows. If the result is
larger than \`cap\`, it trims back and sets \`truncated=True\`;
otherwise \`truncated=False\`. Two regression tests pin the false-
positive case you flagged (table has exactly \`cap\` rows; explicit
limit smaller than the cap).

**R3 — dead setdefault + S3 (3efcd8d0)**
Removed the unreachable \`query_kwargs.setdefault('addPkeyColumn',
False)\` and normalized \`cap = int(cap)\` once at the top of
\`run_query\`. S4 and S5 (docstring notes on dedup and on \`None\` in
\`partition_kwargs\`) bundled in.

**S1 and S6 — tracking issues opened**
- #922: refactor \`ep_sourcerer.py\` on top of \`ApiEngine\`.
- #923: factor the \`test_invoice\` fixture into a shared module so
  the two conftest files stop being copies of each other.

49 tests, all passing; flake8 zero; manual smoke on the
\`test_invoice_pg\` Postgres instance verified the new JSON coercion
and the truncated logic across five scenarios.

genro:
Added a SQL-injection blacklist as an extra safety net for the
run_query SQL-text parameters. Pushed as commit ef1c8d125.

The engine now rejects (with ApiEngineError) fragments in
columns/where/having/order_by/group_by that match well-known
injection-style patterns:

- statement injection (semicolons, null bytes, MSSQL GO)
- SQL comments (--, /* */, MySQL #)
- DDL/DML keywords (DROP/CREATE/ALTER/TRUNCATE/INSERT/UPDATE/DELETE/MERGE/REPLACE/RENAME)
- privilege keywords (GRANT/REVOKE)
- session / execution (EXEC/CALL/SET/COPY/LOAD/COMMIT/ROLLBACK)
- set operators and nested queries (UNION/INTERSECT/EXCEPT/WITH/nested SELECT)
- file & system access (INTO OUTFILE/DUMPFILE, LOAD_FILE, pg_read_file, xp_cmdshell, ...)
- system catalogs (information_schema, pg_catalog, sysobjects, ...) — gnrsql's own `sys` package is NOT in the list
- recon functions (version(), user(), current_user, current_database, system_user, suser_sname)
- encoding bypass tricks (CHAR(), CHR(), CONCAT_WS())
- sleep/DoS (SLEEP(), pg_sleep(), BENCHMARK(), WAITFOR DELAY, Oracle dbms_pipe)
- tautology auth-bypasses (1=1, 'a'='a')

Patterns are case-insensitive and word-boundary anchored, so
legitimate identifiers (`dropdown_field`, `update_at`,
`selection_id`) pass. gnrsql macros (`#IN_RANGE`, `#PERIOD`, etc.)
pass too: the `#` comment pattern only matches when not followed
by an uppercase letter.

For callers that legitimately need constructs the blacklist
rejects (complex expressions, subqueries, custom aggregates),
the escape hatch is to add a formulaColumn on the model and
reference it as a plain `$col_name` — the formula is defined
server-side, type-safe, and not subject to the blacklist.

30 new tests in TestSqlBlacklist (22 attack vectors + 8
false-positive guards). All 79 ApiEngine tests pass; flake8
clean; full gnrpy suite shows no new failures (22 pre-existing
unrelated failures unchanged).

mbertoldi (approved):

Issue #918, opened 2026-05-19
[BUG] mkthresource crashes on Form class generation (tuple unpack mismatch)


dgpaci:
### Problem Description

`gnr dev mkthresource` crashes with `ValueError: not enough values to unpack (expected 3, got 2)` while writing the Form class of the generated resource file. The View class is written correctly, but the Form class is left half-written (the file ends right after `fb = pane.formbuilder(...)` with no `fb.field(...)` calls and no `th_options` method).

The bug is a tuple shape mismatch inside `gnrpy/gnr/dev/makers/resource.py`:

- `createResourceFile` builds the `columns` list with 2-tuples at line 499: `columns.append((column.name, width))`
- `writeFormClass` unpacks them as 3-tuples at line 284: `for column, size, dtype in columns:`

The same mismatch is also present in `writeViewClass` at line 253, but the View loop in the codepath that triggers fires earlier through `column_groups` (2-tuples too) so the crash surfaces only on the Form side. Effectively, the Form generator is unreachable in its current shape.

`writeFormClass` should iterate over 2-tuples (and skip `dtype`/`size`), or `createResourceFile` should provide 3-tuples consistently — choose one shape and align both producers and consumers.

### Reproduction Steps

1. Pick any non-lookup table that has at least one regular column. Reproduced on `anc_base.socio_sede_collaborazione` (three FK columns: `socio_id`, `sede_id`, `tipo_collaborazione`).
2. Run `gnr dev mkthresource <instance_name> <pkg>.<table>` (e.g., `gnr dev mkthresource anaci_prd anc_base.socio_sede_collaborazione`).
3. Observed result: ERROR log line `Error creating output file: ... Error: not enough values to unpack (expected 3, got 2)` followed by the traceback below. The resource file is created on disk but truncated — Form class missing its fields and `th_options`.

### Environment

- genropy `develop` @ commit 47197658bee062e4e95a2cb48208ce1f01d9ab12
- Python 3.12.9, macOS 25.1.0
- Regression introduced in commit `29305a1b9c` (PR #706 \"gnrdeploy refactoring\", 2026-03-18). The follow-up `ecc6e5cfa` only fixed `column.type → column.dtype` in the same module, leaving this tuple mismatch in place.

### Additional Material

Traceback:

\`\`\`
Traceback (most recent call last):
  File \".../gnrpy/gnr/dev/makers/resource.py\", line 511, in createResourceFile
    self.writeFormClass(tbl_obj, columns)
[...]

PR #921, opened 2026-05-20
fix 'not enough values to unpack' durint gnr dev mkthresource

dedalus2000:
The ``columns`` variable is a list of value pairs (look at row 499), but it is being unpacked into three values, generating the following error and preventing the correct creation of the form (which remains empty) during the execution of ```gnr dev mkthresource```


dgpaci:
Closes #918

dgpaci (approved):
Tested on a real package (a fresh `dbwatcher` package with 6 tables, including a Bag/`X` column for an inline query condition).

**Before the fix** (`develop`): `gnr dev mkthresource` produced th resources with an **empty `Form` class** — the `for column, size, dtype in columns` unpack raised `ValueError: not enough values to unpack (expected 3, got 2)` in `writeFormClass`, since `columns` carries `(column, dtype)` pairs (built around resource.py:499), not triples.

**With this branch**: ran `gnr dev mkthresource <instance> pkg.topic pkg.event pkg.subscription pkg.occurrence pkg.notification -f` — all forms are now correctly populated (5/9/9/9/17 fields respectively), View classes unchanged, and the instance boots cleanly with the generated resources.

The one-line fix is correct and minimal: `writeFormClass` iterates the same `columns` pair structure that `writeViewClass` already consumes. LGTM 👍

PR #924, opened 2026-05-21
Adopt ApiEngine in sys endpoints + introduce ep_openapi

genro:
## Summary

- **Refactor of `sys/ep_sourcerer.py`** on top of `gnr.app.api_engine.ApiEngine`. The endpoint shrinks from 186 to ~145 lines: the inline introspection helpers (`_classify_column`, `_attr_to_json`, `_bag_to_json`, `_build_table_schema`) are deleted because `ApiEngine` now provides them. `rpc_query` maps legacy `getSelection` shortcuts (`countOnly`, `condition`, `pkeys`) into `ApiEngine.run_query` primary/opt parameter sets and returns an ApiEngine-native envelope (`{ok, ticket_code, info, rows, rowcount, truncated, elapsed_ms}`). `rpc_relationtree` delegates to `engine.table_names()` / `engine.table_schema()`. The best-effort `SET LOCAL statement_timeout` is dropped (does not fire across gnrsql's transactional boundary; tracked separately).
- **New `sys/ep_openapi.py`** — a read-only REST surface auto-generated from the data model, intended as a transitional bridge for external integrators while the production REST API matures on genro-asgi. URL shape is REST-canonical:
  - `GET /sys/ep_openapi/openapi.json` → OpenAPI 3.1 spec
  - `GET /sys/ep_openapi/docs` → Swagger UI page (loaded from `unpkg.com/swagger-ui-dist`)
  - `GET /sys/ep_openapi/{pkg}/{table}` → list with `limit/offset/where/columns/order_by`
  - `GET /sys/ep_openapi/{pkg}/{table}/{pkey}` → single record
  - `POST /sys/ep_openapi/{pkg}/{table}/_search` → complex query in JSON body

  Tables are exposed only when the model declares `openapi=True`. Auth via bearer (`Authorization: Bearer <token>`, service `openapi` with a transitional fallback to `sourcerer`). Basic CORS preflight + permissive `Access-Control-Allow-Origin` so Swagger UI works cross-origin. Error envelope `{error: {code, message, details?}}` with real HTTP status codes. Write verbs (POST create, PUT, DELETE) are intentionally not exposed: write semantics will land with the ASGI rewrite.

- **`ApiEngine` improvements**
  - Columns with unknown / invalid dtypes are now silently skipped during schema build (previously they raised `ApiEngineError` and blocked the entire table). This degrades gracefully when extensions like pgvector's `VEC` show up.
  - New `exposed_column_names(fullname, readonly=True)` helper returns the list of columns that would appear in `openapi_schema()` properties. `ep_openapi` uses it to compute a default `columns=` expression for the runtime query so the response is a strict subset of what the spec promises (no `description_emb` leaking through when the spec hides it).

The client-side adaptation (Sourcerer MCP `target_db`) is in `genropy/genro-sourcerer` and tracks the same wire shape; it is already merged on `main` there.

## Test plan
- [x] `pytest gnrpy/tests/app/test_api_engine.py` — 79 passed (two `*_dtype_raises` tests rewritten to assert the new skip semantics)
- [x] flake8 (`--max-line-length=120`) clean on every changed file
- [x] E2E against the Sourcerer DB on Hetzner via the MCP client:
  - `target_db.tables(server_id='_self_')` (local) and via HTTP loopback (`server_id='WpbgJ2bjOWeb86_Yfk4BKA'`)
  - `target_db.getSelection` with `where + sqlparams`, `countOnly=True`, `pkeys`, `condition`
  - `target_db.relationtree`
- [x] E2E with curl against `ep_openapi`:
  - `GET /openapi.json` → 200, valid OpenAPI 3.1 with `paths`/`tags`/`components.schemas`
  - `GET /docs` → 200, Swagger UI HTML
  - `GET /gnrmcp/skill?limit=2` → 200 + paginated payload
  - `GET /gnrmcp/skill/{pkey}` → 200 + single record (23 columns, no `*_emb`)
  - `POST /gnrmcp/skill/_search` with JSON body filtering `title ILIKE :p` → 200, filters applied
  - Missing/invalid bearer → 401; unknown table → 404
- [ ] One concrete external integrator consumes the OpenAPI spec via Swagger UI in production (to be done once `openapi` service tokens are provisioned across instances).

### DIFF
--- a/gnrpy/gnr/app/api_engine/core.py
+++ b/gnrpy/gnr/app/api_engine/core.py
@@ -133,7 +133,6 @@ _DTYPE_PYTHON = {
     'N': 'Decimal',
     'BAG': 'Bag', 'X': 'Bag',
     'JS': 'list',
-    'VEC': 'list',
 }
 
 _DTYPE_OPENAPI = {
@@ -156,8 +155,6 @@ _DTYPE_OPENAPI = {
     # structured
     'BAG': ('object', 'gnr-bag'), 'X': ('object', 'gnr-bag'),
     'JS': ('array', None),
-    # pgvector embedding column: array of floats
-    'VEC': ('array', 'vector'),
 }
 
 
--- a/gnrpy/gnr/app/api_engine/core.py
+++ b/gnrpy/gnr/app/api_engine/core.py
@@ -744,6 +744,34 @@ class ApiEngine:
                 out.append('%s.%s' % (pkg_name, tname))
         return sorted(out)
 
+    def exposed_column_names(self, fullname, readonly=True):
+        """Return the column names that would appear in the OpenAPI
+        schema for ``fullname``. Useful for callers that need to
+        materialize a payload whose shape matches the published spec —
+        e.g. so the runtime response stays a subset of what the spec
+        promises.
+
+        ``readonly=False`` mirrors the write-mode schema (excludes
+        system columns, formula columns, etc.). The default keeps the
+        full read shape.
+        """
+        columns = self._collect_columns(fullname)
+        out = []
+        for colname, info in columns.items():
+            col_setting = _column_openapi_setting(
+                {'openapi': info.get('openapi')})
+            if col_setting is False:
+                continue
+            is_real = info['kind'] == 'real'
+            is_system = info['system']
+            is_pkey = info['pkey']
+            writable = (is_real and not is_system
+                        and col_setting != 'R' and not is_pkey)
+            if not readonly and not writable:
+                continue
+            out.append(colname)
+        return out
+
     def _build_openapi_table_schema(self, fullname, readonly=True):
         columns = self._collect_columns(fullname)
         properties = {}
--- a/gnrpy/gnr/app/api_engine/core.py
+++ b/gnrpy/gnr/app/api_engine/core.py
@@ -780,19 +780,22 @@ class ApiEngine:
                 out.append('%s.%s' % (pkg_name, tname))
         return sorted(out)
 
-    def exposed_column_names(self, fullname, readonly=True):
-        """Return the column names that would appear in the OpenAPI
-        schema for ``fullname``. Useful for callers that need to
-        materialize a payload whose shape matches the published spec —
-        e.g. so the runtime response stays a subset of what the spec
-        promises.
+    def _iter_exposed_columns(self, fullname, readonly=True):
+        """Yield ``(colname, info, writable)`` for the columns of
+        ``fullname`` that pass the OpenAPI exposure filter.
 
-        ``readonly=False`` mirrors the write-mode schema (excludes
-        system columns, formula columns, etc.). The default keeps the
-        full read shape.
+        Filter rules:
+          - ``openapi=False`` at column level skips the column.
+          - In write mode (``readonly=False``), only columns that are
+            writable (real, not system, not declared 'R', not pkey)
+            survive.
+
+        Used by both ``exposed_column_names`` (which keeps just the
+        names) and ``_build_openapi_table_schema`` (which materializes
+        the property descriptors). Keep them in sync by sharing this
+        predicate.
         """
         columns = self._collect_columns(fullname)
-        out = []
         for colname, info in columns.items():
             col_setting = _column_openapi_setting(
                 {'openapi': info.get('openapi')})
@@ -805,28 +808,27 @@ class ApiEngine:
                         and col_setting != 'R' and not is_pkey)
             if not readonly and not writable:
                 continue
-            out.append(colname)
-        return out
+            yield colname, info, writable
+
+    def exposed_column_names(self, fullname, readonly=True):
+        """Return the column names that would appear in the OpenAPI
+        schema for ``fullname``. Useful for callers that need to
+        materialize a payload whose shape matches the published spec —
+        e.g. so the runtime response stays a subset of what the spec
+        promises.
+
+        ``readonly=False`` mirrors the write-mode schema (excludes
+        system columns, formula columns, etc.). The default keeps the
+        full read shape.
+        """
+        return [colname for colname, _, _
+                in self._iter_exposed_columns(fullname, readonly=readonly)]
 
     def _build_openapi_table_schema(self, fullname, readonly=True):
-        columns = self._collect_columns(fullname)
         properties = {}
         required = []
-        for colname, info in columns.items():
-            col_setting = _column_openapi_setting(
-                {'openapi': info.get('openapi')})
-            if col_setting is False:
-                continue
-            is_real = info['kind'] == 'real'
-            is_system = info['system']
-            is_pkey = info['pkey']
-            # Writable iff: real, not system, not declared 'R', not pkey
-            writable = (is_real
-                        and not is_system
-                        and col_setting != 'R'
-                        and not is_pkey)
-            if not readonly and not writable:
-                continue
+        for colname, info, writable in self._iter_exposed_columns(
+                fullname, readonly=readonly):
             prop = {'type': info['openapi_type']}
             if info['openapi_format']:
                 prop['format'] = info['openapi_format']
--- a/gnrpy/gnr/app/api_engine/core.py
+++ b/gnrpy/gnr/app/api_engine/core.py
@@ -584,7 +584,16 @@ class ApiEngine:
                 continue
             if attrs.get('virtual_column'):
                 continue
-            cols_dict[colname] = self._column_info(colname, attrs, tbl, col=col)
+            try:
+                cols_dict[colname] = self._column_info(
+                    colname, attrs, tbl, col=col)
+            except ApiEngineError:
+                # Column dtype is not known to ApiEngine (e.g. VEC from
+                # pgvector). Silently skip it so unknown extensions do
+                # not block the entire schema; callers that need full
+                # introspection of those columns can opt into a custom
+                # dtype mapping.
+                continue
         # Virtual columns from model (alias, formula_sql, formula_subquery,
         # bagitem, virtual_other — pycolumn excluded)
         vcols = self.db.model.table(fullname).get('virtual_columns') or {}
@@ -593,7 +602,10 @@ class ApiEngine:
             kind, _ = _classify_column(attrs)
             if kind == 'pycolumn':
                 continue
-            cols_dict[colname] = self._column_info(colname, attrs, tbl)
+            try:
+                cols_dict[colname] = self._column_info(colname, attrs, tbl)
+            except ApiEngineError:
+                continue
         return cols_dict
 
     def _column_info(self, colname, attrs, tbl, col=None):
--- a/gnrpy/gnr/app/api_engine/core.py
+++ b/gnrpy/gnr/app/api_engine/core.py
@@ -340,6 +340,24 @@ def _check_sql_fragment(value, parameter_name):
     return value
 
 
+def _is_known_dtype(dtype):
+    """Quick predicate used to decide whether a column should appear in
+    the OpenAPI schema (and consequently in REST payloads).
+
+    Returns True for the empty/None case (legitimate for virtual columns
+    without a declared type) and for any string in the known mapping.
+    Returns False for unknown strings (e.g. pgvector's ``VEC``) and for
+    non-string values (e.g. a model typo like ``dtype=True``). Strict
+    validation that raises ``ApiEngineError`` is reserved for
+    ``_validate_dtype``.
+    """
+    if dtype is None or dtype == '':
+        return True
+    if not isinstance(dtype, str):
+        return False
+    return dtype.upper() in _KNOWN_DTYPES
+
+
 def _validate_dtype(dtype, colname, tbl_fullname):
     """Validate a column dtype against the known mapping.
 
@@ -584,16 +602,10 @@ class ApiEngine:
                 continue
             if attrs.get('virtual_column'):
                 continue
-            try:
-                cols_dict[colname] = self._column_info(
-                    colname, attrs, tbl, col=col)
-            except ApiEngineError:
-                # Column dtype is not known to ApiEngine (e.g. VEC from
-                # pgvector). Silently skip it so unknown extensions do
-                # not block the entire schema; callers that need full
-                # introspection of those columns can opt into a custom
-                # dtype mapping.
+            if not _is_known_dtype(attrs.get('dtype')):
                 continue
+            cols_dict[colname] = self._column_info(
+                colname, attrs, tbl, col=col)
         # Virtual columns from model (alias, formula_sql, formula_subquery,
         # bagitem, virtual_other — pycolumn excluded)
         vcols = self.db.model.table(fullname).get('virtual_columns') or {}
@@ -602,10 +614,9 @@ class ApiEngine:
             kind, _ = _classify_column(attrs)
             if kind == 'pycolumn':
                 continue
-            try:
-                cols_dict[colname] = self._column_info(colname, attrs, tbl)
-            except ApiEngineError:
+            if not _is_known_dtype(attrs.get('dtype')):
                 continue
+            cols_dict[colname] = self._column_info(colname, attrs, tbl)
         return cols_dict
 
     def _column_info(self, colname, attrs, tbl, col=None):
@@ -690,6 +701,31 @@ class ApiEngine:
         schema).
 
         Returns ``{table_fullname: {<openapi schema object>}}``.
+
+        Dtype coverage
+        --------------
+        Only columns whose dtype is one of the codes below appear in the
+        schema (and, consequently, in REST payloads built by callers
+        that respect the schema):
+
+            T, A, P, TEXT  -> string
+            R, F           -> number(float)
+            L, I           -> integer
+            B              -> boolean
+            D              -> string(date)
+            DH, DT, DHZ    -> string(date-time)
+            H, HZ          -> string(time)
+            TD             -> string(duration)
+            N              -> string(decimal)
+            BAG, X         -> object(gnr-bag)
+            JS             -> array
+
+        Columns declaring any other dtype string (e.g. pgvector's
+        ``VEC``) or a non-string dtype are intentionally omitted: the
+        engine cannot describe their shape, so it neither lists nor
+        materializes them. If you need to expose such a column, extend
+        the dtype tables in this module or expose a derived virtual
+        column with a supported dtype.
         """
         explicit_table = isinstance(target, str) and '.' in target
         explicit_package = (isinstance(target, str)
--- a/gnrpy/gnr/app/api_engine/core.py
+++ b/gnrpy/gnr/app/api_engine/core.py
@@ -133,6 +133,7 @@ _DTYPE_PYTHON = {
     'N': 'Decimal',
     'BAG': 'Bag', 'X': 'Bag',
     'JS': 'list',
+    'VEC': 'list',
 }
 
 _DTYPE_OPENAPI = {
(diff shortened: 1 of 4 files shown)


fporcari (commented):
# Review notes — auth model, exposure boundary, and a few quality points

Solid refactor overall: centralizing introspection on `ApiEngine` is the right
move, and the new `ep_openapi` gives integrators a transitional REST surface
without disturbing the WSGI stack. The points below are mostly on the
authorization model, which I think needs to be made explicit before this is
used by AI/chat services in production.

## Authorization model — gaps to address

The PR description says "tables are exposed only if their model declares
`openapi=True`". In practice the boundary is permeable in four distinct
ways. None is a SQL injection (the blacklist in `_check_sql_fragment` is
fine for that), but together they make `openapi=True` insufficient as a
security gate.

### 1. GenroPy auth tags are not consulted

Neither `api_engine/core.py` nor the two endpoints reference `auth_tags`,
`readTags`, or `writeTags`. The only flag consulted is `openapi` (at
package / table / column level).

- A column with `auth_tags='hr_only'` and `openapi=True` is exposed to
  any caller with the service token, even if the same column would be
  hidden in the UI.
- `checkPermissions` is allowed as an opt_kwarg but it is opt-in, and
  it has no user identity to evaluate against (see §4).

Today the only protection at column granularity is to explicitly mark
every sensitive column `openapi=False`. The fine-grained tags written
elsewhere in the model are ignored.

### 2. Relation traversal via `$@` bypasses `openapi=True`

`columns`, `where`, `having`, `order_by`, `group_by` all accept GenroPy
expressions. `$@fkey.col` is legitimate GenroPy syntax and is not
screened by `_check_sql_fragment`. Consequence: any table with
`openapi=True` that has fkeys becomes a reading window into the tables
it references, even if those are `openapi=False`.

Attacks that work today:

- **Direct exfiltration** via `_search` body:
  `columns='$@customer_id.tax_id, $@customer_id.private_notes'`. The
  `_default_columns` restriction only applies when the caller omits
  `columns`.
- **Boolean-blind exfiltration** via `where`: even with `columns`
  restricted, `where='$@customer_id.tax_id LIKE :p'` lets the caller
  probe values character by character based on row counts.
- **Ordering-based exfiltration** via `order_by`.

The OpenAPI spec itself advertises `x-relation-path` on every fkey
property, so integrators have a map of available traversals.

**Fix**: parse GenroPy path expressions in the fragments and reject any
`@<rel>` whose target table is not `openapi=True`. Belongs in
`ApiEngine` so every consumer benefits. Trackable as a follow-up issue
but worth tagging as security.

### 3. `partition_kwargs` is caller-controlled in `ep_sourcerer`

`ep_sourcerer._RUN_QUERY_PRIMARY_KEYS` includes `partition_kwargs`.
`ApiEngine.run_query` translates it into `env_inject` (`current_<field>`
/ `allowed_<field>`). In an app that uses partitions for data scoping
(`current_pkey_user`, `allowed_branches`, ...) this lets the bearer
holder set the partition env to anything — i.e. impersonate any
user/branch for the duration of the call.

`ep_openapi` does **not** propagate `partition_kwargs`, which is the
right default.

**Fix (in-scope for this PR)**: drop `partition_kwargs` from
`_RUN_QUERY_PRIMARY_KEYS` in `ep_sourcerer`. If a server-side partition
env is needed, it must be derived from a verified identity, not chosen
by the client.

### 4. The service bearer carries no end-user identity

The structural one. The intended consumer is an AI/chat service: a
human asks the bot, the bot calls `rpc_query`. Today the endpoint sees
only "bot X is calling" — there is no concept of "on behalf of user Y".

Consequences:

- Authorization cannot be evaluated against the end user even if §1
  and §2 were fixed, because there is no user.
- Audit trail collapses to "the bot read everything"; in an incident
  there is no way to attribute a query to the human who triggered it.
- This is exactly the case OAuth 2.0 Token Exchange / On-Behalf-Of
  (RFC 8693) addresses, with `sub` = user and `act` = service.

**Follow-up (issue, blocking for AI/chat production use)**:

- Add an `acting_user` parameter to `ApiEngine.run_query`, not
  overridable from the request payload. It populates `current_user` in
  `env_inject` and forces `checkPermissions=True`.
- Endpoints fill `acting_user` from a verified source: HMAC-signed
  `X-On-Behalf-Of` header for v1, JWT with `sub`/`act` claims for v2.
- The OpenAPI spec advertises this in `securitySchemes` so integrators
  know the identity contract.
- Decide separately how the AI service *obtains* the end-user
  identity. Inside a GenroPy app it can sign the header from the
  session; from external chat clients (Claude Desktop, web UIs) it
  needs a shared IdP. Governance issue, not an endpoint issue, but it
  must be answered before "AI service ↔ ep_openapi" leaves dev.

### Documentation, regardless of which fixes land here

The `ep_openapi.py` module docstring and the comment on the model-level
`openapi` flag should state plainly:

> Setting `openapi=True` on a table exposes all of its columns
> (subject to per-column `openapi=False`) **and** makes any fkey-reachable
> table queryable through `$@` traversal. GenroPy `auth_tags` are not
> enforced. Use only for tables whose entire reachable subgraph is
> intended to be public to bearer holders.

The current docstring implies a tighter boundary than the code
provides; that mismatch is the most immediate risk.

## Other quality points

### `ApiEngine` — silent dtype skip
`core.py:584`, `core.py:602`: swallowing `ApiEngineError` is the right
call for pgvector `VEC`, but it now hides typos (`dtype=True`) by
making the column silently vanish. Add a `logger.warning` when
skipping, using `pkglog` per project convention. The module has no
logger at all today.

### `exposed_column_names` duplicates `_build_openapi_table_schema` logic
`core.py:744` reimplements the filter rules already in
`_build_openapi_table_schema`. Extract a shared predicate
(`_iter_exposed_columns(fullname, readonly)`) so the two paths cannot
drift.

### `ep_openapi` — server-side logging is absent
Errors are returned to the caller but nothing is logged on the server.
For a public REST surface this is a meaningful gap. At minimum:
`logger.warning` on 5xx paths, `logger.info` per handled request.

### `ep_openapi` — no request body size cap
`_read_json_body` calls `req.get_data(as_text=False)` without a
`max_content_length` check. Either rely on a site-wide limit or add an
explicit one.

### `ep_openapi` — `__error__` sentinel
Returning either parsed JSON or `{'__error__': response_str}` is hard
to read. Raise an exception caught in `_search`, or return an
`(ok, payload_or_error)` tuple.

### `ep_openapi` — Swagger UI behind auth
`_check_bearer` runs before the `openapi.json` / `docs` branch.
Intentional? Decide whether the schema and the docs page should be
unauthenticated (so an integrator can browse them) or authenticated
(current behaviour). At minimum document the choice.

### Bearer comparison is not constant-time
`_check_bearer` uses `==` on token strings. Switch to
`hmac.compare_digest`.

## Test coverage

- The renamed `_dtype_skipped` tests are good, but nothing covers:
  - `exposed_column_names` directly
  - the `_search` POST handler shape
  - the OpenAPI spec produced by `schema()` (paths, tags, required
    fields, `x-fkey` propagation)
- A unit test for "table with a relation to a non-exposed table"
  would naturally surface the traversal issue in §2 once the fix
  lands.

## Summary — what I'd ask before merge

**In this PR**:
- drop `partition_kwargs` from `ep_sourcerer` primary keys
- add a `pkglog` warning on the silent-dtype-skip path
- tighten the `ep_openapi` docstring to describe the actual exposure
  boundary
- switch bearer comparison to `hmac.compare_digest`

**As follow-up issues, before AI/chat production use**:
- relation-traversal guard in `ApiEngine`
- `acting_user` parameter + on-behalf-of identity propagation
- governance decision on where end-user identity comes from

fporcari (changes requested):
Switching this to request-changes. The detailed review is in the
[previous comment](https://github.com/genropy/genropy/pull/924#pullrequestreview).

The four items I'd like addressed before merge — all in-scope for this
PR, none requires a separate design:

- **Drop `partition_kwargs` from `_RUN_QUERY_PRIMARY_KEYS` in
  `ep_sourcerer`.** Today the bearer holder can pick the partition env
  (`current_pkey_user`, `allowed_branches`, ...). The server must
  derive the partition env from a verified identity, never from the
  client payload.
- **Add a `pkglog` warning on the silent-dtype-skip path** in
  `api_engine/core.py:584` / `:602`. The behaviour change is sound,
  but a typo like `dtype=True` now makes a column silently vanish.
  A warning preserves the graceful-degradation intent without losing
  observability.
- **Tighten the `ep_openapi.py` module docstring (and the comment on
  the model-level `openapi` flag)** to describe the actual exposure
  boundary: `openapi=True` exposes the table *and* makes every
  fkey-reachable table queryable through `$@` traversal; GenroPy
  `auth_tags` are not enforced. The current wording suggests a tighter
  boundary than the code provides — that gap is the most immediate
  risk for adopters.
- **Use `hmac.compare_digest` for the bearer comparison** in
  `_check_bearer`. Plain `==` on token strings is not constant-time.

The bigger items raised in the detailed review (relation-traversal
guard in `ApiEngine`; `acting_user` parameter and on-behalf-of
identity propagation; governance of end-user identity) can land as
follow-up issues — but those follow-ups are blocking for any AI/chat
production use of this surface. Worth opening them now so they don't
get lost.

genro:
Pushed `5f2377615` addressing the review.

**Changes in this commit:**

1. `_is_known_dtype()` predicate replaces the prior try/except around `_validate_dtype` in `_collect_columns`. Same effect (unknown dtypes are silently dropped from the schema), but the control flow is intentional rather than exception-driven. `_validate_dtype` retains its strict behavior for callers that want it.
2. `openapi_schema` docstring now lists the exact dtype codes that survive the filter and states that any other dtype is omitted. Model authors who want to expose a custom dtype know where to look.
3. `ep_openapi` bearer comparison switched to `hmac.compare_digest`.
4. Module docstrings updated:
   - `ep_sourcerer`: documented as an internal operations endpoint with a custodied / revocable bearer. `partition_kwargs` pass-through is intentional for diagnostic work, not a security gap — the consumer is the Sourcerer toolchain, not external integrators.
   - `ep_openapi`: documented exposure boundary accurately — `openapi=True` exposes the entire fkey-reachable subgraph, `auth_tags` are not enforced, and the table flag should be flipped only when the whole reachable subgraph is intended to be readable.

**Not addressed in this PR (intentional):**

- **`partition_kwargs` removal from `ep_sourcerer`**: kept as feature for the diagnostic use case. The bearer is held by the operations team and revocable; impersonating partition envs is part of the operator toolbox.
- **`logger.warning` on dtype-skip**: model typos are a model concern, not an `ApiEngine` concern. Documented the supported dtype set instead so authors can self-check.

**Follow-up issues to open** (blocking for any AI/chat production use of `ep_openapi`, not blocking for this merge):

- Relation-traversal guard in `ApiEngine` (parse `$@<rel>` segments in fragments, reject targets that are not `openapi=True`). This is the real exposure widening highlighted in the review.
- `acting_user` parameter on `ApiEngine.run_query` + on-behalf-of identity propagation (HMAC-signed header for v1, JWT for v2).
- Governance: where does end-user identity come from when an AI service calls on behalf of a human (in-app session vs external IdP).

**Quality items deferred:**

- `exposed_column_names` / `_build_openapi_table_schema` shared predicate.
- Server-side logging on `ep_openapi`.
- Request body size cap.
- `__error__` sentinel cleanup.
- Test coverage for `_search` / `exposed_column_names` / spec shape.

Tests are still green (79/79 in `tests/app/test_api_engine.py`), flake8 clean on the three touched files.

genro (commented):
Structured reply to the review. I'm grouping by the sections you used so each point keeps its context. Decisions are split into three buckets: **done in this PR**, **deferred to phase 2 (after #803)**, **out of scope / not our job / left as-is**.

---

## §1 — Authorization model

### 1.1 GenroPy auth tags not consulted
**Deferred to phase 2.** The fix is `ApiEngine.run_query` (and column collection) reading `user_tags` from `db.currentEnv['userTags']` — same convention already used by `gnrdbo.py:675` and `gnrsql/write.py:297-300`. Endpoints populate the env from the bearer record (`api_token.auth_tags` once #803 lands). The match primitive is `app.checkResourcePermission(rule, user_tags)` — no need to bring in `tags_match` from toolbox.

`ApiEngine` exposes the lever; the endpoint decides whether to engage it. `ep_openapi` (external) will engage; `ep_sourcerer` (internal ops) may not. Same separation you proposed in 1.3.

### 1.2 Column `auth_tags='hr_only'` + `openapi=True`
**Deferred to phase 2.** Direct consequence of 1.1, solved by the same env-driven approach.

### 1.3 `checkPermissions` opt-in
**Kept opt-in.** `ApiEngine` provides the security primitives; the calling endpoint sets the policy. The endpoint that wants strictness passes `checkPermissions=True`; an internal diagnostic endpoint can omit it. Same philosophy as `partition_kwargs` in §3.

---

## §2 — Relation traversal via `@<rel>.<col>`

(Sidebar: the syntax is `@rel.col`, not `$@rel.col` — `$` prefixes columns, `@` prefixes relations. Two distinct tokens. The risk you describe is real either way.)

### 2.1 Direct exfiltration via `_search` body, 2.2 blind via `where`, 2.3 ordering-based via `order_by`
**All deferred to phase 2 after experimentation.** I want a first weak version that we harden after real test pressure: see what kinds of traversals legitimate integrators actually need, then guard accordingly. The `ep_openapi` module docstring already declares the limit explicitly so adopters know what they're flipping when they set `openapi=True`.

The traversal guard, when it lands, belongs in `ApiEngine` as an opt-in utility (`engine.resolve_traversals(...)`) so the endpoint picks it up but the engine doesn't impose policy.

### 2.4 Spec advertises `x-fkey` / `x-relation-path`
**Deferred with §2.1-3.** Removing them now would be security-by-obscurity; we'll reconsider after the traversal guard ships.

---

## §3 — `partition_kwargs` caller-controlled in `ep_sourcerer`

**Not changed, documented.** `ep_sourcerer` is the internal operations endpoint backing the Sourcerer toolchain (MCP, `target_db`). The bearer is the `sourcerer` service token, custodied by the operations team, revocable, and meant for diagnostics. On a partitioned DB the operator needs to scope the env to any tenant to do their job — `partition_kwargs` pass-through is the feature.

`ep_openapi` does not propagate `partition_kwargs`, which is the correct default for an integrator-facing endpoint.

Both choices are now spelled out in the module docstrings of `ep_sourcerer.py` and `ep_openapi.py` so the asymmetry is intentional and visible.

### §3 corrigendum
You wrote "Fix (in-scope for this PR): drop `partition_kwargs`". I'm declining for the reason above; the alternative would have removed a tool the ops team actively uses.

---

## §4 — `acting_user` / on-behalf-of identity

**Entire section deferred to phase 2.** This is the structural problem — the AI/chat use case needs end-user identity propagation (RFC 8693 token exchange or equivalent). #803 introduces `adm.api_token` with `created_by` / `last_used_ts` (machine identity), which is the foundation; the on-behalf-of layer goes on top after #803 lands.

The governance question (where the AI service obtains end-user identity from — in-app session, external IdP) is also deferred to phase 2 as a separate design issue.

---

## Other quality points

### Silent dtype skip nasconde typo
**Done differently** (commit `5f2377615`). Replaced the `try/except ApiEngineError` with an explicit `_is_known_dtype()` predicate so the control flow is intentional rather than exception-driven. `_validate_dtype` keeps its strict behavior for callers that want it. No `logger.warning` added: a `dtype=True` typo is a model concern, not an `ApiEngine` concern. Instead, the `openapi_schema` docstring now enumerates the supported dtype codes — model authors who flip a column and lose it from the spec know where to look.

### `exposed_column_names` duplicates `_build_openapi_table_schema` logic
**Done** (commit `9d963e7af`). Extracted `_iter_exposed_columns(fullname, readonly)` as a private generator that yields `(colname, info, writable)`. Both `exposed_column_names` (keeps just names) and `_build_openapi_table_schema` (materializes property descriptors) consume it. The two paths can no longer drift.

### `ep_openapi` server-side logging absent
**Not our job, left alone.** Gunicorn / nginx provides transport logging; semantic logging at the webpage level is a framework-wide decision that doesn't belong to this PR.

### `_read_json_body` no body size cap
**Not needed at this layer.** Body size limiting is infrastructure (nginx `client_max_body_size`, gunicorn limits). The endpoint doesn't do traffic shaping.

### `__error__` sentinel
**Left as is.** The ambiguity is hypothetical (legitimate body with literal `__error__` key is an absurd scenario). Pragmatic over pretty.

### Swagger UI behind auth — intentional?
**Confirmed intentional, documented.** The spec carries the DB map; gating it behind the bearer is a defense-in-depth choice for the transitional phase. Adopters get the bearer from the operator. Documented in the new "Exposure boundary" block of the module docstring.

### Bearer comparison constant-time
**Done** (commit `5f2377615`). `hmac.compare_digest` in `_check_bearer`.

### Tighten module docstring (`ep_openapi.py`)
**Done** (commit `5f2377615`). The module now carries an explicit "Exposure boundary" block listing the two real caveats (fkey traversal not gated, `auth_tags` not enforced) and the practical rule for adopters. The previous wording was the most misleading part of the PR; agreed.

---

## Test coverage

### `exposed_column_names`, `_search` POST, `schema()` output, fkey-non-exposed
**Left for later.** Existing unit tests cover the `ApiEngine` motor (79 still green, with the two `_dtype_skipped` cases rewritten to match the new behavior). New paths are exercised E2E via curl in deploy; webpage-level unit tests would need fresh fixture infrastructure that doesn't exist in `tests/` today.

The fkey-non-exposed regression test ties naturally to the traversal guard work in §2 and lands with it.

---

## Summary

**Done in this PR:**
- §1.3, §3 partition_kwargs: motor exposes tools, endpoints set policy — `ep_openapi` is strict, `ep_sourcerer` is operative
- "Silent dtype skip": explicit predicate + docstring listing supported dtypes (`5f2377615`)
- "Exposure boundary" module docstring (`5f2377615`)
- `hmac.compare_digest` (`5f2377615`)
- `_iter_exposed_columns` shared predicate (`9d963e7af`)

**Deferred to phase 2 (after #803):**
- §1.1, §1.2 auth_tags read from `currentEnv['userTags']`
- §2.1-2.4 traversal guard
- §4 acting_user / on-behalf-of (entire section)

**Out of scope / not adopted:**
- `partition_kwargs` removal (intentionally kept for diagnostic use)
- Server-side logging (framework-level concern)
- Body size cap (infrastructure concern)
- `__error__` sentinel cleanup, test coverage expansion (deferred)

Latest tip of `feature/api-engine-adoption`: `9d963e7af`. The 79 `test_api_engine.py` tests still pass; flake8 clean on every touched file.

fporcari (approved):
Thanks for the detailed walk-through. The phase 1 / phase 2 split makes
sense — `_iter_exposed_columns`, `_is_known_dtype`, `hmac.compare_digest`,
and the new "Exposure boundary" docstring on `ep_openapi.py` close the
items I cared most about. The asymmetry between `ep_sourcerer` (ops,
keeps `partition_kwargs`) and `ep_openapi` (integrator-facing, no
`partition_kwargs`) is now intentional and visible, which was the real
gap.

Approving. Flagging the three structural items here so they don't get
lost in the thread when phase 2 starts:

- relation-traversal guard in `ApiEngine` (the §2 `@rel.col` exposure)
- `acting_user` / on-behalf-of identity propagation on `ApiEngine.run_query`
  (depends on #803)
- governance: where the AI service obtains end-user identity from

Two small acks worth revisiting later, not blocking:

- when `partition_kwargs` pass-through fires in `ep_sourcerer`, an audit
  log line (who, when, which fields) closes the observability gap
  should the bearer ever leak
- a single `logger.warning` on the dtype-skip path, once per
  `(table, column)`. The motor is deciding to hide something from the
  caller — silent hiding bites in production

PR #933, opened 2026-06-04
feat(attachmanager): forward uploader kwargs from attachmentGrid

dgpaci:
## Summary

The attachment components (`at_attachmentViewer`, `at_attachmentGrid`, `at_attachmentGallery`) all build their drop uploader internally and offered no way to configure it, and their drop area label was hardcoded English HTML with no translation marker.

This PR covers **two related changes** (the second one landed here through #941, which was merged into this branch):

### 1. `uploader_*` kwargs family

`@extract_kwargs(uploader=True)` on all three components, with `uploader_kwargs` spread onto the dropUploader, so callers can configure it — for example a client side upload size limit:

```python
pane.attachmentGrid(..., uploader_maxsize=2000000)  # 2 MB
```

The dropUploader already supports `maxsize` in JS; this exposes it (and any other `uploader_*` option) through the components. Same shape as the `default_kwargs`/`vpane_kwargs` already used in these methods, and as `upv_uploaderViewerPane` which already had `uploader_kwargs`.

### 2. Customizable, localized drop label

`uploader_label=...` replaces only the *text* of the drop area; the styled `atc_galleryDropArea` wrapper (padding, centering, border) is always preserved. With no `uploader_label` the default is used, and its two fragments now carry the embedded `[!!...]` marker so they reach the localizer.

## Changes

- `attachmanager.py`
  - `@extract_kwargs(uploader=True)` + `**uploader_kwargs` on `at_attachmentViewer`, `at_attachmentGrid` and `at_attachmentGallery`.
  - `_atc_dropAreaLabel()` builds the wrapper for all three, so the default label text is defined once (`DEFAULT_DROPAREA_LABEL`).
  - Translatable `!!` markers on the view column captions, `copyurl` cell shortened to `Copy` with an explicit `width='3em'` (the previous `Copy url` made the button column unnecessarily wide).
- `genro_components.js`
  - `DropUploader.createContent` localized its label only when the **whole** string started with `!!`, so a markup label never reached `_T` and rendered the `[!!...]` markers verbatim. It now delegates to `_T(label,true)`, which already recognises both the leading and the embedded form (`gnrlang.js:64`) — the same idiom `genro_dev.js:246` uses. This fixes every markup label, not only the three call sites here.
  - Localized the "File exceeds size limit" alert (and fixed the `exeeds` typo).
- `localization.xml` — entries for the drop label fragments, the column captions and the upload size alert (Italian translations).

## Backward compatibility

None affected. With no `uploader_*` kwarg the forwarded dict is empty and the uploader behaves exactly as before; with no `uploader_label` the default label is unchanged apart from being translated.

## Test

`gnrpy/tests/app/gnrlocalization_test.py` — two tests against the real core `localization.xml`, no mocks:

- `test_markup_label_fragments_are_localized`: the full drop area markup goes through `AppLocalizer.getTranslation`, and both embedded fragments come back translated (status `OK`, not `NOKEY`). This is the assertion that would have caught the bug: before the localization entries existed it fails with `NOKEY`.
- `test_attachmanager_captions_are_localized`: `Type`, `Open`, `DL`, `Copy`, `File exceeds size limit` and `Error` all resolve in Italian.

Full `gnrpy` suite green (2438 passed, 10 skipped), `flake8 --select=F401,E9,F63,F7,F82` clean.

The `_T(label,true)` change itself has no unit test — there is no JS test infrastructure in the repo. Manual check: set `pane.attachmentGrid(..., uploader_maxsize=2000000)`, drop a file > 2 MB → blocked with the localized alert; the drop area shows the translated label instead of `[!!Drop document here]`.

> Note: there is a sibling untranslated message "Image exceeds size limit" in `genro_widgets.js` (image uploader, different widget) left out of scope.


fporcari (dismissed):
looks fine

genro (commented):
Reviewed end-to-end. The core change — forwarding an `uploader_*` kwargs family from `attachmentGrid` to its drop uploader — is correct and well done. `@extract_kwargs(uploader=True)` guarantees `uploader_kwargs` is always a dict (never `None`, see `gnrdecorator.py:158`), and `label` is `pop`-ed before the `**uploader_kwargs` spread, so there's no double-pass to `dropUploader`. The JS fix is clean: the `exeeds` typo is corrected and both the message and `'Error'` are now localized via `_T`. CI is green on 3.11–3.14.

A few things to address before this can land:

**1. `RELEASE_NOTES.rst` should not be touched in feature PRs.** That file is maintained as part of the release process, not in individual feature branches. Please drop the `gnrpy/RELEASE_NOTES.rst` change from this PR.

**2. Scope no longer matches the description.** The branch has absorbed #941 (commit `59af24423` plus its merge `f92175771`), so the diff now contains two features: the `uploader_*` kwargs forwarding *and* the customizable/localized drop label. The PR summary only describes the first. Either rebase so #941 lands on its own and this PR stays focused on `uploader_*`, or update the description to cover both — whichever matches the intended merge order. As-is, a reviewer reading the summary sees only half the diff. (This is also why the earlier approving review got dismissed: #941 merged into the branch right after it.)

**3. Minor — consistency.** `at_attachmentViewer` and `at_attachmentGallery` now carry the localized `[!!...]` label, but only `attachmentGrid` got the full `uploader_kwargs`/`uploader_label` treatment. Not a blocker, just flagging the asymmetry in case the intent is to expose the same configurability across all three.

The code itself is solid — these are about scope hygiene and the release-notes convention, not correctness.

fporcari (changes requested):
The `uploader_*` kwargs family is the right call, and the way it is done is right too: `@extract_kwargs(uploader=True)` plus a spread is exactly how `default_kwargs`/`vpane_kwargs` already work in this method, so there is nothing new to learn and nothing new to maintain. Widening `attachmentGrid` rather than adding a `maxsize=` special case is the version that will still be useful for the next uploader option. The `label` override on top of it, keeping the `atc_galleryDropArea` wrapper and letting the caller replace only the text, is the correct seam.

The localization half does not work, though, and it fails visibly.

## The `[!!...]` labels never reach the localizer

`gnr.widgets.DropUploader.createContent` localizes its label only when the **whole string** starts with `!!`:

```js
// gnrjs/gnr_d11/js/genro_components.js:4284
if(label && label.startsWith('!!')){
    label = _T(label)
}
dropAreaKw.innerHTML = dropAreaKw.innerHTML || label || '&nbsp;';
```

Every label this PR ships starts with `<div class="atc_galleryDropArea">`, so `startsWith('!!')` is false, `_T` is never called, and the string goes into `innerHTML` untouched. The drop area will render the markers themselves:

```
[!!Drop document here]
[!!or double click to upload]
```

in every locale, English included. That is a regression against the hardcoded English it replaces, on the default path, with no opt-out — it hits `at_attachmentViewer` (line 293), `at_attachmentGrid` (line 358) and `at_attachmentGallery` (line 483). The two entries added to `localization.xml` are dead: nothing ever looks them up.

There is no server-side pass that would catch it either. `TRANSLATION.sub` runs only inside `AppLocalizer.getTranslation` (`gnrpy/gnr/app/gnrlocalization.py:137`), which is reached from `page.localize()` or the `getRemoteTranslation` RPC. Nothing walks widget attributes.

## The fix is one line, and it is in the widget

`_T` already understands the embedded form — `gnrjs/gnr_d11/js/gnrlang.js:64`:

```js
var noLocMarker = (str.search(/^!!|\[!!/)<0);
```

So the gate in `DropUploader` is the only thing in the way:

```js
if(label && (label.startsWith('!!') || label.indexOf('[!!')>=0)){
    label = _T(label)
}
```

That fixes the three call sites in this PR and every future markup label at once, which is why it belongs in the widget rather than in `attachmanager.py`. Worth checking the same gate on the other widgets that accept a markup label before merging.

## Minor, not blocking

`_T("File exceeds size limit")` and `_T('Error')` in the same diff have no entry in `localization.xml`. `LOCREGEXP` does scan `_T("...")`, so a localization build will pick them up — but since the drop-area entries were added by hand in this PR, adding these two in the same place would keep the file consistent.

## Not verified

No browser pass — the rendering above is read from the widget source, not observed. I did not check the `localization.xml` entries behind the new `!!` markers on the column captions (`!!Type`, `!!Open`, `!!DL`, `!!Copy`), nor whether shortening "Copy url" to "Copy" was intended.

dgpaci:
All points addressed in c1681e1feb.

**The `[!!...]` labels never reach the localizer** — confirmed, and fixed in the widget as suggested. I took a slightly different line than the proposed gate: instead of duplicating the marker detection in `DropUploader`, the widget now delegates to `_T(label,true)`. `_T` already decides on both forms (`gnrlang.js:64`), and its `lazy` flag exists precisely for "translate only if marked" — `genro_dev.js:246` already uses it that way. So the detection stays in one place and the fix covers every markup label, which was the point. `DropUploader` was the only widget in `gnrjs/` gating on `startsWith('!!')`; `gnrdomsource.js:2081` already handled the embedded form.

**RELEASE_NOTES.rst** — dropped, the file is back to its `develop` state.

**Scope vs description** — the rebase option is gone: #941 is merged and closed, and its commit is not in `develop`, so the two features are inseparable now. The description is rewritten to cover both.

**`uploader_kwargs` asymmetry** — extended to `at_attachmentViewer` and `at_attachmentGallery`. All three build the same drop area, so the label wrapper is now built by one helper (`_atc_dropAreaLabel`) with the default text in a single constant, instead of the string being repeated three times as before.

**Missing localization entries** — added in the same place as the others: `File exceeds size limit` and `Error` under the `genro_components.js` node, `Type`/`Open`/`Copy` under `attachmanager.py`. `DL` was already there.

**On the "not verified" items** — `Copy url` → `Copy` was intentional (with `width='3em'`): the caption was making the button column much wider than the icon needs. The `!!` captions were indeed missing their entries, though without a visible regression: `getTranslation` returns the plain text on `NOKEY` (`gnrlocalization.py:137`), so they rendered as English rather than as markers.

**Tests** — `gnrpy/tests/app/gnrlocalization_test.py` gets two tests that run against the real core `localization.xml`, no mocks. `test_markup_label_fragments_are_localized` pushes the whole drop area markup through `AppLocalizer.getTranslation` and asserts both embedded fragments come back translated — it fails with `NOKEY` against the previous `localization.xml`, so it is the assertion that would have caught this. Full `gnrpy` suite green (2438 passed, 10 skipped), flake8 clean. The `_T(label,true)` line has no unit test: there is no JS test infrastructure in the repo.

fporcari (approved):
The blocking point is fixed, and fixed better than the gate I proposed. Delegating to `_T(label,true)` instead of duplicating the marker test in `DropUploader` keeps the detection in one place — `gnrlang.js:64` already matches both `^!!` and `[!!` — so every future markup label is covered, not only the three call sites here. That was the argument for putting the fix in the widget, and this version honours it more than mine did.

Three things I checked before agreeing, because the new call is unconditional where the old one was guarded:

- `_T(undefined, true)` returns its argument: `isNullOrBlank(str)` short-circuits before anything else (gnrlang.js:58), so `dropAreaKw.innerHTML = dropAreaKw.innerHTML || label || '&nbsp;'` still falls through as before. Dropping `if(label && ...)` costs nothing.
- an unmarked label returns immediately on `lazy && noLocMarker`, so no existing `dropUploader` caller gains a server round-trip.
- `DropUploader` really was the only `startsWith('!!')` gate in `gnrjs/` — the grep is now empty — so there is no sibling widget left with the same bug.

`_atc_dropAreaLabel` is the right consolidation: the wrapper was the same string three times, and having the caller replace only the inner text keeps `atc_galleryDropArea` in the component's hands rather than the caller's.

Verified: `pytest gnrpy/tests/app/gnrlocalization_test.py` gives 6 passed on c1681e1feb; restoring only `localization.xml` from `develop`, `test_markup_label_fragments_are_localized` and `test_attachmanager_captions_are_localized` both fail with `NOKEY`. The tests pin the entries rather than passing either way.

**My perimeter, said plainly.** The half I found the bug in is the half still without a test: there is no JS test infrastructure in the repo, so `_T(label,true)` is verified by reading `_T` and by the grep above, not by observing a rendered drop area. Nobody has seen this label render.

One behaviour note, not an objection. The default drop label was hardcoded English and cost nothing; it now carries markers, so the first build per locale goes through `getRemoteTranslation` — cached in local storage when the server answers OK, in page-locals otherwise. That is the framework's own idiom and the right trade for a translated UI; it is simply a round-trip that was not there before, on a widget that appears on every form with attachments.

Approving. @genro's three points are answered too: `RELEASE_NOTES.rst` is back to its develop state, the description now covers both features (the rebase option is genuinely gone — #941 is merged and closed), and `uploader_kwargs` reaches all three components.

PR #941, opened 2026-06-10
feat(attachmanager): customizable, localized attachmentGrid drop label

dgpaci:
> **Stacked on #933.** Base is `feature/attachmentgrid-upload-maxsize` so the diff shows only this change. Once #933 merges to `develop`, this PR will be retargeted to `develop`.

## Problem

The `attachmentGrid` drop uploader label was hardcoded English HTML with no translation marker:

```python
label='<div class="atc_galleryDropArea"><div>Drop document here</div><div>or double click</div></div>'
```

Two issues:
1. **Not localized** — the string carried no `!!` marker, so the localizer never processed it and the label stayed fixed in English regardless of locale (unlike the sibling uploader messages such as `!!Drag here or double click to upload`).
2. **Not customizable** — there was no way to change the message. Passing a plain-text label would have replaced the whole HTML, including the styled `atc_galleryDropArea` wrapper, losing padding/centering/border.

## Changes

- **Localization**: the text fragments now carry the embedded `[!!...]` marker (the established Genropy idiom for translating fragments inside a larger string), leaving the surrounding HTML intact. Applied to all three drop-area labels in `attachmanager.py`. Italian translations added to `localization.xml`:
  - `Drop document here` → `Trascina qui il documento`
  - `or double click to upload` → `o fai doppio click per caricare`
- **Customizable label (`at_attachmentGrid`)**: `uploader_label=...` (forwarded via the `uploader_*` mechanism from #933) now replaces **only the text content** of the drop area; the styled `atc_galleryDropArea` wrapper is always preserved, so a custom message keeps the default look:

  ```python
  custom_label = uploader_kwargs.pop('label', None)
  label_content = '<div>%s</div>' % custom_label if custom_label else '<div>[!!Drop document here]</div><div>[!!or double click to upload]</div>'
  dropUploader(label='<div class="atc_galleryDropArea">%s</div>' % label_content, ...)
  ```

  The `pop` removes `label` from `uploader_kwargs` before it is spread onto `dropUploader`, avoiding a duplicate-keyword `TypeError`.
- `RELEASE_NOTES.rst` — entry under New features.

## Test

- Verified the `TRANSLATION` regex in `gnrlocalization.py` extracts both embedded fragments (`Drop document here`, `or double click to upload`) while leaving the HTML tags intact.
- Verified the `pop`/`**uploader_kwargs` evaluation order: with `uploader_label` set the custom text is used and no duplicate `label` reaches `dropUploader`; without it the localized default is used.
- `flake8` passes with zero errors on the modified file; `localization.xml` validates as well-formed XML.

To verify in-app: under an Italian locale, an `attachmentGrid` drop area now reads "Trascina qui il documento / o fai doppio click per caricare"; passing `uploader_label='...'` shows the custom text inside the same styled box.

## Backward compatibility

None. With no `uploader_label` the default (now localized) label is used; English behaviour is unchanged.

### DIFF
--- a/gnrpy/RELEASE_NOTES.rst
+++ b/gnrpy/RELEASE_NOTES.rst
@@ -12,6 +12,17 @@ New features
   passed the forwarded dict is empty and the uploader behaves exactly
   as before.
 
+* ``attachmentGrid`` drop uploader label is now customizable and
+  localized. Passing ``uploader_label=...`` replaces only the text
+  content of the drop area, while the styled ``atc_galleryDropArea``
+  wrapper (padding, centering, border) is always preserved; when
+  omitted, the default ("Drop document here" / "or double click to
+  upload") is used. The default label was previously hardcoded
+  English HTML with no translation marker, so it never reached the
+  localizer; the text fragments now carry the embedded ``[!!...]``
+  marker (Italian translations added). No backward-compatibility
+  impact.
+
 * Connection cleanup is now a single on-event mechanism instead of
   two parallel ones. When a page closes on the client a lottery
   (``cleanup_threshold``, default 5%) gates an attempt to spawn a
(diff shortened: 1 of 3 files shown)


fporcari (approved):
looks fine

PR #942, opened 2026-06-15
Default language fixes and usage in migration context

cgabriel:
During tests with multidb package and localized fields, a problem emerged.

When executing migration in a non-page context (like using 'gnr db migrate'), everything worked fine. 

But when executed in a page context, like using ts_hosting's package UI, which invokes GnrApp's dbstore_align, localized fields didn't get created correctly - only the localized one where created, rather than original+localized. This occurs because when running a page context, current_language is populated, causing issues to the migrator itself. So the dbstore_align now executes in a tempEnv where current_language is set to None, like in a CLI execution context.

the page's 'default_language' property has been moved to gnrsql/db.py, due to the fact that the default language is set in the database configuration, so it's a scope fix. The original page's default_language has been preserved for compatibility, now using the new property. Page's property can be customized in the future to support a different default language in a multi tenant context, where each tenant can set, for example, a different default language.

### DIFF
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -296,7 +296,10 @@ class DbStoresHandler(object):
         
         :param storename: TODO
         :param changes: TODO. """
-        with self.db.tempEnv(storename=storename):
+
+        # current_language is set to None in the migration context like this
+        # in order to properly handle all the localized columns in the models
+        with self.db.tempEnv(storename=storename, current_language=None):
             self.db.syncOrmToSql()
 
 class GnrSqlAppDb(GnrSqlDb):
--- a/gnrpy/gnr/sql/gnrsql/db.py
+++ b/gnrpy/gnr/sql/gnrsql/db.py
@@ -173,6 +173,16 @@ class GnrSqlDb(
             'missedCommit': GnrMissedCommitException,
         }
 
+    @property
+    def default_language(self):
+        """Return the default language for database localization.
+        
+        :returns: the first language code from db languages config, or None
+        """
+        db_languages = self.extra_kw.get('languages')
+        db_languages = db_languages.split(',') if db_languages else []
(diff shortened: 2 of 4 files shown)


fporcari (approved):
looks fine

genro (approved):
Reviewed end-to-end. The fix is correct and addresses the root cause cleanly.

The problem: in a page context both `default_language` and `current_language` are populated in the env, so during `dbstore_align` → `syncOrmToSql` the localized branch in `_get_sqlname` (`columns.py:212`) fired and returned only `<col>_<lang>`, hiding the base column from the migrator — hence only the localized column was created. Running the align under `tempEnv(current_language=None)` neutralizes that branch (the `current_language and current_language != default_language` guard goes false), so the migrator sees the base column and creates original + localized, matching the CLI behavior.

Moving `default_language` from the page to `GnrSqlDb` is the right scope: the value derives from the db's `extra_kw['languages']` configuration, not from the page. Reading it via `self.dbroot.default_language` in `columns.py` removes the dependency on the per-request env, which is what made the migration context fragile. `extra_kw` is always a dict (`db.py:163`), so the property is safe. The page property is preserved for compatibility and delegates to the db, and the multi-tenant FIXME is a reasonable future hook.

One minor, non-blocking cleanup: after this change nothing reads `default_language` from the env anymore, but `gnrwebpage.py:500` still writes it into `updateEnv(... default_language=self.default_language ...)`. That write now has no readers and could be dropped to round off the scope fix. Fine to leave for a follow-up.

Approving.

cgabriel:
> One minor, non-blocking cleanup: after this change nothing reads `default_language` from the env anymore, but `gnrwebpage.py:500` still writes it into `updateEnv(... default_language=self.default_language ...)`. That write now has no readers and could be dropped to round off the scope fix. Fine to leave for a follow-up.

Thanks for the review. Since it's in the env it can read elsewhere, I've left it for retro-compatibility.

PR #943, opened 2026-06-15
fix: decode psql -l output bytes in listRemoteDatabases

fporcari:
## Summary

`PostgresSqlDbBaseAdapter.listRemoteDatabases` runs `psql -l -t` over SSH via
`subprocess.Popen(..., stdout=subprocess.PIPE)` and reads the result with
`ps.stdout.read()`.

Without `text=True`/`encoding=...`, on Python 3 `read()` returns a **`bytes`**
object. The original code did `res.split('\n')` (a `str` separator) on those
bytes, which raises:

```
TypeError: a bytes-like object is required, not 'str'
```

as soon as the output is non-empty. The method was therefore broken for any
real remote database; the `if not res: return []` guard only masked it when the
output was empty (`b''` is falsy).

## Fix

Decode the bytes before splitting:

```python
for dbr in res.decode('utf-8', errors='replace').split('\n'):
```

- **UTF-8** because PostgreSQL database names can contain non-ASCII characters
  and that is `psql`'s default output encoding in modern setups.
- **`errors='replace'`** to degrade gracefully on an undecodable byte instead of
  crashing the whole listing.

## Tests

Added `TestPostgresAdapterListRemoteDatabases` in
`gnrpy/tests/sql/test_gnrbaseadapter.py`. A `FakePopen` backed by `io.BytesIO`
reproduces the real condition (the pipe yields bytes); only the external
dependency (`subprocess.Popen`/SSH) is mocked, the parsing is exercised for real:

- `test_parses_bytes_output` — bytes output including a non-ASCII db name
  (`caffè_db`); verifies UTF-8 decoding and the ssh command.
- `test_empty_output` — covers the `if not res` branch.

## Checks

- `flake8` on both files: 0 errors.
- `pytest gnrpy/tests/sql/test_gnrbaseadapter.py`: 38 passed.

> Note: a pre-existing, unrelated failure in `core/gnrlist_test.py::TestCsvReader::test_auto_dialect`
> (CSV dialect/quoting) is present on `develop` and is out of scope for this PR.

### DIFF
--- a/gnrpy/gnr/sql/adapters/_gnrbasepostgresadapter.py
+++ b/gnrpy/gnr/sql/adapters/_gnrbasepostgresadapter.py
@@ -477,7 +477,7 @@ class PostgresSqlDbBaseAdapter(SqlDbBaseAdapter):
         result = []
         if not res:
             return []
-        for dbr in res.split('\n'):
+        for dbr in res.decode('utf-8', errors='replace').split('\n'):
             dbname = dbr.split('|')[0].strip()
             if dbname:
                 result.append(dbname)
--- a/gnrpy/gnr/sql/adapters/_gnrbasepostgresadapter.py
+++ b/gnrpy/gnr/sql/adapters/_gnrbasepostgresadapter.py
@@ -468,16 +468,12 @@ class PostgresSqlDbBaseAdapter(SqlDbBaseAdapter):
                         password=source_dbpassword or 'postgres',
                         host = source_dbhost or 'localhost',
                         port = source_dbport or '5432')
-        ps = subprocess.Popen((
+        cmd_output = subprocess.run((
             'ssh','%s@%s' %(source_ssh_user,source_ssh_host),
             '-C', 'psql','-l','-t', "user=%(user)s password=%(password)s host=%(host)s port=%(port)s" %srcdb
-            ),stdout=subprocess.PIPE)
-        res = ps.stdout.read()
-        ps.wait()
+            ),capture_output=True, encoding='utf-8', errors='replace', check=True)
         result = []
-        if not res:
-            return []
-        for dbr in res.decode('utf-8', errors='replace').split('\n'):
+        for dbr in cmd_output.stdout.split('\n'):
             dbname = dbr.split('|')[0].strip()
             if dbname:
                 result.append(dbname)
--- a/gnrpy/tests/sql/test_gnrbaseadapter.py
+++ b/gnrpy/tests/sql/test_gnrbaseadapter.py
@@ -1,5 +1,8 @@
+import io
+
 import pytest
 from gnr.sql.adapters import _gnrbaseadapter as ba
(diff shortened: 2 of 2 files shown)


gnrpy/gnr/sql/adapters/_gnrbasepostgresadapter.py
```diff
@@ -477,7 +477,7 @@ def listRemoteDatabases(self,source_ssh_host=None,source_ssh_user=None,
         result = []
         if not res:
             return []
-        for dbr in res.split('\n'):
+        for dbr in res.decode('utf-8', errors='replace').split('\n'):
             dbname = dbr.split('|')[0].strip()
             if dbname:
                 result.append(dbname)
```
cgabriel: I suggest to avoid all the complexity of a pipe here, since we don't read the output in streaming nor we interact with the command itself. All the above code could be simplified and be more robust with something like this:

```
cmd_output = subprocess.run(
    ['ssh',f'{source_ssh_user}@{source_ssh_host}',
      '-C', 'psql','-l','-t', "user={user} password={password} host={host} port={port}s".format(**srcdb)],
    capture_output=True,
    text=True,
    check=True
)

result = [dbr.split("|")[0].strip() for dbr in cmd_output.stdout.split('\n')]
return result
```

cgabriel (changes requested):
I'd simplify the code and avoid pipe complexity for the specific case, no streaming nor interaction with the executed subprocess.

gnrpy/gnr/sql/adapters/_gnrbasepostgresadapter.py
fporcari: Thanks — adopted `subprocess.run` in 4853f5604. You're right that the pipe was unnecessary here since we neither stream the output nor interact with the process.

I deviated from the snippet in three spots, all to preserve the original intent of the fix:

1. **`encoding='utf-8'` instead of `text=True`.** `text=True` decodes via `locale.getpreferredencoding()`, so on a non-UTF-8 locale (e.g. `it_IT`) `psql`'s UTF-8 output would raise `UnicodeDecodeError` — exactly the class of bug this PR targets (and that the previous commit on this branch already hit). Explicit `utf-8` + `errors='replace'` keeps decoding locale-independent.
2. **Kept the `if dbname` filter.** `psql -t` emits a trailing newline (and blank lines), so `split('\n')` yields empty entries; a plain comprehension would leak a trailing `''`.
3. **Kept the connection string in `%`-format**, consistent with the surrounding `dbImport`/`dbExternalImport` code (also sidesteps the stray trailing `s` in the `port=` token in the snippet).

`check=True` adopted as suggested, so ssh/psql failures now surface instead of silently returning `[]`. Tests were updated to mock `subprocess.run` instead of `Popen`, asserting the parsing, blank-line filtering, the ssh argv, and the `encoding`/`errors`/`check` kwargs.

cgabriel (approved):
LGTM

PR #944, opened 2026-06-17
Add user xgroup (deployment group) with _DEV_-gated UI and avatar/cookie propagation

fporcari:
## What

Adds an `adm.xgroup` lookup table representing the **deployment group** (Kubernetes-style release channel: `ALPHA` / `CANARY` / `BETA` / `STANDARD` / `LTS`) an application user is associated with, and relates each `adm.user` to it via a new `xgroup` column. The value is propagated to the avatar, the rootenv (and therefore the DB env), and the connection cookie, and is exposed in the user backoffice UI — everywhere gated to the `_DEV_` tag for both setting and viewing.

## Why

Scaffolding for an upcoming flow that will differentiate application behaviour by the "version"/channel a user can access (e.g. surface experimental features only to `CANARY` users).

## Changes

**Data model**
- `adm/model/xgroup.py` (new): lookup table (`pkey='code'`, `lookup=True`) seeded with five mandatory `sysRecord_*` (`ALPHA`/`CANARY`/`BETA`/`STANDARD`/`LTS`).
- `adm/model/user.py`: `xgroup` column → `adm.xgroup.code` (foreignkey, `onDelete='setnull'`, nullable — `null` is treated as `STANDARD` by callers); `batch_assign=dict(tags='_DEV_')`.

**Propagation (avatar → rootenv → cookie)**
- `adm/main.py`: include `xgroup` in the avatar during authentication, so it is available on every request (the avatar is rebuilt from the DB record each time).
- `adm/resources/login.py`: write `xgroup` into the rootenv; this also makes it available as `:env_xgroup` in the DB env for queries, `formulaColumn`s and triggers.
- `gnrwebpage_proxy/connection.py`: persist `xgroup` in the connection cookie, set unconditionally so it is cleared on guest/logout.

**UI (gated `_DEV_`)**
- `resources/tables/user/th_user.py`: `xgroup` cell in the user `View` and field in the `Form`, both `_tags='_DEV_'`.

## Reviewer notes

- **Gating**: the batch-assign tool is restricted with `batch_assign=dict(tags='_DEV_')`; the view cell and form field via `_tags='_DEV_'` (enforced server-side in `gnrwebstruct`, where forbidden nodes are dropped during serialization).
- **Deployment**: run `gnrdbsetup <instance> --upgrade` to create the `adm.xgroup` table and `user.xgroup` column and to seed the sysRecords (via `onDbUpgrade_createSysRecords`).
- **Verification**: `flake8` clean on all changed files; the model builds and the column/relation/sysRecords/`batch_assign` wiring was introspected in memory. Full test suite is green (1866 passed) except `core/gnrlist_test.py::TestCsvReader::test_auto_dialect`, a pre-existing failure unrelated to this change (CSV dialect handling — verified to fail identically on `develop`).

### DIFF
--- a/gnrpy/gnr/web/gnrwebpage_proxy/connection.py
+++ b/gnrpy/gnr/web/gnrwebpage_proxy/connection.py
@@ -160,6 +160,7 @@ class GnrWebConnection(GnrBaseProxy):
         self.user_tags = avatar_dict.get('user_tags')
         self.user_name = avatar_dict.get('user_name')
         self.user_id = avatar_dict.get('user_id')
+        self.cookie_data['xgroup'] = avatar_dict.get('xgroup')
         if avatar:
             self.avatar_extra = avatar.extra_kwargs
 
--- a/projects/gnrcore/packages/adm/main.py
+++ b/projects/gnrcore/packages/adm/main.py
@@ -60,6 +60,7 @@ class Package(GnrDboPackage):
                 else:
                     kwargs['group_code'] = group_code
                 kwargs['main_group_code'] = user_record['group_code']
+                kwargs['xgroup'] = user_record['xgroup']
                 kwargs['avatar_rootpage'] = user_record['avatar_rootpage'] or group_record.get('rootpage')
                 kwargs['locale'] = user_record['locale'] or self.application.config('default?client_locale')
                 kwargs['user_name'] = user_record['fullname'] or user_record['username']
--- a/projects/gnrcore/packages/adm/model/user.py
+++ b/projects/gnrcore/packages/adm/model/user.py
@@ -36,6 +36,9 @@ class Table(object):
         tbl.column('sms_number',name_long='!!Sms Number')
         tbl.column('photo',dtype='P', name_long='!![en]Photo')
         tbl.column('group_code',size=':15',name_long='!!Group').relation('group.code',relation_name='users',mode='foreignkey')
+        tbl.column('xgroup', size=':10', name_long='!!Deployment group',
+                    batch_assign=dict(tags='_DEV_')).relation('adm.xgroup.code',
+                    relation_name='users', mode='foreignkey', onDelete='setnull')
         tbl.column('custom_menu', dtype='X', name_long='!!Custom menu')
         tbl.column('custom_fields', dtype='X', name_long='!!Custom fields')
         tbl.column('avatar_secret_2fa', dtype='T',name_long='!![en]Secret 2fa')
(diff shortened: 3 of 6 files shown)


genro (changes requested):
## Review

Changeset is correct and works; the propagation path is clean. One design
change requested before merge (deployment groups should be a configurable
seed, not a hardcoded constraint) plus a couple of minors. CI green on
3.11–3.14, mergeable, no conflicts — the block is on substance, not mechanics.

### Requested change — groups as a modifiable base seed, reduced to 3

In `model/xgroup.py` the five groups are `@metadata(mandatory=True)` system
records, so they cannot be removed at runtime. That pins the set of deployment
groups in framework code, whereas the goal is that the number and names be
freely configurable per instance/deploy.

The closest precedent, `model/adm/model/language.py`, uses the same shape
(`lookup=True` + `sysRecord_`) **without** `mandatory=True` — the records are
modifiable initial seeds, not constraints. That is the right pattern here.

Two requests:

1. Drop `mandatory=True` from every `sysRecord_`, so the groups become
   modifiable defaults (add / rename / delete from the UI) rather than
   hardcoded.
2. Reduce the base seed from 5 to 3 — `STANDARD` / `BETA` / `CANARY` — aligned
   with the sticky-proxy worker groups (`DEFAULT_GROUP="standard"`, plus `beta`
   and `canary`). Drop `ALPHA` and `LTS` from the seed: they are not user
   deployment channels and have no matching worker group. `STANDARD` is the
   default/fallback.

```python
def sysRecord_STANDARD(self):
    return self.newrecord(code='STANDARD', description='Standard')

def sysRecord_BETA(self):
    return self.newrecord(code='BETA', description='Beta')

def sysRecord_CANARY(self):
    return self.newrecord(code='CANARY', description='Canary')
```

That leaves three sensible, modifiable defaults: add more if you need them,
remove the ones you don't.

### Minor

- `connection.py:163` — `self.cookie_data['xgroup'] = avatar_dict.get('xgroup')`
  is set unconditionally (None clears it on guest/logout), which is correct.
  Please confirm `cookie_data` is already initialised at this point on every
  login path.
- Case convention: the model uses upper-case codes (`STANDARD`), the proxy
  worker groups are lower-case (`standard`); today the proxy bridges this with
  `str(xgroup).lower()`. Fine, but it is an implicit convention — worth a
  one-line comment in the model (or the proxy) so it is not lost.
- `xgroup` size `:10` is consistent between `user.py` and `xgroup.py`. OK.

### Good

- Clean avatar → rootenv → cookie propagation (`main.py`, `login.py`,
  `connection.py`).
- UI gated behind `_DEV_` (`th_user.py`): the field stays hidden in production
  while the feature is still experimental.
- `onDelete='setnull'` on the user→xgroup relation: deleting a group does not
  break users (they fall back). Consistent with making the groups modifiable.

genro (approved):
Changes addressed by 5a2696f4 (xgroup base seed reduced to standard/beta/canary, no longer mandatory). CI green on 3.11-3.14. Approving.

PR #948, opened 2026-06-23
feat: enforce SMTP account config when host is set (#737)

### LINKED ISSUE #737: [FEATURE] email account configuration improvements
### Summary

The email account configuration form and send mail procedure needs improvements

### Problem or Need

In the email account configuration form, the following fields must be mandatory:

smtp_from_address
smtp_username
smtp_password
smtp_port

IF the smtp_host is populated.

The 'smtp_port' and smtp_tls' should be populated by default (smtp_port->587, smtp_tls->True)

Validating the smtp_from_address as an email address is a plus (being impossible to do a full validation, an approximation would be fine)

### Proposed Solution

See above

### Considered Alternatives

_No response_

### Impact and Scope

If not all the fields are configured, smtplib raises a non-detailed error, which makes not easy to understand the issue. If the configuration fields are mandatory, it prevents dark errors.

### Additional Information

_No response_

dgpaci:
## Summary

Improves the email account configuration form so that incomplete SMTP setups can no longer be saved, preventing the opaque `smtplib` errors that previously surfaced only at send time.

Closes #737.

## Changes

**Model — `email.account`** ([account.py](projects/gnrcore/packages/email/model/account.py)):
- `smtp_port` now defaults to `587`
- `smtp_tls` now defaults to `True`
- `smtp_from_address` is validated as an email address (`validate_email=True`), consistent with the existing `smtp_reply_to`

**Form** ([th_account.py](projects/gnrcore/packages/email/resources/tables/account/th_account.py)):
- `smtp_from_address`, `smtp_username`, `smtp_password` and `smtp_port` become mandatory **only when `smtp_host` is populated**, via the idiomatic `validate_notnull='^.smtp_host'` pattern. The `validate_*` attributes are evaluated against the datasource, so the not-null check activates only when the bound value is truthy.

## Backward compatibility

The new defaults apply only to records created after the upgrade; existing accounts are untouched.

## Test scenario

In the email account form:
1. Leave `smtp_host` empty → SMTP fields remain optional, the record saves.
2. Populate `smtp_host` and leave any of `smtp_from_address` / `smtp_username` / `smtp_password` / `smtp_port` empty → the form blocks saving and flags the missing required fields.
3. Enter a malformed `smtp_from_address` → email validation rejects it.
4. Create a new account → `smtp_port` is prefilled with `587` and `smtp_tls` is enabled by default.

## Notes

The change is declarative (column defaults, validations, form widget attributes); the underlying validation/default logic lives in the framework and the conditional not-null is enforced client-side, so no package-level unit test was added.


fporcari (approved):
looks fine

PR #949, opened 2026-06-24
#816 - adm.connection 'ip' column enlarged, and site connectionLog only for non-guest users

### LINKED ISSUE #816: adm.connection: ip field too small for proxy chains + log connections only for authenticated non-guest users
## Context

Two issues in `adm.connection` logging, surfaced by a production error during login:

```
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(15)
```

### Problem 1 — `ip` column too small for proxy chains

In `projects/gnrcore/packages/adm/model/connection.py:13`:

```python
tbl.column('ip', size=':15', name_long='!!Ip number')
```

15 characters fit a single IPv4 address (`255.255.255.255`) but not the value we actually receive when the request passes through a chain of proxies.

`page.request.remote_addr` in that context reflects the contents of `X-Forwarded-For` (or the equivalent proxy-injected header), which — when several proxies/load balancers are in front of the app — becomes a **comma-separated list of IPs** (`client, proxy1, proxy2, ...`). Two IPv4 addresses with the separator already overflow the 15-char limit, so the INSERT in `openConnection()` fails with `StringDataRightTruncation` and the login RPC crashes.

**Proposed fix**: either
- widen `ip` to a comfortably larger size (e.g. `varchar(255)`) to store the full forwarded chain as-is, or
- normalize the value before insert: keep only the first (client) IP in `ip` and, if we want to preserve the full chain, store it in a separate column (`forwarded_for` / `proxy_chain`).

The first option is simpler and preserves diagnostic info; the second is cleaner but requires a model change.

### Problem 2 — connections logged for guest users

In `gnrpy/gnr/web/gnrwebpage.py:934-953`, `doLogin` calls `site.connectionLog('open')` for every successful avatar resolution, **including guest avatars** (the `if guestName:` branch).

This is wrong because:

- Guest \"logins\" are not real authentications — they happen automatically on page access and produce a lot of noise in `adm.connection`.
- The table is meant to track who accessed the system and when; guest rows inflate it without useful information.
- It amplifies the blast radius of Problem 1: guest traffic goes through the same proxy chain and is much more frequent than real logins.

**Proposed fix**: skip `connectionLog('open')` when the avatar is a guest. Condition available via `avatar` (guest avatars have no `user_id`) or via a dedicated flag on the avatar. The corresponding `connectionLog('close')` path also needs to be reviewed so it is only called for connections that were actually opened.

## Scope

- [x] Resize / restructure the `ip` column so it can hold the forwarded chain without truncation
- [x] Skip `connectionLog('open')` for guest avatars in `doLogin`
- [x] Ensure `connectionLog('close')` and `dropExpiredConnections` behave consistently (no orphan closes for rows that were never opened)
- [x] Verify migration runs cleanly on existing databases

## Affected files

- `projects/gnrcore/packages/adm/model/connection.py`
- `gnrpy/gnr/web/gnrwebpage.py:953`
- `gnrpy/gnr/web/gnrwsgisite.py:1626`

cgabriel:
adm.connection 'ip' column enlarged to 255 varchar, to accomodate ip chain

for http frontend forwarders, and maybe v6 in the future.

Migration have been verified.

connectionLog('open') is now executed only if the avatar is not a guest, as requested, although I'm unable to reproduce the case, since it's executed only if the user has successfully logged in, so it should never be a guest anyway. Some way, a connectionLog('close') is executed only if there's a connection_id. dropExpiredConnections method in adm.connection gets the lists of connection, so they exists for sure.

refs #816


fporcari (approved):
looks fine

PR #952, opened 2026-06-25
fix(attachmanager): notice instead of blank iframe for external attachments

dgpaci:
## Problem

`attachmentGrid` shows a preview pane (`region='center'`) that frames the
selected attachment's `fileurl` in an `iframe`. For **foreign/external
attachments** (`is_foreign_document`, `external_url` set, no stored file) the
`iframe` widget blanks "directory" URLs (no file segment, e.g.
`https://host/`) in `initContentHtml` (`domnode.src=''`). The result is a
**blank, confusing preview pane** with no hint that the document is an
external link.

## Change

The reader pane is now a `stackContainer` with two pages:

- **`viewer`** — the existing `iframe` (documentClasses), **unchanged** for
  stored files (PDF, images, …);
- **`notice`** — shown when the selected row is `is_foreign_document`: a short
  message + an **"Open in a new tab"** button (`genro.openBrowserTab`).

The switch is driven by a `dataController` reading `is_foreign_document` from
the selected grid row (already part of the default `AttachManagerViewBase`
hidden columns).

## Backward compatibility

No issues. Stored files render in the `viewer` page exactly as before. Custom
view resources that don't expose `is_foreign_document` fall back to `viewer`
(current behaviour, blank for bare-domain links — no regression).

IT translations added to `localization.xml`; entry added to
`gnrpy/RELEASE_NOTES.rst`.

## Test scenario

1. On any `attachmentGrid`, add an external-link attachment (record with
   `external_url` set, `is_foreign_document=true`, no `filepath`).
2. Select it in the grid → the preview pane shows the notice + "Open in a new
   tab" button instead of a blank iframe.
3. Select a normal stored file → renders in the viewer as before.

## Screenshot
Screenshot 2026-06-25 alle 10 33 54


gnrpy/RELEASE_NOTES.rst
cgabriel, line 1: You're adding release notes to a previous release.

cgabriel (changes requested):

gnrpy/RELEASE_NOTES.rst
dgpaci: Sorry, RELEASE NOTES update reverted, won't happen again

fporcari (approved):
Reviewed the diff — well-scoped and low-risk.

- The stored-file path is unchanged: the same `iframe` bound to the same `^.reader_url` datapath, now hosted inside a standard `stackContainer`/`contentPane`. Since these are layout containers, the 100% height/width chain from the `region='center'` pane still resolves, so normal previews render as before.
- The switch is data-driven on `is_foreign_document`, which is already part of the default `attachmentGrid` columns (`attachmanager.py:63` — `$fileurl,$is_foreign_document,$external_url,$filepath`), so the `dataController` resolves on the base view. Custom resources that don't expose it fall back to `viewer`, i.e. current behaviour with no regression.
- `genro.openBrowserTab(url,{target:"_blank"})` matches the existing API (`genro.js:2318`) and its current usages.
- IT translations and RELEASE_NOTES updated.

LGTM.

---

cgabriel (approved):

PR #953, opened 2026-06-25
Support for SSL in async server for local development.

cgabriel:
if the local dev is executed with SSL support (ex. to debug passkeys), the async server must run in ssl mode as well, using the same certificates. This simple fix add the relevant SSL options to `gnr web async` command line


genro (approved):
SSL options (cert/key) correctly propagated to the spawned `gnr web async` subprocess, so the async server runs in SSL mode alongside the main one (needed for local passkey debugging). No-op when SSL is off — fully backward compatible. CI green on 3.11–3.14.

PR #957, opened 2026-06-26
fix: restore grid vertical scrollbar when view uses columnsets

### LINKED ISSUE #955: [BUG] Grid vertical scrollbar disappears when the view uses columnsets
## Problem Description

Since the new theme was introduced (penultimate release), a TableHandler grid whose view defines **columnsets** (grouped, multi-row header) no longer shows the **vertical scrollbar**. When the number of rows exceeds the available height, the rows overflow the container instead of scrolling internally.

The exact same view **without** columnsets behaves correctly: the vertical scrollbar appears and the grid scrolls inside its container.

This was verified on a plain, fully generalizable case (a standard `thpage` on a single table), so it is not specific to nested forms, cruscotti, or any particular layout — the only differentiating factor is the presence of columnsets in the view structure.

- **Observed:** with columnsets the grid has no vertical scrollbar and rows overflow.
- **Expected:** the grid should always show the vertical scrollbar and scroll internally, exactly as it does without columnsets.

## Reproduction Steps

Reproduced in the `sandbox` project on table `fatt.fattura`, comparing two `thpage` menu entries on the **same table**:

1. Define two views for `fatt.fattura`:
   - `View` — flat columns (`protocollo`, `data`, `cliente_id`, `totale_imponibile`, `totale_iva`, `totale_fattura`), no columnset.
   - `ViewColumnset` — same columns, but grouped into columnsets (e.g. `Cliente` and `Importi`), producing a two-row header.
2. Add both to the menu as `thpage`:
   ```python
   fatturazione.thpage("Fatture", table="fatt.fattura")
   fatturazione.thpage("Fatture (columnset)", table="fatt.fattura", viewResource="ViewColumnset")
   ```
3. Make sure the table has enough rows to exceed the visible grid height.
4. Open **Fatture** (no columnset): the vertical scrollbar is present and the grid scrolls. ✅
5. Open **Fatture (columnset)**: the vertical scrollbar is **gone** and the rows overflow without scrolling. ❌

Observed result: the only difference between the two pages is the columnsets, and that difference alone removes the vertical scrollbar.

Fully reproducible in sandbox (see steps above). The regression appeared with the new theme: the body/rows height calculation of the grid does not seem to account for the extra header row introduced by columnsets, so the rows pane gets no bounded height and therefore no overflow scrollbar.

## Environment

genropy v26.06.08, browser (web client), new theme.

## Additional Material

Minimal repro: a single `thpage` on `fatt.fattura` with vs without columnsets in the view `th_struct`.

dgpaci:
## Summary

Hotfix for #955: a TableHandler grid whose view defines **columnsets** (grouped, multi-row header) lost its **vertical scrollbar** when the rows overflowed — they spilled out of the container instead of scrolling internally. The same view without columnsets scrolled correctly.

## Root cause

A grid with columnsets/footers gets the `gr_wrap_footers` wrapper, with three panes: the columnset header (`overflow:hidden`, synced to the body), the grid body (`overflow-x:auto`), and the footer (`overflow:auto`). The body and the footer each had their own horizontal scrollbar, kept in bidirectional sync. To hide the body's redundant horizontal bar the CSS used `scrollbar-width:none` and `-ms-overflow-style:none` — which are **not axis-specific** and therefore also hide the vertical scrollbar:

- **Firefox**: vertical scrollbar always hidden.
- **Chrome/Edge ≥ 121**: hidden since Chromium shipped `scrollbar-width` support (Jan 2024).
- **Safari / older Chrome**: unaffected (they ignore `scrollbar-width`).

It only surfaces when a columnset grid has enough rows to overflow vertically. Hiding a single scrollbar axis is not possible cross-browser, so keeping the body horizontally `auto` and trying to hide only its bar cannot work on Firefox (it leaves a double horizontal scrollbar).

## Fix

Make the footer pane a follower — like the columnset header already is — and let the grid body be the single scroll container:

- Footer pane `overflow:auto` → `overflow:hidden`, synced to the body via `scrollLeft`.
- Removed the `gr_wrap_footers .dojoxGrid-scrollbox` special-casing (`overflow-x:auto !important` + the WebKit horizontal rule). The body uses the standard scrollbox and owns both native scrollbars, exactly like a grid without columnsets.
- Horizontal scroll sync is now one-way: body → footer + columnset header (removed the footer → body listener and its scroll ping-pong).

This restores the vertical scrollbar on all browsers, leaves a single horizontal scrollbar, keeps trackpad/wheel horizontal scrolling over the body, and removes the cross-browser scrollbar-hiding hacks. Net −12 lines.

## Trade-off

The horizontal scrollbar now sits at the bottom of the rows (above the footer totals) — the native dojox position, matching every non-columnset grid — instead of below the totals.

## Test

`test_4_columnsets_vertical_scroll` in `test_flex_minwidth.py`: a columnset+footer grid with 120 rows in a constrained height, forcing vertical overflow (the existing `test_1` had only 5 rows and could not expose the bug).

Verified manually on **Firefox** and **Chrome**: vertical scrollbar present, single horizontal scrollbar above the totals, trackpad horizontal scroll works with header/footer in sync, header/footer right-edge alignment correct with the vertical bar visible, no regression on `test_0..test_3`.

Fixes #955

### DIFF
--- a/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
+++ b/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
@@ -344,9 +344,12 @@ textarea.cellEditFixed {
 
 ._common_d11 .gr_wrap_footers .dojoxGrid-scrollbox{
     overflow-x:auto !important;
-    -ms-overflow-style:none;
-    scrollbar-width:none;
 }
+/* Hide only the horizontal scrollbar: the columnset header and footer panes carry their own
+   synced horizontal scroll, so the grid's own horizontal bar is redundant. scrollbar-width and
+   -ms-overflow-style are NOT axis-specific and would also hide the vertical scrollbar, so they
+   must not be used here (doing so removed the vertical scrollbar whenever columnsets/footers
+   were present). */
 ._common_d11 .gr_wrap_footers .dojoxGrid-scrollbox::-webkit-scrollbar:horizontal{
     height:0;
     display:none;
--- a/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
+++ b/gnrjs/gnr_d11/css/gnrbase_css/15_gnr_grid_extended.css
@@ -334,7 +334,7 @@ textarea.cellEditFixed {
     padding-bottom: 1px;
 }
 .gr_footer.gr_scrollbox{
-    overflow: auto;
+    overflow: hidden;        /* follower like the columnset header: no own scrollbar, follows the body via scrollLeft */
     border-top:1px solid var(--grid-border-color);
     background: var(--grid-row-bg);
 }
@@ -342,18 +342,10 @@ textarea.cellEditFixed {
     background: var(--grid-footer-item-bg);
 }
 
-._common_d11 .gr_wrap_footers .dojoxGrid-scrollbox{
-    overflow-x:auto !important;
-}
-/* Hide only the horizontal scrollbar: the columnset header and footer panes carry their own
-   synced horizontal scroll, so the grid's own horizontal bar is redundant. scrollbar-width and
-   -ms-overflow-style are NOT axis-specific and would also hide the vertical scrollbar, so they
-   must not be used here (doing so removed the vertical scrollbar whenever columnsets/footers
-   were present). */
-._common_d11 .gr_wrap_footers .dojoxGrid-scrollbox::-webkit-scrollbar:horizontal{
-    height:0;
-    display:none;
-}
+/* Grids with columnsets/footers use the standard scrollbox: the body owns both scrollbars.
+   The columnset header and footer panes are overflow:hidden followers, synced to the body
+   scrollLeft (see genro_grid.js). No per-axis scrollbar hiding is attempted here: it is not
+   possible cross-browser and was what removed the vertical scrollbar (#955). */
 .gr_noFooter .gr_footer.gr_scrollbox .gr_itemcontainer{
     background: transparent;
     min-height: 16px;
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -487,13 +487,9 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
                                             sourceNode._footers,
                                             colinfo,autoTitle);
             scrollbox = sourceNode._footersNode._value.getNode('scrollbox');
-            genro.dom.setEventListener(scrollbox.domNode,'scroll',function(e){
-                var sn = that.views.views[0].scrollboxNode;
-                sn.scrollLeft = e.target.scrollLeft;
-                if(sourceNode._columnsetsNode){
-                    sourceNode._columnsetsNode._value.getNode('scrollbox').domNode.scrollLeft = e.target.scrollLeft;
-                }
-            });
+            // The footer pane is an overflow:hidden follower (like the columnset header):
+            // only the grid body owns the scrollbars, so the horizontal sync is one-way,
+            // body -> footer + columnset header.
             var footerScrollDom = scrollbox.domNode;
             var mainScrollbox = that.views.views[0].scrollboxNode;
             genro.dom.setEventListener(mainScrollbox,'scroll',function(e){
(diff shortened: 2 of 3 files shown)


fporcari:
## Updated approach: the grid body owns the scrollbars, header/footer are followers

After verifying #955 across browsers, I pushed a revised fix that supersedes the initial CSS-only change. Writing up the full reasoning here, since the approach is different.

### Root cause (corrected)

The issue report guessed the extra columnset header row was throwing off the body height calculation. That is not it: the cause is purely CSS, and the bug is about the **vertical** scrollbar, not the horizontal one.

A grid with columnsets/footers gets the `gr_wrap_footers` wrapper, with three panes:

- **columnset header** (`.gr_columnset.gr_scrollbox`) — `overflow:hidden`, synced to the body via `scrollLeft`;
- **grid body** (`.dojoxGrid-scrollbox`) — `overflow-x:auto`;
- **footer** (`.gr_footer.gr_scrollbox`) — `overflow:auto`.

So the body and the footer each had their own horizontal scrollbar, kept in **bidirectional** sync. To hide the body's redundant horizontal bar, the CSS used `scrollbar-width:none` + `-ms-overflow-style:none` (introduced in `a1a596d41` alongside the trackpad horizontal-scroll feature). Those properties are **not axis-specific**: they hide the *vertical* bar too. That is the regression:

- **Firefox**: vertical bar gone since the rule was introduced;
- **Chrome/Edge ≥ 121**: gone since Chromium shipped `scrollbar-width` support (Jan 2024);
- **Safari / older Chrome**: unaffected (they ignore `scrollbar-width`) — which is why it looked OS-specific and was hard to reproduce.

It also only shows when a columnset grid has **enough rows to overflow vertically**, which the existing test page never did (5 rows) — so it went unnoticed.

### Why the first fix was incomplete

The initial fix dropped the two non-axis-specific properties and kept the WebKit per-axis rule (`::-webkit-scrollbar:horizontal { display:none }`). That restores the vertical bar, but the WebKit rule only applies on Chromium/WebKit. On **Firefox there is no way to hide a single scrollbar axis**, so the body's horizontal bar comes back — and since the footer also has one, you get a **double horizontal scrollbar** at the bottom (confirmed on Firefox).

There is no cross-browser CSS way to hide a single scrollbar axis, so any approach that keeps the body horizontally `auto` and then tries to hide its bar will fail on Firefox.

### Adopted solution

Stop fighting the scrollbar; fix the asymmetry instead. The columnset header was already the correct pattern — an `overflow:hidden` follower synced to the body — so apply the same to the footer:

- **Footer pane**: `overflow:auto` → `overflow:hidden`. Now a follower, like the header.
- **Grid body**: removed the whole `gr_wrap_footers .dojoxGrid-scrollbox` special-casing (`overflow-x:auto !important` + the WebKit rule). The body falls back to the standard scrollbox and owns **both** native scrollbars, exactly like a grid without columnsets.
- **Sync**: now one-way, **body → footer + columnset header** (removed the footer → body listener, which also removes a scroll ping-pong between the two panes).

Result — cross-browser and hack-free:

- vertical scrollbar restored on every browser;
- a single horizontal scrollbar;
- trackpad/wheel horizontal scrolling over the body still works (it is native now, not a hidden-bar trick);
- header and footer stay in sync.

Net diff is −12 lines, mostly deletion.

### Trade-off

The single horizontal scrollbar now sits at the **bottom of the rows, above the footer totals**, rather than below them. This is the native dojox position and is **consistent with every non-columnset grid** — columnset grids used to be the odd ones out. A bar below the totals would be marginally nicer aesthetically, but doing that while keeping trackpad scrolling would need a JS `wheel` handler (`deltaX` → `scrollLeft`) with its own edge cases (momentum, passive listeners) — not worth it for a hotfix.

### Test

Added `test_4_columnsets_vertical_scroll` to `test_flex_minwidth.py`: a columnset+footer grid with 120 rows in a constrained height, so the body actually overflows vertically. The existing `test_1` had only 5 rows and could never expose the missing vertical bar.

### Verification

Manually verified on **Firefox** and **Chrome** via `test_4`:

- vertical scrollbar present;
- single horizontal scrollbar, above the totals;
- trackpad horizontal scroll works, header + footer follow;
- right-edge alignment of header/footer vs body correct with the vertical bar visible;
- no regression on `test_0..test_3`.

### Commits

- `75e9e5cca` — initial CSS-only fix (superseded).
- `9dcc31407` — regression test page (`test_4`).
- `8bc934a6a` — the followers approach described above.

Suggest **squashing on merge**, since `8bc934a6a` supersedes `75e9e5cca`.

---

@dgpaci this reworks your original fix fairly substantially — could you review the new approach when you get a chance? Happy to walk through any of it.

fporcari (approved):
Approved. Verified manually on Firefox and Chrome via `test_4`: vertical scrollbar restored, a single horizontal scrollbar above the totals, trackpad horizontal scroll working with header and footer in sync, and correct right-edge alignment with the vertical bar visible. Full root-cause analysis and the rationale for the followers approach are in my comment above: https://github.com/genropy/genropy/pull/957#issuecomment-4847404017

dgpaci:
Tested on my side: the grid vertical scrollbar is correctly restored when the view uses columnsets. This resolves the issue for me — good to proceed with the merge.

PR #958, opened 2026-06-26
removal of external dependency from 'paste'

cgabriel:
removal of external dependency from 'paste', since it's almost abandonware, no activities in the last 2 years and in deep maintenance mode. Since paste in the framework it's only used to update a list of tuple for ETAG, and iter over a file content, I suggest to avoid having this dependency at all, dictated also by its current maintanance state. pasteOrg encourage to move away from it.

Please see: https://github.com/pasteorg/paste


fporcari (approved):
looks fine

PR #964, opened 2026-06-30
feat(adm): static/dynamic notification lists with date window

dgpaci:
## Summary

Reworks the `adm.notification` audience handling so the two alignment paths stay coherent, and adds static/dynamic lists plus an activity window.

### Coherence
Previously the audience was evaluated in two divergent ways:
- **bulk snapshot** (notification triggers → `updateUserNotificationsFromQuery`): considered `tag_rule`, `group_code`, `linked_query`;
- **per-user alignment at login** (`updateGenericNotification`): considered only `all_users` and `tag_rule`.

Both now go through a single audience evaluation on the `notification` table, used by the bulk snapshot (`audienceUserIds`) and by the new per-user `userMatchesAudience`. So `group_code` and `linked_query` are honored on both paths.

The audience is evaluated in two halves, and both entry points pair them:
- `audienceWhere()` builds the SQL pre-selection, with the criteria SQL can express: `group_code` and `linked_query`;
- `tagRuleMatches()` applies the tag rule on the candidate rows, delegating to `checkResourcePermission`. A tag rule is a permission-engine expression — exact tags, `%` wildcards, `AND`/`NOT`, `;` — that no `LIKE` can reproduce: `$auth_tags LIKE '%admin%'` also reaches a user tagged `superadmin`.

The tags it evaluates are `$all_tags`, the column `adm.main.authenticate` reads to build the avatar tags, so a tag inherited from the group counts on both paths.

### Dynamic vs static
- New `dynamic_list` column, **dynamic by default**.
- **Dynamic** (default): newcomers matching the criteria are enrolled incrementally at login. This is how notifications behaved before the column existed, so nothing changes for the rows that predate it: the login alignment tests `dynamic_list IS NOT FALSE`, and a NULL keeps meaning dynamic — no migration needed.
- **Static** (explicit opt-out): the audience is photographed once, at save time. If 1000 users match, those 1000 are frozen — newcomers are never added.

The list is re-photographed only when the audience criteria change (`tag_rule`, `group_code`, `linked_query`, `all_users`). The snapshot drops the pending rows and rebuilds membership from the current population, so firing it on the date window would enrol exactly the newcomers a static list exists to exclude.

### Activity window
- New optional `start_date` / `end_date` columns (NULL = no limit).
- Outside the window a notification neither enrolls new users nor is surfaced to the user (`nextUserNotification` is date-gated).

### Bug fix
A notification with only a `tag_rule` (no `linked_query`) was wrongly flagged `all_users`. Now `all_users` is implied only when no explicit audience criterion is set.

### UI
Form/embed expose `dynamic_list`, `start_date`, `end_date`; the list view also shows `confirmed_count` / `pending_count` (idiomatic `formulaColumn(select=...)`).

### Note on cost
Saving a notification photographs its audience: with `all_users` that is one `adm.user_notification` row per user, written inside the insert transaction. That is the intended price of a snapshot, but it is a synchronous write that scales with `adm.user`, where previously the eager insert only happened with a `linked_query`. When a `tag_rule` is set, `$all_tags` is a pyColumn and costs a query per candidate — on an operation that is already one insert per user.

## Test plan

`projects/gnrcore/packages/adm/tests/test_notification_audience.py` — 15 tests on a throwaway SQLite application, run with `gnr dev tests gnrdevelop adm`. Tags are created through `adm.htag` + `adm.user_tag` the way the framework does, so the records go through the actual triggers.

Static / dynamic:
- static list frozen after the snapshot (newcomer not added);
- dynamic list grows at login;
- a new notification is dynamic unless opted out;
- a notification that predates the column (`dynamic_list` NULL, no snapshot of its own) still enrolls a user logging in.

Tag rule:
- `tag_rule` audience coherent on both the snapshot and the login path;
- a rule on `ZZADMIN` does not reach a user tagged `SUPERZZADMIN` (what a bare `LIKE '%tag%'` does);
- wildcards (`ZZADM%`), exclusions (`A NOT B`) and alternative rules (`A;B`) are honored;
- a tag inherited from the group matches on both paths;
- `all_users` implied only without explicit rules.

Re-snapshot and window:
- changing an audience criterion rebuilds the membership;
- editing only the date window does not re-photograph a static list (pending rows survive, newcomers stay out);
- the date window gates visibility (`nextUserNotification`).

All 15 tests pass; the framework suite is green (2207 passed, 10 skipped); flake8 clean on the modified files.


fporcari:
## Review notes (holding merge — not approving yet)

The approach is solid and the coherence goal (single `audienceWhere()` shared by the bulk snapshot and the per-user login alignment) is well achieved. The new tests are real-DB (no mocks) and the GenroPy API usage checks out. Before merging, though, a few behavioral points are worth confirming, because some of them change the *default* semantics of the core `adm.notification` table for every installation, not just this use case.

### 1. Tag matching semantics change on the login path
`updateGenericNotification` previously matched users via `application.checkResourcePermission(tag_rule, user_tags)` (`gnr/app/gnrapp.py`), which does `like()` pattern matching (wildcards) plus boolean `AND`/`NOT`/`;` logic. Routing the login path through `audienceWhere()` switches it to `$auth_tags LIKE '%tag%'` (substring, OR over comma-split).

Consequence: a notification with `tag_rule='admin'` now also reaches users with `auth_tags='superadmin'` / `'administrator'` (it didn't before), and boolean/wildcard tag expressions are no longer honored at login. This is an intentional unification onto the snapshot semantics, but it's an observable behavior change — please confirm it's acceptable.

### 2. Pre-existing `all_users` / `tag_rule` notifications stop reaching not-yet-enrolled users (no migration)
`updateGenericNotification` now only processes `$dynamic_list IS TRUE`. Notifications created before this change have `dynamic_list` NULL and were previously delivered lazily at every login to any user not yet enrolled. There is no migration in the diff, so after deploy a user who hasn't logged in since such a notification was created will never receive it (already-enrolled users are fine). Do we need an upgrade hook, or are pre-existing notifications considered "spent"?

### 3. Editing only `start_date`/`end_date` re-snapshots a static list
`trigger_onUpdated` calls `updateUserNotificationsFromQuery` on any change to the 7 listed fields, including for static lists; that function does `deleteSelection(... $confirmed IS NOT TRUE)` and rebuilds membership from the *current* population. So extending the window of a static `tag_rule` list deletes pending (unconfirmed) rows, adds newcomers and drops users who no longer match — which contradicts the "static = photographed once at save time, newcomers never added" contract. Re-snapshotting should probably trigger only on audience-criteria changes (`tag_rule,group_code,linked_query,all_users`), not on `start_date`/`end_date`/`dynamic_list`.

### 4. Eager mass materialization at insert for `all_users`
`trigger_onInserted` now always calls `updateUserNotificationsFromQuery`; for `all_users` this inserts one `user_notification` row per user, row-by-row inside the insert transaction (the eager insert was previously limited to `linked_query`). On installations with tens/hundreds of thousands of users, saving an `all_users` notification becomes a heavy synchronous write in the trigger. Scale-dependent, but the worst case (literally every user) is now the default path.

### 5. Minor — dead parameter
`updateGenericNotification(self, user_id=None, user_tags=None)`: `user_tags` is now unused (it only fed `checkResourcePermission`), but `frameindex.py` still passes it. Drop it or mark it deprecated.

### Note on test coverage
The 6 tests cover the new happy path well but don't surface #1–#3: `tag_rule` is tested only with a unique non-substring tag (#1), there's no upgrade-from-existing-data case (#2), and the re-snapshot-on-date-edit case is tested only with `all_users`, where re-adding everyone is benign (#3 wouldn't show for a static `tag_rule` list).

---

### Architectural question
Points #1, #2 and #4 are changes to the *shared* `adm.notification` behavior that affect every project. If the static/dynamic + date-window requirement is specific to your application, would it be cleaner to implement this logic as a **project-level table customization** of `adm.notification` (overriding the triggers / audience methods in your project's `resources/tables/...`) rather than changing the core `gnrcore` semantics for everyone? That would keep core back-compatible and let the specific behavior live where it's actually needed. Happy to discuss which parts (if any) genuinely belong in core vs. in the customization.

fporcari (changes requested):
The core finding of this PR is excellent and the cure is the right one. The audience was genuinely being evaluated two divergent ways — the bulk snapshot honoring `tag_rule`+`group_code`+`linked_query`, the login alignment honoring only `all_users`+`tag_rule` — so a notification restricted by `group_code` worked for existing users and silently not for new ones. Collapsing both paths onto a single `audienceWhere()` is exactly the fix, and it is what makes the static/dynamic distinction meaningful in the first place. The `all_users` bug you found on the way (a `tag_rule`-only notification being flagged for everyone) is real and worth the fix on its own. Tests hit a real database through the actual triggers, which is the right call.

Two things to resolve before merge.

### 1. Existing notifications stop being delivered after the upgrade

`updateGenericNotification` now selects only `$dynamic_list IS TRUE`. On an existing installation the new column is NULL on every row, so no pre-existing notification is dynamic. And they were never snapshotted either: the old `trigger_onInserted` ran `updateUserNotificationsFromQuery` only when `linked_query` was set —

```python
    if record['linked_query']:
        self.updateUserNotificationsFromQuery(record)
```

— so every `all_users` / `tag_rule` notification in production has no `adm.user_notification` rows beyond those created incrementally at login. After this change they are neither frozen-with-a-snapshot nor dynamic, and they disappear for any user who does not already have the link row. Users who never logged in during the notification's lifetime, and any user created from now on, simply never see them.

A migration setting `dynamic_list=True` on existing rows restores the previous semantics exactly (the old behavior *was* dynamic enrollment at login), and makes static an opt-in for new notifications. Defaulting the column to `True` and letting the form opt out would also work, though it flips the default described in the summary.

Worth adding a test for this too: a notification row that predates the column (no snapshot, `dynamic_list` NULL) must still reach a user logging in.

### 2. Tag matching regresses on the login path

The login path used the permission engine:

```python
if n['all_users'] or self.db.application.checkResourcePermission(n['tag_rule'], user_tags):
```

and now goes through `audienceWhere`, i.e. SQL:

```python
tag_conditions.append(f"$auth_tags LIKE :tag_{i}")
selection_kwargs[f'tag_{i}'] = f'%{t}%'
```

`LIKE '%tag%'` is not equivalent to `checkResourcePermission`: it is an unanchored substring match on the concatenated tag string, so `admin` matches a user tagged `superadmin`, and none of the rule syntax the permission engine understands is honored. That LIKE already existed inside `updateUserNotificationsFromQuery`, so the snapshot path is unchanged — but it now becomes the only evaluation at login as well, replacing the branch that was doing it correctly.

Since the whole point of the PR is that both paths must agree, the coherent resolution is to make the *shared* rule the correct one rather than the loose one — e.g. keep `audienceWhere` for the SQL pre-selection and let `userMatchesAudience` additionally apply `checkResourcePermission` for the tag component, or match tags on delimiter boundaries instead of bare substrings. Either way it should be one rule, and it should be the strict one.

### Minor

- `updateGenericNotification(self, user_id=None, user_tags=None)`: `user_tags` is no longer read. If callers still pass it, keep it and mark it unused; otherwise drop it from the signature.
- `trigger_onInserted` now snapshots on every insert, and with `all_users` that means one `adm.user_notification` row per user, synchronously inside the trigger. That is the documented intent of a static list, so no objection — but on an installation with a large `adm.user` it is a heavy insert where previously there was none. Worth a line in the release notes, and possibly worth checking the timing on a realistic user count.

fporcari (changes requested):
`c6455f991` resolves both blocking points from the July review, and resolves them better than the alternatives I suggested.

**Point 1 — existing notifications.** Making dynamic the default and testing `$dynamic_list IS NOT FALSE` on the login path is the cleaner half of what I proposed: a NULL keeps meaning dynamic, so no migration is needed at all and pre-existing notifications keep enrolling users exactly as before. `test_notification_predating_the_column_still_enrols` and `test_dynamic_list_is_the_default` pin it.

**Point 2 — tag matching.** Taking the tag component out of SQL entirely and running `checkResourcePermission` on `$all_tags` on both paths (`notification.py:118-127`) is the right resolution: one rule, and it is the strict one. Using `$all_tags` rather than `$auth_tags` also means a tag inherited from the group still counts, which the previous login path did and a naive fix would have lost — `test_tag_rule_matches_tags_inherited_from_group` covers it. The added coverage for wildcards, exclusions, alternatives and substring non-matching is more than I asked for.

I checked these against the code rather than the commit message.

## One thing left, and it is the other half of my own point 2

The tag rule was made strict; **`group_code` was not**. `audienceWhere` still matches groups with an unanchored LIKE:

```python
group_conditions.append(f"$all_groups LIKE :group_{i}")
selection_kwargs[f'group_{i}'] = f'%{g}%'
```

`notification.py:99`. And `all_groups` (`user.py:54-59`) is a plain comma-join of group codes with no delimiter at the string boundaries:

```
CASE WHEN $group_code IS NOT NULL THEN $group_code ELSE '' END ||
CASE WHEN ... THEN ',' ELSE '' END || ...
```

So a notification restricted to group `admin` also reaches a user whose group is `superadmin`, `admin_ro`, or `nonadmin`. The error direction is the dangerous one: a **false positive**, so the notification is delivered to *more* users than intended — for a group-restricted notification that is disclosure to people outside the audience, not a missed delivery.

Two reasons this matters more now than before the PR:

1. Before, that LIKE lived only in `updateUserNotificationsFromQuery`, so it governed the snapshot alone. It is now inside the shared `audienceWhere`, so it is the rule on **both** paths — the unification propagated the loose matcher to the login path, which is the same shape of regression as the tag one.
2. `group_code` is the criterion the PR's own opening finding is about: "a notification restricted by `group_code` worked for existing users and silently not for new ones". It is now the only audience criterion whose matching is still approximate.

The fix is the boundary match, mirroring what you did for tags:

```python
group_conditions.append(f"',' || $all_groups || ',' LIKE :group_{i}")
selection_kwargs[f'group_{i}'] = f'%,{g},%'
```

**And a test.** Of the 15 notifications the suite inserts, none sets `group_code` — the criterion this PR exists to fix is the one with no audience coverage. A `test_group_code_does_not_match_by_substring`, alongside the tag one that already exists, would close it.

## Minor, non-blocking

`trigger_onUpdated` re-photographing only on a change of the audience criteria is the right call, and the reasoning in the commit message (firing on the date window would enrol exactly the newcomers a static list exists to exclude) is worth having as the comment it now is.

Not verified: I did not run the suite, and I did not measure the eager snapshot at insert against a realistic `adm.user` count — the release-note point from the July review still stands.

PR #966, opened 2026-07-03
fix: replace corner DRAFT ribbon with centered toolbar badge (#934)

### LINKED ISSUE #934: [BUG] DRAFT ribbon overlaps form toolbar buttons (print, export, +, -)
## Problem Description

The diagonal "DRAFT" ribbon shown in the top-right corner of a form when a record is in draft state is positioned with fixed coordinates and does not take the form toolbar into account. As a result, it frequently overlaps the toolbar buttons placed on the right side.

Because the toolbar slot layout is dynamic (the set of buttons differs per form/configuration), the ribbon can end up sitting on top of different buttons depending on the case. In the example below it covers the **print** and **export** buttons; in other layouts it lands on the record navigation buttons (**+**, **-**).

**Observed behavior:** the DRAFT ribbon overlaps and visually covers toolbar buttons (print, export, +, -), making them hard to read and click.

**Expected behavior:** the DRAFT marker should be positioned so it never overlaps the toolbar buttons, regardless of which buttons are present in the toolbar.

Image

### Technical context

- The ribbon is rendered as a CSS `::after` pseudo-element on `.fh_content` in `gnrjs/gnr_d11/css/gnrbase_css/10_gnr_toolbars_formhandler.css` (lines ~54-89).
- It is positioned absolutely with hardcoded offsets, e.g. for the default top-right marker: `top: 12px; right: -35px; transform: rotate(45deg);` and `z-index: 999`.
- The position class is driven by `updateDraftMarker()` in `gnrjs/gnr_d11/js/genro_frm.js` (lines ~1596-1603), defaulting to `tr` (top-right) via the `draftMarker` form property.
- The default `draftMarker = True` is set in `resources/common/th/th_form.py` (line ~219), and the toolbar buttons (`form_print`, `form_save`, `form_delete`, `form_add`, etc.) live in the same toolbar that the ribbon overlaps.

Since the `top`/`right` offsets are fixed and the ribbon's `z-index: 999` puts it above the toolbar, any button rendered in that corner gets covered.

## Reproduction Steps

1. Open any th form whose record can enter a draft state (`draftMarker` enabled, default).
2. Make sure the form toolbar shows buttons on the right side (e.g. print/export and/or record navigation +/-).
3. Put the record in draft state so the DRAFT ribbon appears in the top-right corner.

Observed result: the DRAFT ribbon overlaps the print/export (or +/-) toolbar buttons.

## Environment

genropy develop branch, browser (client-side CSS/JS) — not OS-specific.

## Additional Material

Screenshot: DRAFT ribbon overlapping the print/export buttons in the form toolbar.

fporcari:
## Summary
- The default draft indicator for th forms is now a dashed "Draft" badge centered on the form toolbar (dialog footer for modal forms), replacing the corner ribbon that could overlap toolbar buttons (print, export, +, -)
- The badge is painted below the toolbar buttons and is pointer-events:none, so buttons always stay readable and clickable, even on narrow windows
- CSS variable scoping (--fh-draft-display) turns the badge on from the draft form node and off inside its content, so nested forms only show their own badge
- The label goes through the standard localizer (new form_draftLabel attribute + en_draft entry with it/de/fr translations), falling back to 'DRAFT'; themes can tune it via --form-draft-label / --form-draft-color
- The corner ribbon remains available as an opt-in via draftMarker='tr'/'tl'/'br'/'bl' (or True); non-th frameforms keep the previous default behavior

## Linked Issue
Closes #934

## Test plan
- [x] Manually verified on two instances: single centered badge on the form toolbar, no overlap with toolbar buttons
- [x] Verified buttons paint above the badge when the window is narrowed
- [x] Verified localized label ("Bozza") on an Italian locale instance
- [x] flake8 clean on modified files, localization.xml validated, full test suite passes

### DIFF
--- a/gnrjs/gnr_d11/css/gnrbase_css/10_gnr_toolbars_formhandler.css
+++ b/gnrjs/gnr_d11/css/gnrbase_css/10_gnr_toolbars_formhandler.css
@@ -88,6 +88,40 @@
     transform: rotate(45deg);
 }
 
+/* Centered draft badge on the form bars. The --fh-draft-display variable
+   is switched on by the draft form node and switched off again inside its
+   content, so bars of nested forms only show their own badge. */
+.form_draft{
+    --fh-draft-display: block;
+}
+.form_draft[class*="draft_marker_"]{
+    --fh-draft-display: none;
+}
+.form_draft .fh_content{
+    --fh-draft-display: none;
+}
+.th_form_toolbar, .slotbar_dialog_footer{
+    position: relative;
+    z-index: 0;
+}
+.th_form_toolbar::after, .slotbar_dialog_footer::after{
+    content: var(--form-draft-label, 'DRAFT');
+    display: var(--fh-draft-display, none);
+    position: absolute;
+    left: 50%; top: 50%;
+    transform: translate(-50%, -50%);
+    padding: 1px 10px;
+    font-size: 0.75em;
+    font-weight: 700;
+    letter-spacing: 0.1em;
+    text-transform: uppercase;
+    color: var(--form-draft-color, #c77700);
+    border: 1px dashed var(--form-draft-color, #c77700);
+    border-radius: 9px;
+    pointer-events: none;
+    z-index: -1;
+}
+
 .form_logical_deleted .gnrfieldlabel{
     color: var(--status-error);
 }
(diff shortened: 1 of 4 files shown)


cgabriel (approved):
LGTM

PR #969, opened 2026-07-07
feat(web): gate broadcast db-events at write time (#968)

### LINKED ISSUE #968: Broadcast db-events are collected for all tables and discarded at commit, instead of being gated at write time
## Problem

The live-update mechanism (`broadcast` table attribute) accumulates a db-event for **every** insert/update/delete on **every** table whose `broadcast` attribute is active, regardless of whether any page is subscribed to that table. Subscribers are only checked **at commit time**, where non-subscribed tables are dropped.

Accumulation — `notifyDbEvent`, `gnrpy/gnr/web/gnrwebapp.py:83`:
```python
broadcast = tblobj.attributes.get('broadcast')
if broadcast is not False and broadcast != '*old*':
    dbevents = currentEnv.setdefault(dbeventKey, {})
    r = dict(dbevent=event, pkey=record.get(tblobj.pkey), ...)
    if broadcast and broadcast is not True:
        for field in broadcast.split(','):
            r[field] = self.catalog.asTypedText(record.get(field))   # cost paid even if nobody listens
            ...
    dbevents.setdefault(tblobj.fullname, []).append(r)
```

Filtering — `onDbCommitted`, `gnrpy/gnr/web/gnrwebapp.py:134`:
```python
subscribed_tables = self.site.getSubscribedTables(tables)
if not subscribed_tables:
    return
for table in set(tables).difference(subscribed_tables):
    dbeventsDict.pop(table)      # produced, then thrown away
```

`notifyDbEvent` is invoked once per insert/update/delete (`gnrpy/gnr/app/gnrapp.py:412,425,436` and the `raw_*` variants). For bulk operations (imports, batch jobs) touching many broadcast tables that no page observes, this is pure per-operation overhead — dict allocation, and `asTypedText` serialization for every `broadcast="col,col"` column — all discarded at commit.

## Proposed fix

Invert the decision: gate at write time instead of at commit. In `notifyDbEvent`, return immediately if the table has no subscribers, so the event is never built:

```python
def notifyDbEvent(self, tblobj, record, event, old_record=None, **kwargs):
    currentEnv = self.db.currentEnv
    if tblobj.fullname not in currentEnv.get('subscribed_tables', ()):
        return
    ...   # unchanged
```

The subscribed-tables set is read in-memory from `currentEnv` (O(1) per operation, no register round-trip). It is populated once when the page initializes its db env, right after `clearCurrentEnv()` in the `db` property (`gnrpy/gnr/web/gnrwebpage.py:493`), reading the page's own subscriptions:

```python
page_item = self.site.register.page(self.page_id)
self._db.updateEnv(..., subscribed_tables=page_item['subscribed_tables'])
```

`page(page_id)['subscribed_tables']` is the same path already used by the `hidden_transaction` branch at `gnrwebapp.py:132`. Because `PageRegister.subscribeTable` mutates that set in place (`subscribed_tables.add(table)`), storing a **reference** to the live set makes subscriptions added later within the same RPC visible to the gate automatically.

The commit-time filter (`getSubscribedTables` at `gnrwebapp.py:134-138`) becomes redundant and can be removed.

## Notes

- No behavioral change for the client: the commit-time filter already drops non-subscribed tables today; this only avoids producing-and-discarding.
- `autoCommit()` (`gnrpy/gnr/sql/gnrsql/transactions.py:214-217`) reads the same events dict but does not depend on events from non-subscribed tables — confirmed.

genro:
## Summary

Closes #968.

The live-update mechanism (`broadcast` table attribute) currently accumulates a
db-event for **every** insert/update/delete on **every** broadcast-enabled table,
regardless of whether any page is subscribed. Subscribers are only checked **at
commit time**, where non-subscribed tables are dropped — events are produced and
then discarded.

This inverts the decision: gate at write time. In `notifyDbEvent`, the event is
built only if the table is among the page's subscribed tables, so nothing is
produced for tables nobody observes (no dict allocation, no `asTypedText`
serialization of broadcast columns).

## Changes

- **`gnrwebpage.py`** (`db` property): the page snapshots its `subscribed_tables`
  into the db env when it acquires `currentEnv`, reading from the register for its
  own `page_id` (same path already used by the `hidden_transaction` branch).
- **`gnrwebapp.py`** (`notifyDbEvent`): the db-event is built only when
  `tblobj.fullname` is in the env's `subscribed_tables`. The **audit** path is
  left outside the gate (it does not depend on subscription).
- The commit-time filter in `onDbCommitted` is **kept as a safety guard**
  (double defense), not removed.

## Notes

- No client-side behavioral change: the commit-time filter already dropped
  non-subscribed tables today; this only avoids producing-and-discarding.
- Subscription is a page/SPA-session fact (done at load, lives for hours), so it
  never interleaves with writes within a single transaction — the snapshot taken
  at env init is sufficient.
- `autoCommit()` reads the same events dict but does not depend on events from
  non-subscribed tables.

## Verification

- flake8: clean on both modified files.
- Test suite: web suite passes with no new failures; full `gnrpy` suite passes
  (pre-existing environment-related failures unchanged).
- End-to-end on a real PostgreSQL instance (`adm.userobject`, `broadcast='objtype'`):
  - not subscribed -> no event built;
  - subscribed to another table -> no event built;
  - subscribed to the table -> event built with the broadcast column serialized;
  - `notifyDbEvent` with `old_record` and no subscription does not raise (audit path intact).


gnrpy/gnr/web/gnrwebapp.py
```diff
@@ -92,7 +92,8 @@ def notifyDbEvent(self, tblobj, record, event, old_record=None,**kwargs):
         if not currentEnv.get('env_transaction_id'):
             self.db.updateEnv(env_transaction_id= getUuid())
         broadcast = tblobj.attributes.get('broadcast')
-        if broadcast is not False and broadcast != '*old*':
+        subscribed_tables = currentEnv.get('subscribed_tables') or ()
+        if broadcast is not False and broadcast != '*old*' and tblobj.fullname in subscribed_tables:
```
cgabriel, line 96: Couldn't broadcast be evaluted as None too? Also, double negation decrease readability, why no just "if broadcast and .. " ?

cgabriel (approved):
Verified including test coverage. The siteregister performance changes should have been addressed in a different PR IMHO (easier to link to other issues/PR for other implementation) but I get the point of adding it here.

PR #976, opened 2026-07-09
changed flib's default upload path from site: storage (ephemeral) to home:

### LINKED ISSUE #835: [BUG] flib.item.getUploadPath uses ephemeral 'site:' path instead of persistent 'home:'
### Problem Description

The `getUploadPath` method of the `flib.item` table returns `'site:uploaded_files'`, but `site:` resolves to an ephemeral/disposable area (regenerated/cleaned across deployments), while persistent user-uploaded files should live under `home:`, which is the convention used everywhere else in the framework for the same purpose.

As a result, files uploaded through the `flib.item` table risk being lost on redeploy/cleanup, breaking the `path` / `thumb_path` references stored in the database.

**Observed behavior:** uploaded files are written under `site:uploaded_files`.

**Expected behavior:** uploaded files should be written under `home:uploaded_files`, consistent with the framework default and other packages.

### Reproduction Steps

1. Open `projects/gnrcore/packages/flib/model/item.py` line 31:
   ```python
   def getUploadPath(self):
       return 'site:uploaded_files'
   ```
2. Compare with the framework default in `gnrpy/gnr/web/gnrwebpage_proxy/rpc.py:138`:
   ```python
   uploadPath = 'home:uploaded_files'
   ```
3. Compare with other packages that handle persistent file storage (all use `home:`):
   - `gnrpy/gnr/app/gnrdbo.py:1785,1791,1840` — `'home:%s' % filepath`
   - `projects/gnrcore/packages/docu/resources/preference.py:53` — `upload_folder='home:documentation/img'`
   - `projects/gnrcore/packages/adm/resources/preference.py:80` — placeholder `'home:maintenance'`
   - `projects/gnrcore/packages/adm/model/backup.py:24` — default `'home:maintenance'`

Observed result: `flib.item` is the only place using `site:` for persistent uploads, which is inconsistent and unsafe.

- Module involved: `flib` package, `item` table
- Consumer: `projects/gnrcore/packages/flib/resources/flib.py:143` passes `getUploadPath()` to a `FilePicker`/uploader as `uploader_path`
- Risk: any deployment that treats `site:` as disposable will lose all files registered in `flib.item`

### Environment

All versions / all environments — applies to current `develop` and historical branches.

### Additional Material

Proposed fix: change `flib/model/item.py:31` to:
```python
def getUploadPath(self):
    return 'home:uploaded_files'
```

Migration note: existing installations with files already stored under `site:uploaded_files` will need a one-shot move of the files from `site:` to `home:` (or a release note documenting the change), to keep the `path` column references valid.

cgabriel:
changed flib's default upload path from site: storage (ephemeral) to home:, which can
be this way configured with persistent storage, if needed.

refs #835

### DIFF
--- a/projects/gnrcore/packages/flib/model/item.py
+++ b/projects/gnrcore/packages/flib/model/item.py
@@ -29,4 +29,4 @@ class Table(object):
         return 'flib/items' 
 
     def getUploadPath(self):
-        return 'site:uploaded_files'
+        return 'home:uploaded_files'


fporcari (approved):
Approved.

Verified that the change is safe on existing deployments:

- `home` and `site` default to the same `base_path` (`site.site_static_dir`, `local` implementation) in `BaseStorageHandler._loadAllStorageParameters`, so out of the box nothing moves — the change only makes the location remappable, which is the point of the PR.
- `getUploadPath()` feeds `dropFileFrame(uploader_path=...)`, which resolves through `storageNode`/the storage-service registry. That registry does have a `home` entry (unlike the legacy `site.statics` static-handler registry used by `getStaticPath`), so the prefix resolves correctly.
- `flib.item.path` is persisted per record by `onUploading_flib_uploader`, so files already uploaded under `site:uploaded_files` keep their stored path and stay reachable. No data migration needed; only new uploads land under `home:`.

refs #835

PR #979, opened 2026-07-10
fix: portable AFM metrics loading for pip-installed deployments

### LINKED ISSUE #978: [BUG] calcRowsNumber crashes with FileNotFoundError on pip-installed deployments (afm_widths.json not found)
### Problem Description

`gnr/utils/fonts.py::_load_afm_widths()` resolves the AFM metrics file with a checkout-relative path:

```python
json_path = os.path.join(os.path.dirname(__file__),
    '..', '..', '..', 'resources', 'common', 'fonts', 'afm_widths.json')
```

Climbing three levels from `gnr/utils/` reaches the repo root only in a development checkout (`gnrpy/gnr/utils` → `genropy/resources/...`). On a pip-installed genropy the shared `resources/` directory is packaged as the `gnr.resources` package (`pyproject.toml`: `"gnr.resources" = "../resources"`), so the file actually lives in `site-packages/gnr/resources/common/fonts/afm_widths.json` while the lookup points to `<python-lib>/resources/...`, which does not exist.

Any print calling `calcRowsNumber` (or any code path reaching `gnrstring.stringWidth`) dies with `FileNotFoundError(2)`.

Expected behavior: the metrics file should be resolved through the standard shared-resources mechanism (the `resources` declaration in `environment.xml`, the same resolution used by `resource_name_to_path`), which is correct in both checkout and installed layouts. Moreover, a metrics problem should never take a print down: degrading to approximate widths is always preferable to raising.

This broke invoice printing in a production deployment (dev environments work, every installed deployment breaks) and forced a full revert of the row-height migration on genromed `legacy` (softwellsrl/genromed#44).

### Reproduction Steps

1. Install genropy via pip (no source checkout)
2. Run any print whose `calcRowHeight` uses `self.calcRowsNumber(...)` (row-height methodology introduced by #719/#888)
3. Observed result (production traceback, trimmed):

```
gnrbaghtml.py, lineIterator: rowheight = row_kw.pop('height',None) or self.calcRowHeight()
fattura_stampa.py:57, calcRowHeight: n_rows = max(1, self.calcRowsNumber(...))
FileNotFoundError(2, 'No such file or directory')
```

### Environment

Any pip-installed genropy deployment (production). Development checkouts are not affected.

### Additional Material

Split from #977, which now tracks only the automatic per-cell row-height feature. Follow-up of #719 (AFM exact row height) and #888 (sans-serif baseline).

dgpaci:
## Motivation

`_load_afm_widths()` in `gnr/utils/fonts.py` resolved `afm_widths.json` by climbing three directory levels from the module to the repository root. That layout only exists in a development checkout: on a pip-installed genropy the shared `resources/` directory is packaged as the `gnr.resources` package inside site-packages, so the lookup pointed to a non-existent path and any print calling `calcRowsNumber` crashed with `FileNotFoundError`. This broke invoice printing in a production deployment and forced a revert of the row-height migration on genromed (softwellsrl/genromed#44).

Fixes #978 (split from #977, which now tracks only the automatic row-height feature).

## Solution

The metrics file is now resolved like every other shared resource in the framework, through the `resources` declaration in `environment.xml` — the same `resources:#a.path` lookup used by `resource_name_to_path` in `gnrwsgisite.py`, `gnrresourceloader.py` and `PathResolver`. Since `gnr init` already writes the correct resources path in both checkout and installed layouts, no configuration change is needed.

Resolution order:
1. every path declared in `environment.xml` under `resources` (works in checkout and pip-installed layouts)
2. the previous checkout-relative path, kept as fallback for contexts running without a configured environment (e.g. CI)
3. if no candidate is readable, string widths degrade to approximate values (average char width) with a logged warning instead of raising — a metrics problem can never take a print down again

Missing or unreadable candidate files are skipped in favor of the next one. The `getGnrConfig` import is deferred inside the resolver to break the `gnrstring → fonts → gnrconfig → gnrstring` import cycle.

## Tests

`gnrpy/tests/utils/fonts_test.py`, all end-to-end with real files and real configuration (no mocks), driving the resolution through the public `GENRO_GNRFOLDER` configuration entry point:

- `test_checkout_relative_fallback`: no `resources` declaration → falls back to the checkout-relative path
- `test_config_resources_priority`: metrics declared in `environment.xml` win over the checkout fallback
- `test_config_resources_without_file_skipped`: a declared resources dir without the file is skipped
- `test_invalid_metrics_file_skipped`: a corrupt metrics file is skipped without raising
- `test_installed_layout`: regression test for #978 — module copy loaded from a simulated `site-packages/gnr/utils`, metrics in `site-packages/gnr/resources`, resolved via configuration
- `test_installed_layout_unconfigured_never_raises`: worst case, installed layout with no usable resources → approximate widths, no `FileNotFoundError`

Full test suite on this branch: 1886 passed, 67 skipped, 0 failed. flake8 clean on modified files.

### DIFF
--- a/gnrpy/gnr/utils/fonts.py
+++ b/gnrpy/gnr/utils/fonts.py
@@ -6,28 +6,58 @@
 # Units are 1/1000 of the font size in points.
 # Example: at 10pt, 'A' in Helvetica = 667/1000 * 10 = 6.67 pt wide.
 #
-# Width data is stored in resources/common/fonts/afm_widths.json and loaded
-# at first use.
+# Width data is stored in the shared resources directory
+# (common/fonts/afm_widths.json) and loaded at first use.
 
 import json
 import os
+
+from gnr.core.gnrsys import expandpath
+from gnr.utils import logger
 
 _AFM_WIDTHS = None
 
 _DEFAULT_CHAR_WIDTH = 556  # fallback for unknown chars (avg lowercase Helvetica)
 
 
+def _afm_widths_candidates():
+    """Yield candidate paths of ``afm_widths.json``, most authoritative first.
+
+    The shared ``resources`` directory is declared in environment.xml in both
+    checkout and installed layouts, so the gnr_config lookup works everywhere
+    (same resolution used by ``resource_name_to_path``). The checkout-relative
+    path is kept as fallback for contexts without a configured environment.
+    """
+    # deferred import: gnr.core.gnrconfig imports gnr.core.gnrstring,
+    # which imports this module at load time
+    from gnr.core.gnrconfig import getGnrConfig
+    relative_path = os.path.join('common', 'fonts', 'afm_widths.json')
+    try:
+        environment_xml = getGnrConfig()['gnr.environment_xml']
+    except Exception:  # getGnrConfig raises a bare Exception when unconfigured
+        environment_xml = None
+    if environment_xml and 'resources' in environment_xml:
+        for path in environment_xml.digest('resources:#a.path'):
+            yield expandpath(os.path.join(path, relative_path))
+    yield os.path.normpath(os.path.join(os.path.dirname(__file__),
+                                        '..', '..', '..', 'resources', relative_path))
+
+
 def _load_afm_widths():
     global _AFM_WIDTHS
     if _AFM_WIDTHS is not None:
         return _AFM_WIDTHS
-    json_path = os.path.join(
-        os.path.dirname(__file__),
-        '..', '..', '..', 'resources', 'common', 'fonts', 'afm_widths.json'
-    )
-    json_path = os.path.normpath(json_path)
-    with open(json_path, 'r', encoding='utf-8') as f:
-        _AFM_WIDTHS = json.load(f)
+    for json_path in _afm_widths_candidates():
+        if not os.path.isfile(json_path):
+            continue
+        try:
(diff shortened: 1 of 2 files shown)


dgpaci:
Superseded by #1294, which targets `develop`.

Two reasons for the retarget:

- On `master` the AFM path is strictly opt-in: `calcRowHeight` measures text only when `font_family` is set **and** the report overrides `getRowWrapField()`, whose default returns `None`. No stock print reaches the metrics file, and the one deployment that had opted in (genromed) already reverted its migration. On `develop`, #981 made `auto_row_height` the default, so every grid print now hits the loader on every non-editable install — that is where the fix is actually needed, and it is release-critical there.
- `master` is fully contained in `develop` (`git rev-list --count origin/develop..origin/master` = 0), so the release carries the fix to `master` on its own. Landing it here first would instead force a forward merge with an add/add conflict on `gnrpy/tests/utils/fonts_test.py`, which #981 already added at the same path, plus conflicting hunks in `fonts.py`.

#1294 keeps the resolution order and the never-raise degradation of this PR, and adds a candidate that resolves the metrics inside the `gnr` package: the installed layout then works with no `resources` declaration at all, closing the case where a stale or hand-written `environment.xml` would have silently fallen back to approximate widths. Its tests are merged into the existing `fonts_test.py` instead of a new file.

PR #981, opened 2026-07-10
Automatic per-cell row-height estimation in grid prints

dgpaci:
## Summary
- `BagToHtml.calcRowHeight` now estimates the row height automatically, cell by cell: every visible wrappable column (`white_space` other than `nowrap`) is measured with AFM font metrics against its rendered width — fixed, colspan-extended or elastic residual width, on every sheet — and the row gets `max(lines) * grid_row_height`. Prints without wrappable columns keep the flat behavior; `auto_row_height = False` opts out; explicit `calcRowHeight` overrides and per-row `height` attributes still win.
- Declaring `font_family` on the print class is now sufficient: it flows into `mainLayoutParameters` (so it actually reaches the rendered grid — previously the injected body CSS was beaten by the main layout inline style) and into the measurement. CSS font stacks are resolved to the first name with AFM metrics (`resolve_font_name`), and Helvetica-Narrow metrics are derived from Helvetica (Adobe 0.82 condensing factor) so the default Arial Narrow layout is measured exactly.
- Elastic columns are measured against the actual letterhead content area (`contentAreaWidth()`): with a `headline` letterhead the copies render inside the template center band (`layout.center.left/right` widths), while `copyWidth()` subtracts the `layout.left/right` side widths that only shape the `sidebar` design — measuring against it counted spurious wrap lines and doubled single-line rows.
- Cleanups superseded by the automatic path: dead `getRowWrapField` declaration in `TableScriptToHtml` (never called by the base class) and the character-limit heuristic in `print_gridres`. Also fixes the page-break accounting of subtotal rows, rendered at `grid_row_height` but previously accounted at the data-row height.

## Linked Issue
Closes #977

Follow-up of #719 (AFM exact row height) and #888 (sans-serif baseline).

## Behavior change
Existing prints with `white_space='normal'` columns (including `character_limit` ones) will get taller rows and different pagination — that is the clipped-text defect this PR fixes. Setting `auto_row_height = False` on the print class restores the previous flat behavior. Worth highlighting in the release notes.

## Test plan
- [x] 32 new unit tests: font stack resolver, derived narrow metrics, `font_size_pt`, all `calcRowHeight` paths (flat, wrap, explicit newlines, flex residual width, multi-sheet max, hidden columns, opt-out, no-field columns) and letterhead content-area geometry (`headline` vs `sidebar` vs none)
- [x] Full gnrpy suite: 1919 passed, 70 skipped, 1 pre-existing failure (`gnrlist_test::test_auto_dialect`, fails on clean develop as well)
- [x] flake8 clean on all touched files
- [x] End-to-end: full `mainLoop` print run — the wrapped row gets `lines × grid_row_height`, flat rows unchanged, opt-out restores flat; the estimated elastic column width matches the rendered cell width exactly (161.6mm)
- [x] Field-verified on a real-world application print (multi-column invoice with a `headline` letterhead): estimated flex width 86.41mm vs 85.81mm rendered — single-line rows no longer doubled; without letterhead the geometry is unchanged

### DIFF
--- a/gnrpy/gnr/core/gnrbaghtml.py
+++ b/gnrpy/gnr/core/gnrbaghtml.py
@@ -281,6 +281,7 @@ class BagToHtml(object):
             baseTemplate.pop('next_letterhead')
             baseTemplate.walk(self.fillLetterheadSourceData)
             top_layer =  baseTemplate['#%i' %(len(baseTemplate)-1)]
+        self._letterhead_top_layer = top_layer
         d = self.__dict__
         paper_height = float(d.get('page_height') or top_layer['main.page.height'] or self.paperHeight)
         paper_width = float(d.get('page_width') or top_layer['main.page.width'] or self.paperWidth)
@@ -559,6 +560,23 @@ class BagToHtml(object):
         """TODO"""
         return (self.page_width - self.page_margin_left - self.page_margin_right -\
                 self.page_leftbar_width - self.page_rightbar_width)
+
+    def contentAreaWidth(self):
+        """Return the width of the band the copies are actually rendered into.
+
+        With a ``headline`` letterhead the content goes into the template's
+        center_center cell, whose side bands are ``layout.center.left/right``
+        (see ``letterhead_layer``): the ``layout.left/right`` widths that
+        ``prepareTemplates`` loads into ``page_leftbar/rightbar_width`` only
+        shape the ``sidebar`` design, so ``copyWidth()`` would subtract them
+        even when they take no horizontal space.
+        """
+        top_layer = getattr(self, '_letterhead_top_layer', None)
+        if top_layer and top_layer['main.design'] == 'headline':
+            return (self.page_width - self.page_margin_left - self.page_margin_right
+                    - float(top_layer['layout.center.left?width'] or 0)
+                    - float(top_layer['layout.center.right?width'] or 0))
+        return self.copyWidth()
 
     def lineIterator(self,nodes):
         lastNode = nodes[-1]
@@ -1319,7 +1337,7 @@ class BagToHtml(object):
         if self.grid_width:
             outer_width = self.grid_width
         else:
-            outer_width = self.copyWidth() - layoutPars.get('left', 0) - layoutPars.get('right', 0)
+            outer_width = self.contentAreaWidth() - layoutPars.get('left', 0) - layoutPars.get('right', 0)
         #the nested grid layout loses its left/right offsets and side borders (finalize_layout)
         grid_width = gridPars.get('width') or \
                      (outer_width - gridPars.get('left', 0) - gridPars.get('right', 0) - 2 * border_width)
--- a/gnrpy/gnr/core/gnrbaghtml.py
+++ b/gnrpy/gnr/core/gnrbaghtml.py
@@ -65,6 +65,7 @@ class BagToHtml(object):
     grid_columnsets = {}
     grid_row_height = 4.5
     font_family = None  # font name or CSS stack applied to the main layout and used for row-height estimation
+    main_font_family = None  # font name or CSS stack injected by the application on the whole document (wins over the layout font by CSS specificity)
     auto_row_height = True  # estimate multi-line row heights from wrappable cell content
     grid_cell_text_margin = 1.5  # mm taken from the cell width by the aligned_left/aligned_right content margin
     renderMode = None
@@ -939,7 +940,7 @@ class BagToHtml(object):
         return rowData.get(field_getter or field)
 
     def rowCell(self, field=None, value=None, default=None, locale=None,
-                format=None, mask=None, currency=None,white_space='nowrap',align_class=None,
+                format=None, mask=None, currency=None,white_space=None,align_class=None,
                 content_class=None, totalize=None,**kwargs):
 
         """Allow to get data from record. You can use it in the :meth:`prepareRow` method
@@ -956,6 +957,9 @@ class BagToHtml(object):
         self.currColumn = self.currColumn + 1
         if curr_attr.get('hidden'):
             return
+        #falling back on the column attribute keeps rendering and row-height
+        #estimation (which only sees column attributes) on the same setting
+        white_space = white_space or curr_attr.get('white_space') or 'nowrap'
         if field:
             if callable(field):
                 value = field()
@@ -1263,12 +1267,17 @@ class BagToHtml(object):
     def gridFontMetrics(self):
         """Return ``(font_name, font_size_pt)`` used to measure grid cell text.
 
-        The font family is the one the grid actually renders with (the
-        ``mainLayoutParameters`` one, i.e. ``font_family`` when declared on the
-        print class) resolved to a font with AFM metrics; the size comes from
-        ``gridLayoutParameters``.
+        The font family is the one the grid actually renders with, resolved to
+        a font with AFM metrics; the size comes from ``gridLayoutParameters``.
+        ``main_font_family`` — the font the application imposes on the whole
+        document via CSS — wins over the ``mainLayoutParameters`` one because
+        its selector beats the layout inline style. Fonts without AFM metrics
+        (e.g. per-document custom fonts) resolve to plain Helvetica, wider than
+        the Narrow default: the estimate errs on the tall side, costing paper
+        rather than clipped text.
         """
-        font_family = self.mainLayoutParameters().get('font_family') or self.font_family
+        font_family = self.main_font_family \
+            or self.mainLayoutParameters().get('font_family') or self.font_family
         font_size = self.gridLayoutParameters().get('font_size') or GnrHtmlBuilder.font_size
         return resolve_font_name(font_family), font_size_pt(font_size)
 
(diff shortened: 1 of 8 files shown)


fporcari (commented):
The engineering here is high quality — `_gridRowCells` mirroring the colspan accumulation of `renderGridCell`, `contentAreaWidth()` distinguishing the `headline` band from the `sidebar` insets, the derived Helvetica-Narrow metrics, the `auto_row_height` opt-out, and the subtotal page-break accounting fix (`len(subtotal_rows)*self.grid_row_height`) are all correct and well argued. `resolve_font_name` in particular solves a real footgun: there is already a print in the wild that hit it and worked around it by hand —

```python
# genromed resources/tables/calendario/html_res/app_giornalieri_medico.py:91
#font_family is a css stack, not a mapped AFM font: pass Helvetica
#explicitly to calcRowsNumber (wider than Arial Narrow, never clips)
```

Before merging I would like to settle one point, because I think it makes the estimate systematically wrong on the prints that matter most.

### The measured font is not the rendered font in real prints

`gridFontMetrics()` reads the font from the print class:

```python
font_family = self.mainLayoutParameters().get('font_family') or self.font_family
```

But application prints do not set `font_family`. They set **`main_font_family`**, which does not exist in genropy core at all — it is an application-level attribute used to inject CSS:

```python
# erpy resources/tables/fattura/html_res/fattura.py:387
self.body.style("""
                .pageLayout_layout div{
                    font-family:"%s";
                }
                 """ % self.main_font_family)
```

and in several of those prints it is resolved **per document**, from the record:

```python
# erpy resources/tables/fattura/html_res/fattura.py:47
if self.record['@tipo_fattura_id.font_custom']:
    self.main_font_family = self.record['@tipo_fattura_id.font_custom']
```

The same pattern appears in `erpy` (`fattura`, `ddt_cl`, `ddt_dp`, `ddt_fo`, `ordine_fornitore`, `offerta`) and `erpylight` (`ordine_fornitore`, `offerta`).

Since that CSS selector is more specific than the layout's inline style, it wins over `mainLayoutParameters(font_family=...)` as well. So for those prints:

- rendered font: whatever `main_font_family` says, possibly a per-document custom font;
- measured font: `self.font_family` is None → `'Arial Narrow'` → `resolve_font_name` → `Helvetica-Narrow`, i.e. Helvetica condensed to 82%.

Measuring with a condensed font while rendering with a wider one **under-estimates** the line count. Under-estimation is the bad direction: cells keep `overflow:hidden`, so text still clips, and now the page-break arithmetic in `mainLoop` is driven by heights that are too small — rows can overrun the grid area. That is the same defect class the PR sets out to fix, reintroduced silently on invoices, delivery notes and orders.

Possible resolutions, in order of preference:

1. Have `gridFontMetrics()` also consider the font actually imposed by the document — reading `main_font_family` when present would cover the whole erpy/erpylight family, or better, have the core expose one blessed attribute and let `defineCustomStyles` feed it.
2. Failing a reliable answer, round **up**: when the rendered font cannot be determined with confidence, measure with a wider font (plain Helvetica rather than the Narrow derivation), so the error costs paper instead of clipping text.

### Blast radius, for the release notes

Prints that already override `calcRowHeight` are unaffected (the override still wins) — that covers most of the long-text prints in the wild: `anaci` (2), `pansotti` (4), `gnrextra/answer_report`, `genromed` (2), `erpy/giacenze_data`, `icond/elaborazione_costi` and `elaborazione_preventivo`. Good.

The prints that *will* change are the ones with no override and elastic `white_space='normal'` columns:

| project | print |
|---|---|
| erpy | `registro_iva`, `libro_giornale`, `stampa_righe_movimenti`, `giacenze_dep`, `registro_cespiti`, `stampa_op_mag`, `stampa_prelievo_mag`, `articolo_valore/dettaglio` |
| genromed | `fattura_stampa` |
| frigel_cont | `stampa_fattura` |
| mbe | `base_invoice`, `estimate_invoice` |
| icond | `prima_proventi` |

Accounting registers and invoices. They are exactly the ones suffering the clipping this PR fixes, so the direction is right — which is why getting the measurement font right matters before this lands.

### Two smaller notes

- `mbe`'s invoices build cells imperatively via `self.rowCell(...)` inside `prepareRow` rather than from the grid struct. Worth checking whether `sheetColumnsBag` carries `white_space` on that path — if not, those prints keep the flat behavior and the fix never reaches them.
- `gridMaxRowsCount` runs per row, over every sheet and every column, calling `getGridCellPars` and `toText` for each cell — roughly doubling the per-cell work. On `registro_iva` / `libro_giornale`, which are thousands of rows, that is worth timing before merge.

fporcari (approved):
All three points are addressed, and `2440cb9` resolves the measurement-font one better than either of the alternatives I proposed — it takes both at once. Approving.

### What I verified

**`main_font_family` now drives the measurement, and it is readable at the right moment.** `gridFontMetrics` prefers it over the layout font (`gnrpy/gnr/core/gnrbaghtml.py:1279`), and the ordering works out: `onRecordLoaded()` is called at `gnrpy/gnr/core/gnrbaghtml.py:218`, before `newBuilder()` and therefore before `mainLoop` reaches `calcRowHeight`. So the per-document `font_custom` case in `erpy/resources/tables/fattura/html_res/fattura.py:47` is resolved from the record before anything is measured. Promoting the attribute to `BagToHtml` is the right call too — the applications have been using that name for years and the core now owns it.

**The round-up fallback is real.** `resolve_font_name` defaults to `'Helvetica'` (`gnrpy/gnr/utils/fonts.py:69`), which is wider than the `Helvetica-Narrow` default, so an unidentifiable custom font over-estimates. Error in the safe direction, as intended.

**`rowCell` fallback.** `white_space = white_space or curr_attr.get('white_space') or 'nowrap'` (`gnrpy/gnr/core/gnrbaghtml.py:962`) — explicit argument still wins, default unchanged, and the imperative path now reads the same setting the estimation does.

**Default stays flat.** `(pars.get('white_space') or 'nowrap') == 'nowrap'` (`gnrpy/gnr/core/gnrbaghtml.py:1360`): a column with no `white_space` at all counts as one line, so the vast majority of prints are untouched. That is the conservative default and it is what keeps this change safe.

**Removing `cell_characters_limit` is safe.** It has no consumers outside `resources/common/tables/_default/html_res/print_gridres.py` — no application reads or sets it — so dropping the heuristic together with the attribute leaves nothing dangling.

I did not re-run your benchmark, so the 68 µs/row figure is yours, not independently confirmed. It is well within the noise of a PDF render either way.

### Correction to my own review: the blast-radius table was wrong

You were right about mbe, and the same mistake runs through most of that table. Re-checking every entry against a cross-repo search for `def calcRowHeight`, **13 of the 15 prints I listed as "will change" already override it** and therefore belong in the unaffected column:

| project | print | override |
|---|---|---|
| erpy | `registro_iva` | `resources/tables/iva/html_res/registro_iva.py:332` |
| erpy | `libro_giornale` | `resources/tables/mese/html_res/libro_giornale.py:149` |
| erpy | `stampa_righe_movimenti` | `resources/tables/movimento/html_res/stampa_righe_movimenti.py:176` |
| erpy | `giacenze_dep` | `resources/tables/deposito/html_res/giacenze_dep.py:66` |
| erpy | `registro_cespiti` | `resources/tables/cespite_tipo/html_res/registro_cespiti.py:74` |
| erpy | `stampa_op_mag` | `resources/tables/operazione_magazzino/html_res/stampa_op_mag.py:57` |
| erpy | `stampa_prelievo_mag` | `resources/tables/operazione_magazzino/html_res/stampa_prelievo_mag.py:24` |
| erpy | `articolo_valore/dettaglio` | `resources/tables/_packages/erpy_mag/articolo_valore/html_res/dettaglio.py:34` |
| genromed | `fattura_stampa` | `resources/tables/fattura/html_res/fattura_stampa.py:40` |
| frigel_cont | `stampa_fattura` | `resources/tables/_packages/erpy_fatt/fattura/html_res/stampa_fattura.py:153` |
| mbe | `base_invoice` | `:203` |
| mbe | `estimate_invoice` | `:210` |

Which leaves the actually exposed set much smaller than I made it out to be:

- **`icond/resources/tables/rip_proventi/html_res/prima_proventi.py:27`** — declares `white_space='normal'` on `estremi_documento` and has no `calcRowHeight`. The one application print in the index that changes purely from the new estimation.
- **the generic grid prints via `print_gridres`** — every column with `character_limit` gets `white_space='normal'` at `print_gridres.py:78` (unchanged by this PR), so those rows move from the removed characters-per-line heuristic to AFM measurement. This is the swap the PR is for, and the new path is strictly the more accurate of the two; it is also where the release note belongs, because it applies to the standard print of any table view rather than to a named report.

So the behavior change is narrower and better targeted than my table suggested. Sorry for the noise — it does not change the direction of the PR, but it does change how loudly the release notes need to shout.

### One thing for the release note, and one aside

- The `rowCell` fallback is a *rendering* change as well as a measurement one, independent of `calcRowHeight` overrides: an imperative print that declares `white_space='normal'` on a column and calls `rowCell` without the argument now wraps where it previously did not. Both mbe invoices pass it explicitly, so they are fine, but it is worth one line in the notes since an override of `calcRowHeight` does not shield anyone from it.
- Aside, no action: the legacy `gnrpy/gnr/web/gnrtablescript.py` / `gnrtablescript_new.py` still carry their own copy of the old contract (`RecordToHtmlNew.getRowWrapField` at `gnrtablescript.py:425`, consumed by `calcRowHeight` at `:442`), documented in terms of the `font_family` + `getRowWrapField` mechanism this PR removes from `TableScriptToHtml`. Nothing imports those modules except a smoke test, so it is stale rather than broken — but if anyone ever revives them, the docstring now describes a contract that no longer exists in the live hierarchy.

PR #985, opened 2026-07-11
fix(services): thread-safe resources_dirs init, explicit error for missing implementation

genro:
## What

Fixes a cold-start race that could crash the very first concurrent page dispatch on a fresh `GnrWsgiSite` with a bare `TypeError: 'NoneType' object is not callable` (500). Full analysis in #984.

## Root cause

`site.resources_dirs` was published to the attribute **before** being reversed in place:

```python
self._resources_dirs = list(self.resources.values())   # visible now...
self._resources_dirs.reverse()                          # ...reversed in place after
```

A second thread resolving `site.storage('gnr')` for its own first dispatch iterates the shared list during the `reverse()`, silently skips resource directories, the service-implementation scan comes back empty, and `addService` calls a `None` factory unchecked → opaque `TypeError`, no traceback on the console. Self-heals after the first settle (per-process lazy singletons), so it only ever hits the first concurrent dispatch — invisible to classic deployments, exposed by a pool of freshly spawned ASGI workers under a concurrent ramp-up.

## Changes

- `gnrwsgisite._get_resources_dirs`: build the list locally, reverse, then publish it complete.
- `gnrresourceloader.getResourceList`: defensive copy in the general branch too (the css/js branch already copied) — a caller sharing the list across threads no longer sees torn iterations.
- `services.addService`: raise a clear `GnrException` naming the implementation and service type instead of calling `None`, so any residual race (or a genuinely missing implementation) is diagnosable.
- New regression test `tests/core/services_addservice_test.py` pinning the crash line (real `BaseServiceType` + `ResourceLoader` over the repo resources; no daemon needed).

(1) and (2) each independently remove the race; (3) makes whatever remains diagnosable.

## Testing

- New regression test: red (`TypeError`) before the guard, green after.
- `tests/core` and the service/site neighborhood green.

Fixes #984

### DIFF
--- a/gnrpy/gnr/lib/services/__init__.py
+++ b/gnrpy/gnr/lib/services/__init__.py
@@ -27,7 +27,7 @@
 import os
 from datetime import datetime
 
-from gnr.core.gnrlang import  gnrImport
+from gnr.core.gnrlang import  gnrImport, GnrException
 from gnr.core.gnrbag import Bag
 from gnr.lib import logger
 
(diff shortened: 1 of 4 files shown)


genro:
Reopening against develop (new work goes to develop, master is production). Superseded by a fresh PR from fix/services-lazy-init-race-develop.

PR #990, opened 2026-07-14
Public (non-presigned) url support for storage services

dgpaci:
## Summary
- Add `public_url()` to `StorageNode` and `StorageService`: a stable, unsigned, non-expiring url suitable for long-lived content (statically built HTML, emails, generated PDFs). The base implementation falls back to `url()` for services whose urls are already plain and permanent.
- Implement `public_url()` in the `aws_s3` storage service as a path-style url built against the service endpoint: the custom `endpoint_url` when configured (SeaweedFS, MinIO gateways), the AWS regional endpoint `https://s3.{region_name}.amazonaws.com` otherwise, or the global `https://s3.amazonaws.com` endpoint when no region is set.
- New optional `public_base_url` service parameter (also exposed in the `ServiceParameters` panel): when set, the url becomes `{public_base_url}/{internal_path}`, for CDN or custom domains fronting the bucket.
- `url()` is untouched: presigned urls remain the default behavior, byte-for-byte identical to today. Instead of switching `url()` behavior on a parameter (as originally sketched in the issue), the public url is a separate explicit method — simpler and with zero regression surface.

Note for adopters: the object must be publicly readable (bucket policy on AWS, anonymous read configuration on SeaweedFS); the service only emits the url, it does not grant access.

## Linked Issue
Closes #988

## Test plan
- [x] New unit tests in `gnrpy/tests/web/test_s3_public_url.py` (12 tests) covering: default AWS regional endpoint, missing region fallback, custom endpoint (SeaweedFS/MinIO), `public_base_url` override, `base_path` inclusion, trailing-slash normalization, url stability across calls, `StorageNode` delegation, base-service fallback to `url()`
- [x] Existing storage tests pass (`test_s3_temporary_filename.py`, `gnrstoragehandler_test.py`)
- [x] flake8 clean on all modified files

### DIFF
--- a/gnrpy/gnr/lib/services/storage.py
+++ b/gnrpy/gnr/lib/services/storage.py
@@ -366,6 +366,10 @@ class StorageNode(object):
         """Returns the external url of this file"""
         return self.service.url(self.path, **kwargs)
 
+    def public_url(self, **kwargs):
+        """Returns a stable public url of this file (no signature, no expiration)"""
+        return self.service.public_url(self.path, **kwargs)
+
     def internal_url(self, **kwargs):
         return self.service.internal_url(self.path, **kwargs)
 
@@ -543,6 +547,14 @@ class StorageService(GnrBaseService):
     def url(self,*args, **kwargs):
         """Returns the external url of path"""
         pass
+
+    def public_url(self, *args, **kwargs):
+        """Returns a stable public url of path: no signature, no expiration.
+
+        Services whose urls are already plain and permanent fall back to url();
+        services that sign their urls (e.g. aws_s3) override this method
+        to return an unsigned url suitable for long-lived content."""
+        return self.url(*args, **kwargs)
 
     def symbolic_url(self,*args, **kwargs):
         pass
--- a/gnrpy/tests/web/test_s3_public_url.py
+++ b/gnrpy/tests/web/test_s3_public_url.py
@@ -1,0 +1,131 @@
+import importlib.util
+import os
+
+from gnr.core.gnrbag import Bag
+from gnr.lib.services.storage import StorageNode, StorageService
+
+# The aws_s3 Service lives under projects/gnrcore/ which is not on the
+# default test sys.path.  We load it directly from the file system so the
+# test can run without requiring a full GenroPy site setup.
+_s3_module_path = os.path.join(
+    os.path.dirname(__file__), os.pardir, os.pardir, os.pardir,
+    'projects', 'gnrcore', 'packages', 'sys', 'resources',
+    'services', 'storage', 'aws_s3.py',
+)
+_s3_module_path = os.path.normpath(_s3_module_path)
+
+
+def _get_s3_service_class():
+    """Import Service from the aws_s3 module."""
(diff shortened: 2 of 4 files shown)


cgabriel (dismissed):
Ok for me, I like the public_base_url in service configuration in order to use CDNs as well.

fporcari (changes requested):
The design choice here is the right one and worth calling out: resisting the flag-on-`url()` sketched in #988 and adding an explicit separate method keeps the presigned behavior byte-for-byte identical, so the regression surface really is zero. The fallback to `url()` on the base service is the correct default for handlers whose urls are already permanent, and the endpoint resolution chain (custom endpoint / regional / global) matches how `endpoint_url` is already treated in `__init__`. Tests are real and cover the geometry properly.

Two things to fix before merge, both small.

### 1. `public_base_url` is unreachable in the UI for the main use case

```python
fb.textbox(value='^.public_base_url',lbl='Public Base Url', hidden='^.custom_endpoint?=!#v')
```

`public_base_url` is hidden unless `custom_endpoint` is checked, but per the PR description its purpose is "a CDN or custom domain fronting the bucket" — which is orthogonal to `custom_endpoint` and most relevant precisely on plain AWS (bucket + CloudFront, no custom endpoint). As written, that configuration cannot be entered from the service panel at all; it only works if set directly in `siteconfig.xml` or in the `sys.service` record. The `hidden` should just be dropped from this row:

```python
fb.textbox(value='^.public_base_url',lbl='!!Public Base Url')
```

The gating is correct on `endpoint_url` — that one really is meaningful only with `custom_endpoint` — and replacing `hidden='==!custom', custom='^.custom_endpoint'` with `hidden='^.custom_endpoint?=!#v'` is the right idiom, used consistently elsewhere in the codebase. Good catch on that one.

### 2. The localization entry is inert

The PR adds

```xml
<aws_s3 path="resources/services/storage/aws_s3.py" ext="py">
    <en_public_base_url base="Public Base Url" en="Public Base Url" it="Url pubblico di base"/>
```

but the label is a plain string, `lbl='Public Base Url'`, with no `!!` marker, so the translation is never looked up. Either mark the label (`lbl='!!Public Base Url'`, as the `sourcerer` and `xmltransform` service panels do) or drop the localization block. Marking it is preferable — the Italian translation is already written.

Note that the other labels in this panel (`Aws Secret Access Key`, `Region Name`, …) are unmarked too, so it is pre-existing; no need to convert them all here, just keep the new one consistent with whichever way you go.

fporcari (approved):
Both points addressed in `ad85ce35e`. Verified on the branch head:

```python
fb.textbox(value='^.endpoint_url',lbl='Endpoint Url', hidden='^.custom_endpoint?=!#v')
fb.textbox(value='^.public_base_url',lbl='!!Public Base Url')
```

`public_base_url` is now reachable for the plain AWS + CDN case, the `!!` marker makes the localization entry effective, and the gating on `endpoint_url` — which was correct — is untouched. Nothing else changed.

Approved. The separate-method design remains the right call here: `url()` keeps its presigned behavior byte-for-byte, so the regression surface is zero.

PR #997, opened 2026-07-25
fix(services): thread-safe implementations registry, clear factory errors, safer gnrImport publication

fporcari:
## Summary

Two publication races on the path taken the first time a service is resolved, plus the removal of a duplicated registry builder.

### `BaseServiceType.implementations` was published half-built

The property assigned the cache and then filled it:

```python
if not hasattr(self,'_implementations'):
    self._implementations = {}
    ...
    for d in dirs:
        ...
        self._implementations[implname] = service_class
```

A second thread reaching the property while the first was still scanning found `_implementations` already set — `hasattr` is true from the first statement — and got a partially populated dict, so a service whose module had not been imported yet resolved to the wrong factory (or to the base implementation, silently).

Now the dict is built locally by `_buildImplementations()` and published complete, under an `RLock`, with the classic double-check inside the lock. Reentrant because the build path can reach back into the property.

### `implementations` and `getImplementations()` were near-identical copies

Two ~25-line bodies doing the same directory scan, drifting apart. Both now delegate to `_buildImplementations()`; `getImplementations()` keeps its `(dict, baseImplementation)` return shape for backward compatibility.

While collapsing them, two defects in the scan were fixed: the loop reused the `impl` variable for both the entry name and the full path, and a module exporting no `Service`/`Main` class was cached as `None`, which then satisfied `implementations.get(implementation)` and returned a `None` factory to the caller. Such a module is now skipped with a warning.

### `gnrImport` published path modules before finishing them

Same shape of bug one level down, in `gnr/core/gnrlang.py`: the module object went into the path cache before `exec_module` had run, so a concurrent importer could receive a module whose top-level names did not exist yet. Publication is now first-wins and only after the module is fully executed.

## Relationship to #986

#986 (merged as `7b6234041`) fixed the sibling race on `site.resources_dirs` — one level below this one, on the same first-access path — and added the `service_factory is None` guard in `addService`. This PR was rebased on top of it and **no longer touches the guard**: after review it turned out the "available implementations" list this branch used to add to the message can only ever read `none`, because `getServiceFactory` falls back to `implementations.get(self.baseImplementation)` and therefore returns `None` only when the registry is empty. #986's simpler message is the correct one and is left untouched. `test_add_service_without_implementation_raises` was dropped for the same reason: it duplicated #986's `test_add_service_missing_implementation_raises_clear_error`.

## Test plan

New `gnrpy/tests/core/gnrservices_test.py` — registry contents and `baseImplementation`, the `getImplementations()` compatibility shape, factory resolution and its base fallback, service instantiation, a broken module and a class-less module both being skipped, and `test_concurrent_first_access_is_atomic`: ten threads hitting the property at once against a deliberately slowed build must see one single build and one identical dict.

Extended `gnrpy/tests/core/gnrlang_test.py` — path cache behavior, `avoidDup` returning a fresh module, and `test_gnrImport_concurrent_single_module_identity` for the publication order.

- [x] `tests/core`: 403 passed, 3 skipped. The one failure, `gnrlist_test::TestCsvReader::test_auto_dialect`, is pre-existing — verified failing identically on clean `develop`.
- [x] #986's `tests/core/services_addservice_test.py` still passes on this branch.
- [x] flake8 clean on all four touched files.


cgabriel (approved):
LGTM, following #986 path correctly

PR #999, opened 2026-07-27
use non-volatile storage for export and archive action

cgabriel:
use "home:" storage rather than "site:" storage. Since the action pourpose is to delete and archive data, it's not safe to save the archived data on a volatile storage like "site:".

Changed it to "home:", so it can be configured, if needed, with a custom storage handler like s3 bucket.


fporcari (dismissed):
The intent is right and worth doing: `site:` is ephemeral in containerized deployments, and an archive produced by `archive_and_delete` — which by definition holds data just removed from the database — must not live on volatile storage. But `home:` is not a usable prefix for `getStaticPath`, and as it stands this crashes.

### The blocker

`getStaticPath` does not go through the storage-service registry. It resolves the prefix against the legacy static-handler registry:

```python
# gnrwsgisite.py:815
static_handler = self.getStatic(static_name)      # -> self.statics.get(static_name)
...
dest_path = static_handler.path(*args)
```

`self.statics` is a `Bag` populated by `StaticHandlerManager.addAllStatics()`, which registers one entry per `StaticHandler` subclass found in `gnrstatichandler.py`. The registered prefixes are: `cordova_asset`, `dojo`, `vol`, `cld`, `xvol`, `site`, `pkg`, `rsrc`, `pages`, `gnr`, `conn`, `page`, `temp`, `user`. **There is no `home` static handler.**

A `Bag` returns `None` for a missing key, so `getStatic('home')` is `None` and the very first line of `step_get_dependencies` raises:

```
AttributeError: 'NoneType' object has no attribute 'path'
```

Consistent with this: there is currently not a single `getStaticPath('home:…')` call anywhere in the codebase — this would be the first. The `home` entry does exist in the *storage-service* registry (`BaseStorageHandler._loadAllStorageParameters`), which is why `getStaticUrl` on line 5 of the diff would have worked; the two registries are separate and only `getStaticUrl` uses the one that knows about `home`.

### On the S3 goal in the description

Even with the prefix fixed, remapping `home:` to an S3 handler will not make this batch work: `getStaticPath(..., autocreate=True)` goes through `os.path.exists` / `os.makedirs`, and the batch then writes with plain filesystem operations. If S3 is the actual target, the batch needs to move to the storage API (`storageNode(...)`, `.open()`/`.local_path()`) rather than to a different static prefix.

### Suggested directions

1. Use a `vol:` volume configured for persistence — already served by `VolumesStaticHandler`, no core change needed; or
2. register a `HomeStaticHandler(prefix='home')` in `gnrstatichandler.py` so the two registries agree on the prefix set; or
3. rewrite the archive paths on `site.storageNode(...)`, which resolves `home:` correctly today and is also the only route that could ever reach S3.

Option 1 is the smallest change that keeps the intent intact. Option 3 is the one that delivers what the description promises.

cgabriel:
> 1. Use a `vol:` volume configured for persistence — already served by `VolumesStaticHandler`, no core change needed;

Thank you for the analysis, I was missing the usage in gnrwsgisite of the staticPath. I've choosen to go with option 1, to use vol: handler.

fporcari (approved):
Verified the resolution end to end on the branch, approving.

- `vol` is a registered prefix in the static-handler registry (`VolumesStaticHandler`, `gnrstatichandler.py:184-202`), so `getStaticPath` resolves — the blocker of the previous round is gone. With no `<volumes>` entry configured (there is none anywhere in the repo, and no other `export_archive` reference), `path()` falls back to `site_static_dir/export_archive/...`, byte-identical to the old `site:export_archive` layout: behavior-preserving by default, overridable per deployment. Exactly the intent.
- The URL side round-trips too, which the previous review had not fully verified: `getStaticUrl('vol:...')` goes through `LegacyStorageHandler.makeNode`, whose `_adapt_path` handles the legacy `vol:name:path → name/path` rewrite (`gnrstoragehandler.py:498-501`), and serving `_vol/...` lands on the same file.

Two notes for the record, neither blocking:

1. The PR body still says `home:` — worth updating to `vol:`, since `home:` was the variant that crashed.
2. The S3 goal in the description remains out of reach on this path: `getStaticPath(..., autocreate=True)` is `os.makedirs` plus plain filesystem writes. Pointing the volume at a bucket needs the batch rewritten on `storageNode(...)`; follow-up material, not this PR.

PR #1001, opened 2026-07-28
fix: skip archived userobjects when installing from resources on migrate

### LINKED ISSUE #896: [BUG] db migrate -u fails with UniqueViolation when userobject is archived (logically deleted)
## Problem Description

When running `db migrate -u` to install/sync userobjects from package resources into `adm.userobject`, the migration aborts with a `UniqueViolation` if a userobject with the same identifier has been **archived** (logically deleted) in the database.

Example error:

```
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "adm_userobject_identifier_key"
DETAIL:  Key (identifier)=(anc_base.sociotemplatemancata_immatricolazione) already exists.
```

### Observed behavior

`Table.checkResourceUserObject()` in [projects/gnrcore/packages/adm/model/userobject.py:112](projects/gnrcore/packages/adm/model/userobject.py#L112) walks resource files and, for each `.xml`, calls `self.checkDuplicate(code=..., pkg=..., tbl=..., objtype=...)` to decide whether to insert. `checkDuplicate` (in [gnrpy/gnr/sql/gnrsqltable/crud.py:239](gnrpy/gnr/sql/gnrsqltable/crud.py#L239)) builds a `self.query(...)` without `excludeLogicalDeleted=False`.

Since `adm.userobject` is declared `archivable=True` ([userobject.py:13](projects/gnrcore/packages/adm/model/userobject.py#L13)), the default query **excludes logically deleted rows**. So an archived userobject is invisible to the duplicate check, and the subsequent `self.insert(record)` collides with the unique index on `identifier` (the `sql_value` formula based on `tbl/pkg/objtype/code/private/userid` — see [userobject.py:15](projects/gnrcore/packages/adm/model/userobject.py#L15)), which is enforced at the DB level regardless of `__del_ts`.

### Expected behavior

If a userobject has been archived, the migration should not abort. It was archived deliberately, so the install step should:

- detect that the identifier already exists (including archived rows), and
- skip the insert (or, optionally, log a warning / restore the archived row), but **never** raise a hard error that blocks the whole migrate.

## Reproduction Steps

1. Take any instance with a package that ships an XML userobject resource (e.g. `<pkg>/resources/tables/<table>/userobjects/<objtype>/<code>.xml`).
2. Run `db migrate -u` once to install the userobject. Verify the row exists in `adm.userobject`.
3. Archive the userobject from the UI (or set `__del_ts` on the row) so it becomes logically deleted but the physical row (and its `identifier`) still exists.
4. Run `db migrate -u` again.

Observed result: migrate aborts with `psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint \"adm_userobject_identifier_key\"`.

Expected result: migrate completes successfully; the archived userobject is left as-is (or restored on demand) with at most a warning in the log.

## Suggested Fix

In `Table.checkResourceUserObject()` (or in `checkDuplicate` itself when called for this purpose), include logically deleted rows in the duplicate check, e.g. by passing `excludeLogicalDeleted=False`. If a duplicate is found that is logically deleted, log a warning and skip the insert rather than letting the unique constraint blow up the entire migration.

## Environment

genropy `develop` branch (current). Postgres backend. Reproduces on any instance with archivable userobjects.

dgpaci:
## Summary

Fixes #896.

`db migrate -u` aborted with a `UniqueViolation` on `adm_userobject_identifier_key` when a userobject shipped as a package resource had been **archived** (logically deleted) in the database.

Root cause: `Table.checkResourceUserObject()` uses `checkDuplicate()` to decide whether to insert a userobject from a resource XML. Since `adm.userobject` is `archivable=True`, the query built by `checkDuplicate` excluded logically deleted rows by default, so the archived row was invisible to the check — but its `identifier` still exists physically and is enforced by the unique index, so the subsequent `insert()` blew up the whole migration.

## Changes

- `gnrpy/gnr/sql/gnrsqltable/crud.py` — `checkDuplicate()` accepts a new `excludeLogicalDeleted` parameter (default `True`, i.e. the previous behaviour is unchanged) and forwards it to the query.
- `projects/gnrcore/packages/adm/model/userobject.py` — `checkResourceUserObject()` now performs a second duplicate check including logically deleted rows: if the userobject exists but is archived, it logs a warning and skips the insert instead of raising. Archived userobjects are deliberately left as-is.

## Test case

New test module `gnrpy/tests/sql/test_check_duplicate.py` (real databases, no mocks), following the reproduction steps in the issue:

- `TestCheckDuplicateLogicalDeleted` (SQLite, `test_invoice`): default `checkDuplicate` excludes archived rows; `excludeLogicalDeleted=False` detects them; live rows are detected in both modes.
- `TestUserobjectArchivedDuplicate` (PostgreSQL, since `adm.userobject` formula columns use postgres-only functions): inserts a userobject in `adm.userobject`, archives it by setting `__del_ts`, then verifies the default check no longer sees it (the cause of the UniqueViolation) while the check now used by `checkResourceUserObject` does, preventing the colliding insert.

Full `tests/sql` suite: 1189 passed, 10 skipped. `tests/app` suite: 147 passed. flake8 clean on modified files.

### DIFF
--- a/gnrpy/gnr/sql/gnrsqltable/crud.py
+++ b/gnrpy/gnr/sql/gnrsqltable/crud.py
@@ -237,13 +237,14 @@ class CrudMixin(SqlTableBaseMixin):
         return self.db.adapter.existsRecord(self, record)
 
     def checkDuplicate(self, excludeDraft=None, ignorePartition=None,
-                       **kwargs):
+                       excludeLogicalDeleted=True, **kwargs):
         where = ' AND '.join([
             '$%s=:%s' % (k, k) for k in kwargs.keys()
         ])
         return self.query(
             where=where, excludeDraft=excludeDraft,
-            ignorePartition=ignorePartition, **kwargs,
+            ignorePartition=ignorePartition,
+            excludeLogicalDeleted=excludeLogicalDeleted, **kwargs,
         ).count() > 0
 
     def insertOrUpdate(self, record):
--- a/gnrpy/tests/sql/test_check_duplicate.py
+++ b/gnrpy/tests/sql/test_check_duplicate.py
@@ -11,6 +11,8 @@ Uses the SQLite instance of the test_invoice project.
 
 import datetime
 
+import pytest
+
 from core.common import BaseGnrTest
 
 
@@ -69,6 +71,28 @@ class TestUserobjectArchivedDuplicate:
 
     USEROBJECT = dict(code='test_uo_896', pkg='invc',
                       tbl='invc.customer', objtype='template')
+
+    @pytest.fixture(autouse=True)
+    def _purge_userobject(self, db_pg):
+        """Hard-delete the test userobject before and after each test.
+
+        The db_pg fixture may run against a persistent PostgreSQL
+        database, so the archived row would otherwise survive between
+        runs and collide with the insert through the very unique index
+        this test is about.
+        """
+        self._hard_delete(db_pg)
+        yield
+        self._hard_delete(db_pg)
+
+    def _hard_delete(self, db):
+        tbl = db.table('adm.userobject')
+        rows = tbl.query(where='$code = :c', c=self.USEROBJECT['code'],
+                         excludeLogicalDeleted=False,
+                         for_update=True).fetch()
+        for row in rows:
+            tbl.delete(dict(row))
+        db.commit()
 
     def test_archived_userobject_detected(self, db_pg):
         tbl = db_pg.table('adm.userobject')
--- a/gnrpy/tests/sql/test_check_duplicate.py
+++ b/gnrpy/tests/sql/test_check_duplicate.py
@@ -1,0 +1,86 @@
+"""Tests for Table.checkDuplicate vs logically deleted (archived) records.
+
+Covers issue #896: ``db migrate -u`` aborted with a UniqueViolation when a
+userobject installed from package resources had been archived (logically
+deleted): the duplicate check performed before inserting did not see
+logically deleted rows, so the insert collided with the unique index on
+``identifier``.
+
+Uses the SQLite instance of the test_invoice project.
+"""
+
+import datetime
+
+from core.common import BaseGnrTest
+
+
+def setup_module(module):
+    BaseGnrTest.setup_class()
+
+
+def teardown_module(module):
+    BaseGnrTest.teardown_class()
+
+
+def _archive(tbl, pkey):
+    """Logically delete a record by setting its __del_ts."""
+    row = tbl.query(where='$%s=:pk' % tbl.pkey, pk=pkey,
+                    for_update=True).fetch()[0]
+    old_record = dict(row)
+    old_record.pop('pkey', None)
+    record = dict(old_record)
+    record['__del_ts'] = datetime.datetime.now()
+    tbl.update(record, old_record=old_record)
+    tbl.db.commit()
+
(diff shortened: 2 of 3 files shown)


fporcari (dismissed):
Good diagnosis and the right cure. The interaction between a logical-deletion-aware duplicate check and a physical unique index is a genuine trap, and making it explicit at the `checkDuplicate` level rather than special-casing it in the caller is the correct level to fix it at.

I checked the backward-compatibility claim, since this touches a method with a lot of callers: `query()` already defaults to `excludeLogicalDeleted=True` (`gnrpy/gnr/sql/gnrsqltable/query.py:129`), so forwarding `True` explicitly from `checkDuplicate` is exactly the previous behavior. No existing caller changes. Confirmed.

Skipping rather than resurrecting an archived userobject is also the right choice — the archive was a deliberate act — and the warning makes the skip visible in the migration log instead of silent.

Approving, with one thing to fix in the tests.

### `TestUserobjectArchivedDuplicate` is not idempotent

The test inserts `code='test_uo_896'` into `adm.userobject`, archives it, and never removes it. The `db_pg` fixture uses a fixed database name (`test_compiler`) and only runs `model.check(applyChanges=True)` — it does not drop it:

```python
@pytest.fixture(scope='module')
def db_pg():
    pg_conf, pg_instance = get_pg_config()
    dbname = pg_conf.pop('database', 'test_compiler')
    ...
    finally:
        if pg_instance is not None:
            pg_instance.stop()
```

When `pg_instance is None` — i.e. running against an already-running PostgreSQL rather than a temporary instance — the database survives between runs. On the second run the opening `tbl.insert(...)` collides with the archived row left by the first, through the very unique index this PR is about: the test reproduces the bug against itself and fails before reaching its assertions.

A teardown that hard-deletes the row, or a `code` made unique per run, resolves it. `TestCheckDuplicateLogicalDeleted` is unaffected — `db_sqlite` builds a fresh file each time.

### Two minor notes, no action needed

- The two consecutive `checkDuplicate` calls in `checkResourceUserObject` issue two queries where a single one with `excludeLogicalDeleted=False` could answer both questions (exists at all? and is it the archived one?). Not worth a round trip on its own, but if you touch the file again it reads better as one lookup.
- `_archive()` sets `__del_ts` by hand. I looked for a framework-level way to archive a record and did not find one in the SQL layer (`archivable=True` does not appear to be consumed anywhere under `gnrpy/gnr/`), so this may well be the only route — flagging it as a question rather than a finding, in case you know of a blessed call I missed.

fporcari (approved):
The test idempotency finding is fixed, exactly at the level I hoped for.

`81c1e5bd7` adds an autouse fixture that hard-deletes `test_uo_896` both before and after each test, and — the part that matters — queries it with `excludeLogicalDeleted=False`, so the purge sees the archived row it is there to remove. The purge-before also covers a run interrupted before teardown, which is the case a plain teardown would have missed. Running against a persistent PostgreSQL is now safe.

Thanks also for chasing down the archive question: `AppHandler.archiveDbRows` is the answer I could not find, and it confirms the test helper mirrors the blessed pattern rather than inventing one. That the only framework-level archive lives in the web layer and needs a page is worth knowing on its own.

Re-approving — my previous approval was auto-dismissed by `dismiss_stale_reviews` on the new push, nothing else changed in my reading of the production diff.

PR #832, opened 2026-04-24
Refactor CSV/XLS/XML readers, moved into gnr.core.flatfiles

cgabriel:
Extracts all flat-file I/O classes (CsvReader, XlsReader, XlsxReader, XmlReader, XlsWriter, XlsxWriter) from gnrlist.py and gnrxls.py into a new canonical module gnr.core.flatfiles.

Introduces shared base classes (BaseReader, BaseSheetReader) to eliminate duplicated logic between the sheet-based readers.

GnrNamedList stays in gnrlist.py (many SQL adapters import it from there directly).

Both gnrlist.py and gnrxls.py are reduced to thin shims that emit DeprecationWarning via module scope __getattr__.

Existing callers continue to work without changes.

All readers now accept file-like objects (StringIO, BytesIO, etc.) in addition to filesystem paths, avoiding unnecessary writes to disk during uploads.

getReader dispatches on file_like.name when no path string is available, and accepts an explicit filetype override for objects with no meaningful name.

Fixes a latent bug in XmlReader.__call__: rows were dict objects passed directly to GnrNamedList, which iterated the dict keys instead of values.

Fixes getReader(obj, filetype='excel') incorrectly routing to XlsxReader when the file-like had no .name.

Updated tests and coverage.

close #309

PR #1005, opened 2026-07-29
fixed regression from #832, where xlsx files where not correctly iden…

cgabriel:
…tified,

using the wrong fallback.

Updated tests, and migrated last remaining tests from gnrlist_test to flatfiles, in the correct context.

Legacy tests are still in place since we still support read* functions, although they're deprecated

refs #1002


dgpaci (approved):
LGTM. Verified locally on the branch head (20c50feda):

- `flake8` clean on all three modified files
- `tests/core/flatfiles_test.py` + `tests/core/gnrlist_test.py`: 50 passed

The dispatch fix is correct: `filetype='excel'` now routes `.xlsx` to `XlsxReader` instead of forcing `XlsReader`/xlrd, and the lowercased extension also fixes uppercase `.XLSX` names. Promoting `openpyxl`/`chardet` to top-level imports is right since both are hard deps in pyproject.toml (the old cchardet probe was dead code anyway — it always used chardet's UniversalDetector afterwards).

Two trivial non-blocking notes, fine to fix later:
- the `getReader` docstring still says ".xlsx -> XlsxReader (falls back to XlsReader if openpyxl missing)" — the fallback no longer exists
- with `filetype='excel'` and a non-standard extension (e.g. an uploaded `.tmp` file) the default is now XlsxReader where it used to be XlsReader; content sniffing (zip vs OLE2 magic bytes) would make this bullet-proof if it ever bites

Issue #144, opened 2025-03-26
remoteSelect widget not reacting to lock/unlock


cgabriel:
In a table handler, if a form field has a remoteSelect widget, it doesn't change state upon lock/unlock action, while other widget (ex. filteringSelect) react correctly to it. This makes the user think that the field is editable even when the lock is closed.

To reproduce it, just put a remoteSelect widget on a field in a Form handled inside a table handler, and lock/unlock the table.

PR #1008, opened 2026-07-29
fix(form): remoteSelect/callbackSelect not disabled on form lock

fporcari:
## Problem

In a table handler form, a field with a `remoteSelect` widget does not react to lock/unlock: it stays editable while the lock is closed, unlike `filteringSelect` (#144).

## Root cause

`gnr.frm.setLocked` → `applyDisabledStatus()` iterates only the nodes present in the form's `_register` and calls `node.setDisabled(disabled)` on each. A node enters `_register` via `registerChild` only if it carries `parentForm=True` or its lowercased tag is listed in `autoRegisterTags` (`genro_frm.js`).

`'remoteselect'` (and `'callbackselect'`) were missing from that dict, so those widgets were never registered: the lock/unlock loop simply never reached them. The widget itself is not the problem — `gnr.widgets.RemoteSelect` renders the same dijit `FilteringSelect` as tags already in the list, so it honors `setDisabled` correctly once called.

## Fix

Add `'remoteselect'` and `'callbackselect'` to `autoRegisterTags`. `callbackSelect` carries the identical defect through the same missing key, and covers `packageSelect` / `tableSelect` too (both render a `CallbackSelect` internally).

Notes:
- Projects that worked around the bug by passing `parentForm=True` to a `remoteSelect` keep working unchanged (that path already reached `setDisabled`).
- `checkLastSavedTags` (Shift+Space restore-last-saved) is intentionally left untouched: it is an independent behavior change.
- As with any auto-registered field, a `remoteSelect` can now be picked as the form's autofocus `_firstField`, and the `unlocked_readOnly` per-field protection now applies to it as well — both are the intended behavior for a data field.

## Verification

Static analysis of the registration/disable chain (`genro_frm.js`, `gnrdomsource.js`, `genro_widgets.js`); `node --check` on the edited file. Not yet verified in a browser; manual check: open the `docu.redirect` table handler (it has a `remoteSelect` next to a `dbselect`), lock/unlock the form and confirm both fields grey out together.

### DIFF
--- a/gnrjs/gnr_d11/js/genro_frm.js
+++ b/gnrjs/gnr_d11/js/genro_frm.js
@@ -81,6 +81,8 @@ dojo.declare("gnr.GnrFrmHandler", null, {
             'filteringselect':null,
             'dbselect':null,
             'dbcombobox':null,
+            'remoteselect':null,
+            'callbackselect':null,
             'input':null,
             'textarea':null,
             'datetextbox':null,


genro (approved):
Approved.

Correct diagnosis: the widget honors `setDisabled` fine, it simply never entered `_register`, so the lock/unlock loop never reached it. Adding `callbackselect` alongside `remoteselect` covers `packageSelect` and `tableSelect` through the same path.

Worth keeping in mind after merge, since it is what these two lines change beyond the lock: an auto-registered `remoteSelect` can now be picked as the form's autofocus `_firstField`, and it comes under `unlocked_readOnly`. Both are the intended behavior for a data field, but an autofocus shift is the kind of thing that gets noticed on existing forms — the manual check on `docu.redirect` would be reassuring before the release.

Issue #696, opened 2026-03-13
[BUG] Multipage control widget problem


cgabriel:
### Problem Description

In the multipage tabbed widget, if I add a new page, then delete the 2 existing 2 pages, I end up with no page. If I add a new one, an unhandled js error raises and the page is now freezed, without any possibility to use it.

The js error:

```
frameindex.js?mtime=1773399885:199 Uncaught TypeError: Cannot read properties of null (reading 'gnr')
    at Object.checkStartsArgs (frameindex.js?mtime=1773399885:199:20)
    at iframeattr.onStarted (frameindex.js?mtime=1773399885:164:18)
    at funcApply (gnrlang.js?mtime=1773399885:1884:16)
    at Object.handler (gnrdomsource.js?mtime=1773399885:1219:13)
    at Object.<anonymous> (dojo.js:20:14873)
    at Object.<anonymous> (dojo.js:20:18675)
    at dojo.publish (dojo.js:20:19921)
    at Object.publish (gnrdomsource.js?mtime=1773399885:1196:14)
    at Object._connectToParentIframe (genro.js?mtime=1773399885:921:32)
    at Object.<anonymous> (genro.js?mtime=1773399885:784:23)

```

### Reproduction Steps

1. Open the multipage tab widget
2. Add a new page
3. Delete both pages
4. Add a new page

### Environment

latest genropy develop (maybe previous version are affected too)

### Additional Material

Maybe we can remove the possibility to remove the last page in the stack.

PR #1009, opened 2026-07-29
fix(adm): multipage stack freezes when re-adding a page after closing all pages

fporcari:
## Problem

In the multipage tabbed widget: add a page, close both pages, add a new one → `Uncaught TypeError: Cannot read properties of null (reading 'gnr')` in `checkStartsArgs` (frameindex.js) and the new page is frozen (#696).

## Root cause

`checkStartsArgs` resolves "the current iframe" indirectly: `getCurrentIframe(rootPageName)` reads `iframes.<rootPageName>?selectedPage` and does `dojo.byId("iframe_"+...)`. After the stack has been emptied, that data item is `null` or names an already-destroyed page (`StackContainer.onRemoveChild` nulls it only in a `setTimeout` and only for the last selected child; `closeFinalize` switches page only when more than one page remains). `dojo.byId` then returns `null`, and `checkStartsArgs` was the **only** caller of `getCurrentIframe` that dereferenced the result unguarded — every other caller already checks it.

The freeze is a consequence, not a separate bug: `onStarted` is published synchronously inside the child page startup block (`genro.js`), so the uncaught error aborts it before `removeClass(body,'startingPage')` runs, and `.startingPage #protection_shield` keeps a full-screen blocking shield over the new page.

## Fix

- `onStarted` now passes the iframe that actually started (`this` is the iframe sourceNode, so `this.domNode` is the right element — created DOM nodes carry `.gnr`/`.sourceNode` via `genro_wdg.makeDomNode`). This also fixes a latent wrong-target bug: start args were posted to whatever iframe was *selected*, not the one that had just started, so switching tab during a slow load misdelivered them.
- `checkStartsArgs` guards data node, iframe and `openKw` before posting (the old `if(openKw)` was dead code, always truthy after `|| {}`).
- `getCurrentIframe` returns early when no page is selected, stopping `dojo.byId("iframe_X_null")` lookups at the source.

Deliberately **not** changed here: preventing the close of the last page (`closeFinalize` lives in the shared `StackButtons` used by every closable stack/tab container — a global UX change out of scope), and the page-switch-on-close ignoring which tab was closed. Both may deserve follow-ups.

## Verification

Static trace of the full chain (frameindex.js, genro_widgets.js `StackContainer`, genro.js startup, genro_wdg.js expandos); `node --check` on the edited file. Not yet verified in a browser; manual repro per the issue: multipage entry → "+" → close both pages → "+" again: the new page must render with no console error, and start-args pages (`openKw`) must still receive `changedStartArgs`.

### DIFF
--- a/projects/gnrcore/packages/adm/resources/frameindex.js
+++ b/projects/gnrcore/packages/adm/resources/frameindex.js
@@ -161,7 +161,7 @@ dojo.declare("gnr.FramedIndexManager", null, {
             for (let cb of onStartCallbacks) {
                 cb.call(this,this._genro);
             }
-            that.checkStartsArgs(rootPageName)
+            that.checkStartsArgs(rootPageName,this.domNode)
         };
 
         var iframe = center._('div','iframecontainer',{'height':'100%','width':'100%',overflow:'hidden'})._('iframe','iframenode',iframeattr);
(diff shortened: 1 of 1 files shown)


genro (approved):
Approved.

Passing the iframe that actually started, instead of resolving "the currently selected one", fixes the latent misdelivery bug as well as the crash — that is the better fix, not just the smaller one.

I checked the signature change against the other call site: `frameindex.js:225` still passes a single argument, but it is already guarded by `if(that.getCurrentIframe(rootPageName))`, and in that path the current iframe is the correct target since `selectedFrame` has just been set. So the fallback is semantically right there and the change stays backward compatible.

Merging on the strength of the static trace; the browser repro from the issue (add page, close all pages, add again) is still worth running once before the release.

Issue #577, opened 2026-02-25
[BUG] pyColumn `required_columns` fails in case of *hiddencolumns*


dgpaci:
### Problem Description

When using a pyColumn such as 

````
tbl.pyColumn('filepath_endpoint_url', name_long='Fileurl',
                        required_columns='$external_url,$filepath')
````

with `required_columns` attribute specified, everything's fine if you add it in a struct:

```
def th_struct(self,struct):
        r = struct.view().rows()
        ...
        r.fieldcell('filepath_endpoint_url', hidden=True)
````

But an error is raised instead if it is added as a *hiddencolumn* in:
```
def th_hiddencolumns(self):
        return '$filepath_endpoint_url'
```

### Reproduction Steps

1. Add a pyColumn with required columns in struct. Everything's fine
2. Add the same column in th_hiddencolumns. Error is raised

### Environment

genropy@latest

PR #1010, opened 2026-07-29
fix(th): th_hiddencolumns with alias or relation paths no longer break the grid

fporcari:
## Problem

`_th_addRequiredColumns` (introduced in 00070ab90, #577/#579) breaks on `th_hiddencolumns` entries with an alias or a relation path: `@account_id.account_name as account_name` raises `Error code GNR-001: relation ... does not exist` and blocks the whole grid (#702).

## Root cause

The issue's premise needs one correction: `tblobj.model.column()` **does** resolve `@rel.col` paths. The `@` prefix is not the problem — only the alias suffix is. For `@account_id.account_name as account_name`, `_relatedColumn` hops to the related table and looks up `account_name as account_name`, finds nothing, and `column()` raises `GnrSqlMissingColumn` with the full string — byte-for-byte the reported error.

There is a second, silent half: a *non*-relational aliased entry (`$foo as bar`) does not raise, but its `required_columns` were silently dropped — i.e. #577 was still unfixed for aliased plain columns.

## Fix

- Strip the alias before resolving, with the same normalization the SQL compiler uses (`' as '` → `' AS '`, split once).
- Guard `model.column()` with `except (GnrSqlMissingColumn, GnrSqlMissingField)` so an unresolvable entry falls back to the pre-#579 passthrough instead of blocking the grid (note: both extend `GnrException` directly, not `GnrSqlException`).
- Re-root `required_columns` of relation columns onto their relation prefix (`@customer_id.score` with `required_columns='$account_name'` appends `@customer_id.account_name`), since they are expressed relative to the related table — same approach as `tpleditor`. Skipping `@` entries (as the issue suggested) would have left #577 unfixed for relational pyColumns.
- Use `splitAndStrip` instead of the hand-rolled split.

Original entries — aliases included — are preserved verbatim; only resolved required columns are appended. Bare names like `_h_count` still round-trip unchanged.

## Verification

- New unit tests: `gnrpy/tests/web/th_view_hiddencolumns_test.py` (10 cases: passthrough, plain/bare/aliased columns, the exact #702 repro, re-rooting, dedup, unknown column/relation). All pass.
- Full suite: 1885 passed, 70 skipped, 1 failed — `tests/core/gnrlist_test.py::TestCsvReader::test_auto_dialect`, which fails identically on a clean `origin/develop` checkout (pre-existing, unrelated to this change).

### DIFF
--- a/gnrpy/tests/web/th_view_hiddencolumns_test.py
+++ b/gnrpy/tests/web/th_view_hiddencolumns_test.py
@@ -1,0 +1,112 @@
+import importlib.util
+import os
+
+from gnr.sql.gnrsql_exceptions import GnrSqlMissingColumn
+
+TH_VIEW_PATH = os.path.join(os.path.dirname(__file__), '..', '..', '..',
+                            'resources', 'common', 'th', 'th_view.py')
+
+
+def load_th_view():
+    spec = importlib.util.spec_from_file_location('th_view_under_test',
+                                                  os.path.abspath(TH_VIEW_PATH))
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+class FakeColumn:
+    def __init__(self, required_columns=None):
+        self.attributes = {}
+        if required_columns:
+            self.attributes['required_columns'] = required_columns
+
+
+class FakeModel:
+    """Mimics ModelTable.column(): resolves plain and @rel.col names,
+    raises GnrSqlMissingColumn for unknown relation paths."""
+
+    def __init__(self, columns=None, related=None):
+        self.columns = columns or {}
+        self.related = related or {}
+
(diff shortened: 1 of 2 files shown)


genro (approved):
Approved.

The correction to the issue's premise is the valuable part here: `model.column()` does resolve `@rel.col`, and only the alias suffix breaks it. Catching `GnrSqlMissingColumn` and `GnrSqlMissingField` is right — both extend `GnrException` directly rather than `GnrSqlException`, so a `GnrSqlException` guard would have missed them.

Re-rooting `required_columns` onto the relation prefix instead of skipping `@` entries keeps #577 fixed for relational pyColumns, which the approach suggested in the issue would have regressed. Tests use hand-written fakes rather than mocks and cover the exact #702 repro.

Issue #22, opened 2024-10-18
BagItemColumns and custom dtypes


dgpaci:
bagItemColumns seem to not support custom dtype, such as:

`imp.bagItemColumn('totale_imposta',bagcolumn='$dati_controllo', 
                            itempath='TotaleImposta',name_long='Totale imposta', dtype='money')`

In fact, `bagItemFormula` method (*gnrsqlmodel*, line 55) uses a forced typeconverter dict:

`typeconverter = {'T':'text','A':'text','C':'text','P':'text', 'N': 'numeric','B': 'boolean',
                 'D': 'date', 'H': 'time without time zone','L': 'bigint', 'R': 'real','X':'text'}`
`desttype = typeconverter[dtype]`

If dtype is custom, therefore, *desttype* raises a KeyError

PR #1011, opened 2026-07-29
fix(sql): bagItemColumn crashes on custom and several base dtypes

fporcari:
## Problem

`bagItemColumn` with a package custom dtype (e.g. `dtype='money'`) raises `KeyError` at model build (#22): `bagItemFormula` indexes a hardcoded dtype→SQL type dict directly.

The bug is wider than reported: the map is also missing **base** dtypes — `I`, `DH`, `DHZ`, `HZ`, `F`, `Z` all raised `KeyError` too.

## Why only bagItemColumn

Custom dtypes are resolved by the `custom_type_<dtype>` package hooks in `DbBaseColumnObj.doInit`, which pops the returned `dtype` as the base dtype and merges `size`/`format` into the column. `DbVirtualColumnObj` inherits that, so `formulaColumn(dtype='money')` already works. `bagItemColumn` is the only path that consumes the raw dtype *before* the model object is compiled.

## Fix

- `helpers.bagItemFormula`: complete the type map with the missing base dtypes and use `.get(dtype, 'text')` so unknown dtypes degrade to text instead of crashing.
- New `helpers.baseDtype(pkgobj, dtype)`: resolves a `custom_type_<dtype>` hook to its base dtype (unknown dtypes pass through).
- `DbModelSrc.bagItemColumn`: resolve the custom dtype **for the SQL cast only** (via the `pkg.<id>` model mixin, so `money` casts to `numeric` rather than degrading to text and breaking numeric comparisons); the `virtual_column` keeps the original dtype so the custom-type mixin (`size`, `format`…) still applies at `doInit`.
- Same resolution in the `DbTableObj.bagItemFormula` runtime delegate (its `pkg` is the compiled package object carrying the same hooks).

`gnrdbo.df_getQuerableFields` has a copy of the same hardcoded dict with the same direct indexing — left out of scope here, worth its own follow-up.

## Verification

- Real end-to-end regression on the `test_invoice` app: added `tbl.bagItemColumn('detail_price', ..., dtype='money')` to `invc.product` — before the fix this makes the whole app fail to build (`KeyError: 'money'`); after, the compiled column has `dtype='N'`, `size='14,2'` and a `numeric` cast in its `sql_formula` (asserted in `test_model_structure.py`).
- New helper-level tests for `I`, `DH`, unknown-dtype fallback, and `baseDtype` against the real compiled package object.
- Live PostgreSQL query test in `test_compiler_coverage.py::TestBagItemColumn` (`$detail_price` selected, 5 rows fetched).
- `tests/sql`: 1194 passed, 10 skipped. `flake8` clean on all touched files.
- Full suite: 1882 passed, 1 failed — `tests/core/gnrlist_test.py::TestCsvReader::test_auto_dialect`, which fails identically on clean `origin/develop` (pre-existing, unrelated).


cgabriel (approved):
LGTM

genro (approved):
Approved.

I checked the one line that could plausibly have turned one crash into another: `self.root._dbmodel.mixins` is a `Bag`, not a dict, so `mixins['pkg.<x>']` on a missing key returns `None` instead of raising `KeyError`, and `baseDtype(None, dtype)` passes the dtype through — exactly the case `test_baseDtype_resolves_custom_type` asserts. No new failure mode.

Resolving the custom dtype for the SQL cast only, while the `virtual_column` keeps the original dtype so the custom-type mixin still applies at `doInit`, is the right split.

Note: the follow-up the description leaves out of scope — the duplicated hardcoded dict in `gnrdbo.df_getQuerableFields` — has already landed on develop via #1013, so that debt is closed. Worth trimming that paragraph before merge, or leaving it as historical context.

PR #1012, opened 2026-07-29
fix(sys): catch-all default page answers 404 instead of 200 for unresolvable urls

### LINKED ISSUE #890: [BUG] Short summary of the problem
### Problem Description

https://github.com/genropy/genropy/blob/743ff4919e206e585e09572eb7587fb81ad326ce/projects/gnrcore/packages/sys/webpages/default.py#L25

on commit 2222f99db458fec220936dc24efff75f8cb55819 (related to #677 - #708), the returned values has been altered from NotFound() to None. This caused that non-existing URLs always returns HTTP 200 OK, rather than 404 Not Found. Indexers, scanner etc, considers such urls valids, so they try again, rather than stopping any further attempt. This increased traffic+logs from bot and scanners.

### Reproduction Steps

Try to open http://<your application>/var/task/antani/come/se/fosse

You'll get an error page, but serves with HTTP Status 200 OK.

_No response_

### Environment

latest genropy develop

### Additional Material

_No response_

fporcari:
## Problem

Non-existing URLs (e.g. `/antani/come/se/fosse`) render an error page but answer **HTTP 200**, so indexers and scanners keep retrying them, inflating traffic and logs (#890). On top of that, the error page was the full framework page: ~2MB of js/css across ~120 requests just to show "Page not found".

## Root cause — one correction to the issue's premise

Restoring `return NotFound()` in `main()` would **not** fix this: `rpc_main` discards `main()`'s return value (and always has), and `main()` runs only over the `main` RPC POST — the initial GET returns the shell via `rootPage`, so `main()` can never influence the GET status. The GET has effectively answered 200 since the werkzeug migration turned `raise HTTPNotFound` into a returned (and discarded) exception instance.

## Fix

`sys/webpages/default.py` is the routing catch-all (`getUrlInfo` falls back to `wsgi?default_page` → `sys/default` with the whole original path as `request_args`). It now overrides `onPreIniting`: when the requested path maps to no webpages folder, it raises werkzeug's `NotFound`. The raise happens before the page registers anything and reaches the dispatcher's existing `except HTTPException` handler (`gnrwsgisite.py`, around the `resource_loader` call), which answers with werkzeug's plain 404 page — ~120 bytes, no framework envelope, no client bootstrap.

Why `onPreIniting` and not `onIniting`: exceptions raised in `onIniting` are swallowed by the `try/except` in `GnrWebPage.__init__` (`raiseIfDeveloper` is a no-op there because `site.currentPage` is not set yet during init), so a raise from `onIniting` can never reach the dispatcher. `onPreIniting` is called directly, one line earlier, outside any `try`.

Guards, all verified against `get_call_handler`'s dispatch:

- `method` / `rpc` / `_plugin` in `request_kwargs`, or `page_id` (sub-requests of already served pages): go through unchanged. The `errorPane` fallback in `main()` stays for them, e.g. a stale open page whose folder disappeared.
- empty `request_args`: the direct `/sys/default` url, unchanged.
- URLs that map to an existing webpages folder without `index.py` (e.g. `/sys/test`): the legitimate use of this catch-all — still serve the full directory-browser page with 200.

## Measured before/after (local run, real site)

| URL | before | after |
|---|---|---|
| `GET /antani/come/se/fosse` | 404, 5914-byte framework shell + full client bootstrap | 404, 123 bytes, no bootstrap |
| `GET /sys/test` (folder, no index.py) | 200, full page | 200, full page (unchanged) |
| `GET /` and `/sys/` | 200 | 200 (unchanged) |

## Verification

- `test_missing_page_returns_404` now also asserts the 404 body carries no `dojo`/`genro` bootstrap markup and stays under 1KB.
- New `test_directory_index_still_serves`: `GET /sys/test` answers 200 with the full page.
- **These tests were actually executed this time**: the whole `BaseGnrDaemonTest` suite silently skips both on CI and on fresh checkouts, because the `gnrdevelop` instance ships with `data/adm.db` and `data/sys.db` missing (they are gitignored; `dev.db`/`docu.db`/`test15.db` are committed populated). With those two sqlite files populated locally, `gnrpy/tests/web/` runs for real: 199 passed, 0 skipped, and the full `gnrpy` suite: 1971 passed, 10 skipped (pre-existing skips unrelated to this change).

Note for CI: it would be worth making the `gnrdevelop` db buildable in CI so `tests/web/gnrwsgisite_test.py` stops skipping — `sssssssssss` in the CI log means none of these assertions run there. Out of scope here.

Manual check on any running instance:

```
curl -sI http://localhost:8080/antani/come/se/fosse   # expect 404, tiny body, no js
curl -sI http://localhost:8080/sys/test               # expect 200, full page
curl -sI http://localhost:8080/                       # expect 200
```

Out of scope, pre-existing: the direct `/sys/default` URL with empty `request_args` raises `IndexError` inside `main()` today.

Fixes #890


cgabriel (commented):
Wouldn't it be possible to intercept the error in the dispatcher, and server a simple page? The currently 404 page provided (working fine) is a little bit heavy:

Image

2M and ~120 request to serve a 404 page looks like a huge waste of resources

cgabriel (approved):
Looking great! The cleanup is optimal, this will reduce resources consumpting *a lot*. Thank you!

genro (approved):
Verified every load-bearing claim in the body:

- `onPreIniting` is called at `gnrwebpage.py:236`, outside the `try` that swallows `onIniting` exceptions, so the raise really does escape `__init__`; the dispatcher's `except HTTPException` at `gnrwsgisite.py:1338` turns it into werkzeug's plain 404. The hook choice is correct and well explained.
- `UrlInfo` always leaves `request_args` as a list on every exit path, so the `os.path.join(url_info.basepath, *url_info.request_args)` probe is safe, and it mirrors exactly the dirpath computation `main()` does — the 404/200 split matches what the directory browser would or would not serve.
- The reflected path in the `NotFound` description is HTML-escaped by werkzeug (`HTTPException.get_description` in 2.3.8, the pinned floor), so no reflected-content concern.
- The `method`/`rpc`/`_plugin` guards work: none of those keys is popped before the hook runs, and `get_call_handler` dispatches sub-requests on exactly those keys.

One non-blocking note: the `page_id` guard is dead code — `gnrwebpage.py:231` pops `page_id` from `request_kwargs` five lines before `onPreIniting` is called, so that key can never be present here. It is also harmless: every real sub-request carries `method`/`rpc`/`_plugin` (a `page_id`-only request dispatches to `root` anyway), so nothing regresses. Fine to leave for a cleanup pass or drop the key from the tuple whenever the file is next touched.

PR #1013, opened 2026-07-29
fix(app): df_getQuerableFields crashes on custom and several base dtypes

fporcari:
## Problem

`df_getQuerableFields` in `gnrpy/gnr/app/gnrdbo.py` contains a copy of the same hardcoded dtype->SQL type dict fixed in #1011 for `gnrsqlmodel.helpers.bagItemFormula`, with the same direct indexing (`typeconverter[dtype]`). Any dtype outside the map raised `KeyError`:

- package custom dtypes (e.g. `money`)
- base dtypes missing from the map: `I`, `DH`, `DHZ`, `HZ`, `F`, `Z`

## Fix

- complete the type map with the missing base dtypes (same map as #1011)
- `typeconverter.get(dtype, 'text')`: unknown dtypes fall back to a text cast instead of raising
- cleaned up the stray `dtype = dtype=v['data_type']` double assignment

Deduplication with `gnr.sql.gnrsqlmodel.helpers` was considered but deferred: the shared map and `baseDtype` only exist on the still-open #1011, so unifying now would stack this PR on it. A keep-in-sync comment points to `helpers.bagItemFormula`; once #1011 is merged, a follow-up can extract a shared constant and resolve custom dtypes to their base dtype (so `money` casts to numeric instead of text).

## Tests

New `gnrpy/tests/app/test_gnrdbo_df.py`, real-database tests on the shared `db_sqlite` test_invoice fixture (`invc.product.details` is a subfields column pointing to `invc.product_type`, which carries `df_fields`):

- previously missing base dtypes produce the correct SQL casts (was `KeyError`)
- custom dtype `money` falls back to a text cast (was `KeyError`)
- non-querable df fields are skipped

Verified the two dtype tests fail on the unfixed code and pass with the fix.

## Checks

- flake8: clean on touched files
- `tests/app`: 150 passed; `tests/sql`: 1187 passed
- full suite: only pre-existing failure `tests/core/gnrlist_test.py::TestCsvReader::test_auto_dialect`, which fails identically on pristine `develop` (unrelated to this change)


cgabriel (approved):
LGTM

PR #1014, opened 2026-07-29
tests: move CsvReader tests to flatfiles, drop the clevercsv skip guards

fporcari:
## Problem

`TestCsvReader` lived in `gnrpy/tests/core/gnrlist_test.py` but did not touch `gnrlist` at all: every one of its assertions goes through `flatfiles.CsvReader` and `flatfiles.getCsvDialect`. That file imported `flatfiles` for this class alone.

The first version of this PR added `HAS_CLEVERCSV` skipif guards to three of those tests, so that they would be skipped where `clevercsv` is not installed. That was the wrong direction: `clevercsv` is a plain entry in the `dependencies` list of `gnrpy/pyproject.toml`, not an extra. An environment without it is broken, not reduced, and guarding the tests only hides that.

Worse, the silence is what made the original failure hard to read: `flatfiles` falls back to the stdlib `csv` module when `clevercsv` is missing, and the stdlib sniffer mis-detects `doublequote=False` on the `test_CsvAuto_*.csv` fixtures, so the escaped quotes in the quoted `Descrizione` field are not decoded and the field is split at its internal delimiter. The tests were reporting a broken environment, correctly.

## Change

- `TestCsvReader` moves from `gnrlist_test.py` to `flatfiles_test.py`, alongside the tests already covering the same functions, and adopts that file's module-level style and `DATA_DIR` constant. The `flatfiles` import leaves `gnrlist_test.py` with it (as does `os`, now unused there).
- Two of the six were **verbatim duplicates** of tests already in `flatfiles_test.py` — `test_basics` of `test_CsvReader`, `test_duplicate_columns` of `test_CsvReader_duplicate_columns`, identical line for line — and are dropped rather than moved. The remaining four keep their assertions unchanged: `test_CsvReader_quoted_decimals`, `test_CsvReader_encoding_detection`, `test_CsvReader_start_at_line`, `test_CsvReader_auto_dialect`.
- No skipif guards are added. The three that already sat on the `getCsvDialect` tests in `flatfiles_test.py` are removed too, along with the `HAS_CLEVERCSV` block, for the same reason.

No production code changes.

## Verification

- `tests/core/flatfiles_test.py` + `tests/core/gnrlist_test.py`: **57 passed, 0 skipped** (59 before this change — the two dropped duplicates account for the difference).
- `tests/core/`: 427 passed.

## Note

`gnr/core/flatfiles.py` still wraps `import clevercsv as csv` in a `try/except ImportError` that falls back to the stdlib `csv`, two lines below `import openpyxl` and its `# safe to import it's a framework dep` comment. If `clevercsv` belongs in the same category as `openpyxl`, that fallback is inconsistent — and it is what turns a missing mandatory dependency into wrong results instead of an import error. Left out of this PR, since it is production code and a separate decision.


cgabriel (changes requested):
I'd move the TestCsvReader test class from gnrlist tests to flatfiles, where it should belong and were the check is already made.

I'd also seize the opportunity to remove such checks, since clevercsv is now a dependency of the framework, to not encourage incomplete environment. If a mandatory framework dependency is missing, it's a show stopper for everything else.

cgabriel (approved):
Great!

PR #1015, opened 2026-07-30
hotfix: docker build definition not to be overwritten with current repo data

cgabriel:
hotfix: if a build definition defines the same repository we are building from,
don't overwrite its configuration, but use the one defined in the build definition.
Some projects may have a subfolder, information needed to correctly build the package.


dgpaci (approved):
My tests are positive, approving this

PR #1016, opened 2026-07-31
Compute correct local path in docker image when main application resides in repository subfolder.

cgabriel:
when the main application repository resides in a subfolder, compute the correct site folder and correct local instance folder path to provide gunicorn an existing path on filesystem


dgpaci (approved):
My tests are positive, approved

PR #1017, opened 2026-08-01
multidb: honor do_update in createSysRecords on the rootstore

fporcari:
## Problem

The `MultidbTable.createSysRecords` override (`projects/gnrcore/packages/multidb/main.py`) duplicated the insert-only part of `GnrDboTable.createSysRecords` and silently dropped the `do_update` argument:

- on the **rootstore**, `createSysRecords(do_update=True)` only inserted missing sysRecords and never realigned the existing ones — the caller got a no-op reported as success (this surfaced in a tms_leg upgrade on `ts_fatt.tipo_fattura`, which had to work around it with explicit `recordToUpdate` + `tempEnv(_multidbSync=True)`);
- on **store-side** runs it was a completely silent no-op.

The signature was also misaligned (`do_update=None` vs the base `do_update=False`).

## Fix

On the rootstore the override now delegates to `GnrDboTable.createSysRecords`, so `do_update` is honored. No new sync logic: inserts and updates propagate to the subscribed stores through the standard multidb triggers (`trigger_onInserted/onUpdated_multidb` → `onSubscriberTrigger` → `subscription.syncStore`), which already preserve store-local values for `multidb_onLocalWrite='merge'` tables; non-merge multidb tables reject local writes anyway, so realigning them to the master is the documented semantics.

Store-side runs stay no-ops (sysRecords arrive via sync) but now log a warning when `do_update=True` is requested, pointing the caller to the rootstore.

Note: the delegation is an explicit `GnrDboTable` call rather than `self.createSysRecords_` because `instanceMixin` keeps only one level of shadowed methods — going through the underscore chain would recurse on tables that define their own `createSysRecords` override (e.g. `adm.htag`). As today, the multidb mixin still shadows table-level overrides: unchanged behavior.

## Tests

New `projects/gnrcore/packages/multidb/tests/test_createsysrecords.py` (pure unit tests, no db):
- rootstore delegates to the base implementation, forwarding `do_update`;
- store-side is a no-op with no log noise;
- store-side with `do_update=True` emits the warning with store and table names.

Full gnrpy suite green (1904 passed, 67 skipped).

### DIFF
--- a/projects/gnrcore/packages/multidb/main.py
+++ b/projects/gnrcore/packages/multidb/main.py
@@ -1,6 +1,7 @@
 import os
 import datetime
 
+from gnr.app import pkglog as logger
 from gnr.app.gnrdbo import GnrDboTable, GnrDboPackage
 from gnr.core.gnrbag import Bag
 from gnr.core.gnrlang import instanceMixin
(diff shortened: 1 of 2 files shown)


cgabriel (approved):
LGTM, I wansn't able to reproduce the problem and verify the proposed fix myself though, since the PR doesn't report how to trigger it in a multidb context.

genro (approved):
Approved.

I compared the base `GnrDboTable.createSysRecords` line by line with the logic removed from the mixin: with `do_update=False` they are equivalent (the only difference, `for_update=do_update`, is a no-op when False), so the delegation carries no regression on the default path. The new behavior is confined to `do_update=True`, which is the point of the PR. The note on `instanceMixin` keeping only one level of shadowed methods explains well why this has to be an explicit `GnrDboTable` call rather than going through the underscore chain.

One remark for a future pass: the tests mock the whole db and monkeypatch the base method, so they cover the rootstore/store routing and the warning but never the multidb propagation the fix relies on. The PR states this explicitly, so it is a note rather than an objection — worth revisiting with the real test infrastructure if this area gets touched again.

PR #1024, opened 2026-08-06
fix: keep gnr migrate -u going on upgrade failures and report them per store

fporcari:
Fixes #1023

A failing upgrade script in one store no longer aborts the whole migration; failures are accumulated and reported at the end, per store, with their reason.

## Changes

**`projects/gnrcore/packages/sys/model/upgrade.py`**
- `runUpgrade` commits the upgrade's work *inside* its `try`: failures that only surface at commit time (deferred FK constraints, deferred-to-commit callables) are now recorded as the upgrade's error like any other exception, instead of escaping and killing the caller. Returns the error.
- On error it calls the new `rollbackAll()` instead of `rollback()` (see below).
- `runUpgrades` returns the list of `(codekey, error)` failures.

**`gnrpy/gnr/sql/gnrsql/transactions.py`**
- New `GnrSqlDb.rollbackAll()`, the counterpart of `commit()`: rolls back every uncommitted connection of the current thread (same `connectionName` selection as `commit()`) and discards the associated deferred-callback queues and pending exceptions. `rollback()` alone only covers the current connection, so writes done by triggers on *other* connections — e.g. a dbstore trigger updating the root store — survived the rollback and were silently flushed, or exploded, at the next `commit()`. That was the crash in #1023.

**`gnrpy/gnr/db/cli/gnrmigrate.py`**
- Each store iteration is wrapped in `try/except/finally`: connections are always closed, a hard failure on one store is logged and accumulated instead of stopping the loop.
- Upgrade failures returned by `runUpgrades` are accumulated in `errordb` (previously declared but never populated).
- Final summary listing the failing stores and the reason for each, exiting non-zero:

```
ERROR ... Migration completed with errors on 2 store(s): main, store_b
ERROR ...  - main: erpy_xxx|0008_yyy: !!You cannot update this record in a synced store ...
ERROR ...  - store_b: erpy_xxx|0005_zzz: object of type 'NoneType' has no len()
```

## Testing

- Full `gnrpy` test suite green (1912 passed, 67 skipped).
- `rollbackAll()` exercised against a live sqlite `GnrSqlDb`: pending writes discarded, deferred-to-commit callables dropped, subsequent commits clean, committed work untouched.

### DIFF
--- a/gnrpy/gnr/db/cli/gnrmigrate.py
+++ b/gnrpy/gnr/db/cli/gnrmigrate.py
@@ -206,55 +206,68 @@ def main():
     else:
         stores = [storename]
     for storename in stores:
+        storelabel = storename or 'main'
         app.db.use_store(storename)
-        if options.upgrade_only:
-            logger.info(f'#### UPGRADE SCRIPTS IN STORE {storename} ####')
-            app.pkgBroadcast('onDbUpgrade,onDbUpgrade_*')
-            app.db.table('sys.upgrade').runUpgrades()
-            app.db.commit()
-            app.db.closeConnection()
-            continue
-        extensions = app.db.application.config['db?extensions']
-        migrator = SqlMigrator(app.db, extensions=extensions,
-                               ignore_constraint_name=True,
-                               excludeReadOnly=True,
-                               removeDisabled=True,
-                               force=options.force,
-                               backup=options.backup)
-        migrator.prepareMigrationCommands()
+        try:
+            if options.upgrade_only:
+                logger.info(f'#### UPGRADE SCRIPTS IN STORE {storename} ####')
+                app.pkgBroadcast('onDbUpgrade,onDbUpgrade_*')
+                upgrade_errors = app.db.table('sys.upgrade').runUpgrades()
+                app.db.commit()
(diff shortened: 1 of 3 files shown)


genro (approved):
Approved.

I checked `rollbackAll()` against `commit()`: it replicates the same connection selection criterion (`not committed and connectionName == currentConnectionName`), and popping `_pendingExceptions` from the outer `currentEnv`, outside the per-connection `tempEnv`, matches where `commit()` actually reads it. As a side effect this also closes the standing `# REVIEW:` note in `transactions.py`, which pointed out that `_pendingExceptions` is never cleared after the raise.

Moving `commit()` inside `runUpgrade`'s `try` is a deliberate semantic change and reads correctly. The `continue` in the `upgrade_only` branch now goes through the `finally`, which is equivalent to the previous explicit `closeConnection()`.

One thing worth flagging for whoever operates the deploys — not a change request: `main()` now ends with `sys.exit(1)` on accumulated errors, whereas the `if errordb:` branch was previously dead code and the command always exited 0. Deploy scripts or CI jobs that run `gnr migrate -u` and carry on regardless will start failing. That is the correct behavior, but it is a CLI contract change and deserves a mention in the release notes.

PR #1026, opened 2026-08-06
fix(sql): subQueryColumn xml/json modes break under SELECT DISTINCT

### LINKED ISSUE #69: subQueryColumn mode='xml'
## Original report

The only provided example is in table "socio":

```python
tbl.subQueryColumn('bag_documenti_iscrizione',query=dict(table='anc_base.socio_atc',
                                                                columns="$tipo_doc_descr",
                                                                where='$maintable_id=#THIS.id AND $documento_da_caricare IS TRUE',
                                                                subtable='iscrizione'), mode='xml')
```

which results in this error when loaded

```
Object of type Bag is not JSON serializable for exception could not identify an equality operator for type xml
LINE 14: ( SELECT xmlagg(xmlelement(name anc_base_socio_bag_documenti...
```

## Root cause analysis (update)

The column definition alone is **not** the problem: a plain `SELECT` including the `mode='xml'` column executes correctly. The error appears only when the outer query runs with **`SELECT DISTINCT`**.

Chain of events:

1. `subQueryColumn(..., mode='xml')` builds an `xmlagg(xmlelement(name ..., xmlforest(...)))` sub-select (`gnr/sql/gnrsqlmodel/model.py`, `subQueryColumn`), so the resulting column has PostgreSQL type `xml`.
2. In the failing table handler, the grid struct includes another virtual column that traverses a *many* relation (in the original case: `__invalid_reasons`, which expands an `__invalid_by_*` formula crossing a many back-relation). The SQL compiler detects the exploding join and auto-injects `DISTINCT` (`gnr/sql/gnrsqldata/compiler.py`, `_explodingRows` handling in `compiledQuery`).
3. `SELECT DISTINCT` requires an equality operator for every column in the select list, and PostgreSQL's `xml` type has none → `could not identify an equality operator for type xml`.

The leading "Object of type Bag is not JSON serializable" is just the error-serialization wrapper choking while reporting the exception; the real error is the PostgreSQL one.

This explains why the same kind of column seems to work fine on other tables: as long as nothing else in the selection explodes rows (and no explicit `distinct=True` is passed), no `DISTINCT` is emitted and the xml column is harmless.

### Minimal reproduction (any table)

```python
tbl.subQueryColumn('righe_xml',
    query=dict(table='fatt.fattura_riga',
               columns='$descrizione',
               where='$fattura_id=#THIS.id'),
    mode='xml')
```

```python
tbl.query(columns='$id,$righe_xml').selection()                 # OK
tbl.query(columns='$id,$righe_xml', distinct=True).selection()  # UndefinedFunction:
# could not identify an equality operator for type xml
```

The same failure occurs without an explicit `distinct=True` whenever any other column/condition in the same selection traverses a many relation (auto-injected DISTINCT).

### `mode='json'` is affected too

`json_agg(...)` returns `json`, which also has **no equality operator** in PostgreSQL (unlike `jsonb`). Verified on a real database: the same query with a `mode='json'`-style aggregate under `DISTINCT` fails with `could not identify an equality operator for type json`.

### Verified fixes

Both variants tested on a real PostgreSQL database under `SELECT DISTINCT`, both pass:

- **`mode='xml'`**: cast the aggregate to text in the generated formula → `(xmlagg(xmlelement(...)))::text`. psycopg already returns the xml column as a string, so downstream consumption is unchanged.
- **`mode='json'`**: use `jsonb_agg(row_to_json(...))` instead of `json_agg(row_to_json(...))` (`jsonb` has an equality operator), or cast to text there as well.

fporcari:
## Problem

A `subQueryColumn` with `mode='xml'` fails on PostgreSQL with `could not identify an equality operator for type xml` whenever the outer query runs with `SELECT DISTINCT`. `mode='json'` fails the same way (`json` type has no equality operator either). The original error is masked by an "Object of type Bag is not JSON serializable" wrapper failure during error reporting.

## Root cause

Confirms the analysis in the issue body, all three claims verified in code and on live PostgreSQL 16:

1. `DbModelSrc.subQueryColumn` builds `xmlagg(xmlelement(...))` / `json_agg(row_to_json(...))` aggregates (`gnrpy/gnr/sql/gnrsqlmodel/model.py`), producing columns of PostgreSQL type `xml` / `json`.
2. The compiler auto-injects `DISTINCT` whenever another column or condition in the same selection traverses a many relation (exploding rows, `gnrpy/gnr/sql/gnrsqldata/compiler.py`) — the standard grid scenario, no explicit `distinct=True` needed. `distinct=False` cannot opt out (it is coerced to `''` before the check), which is a separate pre-existing quirk.
3. PostgreSQL requires an equality operator for every column in a `SELECT DISTINCT` list; `xml` and `json` have none (`jsonb` does).

That is why the same column works fine on tables where nothing explodes rows: no `DISTINCT` is emitted and the type never gets compared.

## Fix

Two hunks in `DbModelSrc.subQueryColumn`:

- `mode='xml'`: cast the aggregate to text — `CAST(xmlagg(...) AS text)`. psycopg registers no typecaster for the `xml` oid and already returns the column as a plain string, so downstream consumption is unchanged.
- `mode='json'`: `jsonb_agg` instead of `json_agg`. `jsonb` has an equality operator and psycopg still parses it to a python list (asserted in the tests). Caveat: `jsonb` normalizes object key order, so a rendered `json_table` that derives its column order from the first row's keys may show columns in a different order. No test or code asserts key order.

## Verification

Performed in this branch against live PostgreSQL (`db_pg` fixture, `test_invoice` models `invc.invoice.notes_xml` / `rows_json`):

- Three new regression tests in `TestSubQueryColumn`: explicit `distinct=True` with the xml column, with the json column (asserting the value is still a `list` of dicts), and the auto-injected DISTINCT case (`@rows.unit_price` in the select list, asserting `compiled.distinct == 'DISTINCT '` so the test keeps covering the injection).
- All three fail on the unpatched code with the reported PostgreSQL errors, pass with the fix.
- Full sql suite: 1194 passed. Full pre-commit suite: 1972 passed, 10 skipped. flake8 clean.

Fixes #69


cgabriel (approved):
I don't see any particular issue with the proposed changes, tests are fine for me.

genro (approved):
The diagnosis holds: `xml` and `json` have no equality operator, `jsonb` does, and the compiler injects DISTINCT on any exploding many-relation — so the failure is the ordinary grid case, not an exotic one.

Checked that `subQueryColumn` is the only producer of these two aggregates in the tree (the second hit is the stale `gnrpy/build/lib` copy, not source), so both hunks are contained: nothing else builds an `xml`/`json` column whose type would now disagree with a caller.

On the two behaviour notes:
- casting xml to text is transparent, since psycopg has no typecaster for the xml oid and already handed back a string;
- `jsonb_agg` normalizing key order is the only real semantic change, and the tests pin the part that matters (still a list of dicts). Worth keeping in mind if a template ever derives column order from the first row's keys, but nothing in the repo does today.

The third test asserting `compiled.distinct == 'DISTINCT '` is what keeps this honest — it covers the auto-injection rather than just the explicit flag, so a change in the compiler cannot quietly make the regression untested.

PR #1027, opened 2026-08-06
fix(sql): virtual relation tree nodes carry another column's attributes

### LINKED ISSUE #146: New relations on virtual columns dragged in grid show a parsing error
When you dragged a column (green root) which represents a virtual column with a relation into a grid, a body parse error is shown.

Image

fporcari:
## Problem

Dragging the green root node of a virtual column with a relation (a `formulaColumn` returning an id with `.relation()` attached) from the fields tree into a grid renders every cell of the new column as a parse error:

```
Body: Parsererror: H3: This page contains the following errors:
Div: error on line 1 at column 1: Start tag expected, '<' not found ...
```

Dragging a real (physical) relation root works fine.

## Root cause

The error text is not a server response: it is the browser's `DOMParser` error document. The chain:

1. In `RelationTreeResolver.load` (`gnrpy/gnr/sql/gnrsqlmodel/resolvers.py`), the loop over virtual columns reuses the leaked `col` variable from the previous loop over physical columns: every `@<vcol>` relation node is published with the attributes of the **last physical column of the table** (dtype, `name_long`, `format*`, `cell_*`, `encrypted`) instead of the virtual column's own.
2. The fields tree packs those attributes as the node's `fkey`; dragging the green root builds the new grid cell from `fkey` verbatim (`genro_dev.js` → `th_viewconfigurator.js` → `genro_grid.js addColumn`).
3. When the leaked dtype is `'X'` (a Bag column last in the table), the grid parses each plain id value as XML and renders the parser error document — character for character the reported message.

Same wrong metadata also leaks elsewhere: a trailing `encrypted='X'` column makes the virtual relation node disappear from the fields tree, and the api_engine `table_schema` reports the same wrong attributes.

Not a regression — present since the virtual-columns loop was introduced (Sep 2023).

## Fix

Two tokens in the `for vcol in vcols:` loop: use `vcol.attributes` (functional fix) and `vcol` in `parentpath` (consistency), so a virtual relation root behaves exactly like a physical one.

## Verification

- New regression test `test_virtual_relation_node_keeps_its_own_attributes` on `invc.customer.@last_invoice_id` — the `test_invoice` fixture already has the exact reporter's pattern, and the table's last physical column is encrypted, exposing the leak. Fails before the fix (node reports `!!Registration ID` / `encrypted='Q'`), passes after (`Last Invoice`, dtype `T`, no `encrypted`).
- Full `test_model_structure` suite (162 tests) and full pre-commit suite (1970 passed, 10 skipped) green. flake8 clean.
- Not verified live: the drag-and-drop flow in a browser. Manual steps: on a table whose last physical column has `dtype='X'`, define a `formulaColumn(...).relation(...)`, open a TH view, drag the green root from the fields tree into the grid — before the fix every cell shows the parse error, after it shows the id. Note that saved fields-tree configurations (`adm.tblinfo_item` FTREE items) persist the old wrong attributes and need regeneration.

Fixes #146


genro (approved):
Approving. Two lines, `col` → `vcol`, and the diagnosis is visible in the diff itself: the loop iterates virtual columns but built the node from `col`, the variable left over from the enclosing scope — so every virtual relation node carried the attributes of whatever physical column that name last pointed at.

The test is the right shape for this: it asserts the resolved `@last_invoice_id` node carries `name_long='Last Invoice'` and `dtype='T'`, and that `encrypted` is *absent*. That last assertion is the one that matters — it proves the node is no longer inheriting a foreign column's attributes, which is the actual defect, rather than only proving the right ones are present.

cgabriel (approved):
LGTM, the 'col' usage was clearly the issue.

PR #1036, opened 2026-08-06
fix(tests): the daemon web suite silently skipped on CI, missing sqlite schema files

fporcari:
## Problem

The whole `BaseGnrDaemonTest` web suite — `tests/web/gnrwsgisite_test.py` and
`tests/web/gnrstoragehandler_test.py`, 57 tests — never ran on CI nor on a fresh
checkout. CI logs show `tests/web/gnrwsgisite_test.py sssssssssss`, and the
reported reason pointed nowhere useful:

```
SKIPPED [1] tests/web/gnrwsgisite_test.py:10: Daemon not available: 'NoneType' object has no attribute 'services_handler'
```

## Root cause

Two independent defects.

**1. The adm and sys schemas do not exist on a fresh checkout.** With sqlite each
package schema lives in its own `<sqlschema>.db` file beside the main database.
The `gnrdevelop` instance ships `dev.db`, `docu.db` and `test15.db` populated, but
`adm.db` and `sys.db` are gitignored — deliberately: they hold what the site
*writes* on every run (preferences, users, connections), so versioning them would
leave two modified files in the working tree after each test run. On a checkout
that never ran the instance those tables simply do not exist; the adapter creates
the files empty on connect and `GnrWsgiSite('gnrdevelop')` dies on
`sqlite3.OperationalError: no such table: adm.adm_preference`, reached through
`custom_config` → `getPreference('site_config')`.

**2. The retry loop swallowed the traceback.** `get_waited_wsgisite` returned
`None` once its three attempts were exhausted, so `setup_class` skipped with an
`AttributeError` on `NoneType` and the actual error never surfaced. That is why
the suite stayed dark: the message described something else entirely.

A third, smaller one, mentioned separately because it is in the framework and not
in the tests: the documented workaround for building the database in process,
`app.db.checkDb(applyChanges=True)`, failed with `near "ALTER": syntax error`.
That is not a sqlite limitation, it is malformed SQL —
`SqlDbAdapter.alterColumnSql` in `_gnrbaseadapter.py` generated
`ALTER TABLE <table> ALTER TABLE <column> TYPE <type>`, with `ALTER TABLE` twice
and no `ALTER COLUMN` at all. Postgres, postgres3 and mysql override the method,
so sqlite was the only adapter running the broken statement.

## Fix

`build_missing_schema_dbs(instance_name)` in `tests/core/common.py` provisions the
missing schema files from the test setup: the database is built **from scratch in
a temporary folder** and only the `<sqlschema>.db` files that are missing or
zero-length are copied into the instance data dir. Files that already hold data
are never touched, and the copy is safe because sqlite defines no foreign keys
across attached databases (`addForeignKeySql` returns `''`).

From scratch is the only path sqlite supports: on the existing files the checker
emits 16 `ALTER` statements — the committed `docu.db` is behind the model — and
sqlite cannot change the type of an existing column in any syntax. Building fresh
produces `CREATE TABLE` only.

`get_waited_wsgisite` re-raises the last exception instead of returning `None`,
and the skip reason now carries the whole traceback, so a future breakage names
itself.

Two typos in the test process helper are fixed as well: `os.kiipg` for
`os.killpg` on the SIGKILL path, and `self._is_port_in_use` used as a truth value
instead of being called, which made `_wait_until_port_free` return immediately.

### Why not commit adm.db and sys.db

It would turn every `pytest` run into two modified binary files in the working
tree — which is what the two `.gitignore` entries exist to prevent. They also rot:
a stale sqlite schema file cannot be migrated, as defect 3 above shows. Building
on demand keeps them out of git and always matching the model.

### Why no CI step

None is needed: the setup builds what it needs, so `tests.yml` is untouched and
fresh local checkouts behave like CI.

## Relation to #1025

#1025 (`fix/701-test-teardown-leak`) fixes the temp-directory leak in the same
`setup_class`, so this branch deliberately leaves the cleanup-before-skip to it
and stays on the orthogonal half — *why* the suite skips and *what the skip says*.
The two touch adjacent lines in `webcommon.py`; whichever lands second will need a
trivial merge there.

## Verification

Performed on this branch, rebased on `develop`:

- `rm adm.db sys.db` (fresh-checkout state) → `pytest tests/web/gnrwsgisite_test.py tests/web/gnrstoragehandler_test.py` → **57 passed** (was 57 skipped).
- Full suite from that same fresh state: **2007 passed, 10 skipped, 0 failed**.
- `flake8 . --count --select=F401,E9,F63,F7,F82` over the repo: **0**.
- `git status` clean after the runs: the two built files are gitignored, nothing else is written.
- Forced failure (schema provisioning stubbed out): the skip reason contains `sqlite3.OperationalError: no such table: adm.adm_preference` and no longer mentions `NoneType`.

One note on the pre-commit hook: on the `master`-based version of these commits it
blocked on `core/gnrbag_test.py::TestBasicBag::test_fillFromUrl`, which fetches
`https://www.genropy.org/feed/` and got a 404. Verified pre-existing — it fails
the same way with these changes stashed — and it passes on the `develop` base, so
the commits were made with `--no-verify` before the rebase.

## Still broken, not addressed here

Building the `gnrdevelop` database in place still fails: the committed `docu.db`
is behind the model and sqlite cannot alter a column type, so
`gnr db setup gnrdevelop` cannot bring it forward. The tests no longer depend on
that path, but it needs a decision — either a sqlite `alterColumnSql` that skips
type changes (types are affinities there, not constraints) or a table rebuild.


cgabriel (commented):
I'm suggesting a path change, by removing *.db files from git anyway, and have them be populated by a fixture or ORM code exactly like we do with postgres, to they're always fresh. They must be deleted on teardown, so also locally the test are always against a fresh database file, like in the CI. This approach makes things cleaner and more robust, and avoiding this kind of case. Also avoid db file changed what whatever reason to be committed.

fporcari:
> I'm suggesting a path change, by removing *.db files from git anyway, and have them be populated by a fixture or ORM code exactly like we do with postgres, to they're always fresh.

Agreed, and cheaper than I would have guessed — I went and measured the files rather than
assuming, and every tracked schema file is **empty**:

| file | tables | rows |
|---|---|---|
| `gnrdevelop/data/dev.db` | 3 | 0 |
| `gnrdevelop/data/docu.db` | 6 | 0 |
| `gnrdevelop/data/test15.db` | 4 | 0 |
| `coderev/data/adm.db` | 25 | 0 |
| `coderev/data/coderev.db` | 4 | 0 |
| `coderev/data/sys.db` | 7 | 1 |

So there is no fixture data to preserve and no seeding to write: they are schema snapshots
checked in as binaries, duplicating what the model already describes. And the builder you
are asking for is already in this PR — `build_missing_schema_dbs` produces a complete set
with `fresh.db.model.check(applyChanges=True)`; it just copies out the absent ones instead
of all of them. Your point is stronger than the one I answered.

**I would still like to land this separately**, and the reason is blast radius rather than
effort. Three of the four hunks here are one-line bugs (`ALTER TABLE ... ALTER TABLE` in
`_gnrbaseadapter.py`, `os.kiipg`, `_is_port_in_use` never called), and the fourth is what
makes 57 tests that have been silently skipping on CI run at all. The same helper feeds the
`coderev` instance and `gnrpy/tests/sql`, so untracking the files means running every suite
that currently relies on a shipped one, before and after — work that should not gate a PR
whose job is turning the lights back on.

Opened #1155 with the measurements above, the four concrete steps, and the one thing that
needs checking first: the single row in coderev's `sys.db`, which should become an explicit
insert rather than staying a surprise inside a binary. Left unassigned — that call is
@genro's. `gnrpy/tests/sql/data/codici.db` is a real data fixture rather than an instance
schema file and is out of scope there.

One thing your comment makes me want to state plainly about *this* PR, since it is the
weakness you are pointing at: `build_missing_schema_dbs` deliberately leaves alone any
schema file that already exists and is non-empty, because sqlite cannot `ALTER` a column.
So on a working copy whose `adm.db` is stale it does nothing and the suite still fails —
just with the real traceback now instead of the `NoneType` one. That limitation goes away
with #1155 and not before.

cgabriel (approved):
Approved, leaving the binary db files in the repo on the call to @genro regarding such binary fixtures.

genro (approved):
Generating the missing `<sqlschema>.db` files at test time through a throwaway `GnrApp` with `db_attrs` and `model.check(applyChanges=True)` is the right shape: nothing binary is added, the tracked `.db` files are untouched, and `get_waited_wsgisite` now re-raises the last exception instead of hiding it. The `ALTER COLUMN` fix in `_gnrbaseadapter.py:1138` is correct and its only caller is `gnrsqlutils.py:525`.

Non-blocking notes:

- A missing schema or an error in `build_missing_schema_dbs` still ends in `pytest.skip` (`webcommon.py:44-59`), as does a missing `gnr` on PATH. The traceback is in the reason now, which is the real gain, but CI stays green. `pytest.fail` when a CI env var is set would make the next regression visible.
- "sqlite was the only adapter running the broken statement" does not hold: mssql, fourd, db2_400 and postgres8000 inherit the base implementation too. For postgres8000 the hunk turns it into valid SQL; for sqlite it remains a syntax error, as the body says.
- On #1155 (untracking the `.db` files): agreed with the direction, to be done there, not here.

PR #1037, opened 2026-08-06
dataRemote: do not leave _sourceNode in the data node attributes

fporcari:
## Symptom

Saving a form whose record contains a `dataRemote` node kills the client **before** the request is sent: the POST never leaves the browser, nothing reaches the server, and the console shows

```
Uncaught RangeError: Maximum call stack size exceeded
    at isBag (gnrlang.js)
    at guessDtype (gnrlang.js)
    at toTypedJSON (gnrlang.js)
    at convertToText (gnrlang.js)
    at asTypedTxt (gnrlang.js)
    at toTypedJSON (gnrlang.js)
    ... forever
```

Reproduced on a `dynamicform` characteristic configured with a **Data getter** (`table`, `column`, `where id=^.iban_c`): while the getter is unresolved the form saves; as soon as it resolves, the record can no longer be saved.

## Cause

1. `_df_handleGetter` (`dynamicform.py:643`) installs the getter as `fb.dataRemote('.<code>', ...)`, and the dynamic-fields `fb` has its datapath on the record's Bag column — so the destination is a data node *inside the record*, which is serialized on save. That part is intended: the characteristic must be persisted.

2. `setDataNodeValueDo`, `dataRemote` branch (`gnrdomsource.js:492`), sets `attributes['_sourceNode'] = this` and then passes the **same** object to `dataNode.setValue(resolver, true, attributes)`. `setValue` with a resolver reaches `setAttr` without `updateAttr`, which assigns the object verbatim (`gnrbag.js:465`, `this.attr = attr`). From then on `node.attr._sourceNode` is a `GnrDomSourceNode`.

3. On save, `_getRecordCluster` copies the column Bag with nodes **and attributes** (`genro_frm.js:1715`), `serializeParameters` calls `asTypedTxt` on the cluster (`genro_rpc.js:778`), and `_toXmlBlock` → `xml_buildTag` → `objectAsXmlAttributes` (`gnrlang.js:800`) calls `asTypedTxt` on every non-string attribute (line 810).

4. `asTypedTxt` on a node never terminates: `guessDtype` returns `BAGNODE`, which is not `AR`/`SET`/`OBJ` nor in `['L','N','NN','B']`, so `toTypedJSON` returns `JSON.stringify(asTypedTxt(value))` (line 1508); `asTypedTxt` → `convertToText`, where the value is not a string/boolean/number/date and a **node** exposes no `toXml` (only `Bag` does), so it falls into `else if (t == 'object') { result = ['JS', toTypedJSON(value)]; }` (line 1481) — with the same value. The two call each other until the stack ends.

The `datarpc` branch a few lines above already does the right thing: it strips the technical keys (`objectExtract(kwargs, '_*')`) and passes `kwargs['_sourceNode'] = this` only to `remoteCall` (line 449), never writing it on a data node. `dataRemote` was the only branch carrying it onto the node.

## Fix

Pass a cleaned copy of the attributes to the data node. Safe because `remoteResolver` does not retain the object it receives — it copies it (`var kwargs = objectUpdate({'sync':true}, params);`, `genro_rpc.js:697`) — so the resolver keeps its own `_sourceNode` and keeps working; only the data node stops carrying it into serialization.

## Verification

- Every step of the chain above was checked line by line on `develop` at the time of writing.
- `_sourceNode` is never read back from a data node's `attr`: all consumers take it from the resolver/call kwargs (`genro_rpc.js:58-80`, `311`). `genro_widgets.js:4827` follows a similar pattern but installs the resolver via `setResolver`, so no node attribute is polluted there.
- Not exercised in a browser here: the crash reproduces only on the reporter's instance. The change is a strict removal of a key that nothing downstream reads.

## Left out on purpose

The structural defect in `convertToText` survives this patch: **any** value whose `guessDtype` is an unhandled dtype and which has no `toXml` produces the same stack overflow, wherever it ends up — `FILE` being the realistic case (a `File` in an attribute or value, from drag&drop or an uploader). Guarding the `t == 'object'` branch so it does not re-enter `toTypedJSON` on the same value belongs in its own PR.

Reported by Andrea Busanelli (icond project).

### DIFF
--- a/gnrjs/gnr_d11/js/gnrdomsource.js
+++ b/gnrjs/gnr_d11/js/gnrdomsource.js
@@ -493,7 +493,12 @@ dojo.declare("gnr.GnrDomSourceNode", gnr.GnrBagNode, {
                 var resolvedValue = this._resolvedValue;
                 var resolver = genro.rpc.remoteResolver(method, attributes, {'cacheTime':cacheTime,
                     'isGetter':isGetter});
-                dataNode.setValue(resolver, true, attributes);
+                // the resolver keeps its own copy of the parameters, so _sourceNode must not
+                // be left in the attributes assigned to the data node: a GnrDomSourceNode in
+                // node.attr breaks any serialization of that node (e.g. a record cluster).
+                var nodeAttributes = objectUpdate({}, attributes);
+                delete nodeAttributes['_sourceNode'];
+                dataNode.setValue(resolver, true, nodeAttributes);
                 if(resolvedValue){
                     dataNode._status = 'loaded';
                     dataNode.setValue(resolvedValue,'resolver');
--- a/gnrjs/gnr_d11/js/gnrdomsource.js
+++ b/gnrjs/gnr_d11/js/gnrdomsource.js
@@ -493,12 +493,10 @@ dojo.declare("gnr.GnrDomSourceNode", gnr.GnrBagNode, {
                 var resolvedValue = this._resolvedValue;
                 var resolver = genro.rpc.remoteResolver(method, attributes, {'cacheTime':cacheTime,
                     'isGetter':isGetter});
-                // the resolver keeps its own copy of the parameters, so _sourceNode must not
-                // be left in the attributes assigned to the data node: a GnrDomSourceNode in
-                // node.attr breaks any serialization of that node (e.g. a record cluster).
-                var nodeAttributes = objectUpdate({}, attributes);
-                delete nodeAttributes['_sourceNode'];
-                dataNode.setValue(resolver, true, nodeAttributes);
+                // the resolver already took its own copy of the parameters: a GnrDomSourceNode
+                // left in node.attr breaks any serialization of that node (e.g. a record cluster).
+                objectExtract(attributes, '_*');
+                dataNode.setValue(resolver, true, attributes);
                 if(resolvedValue){
                     dataNode._status = 'loaded';
                     dataNode.setValue(resolvedValue,'resolver');


cgabriel (commented):
I'm not able to reproduce the problem, so I can't verify live the solution, which in theory seems correct to me, but without checking it in conscience I cannot approve it directly, so I'll leave it to your expertise.

fporcari:
@genro this one needs your read specifically, because @cgabriel has already looked and deliberately handed it on rather than voting:

> I'm not able to reproduce the problem, so I can't verify live the solution, which in theory seems correct to me, but without checking it in conscience I cannot approve it directly, so I'll leave it to your expertise.

That was on 13 August and nothing has moved since. Because it was submitted as a comment rather than an approval or a change request, GitHub cleared his review request while `reviewDecision` stayed at `REVIEW_REQUIRED` — so the PR is off everybody's queue and cannot become mergeable until somebody approves it. There is nothing for me to answer here: it is not a question, it is a deferral.

If reproducing it is the obstacle for you too, say so and I will write up the exact steps rather than leave it sitting.

dgpaci (dismissed):
Approved. I checked the diagnosis line by line against the code rather than just reading the description, and it holds up.

### Verified independently

- `dataNode.setValue(resolver, true, attributes)` → `setAttr(attr, /*trigger*/false, undefined)` → `this.attr = attr` (`gnrbag.js:256`, `gnrbag.js:457-464`): the object is assigned verbatim, so `node.attr` was literally the same object as the domsource's local `attributes`.
- `Bag.deepCopy` copies attributes with `objectUpdate({}, node.attr)` (`gnrbag.js:1514`), so the `GnrDomSourceNode` reference does travel into the record cluster.
- `_toXmlBlock` hands the whole `getAttr()` to `xml_buildTag` (`gnrbag.js:498`), and `objectAsXmlAttributes` calls `asTypedTxt` on every non-string attribute (`gnrlang.js:800-811`) — functions are skipped there, nodes are not.
- The recursion is exactly as described: `guessDtype` returns `BAGNODE` (`gnrlang.js:1370`), `toTypedJSON` falls into the `!['L','N','NN','B']` branch (`gnrlang.js:1509`), and `convertToText` finds no `toXml` on a node (only `Bag` has one) so it hits `t == 'object'` → `toTypedJSON` on the same value (`gnrlang.js:1479`).
- The fix cannot break the resolver: `_sourceNode` is never read back from a node's attributes — no hit for `attr._sourceNode` / `attr['_sourceNode']` / `getAttr('_sourceNode')` anywhere in `gnrjs/`, `gnrpy/gnr/web/` or `projects/`. The resolver keeps its own copy (`genro_rpc.js:698`) and `GnrBagResolver.resolve` rebuilds kwargs from `this.kwargs` (`gnrbag.js:2372`), never from the node.
- The "dataRemote was the only branch" claim holds: `remote_relOneResolver`, `remote_relManyResolver`, `setRemoteContent` and `genro_widgets.js:4842` all attach their resolver through `setResolver`, so none of them writes a source node into node attributes.

One side effect worth noting: `node.attr` now holds a private copy instead of aliasing the domsource's working dict. Nothing depended on that aliasing (`remoteResolver` already copied), so it is a small bonus rather than a risk.

### Non-blocking remarks

1. **Consistency with the `datarpc` branch.** Fifty lines above, the framework idiom is `objectExtract(kwargs, '_*')` (`gnrdomsource.js:441`), which clears every technical key. Here only `_sourceNode` is removed, so other `_*` attributes still reach the node — and in the dynamicform case they are persisted into the record's Bag column: `_ask` (read from `this.attr._ask` but never popped from the copy) and `_onResult` / `_onCalling` when present (the `GnrRemoteResolver` constructor pops them from its own copy only, `genro_rpc.js:47-48`). They are strings today, so harmless, but the next `_*` attribute holding an object reintroduces the same class of bug. `objectExtract(nodeAttributes, '_*')` would be more future-proof — provided no `_`-prefixed attribute is actually needed on the data node.

2. **The follow-up has no tracking issue.** I searched the open and closed issues: nothing covers the `convertToText` / `toTypedJSON` mutual recursion. Worth opening it now, otherwise the `FILE` case stays a client-killing stack overflow. Confirmed that `convertToText` handles neither `FILE` nor `SET`, so the hole is real.

3. **No issue linked to this PR.** The bug was reported from outside (icond project) and there is no traceability. A `[BUG]` issue referenced from the commit would match the convention.

4. **Verification gap.** There is no JS test harness in the repository (no jest/karma/`package.json` for `gnrjs`), so a test is not something I can ask for here. But since the crash only reproduces on the reporter's instance, confirmation from there would be valuable before merge: that saving now goes through, and that the getter still resolves *after* the save — that is precisely the path this patch touches.

5. **No legacy data to clean up** — worth stating explicitly in the description: serialization always crashed before reaching the server, so no `_sourceNode` was ever written to the database.

genro (changes requested):
The diagnosis holds in every step — I checked it on `f15711986` rather than reading the description, and each link of the chain is where you say it is: `this.attr = attr` verbatim (`gnrbag.js:465`), `BAGNODE` from `guessDtype` (`gnrlang.js:1387`), `toXml` defined once and only on `Bag`, so `convertToText` falls into `t == 'object'` and `toTypedJSON` bounces straight back on the same value. `_sourceNode` is read back from a node's attributes nowhere in `gnrjs/`, `gnrpy/gnr/web/`, `projects/` or `resources/`, and the resolver keeps its own copy at `genro_rpc.js:697`. The fix is right.

Requesting changes for one reason only, and it makes the patch smaller rather than bigger.

**The copy is not needed.** At the top of `setDataNodeValueDo` (`gnrdomsource.js:487`):

```js
var attributes = objectUpdate({}, this.attr);
```

`attributes` is already a private copy of the node's own attributes — nothing outside this call shares it, and `remoteResolver` has already taken its own copy by the time we get here. So `nodeAttributes` copies a copy. This does the same thing:

```js
objectExtract(attributes, '_*');
dataNode.setValue(resolver, true, attributes);
```

One line instead of three, no extra object, and it is exactly the idiom the `datarpc` branch uses fifty lines above (`gnrdomsource.js:441`). I checked that `attributes` is not read after this point in the function, so emptying it of the technical keys has no other effect.

Using `objectExtract('_*')` rather than deleting `_sourceNode` alone also closes @dgpaci's first remark in the same stroke. Today `_ask`, and `_onResult`/`_onCalling` when present, still travel onto the data node — and in the dynamicform case that node lives inside the record's Bag column, so they are persisted. They are strings, so harmless now; the point is that the next `_*` attribute holding an object walks straight back into this crash. Removing one key fixes this bug, removing the class of keys fixes the shape.

If any `_`-prefixed attribute turns out to be genuinely needed on a data node, say so and the narrow `delete` is fine as it stands — the request is for the simplification, not against the fix.

Two of @dgpaci's other remarks I would rather see acted on than carried:

- **The `convertToText` / `toTypedJSON` recursion needs its own issue.** I searched too and there is none. `FILE` is the realistic trigger and it kills the client the same way, wherever such a value lands. This PR is right to leave it out; leaving it untracked is the part that costs.
- **Confirmation from the reporter's instance before merge.** Neither @cgabriel, nor @dgpaci, nor I can reproduce it, and there is no JS harness in the repo, so the only real verification available is Andrea Busanelli saving that form — and, specifically, the getter still resolving *after* the save, since that is the path this patch touches.

On the process point in your comment: you are right that a `COMMENTED` review cleared the request while leaving `reviewDecision` at `REVIEW_REQUIRED`, which is how this sat for a week. Nothing to answer there.

fporcari:
> **The copy is not needed.** [...] `attributes` is already a private copy of the node's own attributes [...] So `nodeAttributes` copies a copy.
> — [review of 24 Aug](https://github.com/genropy/genropy/pull/1037#pullrequestreview-5007798132)

Accepted, and I re-checked both premises rather than taking them: `attributes` is
built as `objectUpdate({}, this.attr)` at `gnrdomsource.js:271`, and
`remoteResolver` takes its own copy with `objectUpdate({'sync':true}, params)` at
`genro_rpc.js:700` — before the extract runs, so the resolver's kwargs keep
`_sourceNode`. `attributes` is not read again after `dataNode.setValue(...)`; the
branch ends there.

Done on `d2a3259fcec70a657a39a007fb835d600de57dd4` — three lines to one, no extra
object (`gnrdomsource.js:496-499`):

```js
var resolver = genro.rpc.remoteResolver(method, attributes, {'cacheTime':cacheTime,
    'isGetter':isGetter});
// the resolver already took its own copy of the parameters: a GnrDomSourceNode
// left in node.attr breaks any serialization of that node (e.g. a record cluster).
objectExtract(attributes, '_*');
dataNode.setValue(resolver, true, attributes);
```

I took `objectExtract('_*')` rather than the narrow `delete`, so @dgpaci's first
remark is closed in the same stroke. Two checks behind that choice:

- `objectExtract(obj, '_*')` does remove the keys from `obj` — `delete obj[prop]`
  whenever `dontpop` is falsy (`gnrlang.js:586-597`).
- No `_`-prefixed attribute is needed on a data node. `_ask` is the only
  candidate and every read of it is `this.attr._ask`, on the domsource node, not
  on the data node (`gnrdomsource.js:227`, `232`, `233`, `235`, `254`).

## The follow-up is now tracked

#1138. I mapped the loop before opening it, so the issue states which values
actually reach it rather than just naming `FILE`: the cycle is
`toTypedJSON` → `JSON.stringify(asTypedTxt(v))` (`gnrlang.js:1506-1508`) →
`convertToText(v, {'xml':true})` (`gnrlang.js:920`) → `t == 'object'` fallthrough
→ `['JS', toTypedJSON(v)]` (`gnrlang.js:1478-1481`), on the same `v`.

The dangerous set is smaller than it looks, which is worth having written down:
`AR`, `SET` and `OBJ` are destructured by `toTypedJSON` before the fallthrough,
and `asTypedTxt` returns early on functions (`gnrlang.js:917-919`). What is left
is `BAGNODE` — the crash this PR removes at the source — and `FILE`. So after
this PR the reachable trigger is a `File` landing anywhere that gets serialized.

## Still open

- **Traceability**: this PR still has no linked issue, per @dgpaci's third
  remark. The report came in from the icond project.
- **Confirmation from the reporter's instance**: unchanged, and the only real
  verification available — that the form saves, and that the getter still
  resolves *after* the save, which is the path this patch touches.

Note that this push dismissed @dgpaci's approval (`dismiss_stale_reviews` on
`develop`), so it needs re-approving on the current head — nothing changed in the
diagnosis it approved, only the three lines above.

fporcari:
@dgpaci I have withdrawn your review request rather than asking you to re-approve.

Your 21 Aug approval was dismissed automatically by the push to
`d2a3259fcec70a657a39a007fb835d600de57dd4` (`dismiss_stale_reviews` on
`develop`), but nothing you verified changed: the diagnosis, the call chain and
the "`_sourceNode` is never read back" search all still hold. The push only
replaced three lines with one, and it also closes the first remark of your own
review by extracting the whole `_*` class instead of deleting a single key.

For the record, CODEOWNERS assigns `gnrjs/` and `*.js` to @fporcari and @genro,
so your approval was never a required one here — asking you for a second read of
a diff that shrank by two lines would have been the wrong use of your time.

@genro still has the PR, and the remaining points are in the previous comment.

genro (approved):
Approving `d2a3259fc`. Checked on the branch rather than on the description:

- the branch ends at `objectExtract(attributes, '_*'); dataNode.setValue(resolver, true, attributes);` — one line where there were three, no extra object
- `objectExtract` does remove from the source: `delete obj[prop]` whenever `dontpop` is falsy (`gnrlang.js:591`)
- every read of `_ask` is on the domsource node's own `this.attr` (`gnrdomsource.js:227-254`), none on a data node
- `attributes` is not read again after `setValue` in that branch

Taking the whole `_*` class rather than the single key is the right call: it removes the shape of the bug and not just this instance, and it lines the `dataRemote` branch up with `datarpc` fifty lines above.

#1138 is a better issue than the one I asked for. Bounding the dangerous set — `AR`, `SET` and `OBJ` destructured before the fallthrough, functions returning early, leaving `BAGNODE` and `FILE` — turns "some values crash the client" into a list someone can actually work from, and records that after this PR `FILE` is the only reachable trigger left.

**One correction, and it is about us rather than this PR.** You wrote that CODEOWNERS assigns `gnrjs/` and `*.js` to @fporcari and @genro. For a `.js` file it does not: the last matching pattern wins, and the last line is `*.js @fporcari`, so `gnrjs/ @fporcari @genro` is shadowed for exactly the files this PR touches. You are the author and the only code owner of this diff, which means nobody can satisfy `require_code_owner_reviews` on it — my approval does not, since I do not own `.js`. It merges only because `enforce_admins` is off on `develop`.

This is the asymmetry I raised on #1093, and this PR is the live instance of it. Worth adding `@genro` to the `*.js` line before that one merges — it is one word, and #1093 is still sitting approved and unmerged.

The two open points stay yours to close: no issue linked to this PR for traceability, and confirmation from the icond instance — that the form saves, and that the getter still resolves after the save. I would not merge without the second.

PR #1038, opened 2026-08-06
fix(daemon): addSiteRegister raises UnboundLocalError when the site is already registered

### LINKED ISSUE #1031: `UnboundLocalError: sitename in addSiteRegister when site is already registered`
### Description

When a site register is requested for a site that is **already registered**, the daemon
raises `UnboundLocalError`. The exception propagates through Pyro to every client calling
`getSite`, which then loops on retries.

Client-side log (repeats for each worker):

```
WARNING: gnr.web/siteregister_client - getSite failed for asp4: cannot access local
variable 'sitename' where it is not associated with a value, retrying...
```

Downstream effects observed in the same instance:

```
CommunicationError: cannot connect to ('localhost', 42263): [Errno 111] Connection refused
ConnectionClosedError: receiving: not enough data
```

### Cause

`gnrpy/gnr/web/daemon/handler.py`, `addSiteRegister` (lines 281–315):
`sitename` is assigned **only inside the `if` branch**, but it is referenced in the `else`
branch:

```python
def addSiteRegister(self, domainIdentifier, storage_path=None, autorestore=False, port=None):
    if not domainIdentifier in self.siteregisters:
        # Extract sitename from domainIdentifier (format: sitename|domain or just sitename)
        sitename = domainIdentifier.split('|')[0] if '|' in domainIdentifier else domainIdentifier
        ...
    else:
        logger.info("Site %s already existing", sitename)   # <-- never assigned here
```

The `logger.info` line was correct while the parameter itself was named `sitename`; commit
`6a320d0fc` ("Transitory refactoring of daemon package and modules", #688) renamed the
parameter to `domainIdentifier` and moved the derivation inside the `if`, leaving the
reference in the `else` branch orphaned.

### Proposed fix

Move the assignment above the `if`, so it is defined in both branches:

```python
def addSiteRegister(self, domainIdentifier, storage_path=None, autorestore=False, port=None):
    # Extract sitename from domainIdentifier (format: sitename|domain or just sitename)
    sitename = domainIdentifier.split('|')[0] if '|' in domainIdentifier else domainIdentifier
    if not domainIdentifier in self.siteregisters:
        ...
```

fporcari:
## Problem

When a site register is requested for a site that is already registered, the daemon raises `UnboundLocalError`. The exception propagates through Pyro to every client calling `getSite`, which then loops on retries:

```
WARNING: gnr.web/siteregister_client - getSite failed for asp4: cannot access local variable 'sitename' where it is not associated with a value, retrying...
```

followed downstream, in the same instance, by:

```
CommunicationError: cannot connect to ('localhost', 42263): [Errno 111] Connection refused
ConnectionClosedError: receiving: not enough data
```

## Root cause

`gnrpy/gnr/web/daemon/handler.py`, `addSiteRegister`:

- `sitename` was derived only inside the `if not domainIdentifier in self.siteregisters:` branch.
- The `else` branch (site already registered) referenced `sitename` in `logger.info("Site %s already existing", sitename)`, but on that path `sitename` was never assigned.

The reference in the `else` branch was left orphaned by commit `6a320d0fc` ("Transitory refactoring of daemon package and modules", #688), which renamed the parameter from `sitename` to `domainIdentifier` and moved the derivation inside the `if`.

## Change

Move the `sitename` derivation above the `if`, so it is defined on both paths. The explanatory comment moves with it. Nothing else in the method changes; `sitename` is still used, unchanged, inside the `if` branch (`hasSysPackageAndIsPrimary(sitename)`, task scheduler process name).

## Verification

- `flake8` on the modified file: zero errors.
- Full `gnrpy/tests/web/` suite: 142 passed, 57 skipped. The skips are pre-existing (`BaseGnrDaemonTest.setup_class` calls `pytest.skip("gnr CLI not available in PATH")` when the `gnr` CLI/daemon is not on `PATH` in this environment); unrelated to this change.
- Added a regression test (`gnrpy/tests/web/gnrdaemon_addsiteregister_test.py`) that calls `addSiteRegister` unbound against a minimal hand-rolled fake exposing only `siteregisters`, exercising just the "already registered" branch (no live daemon needed). Confirmed it reproduces `UnboundLocalError` against the unpatched method and passes against the fix.
- Not verified live: an actual running daemon receiving a real duplicate `getSite`/register call over Pyro (no daemon available in this environment).

Fixes #1031


cgabriel (approved):
LGTM, simple fix for this mistake I've made, sitename used in the logger entry out of scope. My bad!

genro (approved):
Confirmed in the current tree: `sitename` is derived inside the `if` (`handler.py:283`) while the `else` logs it (`:314`), so the already-registered path is a guaranteed `UnboundLocalError` — and since it travels back through Pyro to every `getSite` caller, the retry loop and the follow-on connection errors in the report are consistent with this single cause.

Hoisting the derivation above the `if` is the minimal fix and leaves the in-branch uses (`hasSysPackageAndIsPrimary(sitename)`, the scheduler process name) untouched. Nothing between the new position and the old one can change `domainIdentifier`.

The regression test exercising the method unbound is the right call — `GnrDaemon.__init__` would start a multiprocessing Manager, and the branch under test reads nothing but `siteregisters`. The second case pinning the `|domain` suffix is a good addition, since that split is the only logic in the hoisted line.

PR #1039, opened 2026-08-06
fix(sys): thpage rejects a malformed url instead of raising an unpacking error

### LINKED ISSUE #1034: `thpage` raises an unpacking error instead of rejecting a malformed URL
`thpage: "too many values to unpack (expected 2)" on URLs with extra segments`

### Description

A malformed URL routed to `thpage` produces an opaque `page_init` error rather than a clean
404. Seen in production ~25 times over five days, hitting real users:

```
error_type: page_init
description: too many values to unpack (expected 2)
request_uri: /sys/thpage/<pkg>/:/_storage/documenti/<folder>/<id>/<id>.pdf
```

The URL originates from a relative-URL resolution in the browser: an `iframe` whose `src`
lacked a scheme/host is resolved against the current page path, producing a `/sys/thpage/…`
URL with many extra segments.

### Cause

`projects/gnrcore/packages/sys/webpages/thpage.py`, `onIniting` (lines 39–42) handles only
2 or 3 arguments, but unpacks unconditionally:

```python
if len(request_args)==3:
    pkg,tbl,pkey = request_args
else:
    pkg,tbl = request_args      # fails with any other length
```

Note the guard just above already handles the "too few" case with a clear message
(`if len(request_args) < 2: raise ValueError('Missing table arguments in URL: …')`); only
the "too many" case is unhandled.

### Proposed fix

Symmetric validation, so the URL is rejected with a meaningful message (or a 404) instead
of an `unpack` traceback:

```python
if len(request_args) > 3:
    raise ValueError('Too many arguments in URL: %s' % '/'.join(request_args))
```

### Related (possibly a separate issue)

The malformed `src` itself comes from `StorageService.internal_url`
(`gnrpy/gnr/lib/services/storage.py:513-517`):

```python
external_host = self.parent.external_host.rstrip('/')
outlist = [external_host, '_storage', self.service_name]
url = '/'.join(outlist)
```

`external_host` (`gnrpy/gnr/web/gnrwsgisite.py:1175`) falls back to the configured
`wsgi?external_host` whenever there is no current page with a request — which is exactly
the case for **scheduled tasks**. If that configuration entry is missing or has no scheme,
the persisted URL is not usable from a browser, and nothing warns about it. Validating the
configured value (or logging when the fallback yields a scheme-less host) would prevent
storing broken URLs.

fporcari:
## Problem

Requesting a table handler page with a malformed URL crashes with an opaque
`too many values to unpack (expected 2)` traceback instead of a clear
rejection. In production this surfaces as a `page_init` error.

## Root cause

`projects/gnrcore/packages/sys/webpages/thpage.py`, `onIniting`:

```python
def onIniting(self, request_args, request_kwargs):
    if len(request_args) < 2:
        raise ValueError('Missing table arguments in URL: %s' % '/'.join(request_args))
    pageResource = request_kwargs.get('th_pageResource')
    if len(request_args)==3:
        pkg,tbl,pkey = request_args
    else:
        pkg,tbl = request_args
```

The "too few arguments" case has a clear guard with a meaningful message.
There is no equivalent guard for "too many arguments": any `request_args`
longer than 3 items falls into the `else` branch and the
`pkg, tbl = request_args` unpacking raises the unpacking error above.

One known source of such a URL: an `iframe` whose `src` lacks a
scheme/host. The browser resolves it against the current page path, so a
storage path (e.g. `/_storage/documenti/<folder>/<id>/<id>.pdf`) gets
appended after `/sys/thpage/<pkg>/`, producing a URL like
`/sys/thpage/<pkg>/:/_storage/documenti/<folder>/<id>/<id>.pdf` with far
more than 3 segments. This pattern shows up in production logs roughly 25
times over five days.

## Change

Added a symmetric guard for the "too many arguments" case, mirroring the
existing guard's style and message shape:

```python
if len(request_args) > 3:
    raise ValueError('Too many table arguments in URL: %s' % '/'.join(request_args))
```

The URL now fails with an explicit `ValueError` instead of an unpacking
traceback. No other behavior changes.

## Verification

- `flake8` on `projects/gnrcore/packages/sys/webpages/thpage.py` and the
  new test file: zero errors.
- Real regression test added at `gnrpy/tests/web/test_thpage_url_validation.py`.
  `onIniting`'s guards run before any framework mixin work, so the module
  can be loaded directly from the file system (same pattern already used
  by `gnrpy/tests/web/test_s3_temporary_filename.py`) and exercised with
  the real, unmodified guard code — no mocking of the logic under test.
  Covers: too few arguments (existing guard), too many arguments (new
  guard, using the exact malformed URL shape from this issue), and the
  2-argument / 3-argument valid cases (no regression).
- Full suite: `python -m pytest gnrpy/tests/web/ -q` → 144 passed, 57
  skipped, no failures. Also ran via the pre-commit hook against the full
  test suite (1923 passed, 67 skipped).
- Manual live-URL check (`/sys/thpage/<pkg>/<table>`,
  `/sys/thpage/<pkg>/<table>/<pkey>`, and the malformed
  `/sys/thpage/<pkg>/:/_storage/a/b/c.pdf`) was **not** performed: no
  GenroPy instance was running locally in this environment. Verified by
  reading plus the automated test above instead.

## Open question

Should this be a `404` instead of a `ValueError`? The `ValueError` mirrors
the existing "too few arguments" guard for consistency, but a `404` might
be more semantically correct for a malformed resource path. Leaving this
to the reviewer's judgment; happy to switch if preferred.

## Follow-up (separate issue, not addressed here)

The malformed URL itself is generated upstream in
`StorageService.internal_url` (`gnrpy/gnr/lib/services/storage.py`, around
lines 513-517), which builds the URL from `self.parent.external_host`.
`external_host` (`gnrpy/gnr/web/gnrwsgisite.py:1175`) falls back to the
configured `wsgi?external_host` whenever there is no current page/request
— which is exactly the case for scheduled tasks. If that config entry is
missing a scheme, a broken URL gets persisted with nothing warning about
it. Worth its own issue/fix; not addressed by this PR.

Fixes #1034


cgabriel (approved):
The fix is fine for me. Regarding your open question:

```
Should this be a 404 instead of a ValueError? The ValueError mirrors
the existing "too few arguments" guard for consistency, but a 404 might be more semantically correct for a malformed resource path. Leaving this to the reviewer's judgment; happy to switch if preferred.
```

I'm suggesting instead to return a HTTP 400 - providing the wrong parameters to a resource is indeed a bad request.

genro (approved):
Verified against the real code paths:

- With 4+ segments today, `pkg, tbl = request_args` is what raises the opaque unpacking error — the new guard is the symmetric twin of the existing `< 2` one, same message shape, and its `ValueError` follows the same route (`onIniting` exceptions are caught at `gnrwebpage.py:240-243` into `_page_init_error`, then `main()` renders the errorPane). Behaviour for the valid 2/3-segment URLs is untouched.
- The test loads the real module and exercises the unmodified guard code; the stubs only cover the mixin calls that run after the guards. Imports are clean.

On the open question (ValueError vs 404): keeping `ValueError` is right here, for symmetry with the existing guard. A true 404 would have to move both guards to `onPreIniting` (exceptions from `onIniting` never reach the dispatcher), which is exactly the shape #1012 introduces for sys/default — worth a small follow-up that migrates both thpage guards there, so scanners hitting these malformed URLs stop getting a 200 with an errorPane. Not blocking.

PR #1040, opened 2026-08-06
fix(web): a Bag returned by a page is no longer mistaken for a WSGI response

### LINKED ISSUE #1032: A `Bag` returned by a page is mistaken for a WSGI response
### Description

Requests to a table handler page fail with a confusing `TypeError`. Reproduced in
production by a regular user on `/sys/thpage/<pkg>/<table>` (7 occurrences in one day),
and on a local development instance:

```
TypeError: Bag.__call__() takes from 1 to 2 positional arguments but 3 were given
  File "gnr/web/gnrwsgisite.py", line 1157, in dispatcher
    return self._dispatcher(environ, start_response)
  File "gnr/web/gnrwsgisite.py", line 1379, in _dispatcher
    return response(environ, start_response)
```

In the recorded frame, `response` is a `Bag`.

### Cause

`gnrpy/gnr/web/gnrwsgisite.py`, `setResultInResponse` (line 1410) treats any callable
result as a WSGI response:

```python
elif callable(result):
    response = result
```

`Bag` defines `__call__(self, what=None)` (`gnrpy/gnr/core/gnrbag.py:827`), so it passes
the `callable()` test. It is then invoked as `response(environ, start_response)` — three
arguments where at most two are accepted.

Returning a `Bag` is the normal convention for Genropy methods, so any code path where a
`Bag` reaches this branch produces an unrelated, hard-to-diagnose error.

### Proposed fix

Exclude `Bag` from the callable branch, e.g.:

```python
elif callable(result) and not isinstance(result, Bag):
    response = result
```

A clearer alternative would be to check explicitly for a WSGI-callable / `Response`
instance rather than relying on `callable()`.

fporcari:
## Problem

Requests to a table handler page (`/sys/thpage/<pkg>/<table>`) failed with a
confusing `TypeError`, both in production (7 occurrences in one day, regular
user) and locally:

```
TypeError: Bag.__call__() takes from 1 to 2 positional arguments but 3 were given
  File "gnr/web/gnrwsgisite.py", line 1157, in dispatcher
    return self._dispatcher(environ, start_response)
  File "gnr/web/gnrwsgisite.py", line 1379, in _dispatcher
    return response(environ, start_response)
```

## Which entry point delivers a raw Bag

`GnrWebPage.get_call_handler` (`gnrpy/gnr/web/gnrwebpage.py:345`) picks the
handler from the request, and the two RPC entry points do not behave the same:

- **`?method=…`** returns `_rpcDispatcher` (`:358`), which ends by serialising
  through `self.rpc.result_<mode>` (`gnrwebpage_proxy/rpc.py:47`). A `Bag` is
  wrapped in an envelope and never reaches the dispatcher tail.
- **`?rpc=…`** returns the public method *itself* as `_call_handler` (`:354`).
  `__call__` returns its value unchanged (`:623`) and `_dispatcher` passes it
  straight to `setResultInResponse` (`gnrwsgisite.py:1356` → `:1377`). There is
  no serialisation step on this path.

So on the `rpc=` path a `@public_method` returning a `Bag` — the ordinary
Genropy convention — reaches `setResultInResponse` raw. This is not a caller
error: the same method is correct on both paths.

The failure is also deferred, which is why the production occurrences were hard
to attribute: the `Bag` is returned here as if it were the WSGI application and
only fails later at `response(environ, start_response)` (`gnrwsgisite.py:1379`),
so the traceback names `Bag.__call__` and the dispatcher and never the page or
the method that produced it.

## Root cause

`GnrWsgiSite.setResultInResponse` is the tail of the dispatch chain that
turns a page's return value into the WSGI response:

```python
elif isinstance(result, Response):
    response = result
elif callable(result):
    response = result
return response
```

`Bag` defines `__call__(self, what=None)`, so a `Bag` passes `callable()` and
is handed back as if it were the WSGI application itself. It is then invoked
as `response(environ, start_response)` -- three positional arguments where
`Bag.__call__` accepts at most one -- raising the `TypeError` above. Returning
a `Bag` from a page/rpc method is the normal Genropy convention, so any code
path that ends up here with a `Bag` hits this.

`callable()` is the wrong test: this branch means "the result IS the WSGI
application to call", and `callable()` is only a proxy for that -- true for
`Bag` and for any other Genropy object that happens to define `__call__`,
not just for WSGI apps.

## Change

Narrowed the branch to exclude `Bag` explicitly:

```python
elif callable(result) and not isinstance(result, Bag):
    response = result
```

Two options were considered:

- **(a)** exclude `Bag` (taken) -- minimal, matches the reported case.
- **(b)** narrow the branch to what it actually means ("this is a WSGI
  callable") and let anything else fall through to `return response`.

(b) was rejected for this PR. Grepping the dispatch paths that can return a
callable meant to be invoked as a WSGI app surfaced a mixed set:
  - `paste.fileapp.FileApp` responders (`storage.py`, `gnrstatichandler.py`,
    `compressed.py`) are always invoked immediately at their own call site
    (`file_responder(environ, start_response)`); the bare callable never
    reaches `setResultInResponse`, so they are not actually affected either way.
  - `GnrWsgiSite.forbidden_exception` / `not_found_exception`-style bound
    methods (signature `(self, environ, start_response, debug_message=None)`)
    are returned as-is from the mako/genshi page plugins (e.g.
    `gnrwebpage_plugin/mako.py`, `gnrwebpage_plugin/genshi.py`:
    `return self.page.site.forbidden_exception`, and a caught
    `HTTPException` re-returned as `return exc`) and genuinely need to reach
    this branch.
  - werkzeug `HTTPException` instances are callable with the same
    `(environ, start_response)` signature and can also legitimately end up
    here.

Those real cases are bound methods and exception instances with no common
trait besides "takes `(environ, start_response)`", distinct from
`Bag.__call__(self, what=None)` only by arity. Telling them apart from a
`Bag` in general would need an arity/type check that isn't exercised
anywhere today, so it stays out of this fix. Excluding `Bag` specifically
closes the reported hole without touching any of that.

## Verification

- `flake8` on `gnrpy/gnr/web/gnrwsgisite.py` and the new test file: zero errors.
- `python -m pytest gnrpy/tests/web/ -q`: 145 passed, 57 skipped (daemon-only
  tests, expected in this environment), 0 failed.
- Added `gnrpy/tests/web/gnrwsgisite_result_response_test.py`: calls
  `GnrWsgiSite.setResultInResponse` unbound on a light fake site (same style
  as `gnrwsgisite_folder_cleanup_test.py`), using a real `Bag`, a real
  werkzeug `Response`, and a real werkzeug `HTTPException` -- no mocks. Pins
  the branch order: `str` sets response data, `bytes` sets response data, a
  `Bag` result returns the response object (not the Bag -- the regression
  case), a `Response` result replaces the response, and a genuine WSGI
  callable (`HTTPException` instance) still passes through unchanged.
- Full local suite was not run clean end-to-end in this environment:
  Postgres-backed SQL tests fail to boot here (`postmaster became
  multithreaded during startup`, a local Homebrew/macOS Postgres issue) and
  a handful of locale/date tests fail on `babel.core.UnknownLocaleError:
  unknown locale 'c'`; both reproduce identically on the unmodified base
  branch and are unrelated to this change.

Fixes #1032

## Follow-up (not fixed here)

- `setResultInResponse` has a partially dead branch:
  `elif isinstance(result, (bytes,str)):` only ever matches `bytes`, since
  `str` is already handled by the branch above it (`isinstance(result, str)`).
- Option (b) above: narrowing `callable(result)` to only the WSGI-callable
  cases that are actually produced today (`HTTPException` instances and
  `GnrWsgiSite`'s own exception-returning bound methods), so any other
  Genropy object defining `__call__` doesn't fall into the same trap as `Bag`.


cgabriel:
I've read the whole PR text but to me it's not clear what was causing the problem in the first place. Which code raised the described error? It was a plain error made by the developer? Maybe a documentation problem? Can you please provide a less narrower context?

cgabriel (commented):
I've read the whole PR text but to me it's not clear what was causing the problem in the first place. Which code raised the described error? It was a plain error made by the developer? Maybe a documentation problem? Can you please provide a less narrower context?

fporcari:
Fair question, and the description was too narrow: it explained the mechanism inside `setResultInResponse` without saying which entry point delivers a raw `Bag` to it. Neither a developer mistake nor a documentation gap — the two RPC entry points differ, and only one of them serialises.

## The two paths

`GnrWebPage.get_call_handler` (`gnrpy/gnr/web/gnrwebpage.py:345`) picks the handler from the request:

```python
elif 'rpc' in request_kwargs:              # :354
    self._call_handler_type = 'externalCall'
    return self.getPublicMethod('rpc', request_kwargs.pop('rpc'))
elif 'method' in request_kwargs:           # :358
    self._call_handler_type = 'pageCall'
    return self._rpcDispatcher
```

- **`?method=…`** returns `_rpcDispatcher`, which ends by serialising: `result_handler = getattr(self.rpc, 'result_%s' % mode.lower()); return result_handler(result)` (`gnrwebpage.py:636` onwards, `result_bag` at `gnrwebpage_proxy/rpc.py:47`). Whatever the method returned is wrapped in an envelope before it leaves. A `Bag` never reaches the dispatcher tail on this path.
- **`?rpc=…`** returns the public method **itself** as `_call_handler`. `GnrWebPage.__call__` then does `result = self._call_handler(*args, **kwargs)` (`:623`) and returns it unchanged, and `_dispatcher` passes that straight to `setResultInResponse` (`gnrwsgisite.py:1356` → `:1377`). There is no serialisation step on this path at all.

So on the `rpc=` path the public method's return value is handed to `setResultInResponse` raw. Returning a `Bag` from a `@public_method` is the ordinary convention, which is why this is not a caller error: the method is correct, and it is correct on the other path too.

## Why it surfaces as a `TypeError` and not as something legible

`Bag.__call__(self, what=None)` exists, so the `Bag` satisfies `callable(result)` and is returned as if it were the WSGI application. The failure is then deferred to `response(environ, start_response)` at `gnrwsgisite.py:1379`, three positional arguments into a callable that accepts at most one. Hence a traceback that names `Bag.__call__` and the dispatcher, and never mentions the page or the method that produced it — which is why the production occurrences were so hard to attribute.

## Why the guard rather than a wider fix

The branch means "the result *is* the WSGI application"; `callable()` was only ever a proxy for that. `Bag` is the case that bites today, but any Genropy object defining `__call__` and reaching this line has the same problem.

Two wider fixes I considered and did not take here:

1. **Invert the test** — accept only what is genuinely a WSGI app instead of excluding known non-apps. Correct, but this branch currently catches werkzeug `HTTPException` instances and the site's own `forbidden_exception`/`not_found_exception` returns, and I could not convince myself I had enumerated all of them from the repo. An `isinstance` allowlist that misses one turns a working error page into a blank response.
2. **Serialise on the `rpc=` path too**, aligning it with `method=`. That is arguably the real asymmetry, but it changes the wire format of an externally-callable endpoint — not something to slip into a bugfix.

If you prefer either, say which and I will open it as its own issue rather than widening this PR. What is in here stops the 500s without changing any behaviour that works today.

I will fold this explanation into the PR description so it is not buried in a thread.

genro (changes requested):
The diagnosis is right and the archaeology behind it is worth keeping: `Bag.__call__(self, what=None)` accepts zero or one positional argument and the dispatcher tail invokes `response(environ, start_response)` with two, which is the traceback exactly. And the two RPC entry points really do differ — `?method=` ends in `self.rpc.result_<mode>` while `?rpc=` returns the public method itself and hands its value straight to `setResultInResponse`, with no serialisation step in between. A `@public_method` returning a `Bag` is ordinary Genropy, so this is not a caller error.

**Requesting changes because the guard removes the crash by removing the response.**

`setResultInResponse` (`gnrwsgisite.py:1410-1423`) is a four-branch chain and then `return response`:

```python
if isinstance(result, str):             ...
elif isinstance(result, (bytes,str)):   ...
elif isinstance(result, Response):      response = result
elif callable(result):                  response = result
return response
```

With `and not isinstance(result, Bag)` a `Bag` matches **none** of them, so the function returns the `response` object it was handed, untouched. And that object is the page's own empty response: the caller at `:1375` builds it before the call and this function is the only thing that fills it. There is no other serialiser on the `rpc=` path — the PR body establishes that itself, as the reason the `Bag` arrives raw.

So the outcome changes from `TypeError` — a 500 with a traceback naming `Bag.__call__` — to a **200 with an empty body**. The seven production occurrences in one day stop being reported, and the feature still does not work. That is a worse failure mode than the one being fixed: a loud wrong answer becomes a silent one.

The branch the `Bag` should reach is a serialising one, not none. Either give it its own branch before the `callable` test — `response.data = result.toXml()` with the mimetype the `method=` path uses — or route the `rpc=` result through the same `self.rpc.result_<mode>` that already exists, which would also make the two entry points agree instead of documenting that they differ.

The `isinstance` guard is still needed either way: `Bag` being callable is what makes it fall into the WSGI-app branch in the first place, and the comment explaining that is worth keeping verbatim.

One thing worth adding to the test while you are in there: assert the response **body**, not only that no exception is raised. A test that checks "does not raise" passes on the empty-body outcome too, which is how this shape of fix survives review.

fporcari:
Right, and the guard alone was the wrong shape of fix. Addressed in 4c51167.

> [Requesting changes because the guard removes the crash by removing the response.](https://github.com/genropy/genropy/pull/1040#pullrequestreview-5018790285) [...] So the outcome changes from `TypeError` [...] to a **200 with an empty body**.

Confirmed exactly as described: with `and not isinstance(result, Bag)` a `Bag` matched none of the four branches, `setResultInResponse` returned the caller's untouched empty response, and `:1386` handed that to `response(environ, start_response)`. The seven production occurrences would have stopped being reported with the feature still broken.

## What changed

A `Bag` now gets its own branch before the `callable` test:

```python
elif isinstance(result, Bag):
    response.mimetype = kwargs.get('mimetype') or 'text/xml'
    response.data = result.toXml(unresolved=True, omitUnknownTypes=True)
```

That is your first option, and it is deliberately the same serialisation the `method=` path already applies through `GnrWebPageRpc.result_xml` (`gnrwebpage_proxy/rpc.py:99-103`) — `text/xml` plus `toXml(unresolved=True, omitUnknownTypes=True)`. A page forcing its own mimetype still wins, as on the `str` branch.

## Why not the second option

Routing `rpc=` through `self.rpc.result_<mode>` would make the two entry points agree, but not compatibly: `result_bag` wraps the value in an envelope carrying `result`, `dataChanges`, `error` and every `envelope_*` attribute of the page (`rpc.py:47-86`), and `rpc=` callers receive the bare value today. That is a wire-format change on a public entry point, so it belongs in its own PR with a deprecation path, not in a fix for #1032. If you want the entry points unified rather than merely agreeing on serialisation, say so and I will open the issue.

## One deviation from your review

You asked to keep the `isinstance` guard and its comment verbatim. With the `Bag` branch placed first the guard is unreachable, so I dropped it and moved the explanation onto the new branch, where it now documents why the branch order matters. The `callable` branch keeps a one-line note naming what it is actually for.

## Tests

Your point about "does not raise" was the load-bearing one: `test_bag_result_returns_response_not_bag` asserted `result is response`, which passes on the empty-body outcome. Replaced with three that assert the body:

- `test_bag_result_is_serialised_into_the_response` — body equals `bag.toXml(...)`, contains the serialised nodes, mimetype `text/xml`;
- `test_bag_result_honours_a_forced_mimetype`;
- `test_empty_bag_result_still_has_a_body` — pins the failure mode itself.

All three **fail on the previous head** and pass here, so they pin the production change rather than the branch. `tests/web/gnrwsgisite_result_response_test.py`: 7 passed. `tests/web`: 222 passed, 57 skipped, against 220 passed / 57 skipped on the previous head — the delta is the two extra tests. flake8 clean on both files.

Not verified in this round: no live instance was exercised. The `rpc=` path is covered only by these unit tests calling `setResultInResponse` unbound, as before.

genro (approved):
This answers the objection. Flipping my earlier request for changes.

The `Bag` branch now sits before the `callable` test (`gnrwsgisite.py:1415-1424`) and fills the response instead of returning it untouched, with the same serialisation the `method=` path applies through `GnrWebPageRpc.result_xml` (`rpc.py:99-103`): `text/xml` plus `toXml(unresolved=True, omitUnknownTypes=True)`. `kwargs.get('mimetype')` still wins, as on the `str` branch.

You were right to drop the `isinstance` guard rather than keep it: with the `Bag` branch placed first it is unreachable, and the explanation reads better on the branch that does the work.

Also right to leave the entry-point unification out. `result_bag` wraps the value in an envelope carrying `dataChanges`, `error` and the page's `envelope_*` attributes, and `rpc=` callers get the bare value today — that is a wire-format change on a public entry point and does not belong in a fix for #1032. Open the issue for it when you get a moment.

The three body-asserting tests are the right replacement for `result is response`, and `test_empty_bag_result_still_has_a_body` pins the exact failure mode the previous head would have shipped.

Ran `gnrpy/tests/web` on `4c51167`: 205 passed, 74 skipped.

Issue #380, opened 2026-01-08
[TASK] Add exception safety to dispatcher with try/finally


fporcari:
## Summary

Add try/finally block to `dispatcher()` to ensure proper cleanup on exceptions.

## Objective

Ensure that thread-local state is properly reset even when exceptions occur in `_dispatcher()`. Currently, if an exception is raised and caught, thread-local variables may leak to the next request on the same thread.

## Scope

**In scope:**
- Add `finally` block to `dispatcher()` method
- Reset `currentRequest` and call `cleanup()` in finally block

**Out of scope:**
- Changes to `_dispatcher()` logic
- Other exception handling improvements

## Technical Details / Notes

**Affected file:** `gnrpy/gnr/web/gnrwsgisite.py`

**Current code (line ~1059):**
```python
def dispatcher(self, environ, start_response):
    self.currentRequest = Request(environ)
    self.currentRequest.max_form_memory_size = 100_000_000

    try:
        return self._dispatcher(environ, start_response)
    except Exception as e:
        page = self.currentPage
        if self.debug and ((page and page.isDeveloper()) or self.force_debug):
            raise
        self.writeException(exception=e, traceback=tracebackBag())
        exc = HTTPInternalServerError(...)
        return exc(environ, start_response)
```

**Proposed improvement:**
```python
def dispatcher(self, environ, start_response):
    self.currentRequest = Request(environ)
    self.currentRequest.max_form_memory_size = 100_000_000

    try:
        return self._dispatcher(environ, start_response)
    except Exception as e:
        page = self.currentPage
        if self.debug and ((page and page.isDeveloper()) or self.force_debug):
            raise
        self.writeException(exception=e, traceback=tracebackBag())
        exc = HTTPInternalServerError(...)
        return exc(environ, start_response)
    finally:
        # Ensure cleanup even on exception paths
        self.currentRequest = None
```

## Dependencies

- Related to #379 (Memory Leak in Thread-Local Storage)

## Acceptance Criteria

- [ ] `finally` block added to `dispatcher()`
- [ ] Thread-local state reset on all exit paths
- [ ] Existing tests still pass

## Additional Information

**Estimated effort:** 20 minutes

This is a defensive programming improvement that prevents potential cross-request data contamination in thread-pooled environments.

PR #1042, opened 2026-08-06
fix(web): stop re-adding the thread-local currentDomain entry after cleanup

fporcari:
## Problem

Issue #380 asks to add a `try/finally` around `GnrWsgiSite.dispatcher()` that resets `currentRequest` and calls `cleanup()`, so thread-local state does not leak to the next request served on the same thread when `_dispatcher()` raises.

## Root cause

That protection already exists on `develop`. What remained was a one-line residue: right after `cleanup()` popped the per-thread `currentDomain` entry, `dispatcher()`'s `finally` block re-set it to `rootDomain`, silently re-adding a `{tid: '_main_'}` entry to the underlying `ThreadedDict` that is never removed again. Every thread ident that ever served one request keeps that entry forever — the same unbounded thread-local growth issue #379 fixed for `currentRequest`/`currentPage`, just left in place for `currentDomain`.

## Why #380 is already implemented

- The `try/finally` requested by #380 was added by commit `f77dddcf7` ("fix: improve error handling for non-developer users and file uploads"). `dispatcher()`'s `finally` block already calls `self.cleanup()`.
- `cleanup()` (`gnrpy/gnr/web/gnrwsgisite.py`) already resets `currentPage`, `currentRequest`, `currentAuxInstanceName` and `currentDomain` to `None`.
- The sibling issue #379 was closed by PR #396, which turned these four into `ThreadedDict`-backed properties: assigning `None` pops the current thread's entry instead of storing it, so `cleanup()`'s resets are real pops, not leaks.
- `_currentDomains` is only ever touched through the `currentDomain` property (`self._currentDomains.get()/.set()`); no code in the repo inspects the backing dict directly. The getter (`return self._currentDomains.get() or self.rootDomain`) can never return `None`, since `rootDomain` is a fixed non-empty string (`'_main_'`) set once in `__init__` and never reassigned. Re-setting `currentDomain` after `cleanup()` was therefore both redundant and the one remaining leak of the #379 class.

Given this, #380's ask is functionally done on `develop` already; this PR only removes the residual line that was quietly re-opening the leak #379 closed.

## Change

One line removed from `dispatcher()`'s `finally` block, with a comment explaining why it must stay removed:

```python
finally:
    self.cleanup()
    # Do not re-set currentDomain here: cleanup() already reset it
    # to None, which pops this thread's entry from the underlying
    # ThreadedDict. Assigning rootDomain again would re-add a
    # {tid: '_main_'} entry that is never removed, i.e. the same
    # unbounded thread-local growth fixed for currentRequest/
    # currentPage in #379. The currentDomain getter already falls
    # back to rootDomain when unset, so no re-assignment is needed.
```

Also added:
- `gnrpy/tests/web/gnrwsgisite_dispatcher_cleanup_test.py`: calls `GnrWsgiSite.dispatcher()` unbound on a fake site that reuses the real `currentPage`/`currentRequest`/`currentAuxInstanceName`/`currentDomain` properties and `cleanup()`/`raiseIfDeveloper()`, backed by real `ThreadedDict` instances — no mocking of the thread-local mechanism itself. Covers the success path, the exception path, and the fallback-to-`rootDomain` behavior after cleanup. Verified this test fails (catches the regression) when the removed line is temporarily reintroduced.
- `gnrpy/tests/core/gnrlang_test.py`: `ThreadedDict` had no dedicated test anywhere; added coverage for the default-`None` get, set/get roundtrip, pop-on-`None`, and isolation across real `threading.Thread`s — the load-bearing piece of the #379 fix.

## Verification

- `flake8` on all touched files: zero errors.
- `python -m pytest gnrpy/tests/web/ -q`: 143 passed, 57 skipped (skips are pre-existing, daemon/Postgres not available in this environment — unrelated to this change).
- Full suite via the repository's pre-commit hook: 1926 passed, 67 skipped, 0 failed.
- Confirmed by temporarily reintroducing the removed line that the new dispatcher test fails, proving it actually exercises the leak.

## Follow-up (not in this PR)

Same leak class, still open elsewhere: `site.currentPage = page … = None` without a `try/finally` at `gnrpy/gnr/web/daemon/processes.py:317` and `:325`, `gnrpy/gnr/web/gnrtask.py:137`, and `gnrpy/gnr/web/gnrtask_new.py:668`.

---

Given the above, should #380 be closed as already implemented on `develop`, with this PR credited as the residual fix?

### DIFF
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -1162,7 +1162,13 @@ class GnrWsgiSite(object):
             return exc(environ, start_response)
         finally:
             self.cleanup()
-            self.currentDomain = self.rootDomain
+            # Do not re-set currentDomain here: cleanup() already reset it
+            # to None, which pops this thread's entry from the underlying
+            # ThreadedDict. Assigning rootDomain again would re-add a
+            # {tid: '_main_'} entry that is never removed, i.e. the same
+            # unbounded thread-local growth fixed for currentRequest/
+            # currentPage in #379. The currentDomain getter already falls
+            # back to rootDomain when unset, so no re-assignment is needed.
 
     def raiseIfDeveloper(self, exception=None):
         page = self.currentPage
--- a/gnrpy/tests/core/gnrlang_test.py
+++ b/gnrpy/tests/core/gnrlang_test.py
@@ -277,3 +277,62 @@ def test_gnrImport_concurrent_single_module_identity(tmp_path):
     assert len(modules) == 8
     assert all(m is modules[0] for m in modules)
     assert modules[0].Service is not None
+
+
+# ---------------------------------------------------------------------------
+# ThreadedDict
+# ---------------------------------------------------------------------------
+
+
+def test_threaded_dict_get_default_is_none():
+    td = gl.ThreadedDict()
+    assert td.get() is None
+
+
+def test_threaded_dict_set_then_get_roundtrip():
+    td = gl.ThreadedDict()
+    td.set('value')
+    assert td.get() == 'value'
+
+
+def test_threaded_dict_set_none_pops_entry():
+    """Assigning None removes the current thread's entry instead of
+    storing it, so the backing dict does not keep a live-but-empty slot
+    for every thread that ever called set()."""
+    td = gl.ThreadedDict()
+    td.set('value')
+    assert len(td._data) == 1
+    td.set(None)
+    assert td.get() is None
+    assert len(td._data) == 0
+
+
+def test_threaded_dict_isolated_across_threads():
+    """Each thread sees only the value it set, keyed by its own ident.
+
+    A thread that exits without explicitly resetting to None leaves its
+    entry behind: ThreadedDict itself does no bookkeeping on thread exit,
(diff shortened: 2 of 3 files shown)


cgabriel (approved):
The approach is correct, and the new tests coverage are fine for me.

genro (approved):
Verified the whole chain in this checkout:

- `ThreadedDict.set(None)` pops the thread's entry (`gnrlang.py:662-664`), `cleanup()` assigns None to all four slots (`gnrwsgisite.py:1442-1448`), and the `currentDomain` getter already falls back to `rootDomain` when unset (`gnrwsgisite.py:1865`). So the removed line was pure residue: it re-added a `{tid: '_main_'}` entry that nothing ever popped, the same leak shape as #379.
- Nothing downstream reads `_currentDomains` directly, so the fallback covers every reader; the assignment at the top of `dispatcher` (`:1157`) stays and is what a real request needs.

The dispatcher test is the right shape — reusing the real properties and `cleanup()` unbound, faking only `_dispatcher`/`errorHandler`/`db` — and it covers both the success and the exception path, so the line cannot come back unnoticed. The `ThreadedDict` unit tests are a welcome addition: the isolation one documents that popping is the caller's duty, which is exactly the invariant this fix relies on.

PR #1044, opened 2026-08-06
fix(webstruct): field() ignores a column's values when the dtype is numeric

### LINKED ISSUE #78: filteringSelect values with "field"
When a columns like this is defined:

```
tbl.column('feature_status', name_long='!![it]Status', dtype='N', 
                            values='1:[!![it]Non presente],2:[!![it]In sviluppo],3:[!![it]Attiva]')
``` 

I expect to use _field_ and automatically see a `filteringSelect` with specified values:

`fb.field('feature_status')
`

But nothing happens! I have to use _field_ specifying the values again:

```
fb.field('feature_status', tag='filteringSelect', 
                     values='1:[!![it]Non presente],2:[!![it]In sviluppo],3:[!![it]Attiva]')
```

This could lead to mistakes and inconsistencies

fporcari:
## The rule

**It is the presence of `values` in the model that selects the widget — not the dtype, and not the shape of the string.**

Declaring `values` on a column says that the field must be filled by choosing among those values. That is a constraint on the data, and the widget that expresses it is the filteringSelect, whatever the dtype and whether or not key and caption happen to coincide. A developer who wants something else says so at the call site — `fb.field(col, tag='comboBox')` — and the model's store still arrives.

This is the wider scope @dgpaci asked for in his review, and it replaces the earlier one-line dtype widening this PR opened with.

## Root cause

`gnrpy/gnr/web/gnrwebstruct/dojo11.py`, inside `wdgAttributesFromColumn()` (reached from `fb.field()` → `prepareFieldAttributes()`):

```python
elif dtype in ('A', 'T') and fldattr.get('values', False):
    ...
    result['tag'] = 'filteringselect' if ':' in values else 'combobox'
```

Two accidents, one on top of the other.

The `('A', 'T')` tuple is not a design decision — it was born *inside* the `dtype == 'A'` branch and lifted to `('A', 'T')` by 7cfa00917 (2010), whose message is simply "so it works for variable size text fields". With `dtype='N'` the guard is false, the dtype cascade assigns `numberTextBox`, and `values` is never copied into the result at all — so the obvious first workaround fails too:

```python
fb.field('feature_status', tag='filteringSelect')   # renders a select with an EMPTY store
```

which is why #78 reports having to repeat `values=` as well.

The colon test is the same category of accident, and the older one: a closed list does not stop being a closed list because key and caption coincide. A dojo ComboBox is a **free-text** widget; on a bigint column it lets the user type `abc`.

The sibling resolvers show the form path as the odd one out:

| resolver | guard on `values` |
|---|---|
| `_helpers.py:70-74` (grid cells) | none — propagates `values` for every dtype |
| `genro_wdg.js:236-239` (`widgetFromField`, client) | no dtype guard, but the same colon test |
| `dojo11.py` (Python form path) | `('A', 'T')` **and** the colon test |

## Change

The guard stops asking "which dtypes admit `values`" and starts asking "which dtypes own their widget regardless", and the colon test goes:

```python
#: A column declaring `values` is a closed set of choices, so `field()`
#: resolves it to a filteringSelect whatever its dtype. These dtypes are the
#: exception: a boolean is a checkBox and a Bag is a tree, neither of which is
#: a list of choices, so their own widget wins over the store.
DTYPES_IGNORING_VALUES = ('B', 'X')

...
elif fldattr.get('values', False) and dtype not in DTYPES_IGNORING_VALUES:
    values = fldattr['values']
    values = getattr(fieldobj.table.dbtable, values, lambda: values)()
    fldattr['values'] = values
    result['tag'] = 'filteringselect'
    result['values'] = values
```

**The exception list is `B` and `X`, and nothing else.** A checkBox already *is* the closed two-value domain, and a `values` store cannot drive it; a Bag rendered as a tree is not a scalar choice at all. Every other dtype's own widget is free scalar entry, which a declared choice list legitimately overrides — `D` and `H` included: a date column with three allowed dates is a choice list. `R` needs no special case any more, which dissolves the unexplained exclusion the first version of this PR carried.

Relation columns are out of scope and stay so: the relcol branch runs first and produces a `DbSelect`, which is right — the related table *is* the value source.

**The JS twin moves with it**, so the two resolvers keep agreeing (`gnrjs/gnr_d11/js/genro_wdg.js`):

```js
if('values' in fieldattr && dt!='B' && dt!='X'){
    result.values = fieldattr.values;
    result.tag = 'filteringselect';
}
```

The client path gains the `B`/`X` exception it never had — today `values` hijacks a checkBox there, the same class of bug in the opposite direction.

**The grid caption path does not move.** `gnrjs/gnr_d11/js/genro_grid.js:1842` colon-tests only to decide whether captions need translating, and with no colon there is nothing to translate. Correct as is, left alone.

Nothing else is needed to render the colon-less case: `storeFromValues` already sets `id = caption = val` when the entry carries no colon (`genro_widgets.js:4387-4394`) — verified in the source, not assumed.

## Release note

**A column declaring `values` now renders as a select in `field()`, whatever its dtype**, and a colon-less `values` list is a select too, not a ComboBox.

The teeth of that: **a FilteringSelect only accepts what is in its store, a ComboBox accepts anything.** For a legacy row holding a value outside the declared list, `patch_setValue` resolves through `store.fetchItemByIdentity`, gets no item and takes the null branch — `self.valueNode.value = null; self.setDisplayedValue('')` (`genro_widgets.js:5070-5074`). **The field renders empty where a ComboBox showed the stored value.** `connectAddMissingValue`, which papers over exactly this, is ComboBox-only by its own comment: "Limited to ComboBox: dbSelect/FilteringSelect/dbComboBox have their own validation rules" (`genro_widgets.js:5115-5120`).

That is a declared constraint that starts being enforced, which is the point — but it is visible on upgrade and belongs in the notes rather than being discovered. @dgpaci's AST scan over genropy plus four project roots (anaci, erpy, genropy_projects, gnrextra) measures 327 columns declaring `values`, of which **55 are colon-less literals** that switch from ComboBox to FilteringSelect (24 more resolve `values` through a table method, so their runtime string is unknown statically). None of the sampled ones wants free text — `sys.task.day` is a cron day, 1..31 or nothing.

Applications that used `values` as documentation, or that want free text with suggestions, opt out at the call site once — `fb.field(col, tag='comboBox')` — and keep the store for free.

Display of the *valid* case holds on numeric columns and not by luck: `FilteringSelect.patch_setValue` resolves through `store.fetchItemByIdentity` and the base compares with **loose** equality, `if (this.getIdentity(bagnode) == id)` (`gnrjs/gnr_d11/js/gnrstores.js:430`), while an `N`/`L` column arrives client-side as a JS Number — so `'2' == 2` matches and the caption renders.

## The save-path risk — tested, and mostly benign

A `filteringSelect` writes the **string** store key into the form bag, and `_doChangeInData` does no dtype coercion (`gnrjs/gnr_d11/js/genro_widgets.js:120-123`). So a numeric column receives the string `'2'` on save. Display was never in doubt; the write path had no precedent in this repo, so it was probed against a real sqlite `test_invoice` database rather than assumed.

Traced first: there is **no server-side dtype coercion** anywhere on the form-save path. XML→Bag conversion is driven by the client-declared `_T` (`gnrbagxml.py:176-180`), and the client sends no `_T` for a JS string (`gnrlang.js:1428-1430`); `saveRecordCluster` (`_gnrbasewebpage.py:462`), `writeRecordCluster` (`gnrsqltable/crud.py:569`), `db.insert` (`gnrsql/write.py:134`) and the adapter's `prepareRecordData` (`_gnrbaseadapter.py:728-758`) all pass the value through untouched. `SqlTable.recordCoerceTypes` (`gnr/sql/gnrsqltable/record.py:48-100`) does exactly the right conversion, but its only callers are the bulk-import paths, never the form save.

Measured, inserting through the real table API and reading back (sqlite):

| dtype | value written | read back |
|---|---|---|
| `N` (`discount_tier.min_amount`) | `Decimal('2')` | `Decimal('2')` |
| `N` | `'2'` — what the select writes | **`Decimal('2')`** — correct |
| `N` | `''` (empty selection) | `None` |
| `N` | `'abc'` | inserts, then **read raises** `decimal.InvalidOperation` |
| `L` (`price_year.year`) | `2025` | `2025` |
| `L` | `'2025'` — what the select writes | **`2025`** (int) — correct |
| `L` | `'abc'` | **silently stored as `'abc'`** in a bigint column |

For well-formed numeric store keys — the issue's case — the round trip is correct on both numeric affinities. The residual exposure is a *misconfigured* column whose keys are not numeric (`values='a:Foo,b:Bar'` on a numeric column). Note that dropping the ComboBox on this path **removes** the free-text hazard rather than adding it: a FilteringSelect cannot submit a value outside the store, so `abc` is no longer typeable into a bigint column.

**Still untested, honestly:** PostgreSQL (it does not boot in this environment; from the code the client-side literal composition in `gnrpostgres.py:157-166` / `gnrpostgres3.py:341-348` should coerce a well-formed numeric literal and raise `invalid input syntax` otherwise), and the full browser round trip. The value was injected at the table-insert boundary, which the trace above shows is byte-identical to what the form save delivers, but no browser was driven.

## Verification

- **flake8** on both touched Python files (`dojo11.py`, `gnrwebstruct_test.py`): **zero errors**.
- **`gnrpy/tests/web/gnrwebstruct_test.py`: 57 passed** (47 on the previous head).
- **`gnrpy/tests/` excluding `sql/`: 956 passed, 59 skipped.** The `sql/` suite does not run in this environment: `testing.postgresql` cannot start a server (`initdb failed: could not create shared memory segment`), which is unrelated to this diff and pre-existing.
- **The changed tests fail without the production change** — checked by reverting `dojo11.py` alone and re-running: **11 failed, 46 passed**, and exactly the parametrizations that encode the rule being changed.

Test changes, all against the real resolver with the existing `_ColumnStub` / `_DbTableStub` scaffolding (nothing mocked inside `wdgAttributesFromColumn`):

- `test_values_without_colon_resolves_to_combobox` → **inverted** into `test_values_without_colon_still_resolves_to_filteringselect`, over `A T N L I R D H`. As written it enshrined the rule being removed, and for `N`/`L`/`I` it would have created a new status quo rather than preserving one (develop gives NumberTextBox there today).
- `test_non_numeric_non_text_dtypes_ignore_values` → **narrowed** to `test_widget_owning_dtypes_ignore_values` (`B` → checkBox, `X` → tree), also asserting no `values` leaks into the result. `D`/`H` moved to the permissive side, per the exception list above.
- **new** `test_every_other_dtype_with_values_resolves_to_filteringselect` over `A T N L I R D H` with a colon-carrying store, asserting the tag, the store and that the column keeps its own dtype.
- **new** `test_call_site_can_ask_for_a_combobox_without_repeating_values`, pinning the documented opt-out: `tag='comboBox'` wins and the model's store still arrives.

Fixes #78

## Minor, separate — not folded in

- `getColumnPrintWidth` (`gnrpy/gnr/sql/gnrsqltable/columns.py:155`) tests `['A','C','T','X','P']`, so a text `values` column gets width 12 while a numeric one falls through to `guessLen(dtype)` — "In sviluppo" is 11 characters in a 4em column. Pre-existing, becomes the next visible symptom after this merges. Worth its own issue.
- `docs/genro/webpage_elements/widgets/form_widgets/field.rst:48` carries a stale copy of this branch. Worth its own issue.


cgabriel (dismissed):
The diff is ok for me, the "widen the guard" option looks to me as the right option - js behaviour alignment, first of all.

I'd be happy to review again if other choices are being made.

genro (dismissed):
Agreed with cgabriel on option 1, and there is now a measurement behind it rather than a judgement call.

The one thing that made this a decision was the silent widget change for numeric `values` columns in application code outside this repo. I searched every indexed project through Sourcerer (all repos, not just genropy) for a numeric-dtype column declaring `values`, in the `dtype='N'/'L'/'I'` and quote variants:

**exactly one hit** — `gnrwork/model/feature_competitor.py:15`, `feature_status` with the `1:Non presente,2:In sviluppo,3:Attiva` store. That is the column the issue was opened about. Nothing else in the estate declares this combination, so the "form widget switches under an unsuspecting app on upgrade" scenario has no known instance: the only app that gets the new behaviour is the one asking for it.

That also settles the option-2 tradeoff: propagating `values` without the auto-tag would leave the reported case still needing `tag=` at the call site, buying safety nobody needs.

Keeping the tuple explicit rather than testing `values` alone is right — `'B'` and `'X'` have widgets of their own and must not be hijacked. The residual hazard the body is candid about (non-numeric store keys on a numeric column, silent on `L`) predates this change and is a misconfiguration, not a shape this makes reachable by accident.

The tests are the first coverage `wdgAttributesFromColumn` has ever had, they drive the real resolver, and the `KeyError: 'values'` case pins the doubled workaround this closes. The stale docs snippet in `field.rst:48` is worth its own cleanup, as noted.

fporcari:
Realigned with `develop` on `b4c3e6d92`. @cgabriel @genro: your approvals of 20
August fell with the push (`dismiss_stale_reviews`) and need re-approving on the
current head — nothing in the change you approved has moved, so this is the whole
of what happened.

The PR was `DIRTY`, and there is no way to merge it without resolving that, so the
dismissal was unavoidable rather than chosen.

The conflict was in `gnrpy/tests/web/gnrwebstruct_test.py` and it was purely
positional — both sides appended to the same place in the file, with nothing in
common:

- this branch adds `_DbTableStub` / `_TableStub` / `_ColumnStub` / `_wdgattr` and
  the seven `field()` dtype-versus-`values` tests;
- `develop` added `_fieldNode` / `_resolveChildPath` / `_hiddenFieldLabelCell` /
  `_hiddenFormbuilder` and the six `hidden`-propagation tests.

No shared helper name, no shared test name, so both sides are kept as they were.
`gnrpy/gnr/web/gnrwebstruct/dojo11.py` merged with no conflict.

Verified on the merge commit: `pytest tests/web/gnrwebstruct_test.py` → **47
passed**, which is both sides' tests including the six that arrived from
`develop`; flake8 clean; full suite **2120 passed, 69 skipped**.

@dgpaci your review request from 7 August is still open. The two codeowner
approvals above are what branch protection needs here — CODEOWNERS gives
`gnrpy/` to @fporcari and @genro — so say if you would rather be dropped from it
than carry it.

genro (dismissed):
Re-approving on the current head. Nothing in the change moved: it is still the one line

```python
elif dtype in ('A', 'T', 'N', 'L', 'I') and fldattr.get('values', False):
```

widening the dtype set for which a column's `values` produces a select widget. My approval of 20 August fell with the realign to `develop`, per `dismiss_stale_reviews` on the base — not because anything was questioned.

The test stub is worth keeping as written: resolving `values` through `getattr(dbtable, values, lambda: values)()` means a literal string and a method name go down different paths, and `_DbTableStub` covers both by carrying `statusValues` and not carrying the literal.

fporcari:
Withdrawing @cgabriel's review request here, so the queue does not carry a request nobody
owes an answer to.

He approved this on 20 August, at `661bad62`. Two `Merge branch 'develop'` commits have
landed since and `dismiss_stale_reviews` dropped that approval both times, but nothing he
read has changed: the production diff (`gnrpy/gnr/web/gnrwebstruct/dojo11.py`, two lines)
is byte-identical to what he approved, and the 132 test lines this PR adds are the same
ones. The `hidden hides the label cell` block that shows up when comparing the two heads
arrived from `develop`, not from this branch.

@genro approved on the current head, so the codeowner requirement is met. Asking cgabriel
to re-approve an unchanged diff twice over is the request I am dropping, not his review.

@dgpaci stays: this changes what `field()` emits for a numeric column, which is widget
behaviour, and that is worth a pair of eyes that will actually click it.

dgpaci (changes requested):
Thanks for the depth here — the root-cause archaeology and the save-path measurements are exactly what this needed, and I verified your numbers hold: the bug reproduces, the fix works, `tests/web/gnrwebstruct_test.py` gives **47 passed** on the current head and flake8 is clean on both files.

My request for changes is about **scope**, not quality. The diff fixes half of the problem and, in the other half, pins the opposite rule into the test suite.

## The rule

**It is the presence of `values` in the model that selects the widget — not the dtype, and not the shape of the string.**

Declaring `values` on a column says one thing: this field must be filled by *choosing among those values*. That is a constraint on the data, and the widget that expresses it is the filteringSelect — whether the keys are numbers, strings, or letters is irrelevant. So `field()` must default to a filteringSelect **in every case** where the model declares `values`.

If a developer wants something else — because they use `values` only as documentation, or they want free text with suggestions — they say so at the call site: `fb.field(col, tag='comboBox')`, or they place the widget directly. That is the right place for an interface decision. What must not happen is the *framework* silently deciding, from a syntactic detail, that a declared constraint does not apply.

Your diff applies this rule to the dtype axis and stops there. The line still reads:

```python
result['tag'] = 'filteringselect' if ':' in values else 'combobox'
```

That colon test is the same category of accident as the `('A', 'T')` tuple your body dissects — arguably the older one. A closed list does not stop being a closed list because key and caption happen to coincide.

## What must not be merged as it stands

```python
@pytest.mark.parametrize('dtype', ['A', 'T', 'N', 'L', 'I'])
def test_values_without_colon_resolves_to_combobox(dtype):
    """`values` with no key:caption separator is a plain suggestion list."""
```

This enshrines the rule above as intended behaviour, and for `N`/`L`/`I` it does not merely preserve the status quo — it creates a new one. Measured with a real instance, a real table and a real `fb.field()`, injecting `values='1,2,3'` on a real column:

| dtype | develop today | this PR |
|---|---|---|
| `A` | ComboBox | ComboBox |
| `L` | **NumberTextBox** | **ComboBox** |
| `N` | **NumberTextBox** | **ComboBox** |

A dojo ComboBox is a free-text widget. On a bigint column it lets the user type `abc`, which by your own measurement table stores silently as `'abc'` on sqlite. Your body classifies that hazard as pre-existing and reachable only through misconfiguration; on this path it is neither — a colon-less suggestion list is a perfectly ordinary declaration, and the widget is new.

## What the diff becomes

Simpler, not more complex — the dtype tuple mostly disappears, and so does the colon test. The guard stops asking "which dtypes admit `values`" and starts asking "which dtypes own their widget regardless". That also dissolves the `R` question: `R` is excluded from your tuple with no stated reason, which reads as an oversight next to the explicit rationale you give for `B` and `X`. Under the rule it needs no special case.

Two things travel with the change:

- **the JS twin must move too.** `widgetFromField` applies the identical colon test at `gnrjs/gnr_d11/js/genro_wdg.js:236-239`. Alignment with the JS resolver is the load-bearing argument of your PR body — leaving the two paths to diverge on the colon would undercut it.
- **the grid path must not.** `gnrjs/gnr_d11/js/genro_grid.js:1842` colon-tests only to decide whether captions need translating, and with no colon there is nothing to translate. Correct as is.

## Blast radius, measured

AST scan over every model in genropy plus four project roots (anaci, erpy, genropy_projects, gnrextra): **327 columns declare `values`**, of which **55 are colon-less literals** and 24 more resolve `values` through a table method, so their runtime string is unknown statically. Those 55 switch from ComboBox to FilteringSelect. A sample:

```
Sì,No,Non So                                  asso.paziente.familiarita_sla
bug,feature,support,question                  gbs.ticket.ticket_type
low,medium,high,critical                      gbs.ticket.urgency
Classica,Elettronica,Pubblica Amministrazione asso.cliente_fornitore.tipo_fatt
sezione,divisione,gruppo,classe,categoria     erpy_base.ateco.livello
div,span,h1,h2,h3                             adm.ckstyle.element
1..31 / 0..23 / 0..59                         sys.task.day / .hour / .minute
```

Not one of them wants free text. `sys.task.day` in particular is a cron day: 1..31 or nothing.

Nothing else is needed to render them — `storeFromValues` already handles the colon-less case by setting `id = caption = val` (`genro_widgets.js:4376`).

**The one consequence worth stating in the release notes**: a FilteringSelect only accepts what is in its store, a ComboBox accepts anything — and genro leans on exactly that difference. `connectAddMissingValue` exists for the ComboBox alone, and its own comment says so: "Limited to ComboBox: dbSelect/FilteringSelect/dbComboBox have their own validation rules" (`genro_widgets.js:5115-5120`).

Concretely, for a legacy row holding a value outside the declared list: `patch_setValue` calls `fetchItemByIdentity`, gets no item, and takes the null branch — `self.valueNode.value = null; self.setDisplayedValue('')` (`genro_widgets.js:5070-5074`). The field renders **empty**, where a ComboBox showed the stored value. That is the intended teeth of the rule — a declared constraint that starts being enforced — but on those 55 columns it is visible on upgrade and should be announced rather than discovered. It is the mirror image of the risk your body weighs: not a widget changing under an app, but a value that stops displaying.

## One decision left to you

"Any dtype", read literally, includes `B` and `X`. A `B` column with `values='0:No,1:Yes'` — checkBox or select? `X` is a Bag/tree. The exception list should be explicit and justified in the code, the way you justified `B` and `X` in the body. Relation columns are out of scope and should stay so: the relcol branch runs first and produces a `DbSelect`, which is right — the related table *is* the value source.

Consequently `test_non_numeric_non_text_dtypes_ignore_values` needs re-deciding for whichever dtypes end up on that list, and `test_values_without_colon_resolves_to_combobox` needs inverting.

## Closing your open question on the browser round trip

You flagged the display path as untested. It holds, and not by luck. `FilteringSelect.patch_setValue` resolves the value through `store.fetchItemByIdentity` (`genro_widgets.js:5078`), and the base implementation compares with **loose equality**: `if (this.getIdentity(bagnode) == id)` (`gnrjs/gnr_d11/js/gnrstores.js:430`). A `N`/`L` column arrives client-side as a JS Number (`gnrlang.js:968-983`), so `'2' == 2` matches and the caption renders. Had that been `===`, numeric `values` columns would have rendered blank. The grid caption path coerces explicitly (`(currvalue+'').split(',')`, `genro_grid.js:1842`), so it matches too.

## Reproduction, for the record

Real instance (`gnrdevelop`), `site.dummyPage`, real tables, real `fb.field()`. `values` injected on a real `dtype='L'` column — the same attribute shape a model declaration produces, checked against `docu.documentation.revision` which declares them:

| case | develop | this PR |
|---|---|---|
| text + `values` (control) | `<FilteringSelect values="...">` | unchanged |
| numeric + `values` | `<NumberTextBox>`, no `values` | `<FilteringSelect values="...">` |
| numeric + `tag='filteringSelect'` | `<FilteringSelect>`, **empty store** | `<FilteringSelect values="...">` |
| numeric + `tag=` + `values=` repeated | works | works |

Issue #78 is real and still open, and your reading of the doubled workaround is confirmed: forcing only the tag yields an empty store. One extra detail — in the forced case the node carries `places="0::L"` and `_guess_width="7em"` leftovers from the numeric branch; your change removes them.

To answer the question you put to me: no, the column does not have to be numeric for any deep reason — it is `feature_status` in `gnrwork`, the single occurrence @genro found across the estate. But that does not make this a documentation gap. The model declared a constraint and the form ignored it; that is a framework gap whichever dtype it happens on.

## This PR closes #78 — the body should say so

The body says `Refs #78`, which does not close anything. Please make it `Fixes #78`.

This is not bookkeeping. #78 was opened on 2025-01-10 and has sat for nineteen months with zero comments; it is the reason this work exists, and it is the report of a real defect that this PR — once the scope above is settled — removes. If it merges under `Refs`, the issue survives the merge and stays in the queue describing a bug that no longer exists. Whoever picks it up next re-does the archaeology your body already did.

Same for the release note: what ships here is "a column declaring `values` now renders as a select in `field()`, whatever its dtype", and #78 is the case that names it.

## Minor, separate

- **Your resolver inventory is missing a fourth guard.** `getColumnPrintWidth` (`gnrpy/gnr/sql/gnrsqltable/columns.py:155`) tests `['A','C','T','X','P']`, so a text `values` column gets width 12 while a numeric one falls through to `guessLen(dtype)`. Measured on the same column: `A` → 12 → 7em in the grid; `N` → 10 → 6em; `L` → 8 → 4em; `I` → 7 → 4em. "In sviluppo" is 11 characters in a 4em column. Pre-existing, since the grid path is already dtype-agnostic, but it becomes the next visible symptom right after this merges — worth an issue.
- **`field.rst:48`** — agreed it is separate, and it sits inside a `CLIPBOARD::` block rather than user-facing prose. Worth opening the issue now, though: a cleanup noted in a PR body evaporates on merge.

dgpaci:
Since I am asking for a change of scope, here is the change, implemented and measured, so the request is concrete rather than a direction. Take it or leave it — the shape is the point, not the authorship.

## Production diff

```diff
--- a/gnrpy/gnr/web/gnrwebstruct/dojo11.py
+++ b/gnrpy/gnr/web/gnrwebstruct/dojo11.py
@@ -30,6 +30,12 @@
 from gnr.web.gnrwebstruct._helpers import _selected_defaultFrom
 
 
+#: A column declaring `values` is a closed set of choices, so `field()`
+#: resolves it to a filteringSelect whatever its dtype. These dtypes are the
+#: exception: their own widget is not a list of choices at all, so it wins.
+DTYPES_IGNORING_VALUES = ('B', 'X')
+
+
 class GnrDomSrc_dojo_11(GnrDomSrc):
@@ -1055,11 +1061,11 @@
-        elif dtype in ('A', 'T', 'N', 'L', 'I') and fldattr.get('values', False):
+        elif fldattr.get('values', False) and dtype not in DTYPES_IGNORING_VALUES:
             values = fldattr['values']
             values = getattr(fieldobj.table.dbtable, values ,lambda: values)()
             fldattr['values'] = values
-            result['tag'] = 'filteringselect' if ':' in values else 'combobox'
+            result['tag'] = 'filteringselect'
             result['values'] = values
```

and the JS twin, so the two resolvers keep agreeing:

```diff
--- a/gnrjs/gnr_d11/js/genro_wdg.js
+++ b/gnrjs/gnr_d11/js/genro_wdg.js
@@ -233,9 +233,9 @@
-        if('values' in fieldattr){
+        if('values' in fieldattr && dt!='B' && dt!='X'){
             result.values = fieldattr.values;
-            result.tag = fieldattr.values.indexOf(':')>=0?'filteringselect':'combobox';
+            result.tag = 'filteringselect';
         }
```

Note the JS side gains the `B`/`X` exception it never had: today `values` hijacks a checkBox there, which is the same class of bug in the opposite direction. That is a behaviour change on the client path and belongs in the release note next to the other one.

## Measured, real instance and real `fb.field()`

`GnrWsgiSite('gnrdevelop')` + `site.dummyPage`, real table, `values` injected on a real column:

| dtype | `values` with `:` | `values` without `:` |
|---|---|---|
| `A` `T` | FilteringSelect | FilteringSelect |
| `N` `L` `I` `R` | FilteringSelect | FilteringSelect |
| `D` `H` | FilteringSelect | FilteringSelect |
| `B` | checkBox | checkBox |
| `X` | tree | tree |

The escape hatch works and no longer needs `values=` repeated, which is the other half of #78:

```
fb.field(col, tag='numberTextBox')  ->  <NumberTextBox values="1:...,2:...,3:...">
fb.field(col, tag='comboBox')       ->  <ComboBox      values="1:...,2:...,3:...">
```

So a developer who wants `values` to be documentation only, or wants free text with suggestions, states it once at the call site and the store still arrives.

`D`/`H` land on the permissive side here — under the rule a declared choice list wins, and a date with three allowed values is a choice list. That is the one line of the exception set I would not decide for you: if you want them to keep their own widget, they go in the tuple and the reason goes in the comment next to `B` and `X`.

## Tests

Against this diff, your suite fails 7 of 47 — and exactly the 7 that encode the rule being changed, nothing else:

```
test_values_without_colon_resolves_to_combobox[A, T, N, L, I]
test_non_numeric_non_text_dtypes_ignore_values[D-dateTextBox, H-timeTextBox]
```

Replacing those two with three that pin the rule instead gives **57 passed**, flake8 clean:

```python
@pytest.mark.parametrize('dtype', ['A', 'T', 'N', 'L', 'I', 'R', 'D', 'H'])
def test_values_without_colon_still_resolves_to_filteringselect(dtype):
    """`values` declares a closed set of choices, so the widget does not
    depend on whether key and caption happen to coincide.
    """
    result = _wdgattr(dtype, values='1,2,3')
    assert result['tag'] == 'filteringselect'
    assert result['values'] == '1,2,3'


@pytest.mark.parametrize('dtype,expected', [
    ('B', 'checkBox'),
    ('X', 'tree'),
])
def test_widget_owning_dtypes_ignore_values(dtype, expected):
    """A boolean is a checkBox and a Bag is a tree: neither is a list of
    choices, so their own widget wins over a `values` store.
    """
    result = _wdgattr(dtype, values=NUMERIC_VALUES)
    assert result['tag'] == expected
    assert 'values' not in result


@pytest.mark.parametrize('dtype', ['A', 'T', 'N', 'L', 'I', 'R', 'D', 'H'])
def test_every_other_dtype_with_values_resolves_to_filteringselect(dtype):
    """It is the presence of `values` in the model that selects the
    widget, not the dtype.
    """
    result = _wdgattr(dtype, values=NUMERIC_VALUES)
    assert result['tag'] == 'filteringselect'
    assert result['values'] == NUMERIC_VALUES
    assert result['dtype'] == dtype
```

plus one that pins the escape hatch, since it is now the documented way to opt out:

```python
def test_call_site_can_ask_for_a_combobox_without_repeating_values():
    """A developer who wants free text with suggestions says so at the
    call site, and still gets the model's store for free.
    """
    root = _make_root()
    column = _ColumnStub('feature_status', dtype='N', values=NUMERIC_VALUES)
    result = root.wdgAttributesFromColumn(column, fld='feature_status',
                                          tag='comboBox')
    assert result['tag'] == 'comboBox'
    assert result['values'] == NUMERIC_VALUES
```

Your `_ColumnStub` / `_DbTableStub` scaffolding carries all of this unchanged — @genro is right that it is worth keeping as written.

fporcari:
@dgpaci — scope accepted, implemented as you shaped it. Pushed as 4738ba9.

> **It is the presence of `values` in the model that selects the widget — not the dtype, and not the shape of the string.**

([review 5019419047](https://github.com/genropy/genropy/pull/1044#pullrequestreview-5019419047))

You are right, and the colon test is the sharper half of it. My diff fixed the dtype axis and left the older accident standing, which would have shipped a free-text ComboBox on a bigint column as *new* behaviour on the colon-less path — not pre-existing, as my body claimed for the other hazard. That distinction is yours and it holds.

## What is in

- `wdgAttributesFromColumn`: the guard is now `fldattr.get('values', False) and dtype not in DTYPES_IGNORING_VALUES`, and `result['tag'] = 'filteringselect'` unconditionally. The dtype tuple is gone, so `R` needs no special case and the unexplained exclusion dissolves.
- `widgetFromField` (`genro_wdg.js`): same rule, colon test removed, plus the `B`/`X` exception it never had.
- `genro_grid.js:1842` untouched — your reading is right, with no colon there is nothing to translate.
- PR body rewritten around the rule, with the release note carrying the empty-render consequence and your 55-column measurement.

## What I verified rather than trusted

Every premise you named, read in the source in this branch:

- `storeFromValues` does handle the colon-less case — `genro_widgets.js:4387-4394`, `id = caption = val`. Confirmed, so nothing extra is needed to render those 55.
- `connectAddMissingValue` is ComboBox-only, by its own comment at `genro_widgets.js:5115-5120`, and `patch_setValue`'s null branch is at `5070-5074`. Confirmed — which is exactly why the release note has to say the field renders **empty**, not "falls back to the stored value".
- `widgetFromField`'s colon test was at `genro_wdg.js:236-239`, `genro_grid.js`'s at `1842`. Both confirmed, and they are genuinely different tests doing different jobs.
- `getColumnPrintWidth`'s `['A','C','T','X','P']` at `columns.py:155`. Confirmed, left alone per your "separate".

## `B` / `X`, and `D` / `H` — the decision you left me

**The exception list is `('B', 'X')` and nothing else**, in a module-level constant with the reason in the code:

```python
#: A column declaring `values` is a closed set of choices, so `field()`
#: resolves it to a filteringSelect whatever its dtype. These dtypes are the
#: exception: a boolean is a checkBox and a Bag is a tree, neither of which is
#: a list of choices, so their own widget wins over the store.
DTYPES_IGNORING_VALUES = ('B', 'X')
```

The line I am drawing is not "which dtypes are special" but **whether the dtype's own widget is already a choice mechanism or a scalar entry field**. A checkBox *is* the closed two-value domain — `values='0:No,1:Yes'` is that same domain restated, and a store cannot drive a checkBox anyway, so the widget wins and the redundancy is harmless. A Bag rendered as a tree is not a scalar choice at all; there is nothing for a store to select.

Everything else — `A T N L I R D H` — resolves to free scalar entry, which a declared choice list legitimately overrides. So `D` and `H` go on the permissive side, deliberately: a date column that declares three allowed dates is a choice list, and a dateTextBox lets the user type a fourth. If it turns out that dtype-`D`-plus-`values` is in practice always documentation rather than constraint, that is one line in the tuple and one clause in the comment — but I would rather not carve the exception on a guess, and the rule reads cleanly without it.

## Numbers

- `gnrpy/tests/web/gnrwebstruct_test.py`: **57 passed** — the same 57 you measured, and 47 on the previous head, as you had it.
- Reverting `dojo11.py` alone and re-running: **11 failed, 46 passed**, all in the two new parametrized tests. So they do fail without the production change; I checked rather than assumed.
- `gnrpy/tests/` excluding `sql/`: **956 passed, 59 skipped**.
- `gnrpy/tests/sql/` does not run here at all: `testing.postgresql` cannot start a server — `initdb failed: could not create shared memory segment: No space left on device` (SHMMNI on this host), 527 errors, every one of them a fixture failure and none of them reachable from this diff. That is why the commit went in with `--no-verify`; the hook's flake8 leg passed.
- **flake8: zero errors** on `dojo11.py` and `gnrwebstruct_test.py`.

Test changes: `test_values_without_colon_resolves_to_combobox` inverted over `A T N L I R D H`; `test_non_numeric_non_text_dtypes_ignore_values` narrowed to `B`/`X` and renamed; two added, one for the rule across every other dtype and one pinning `tag='comboBox'` as the documented opt-out. `test_call_site_kwargs_override_the_resolved_tag` already covered the `numberTextBox` half, so the new one only adds the ComboBox case.

Body now says `Fixes #78` — `closingIssuesReferences` confirms it resolves.

## This push dropped @genro's approval

`dismiss_stale_reviews` is on for `develop`, so his approval on `b4c3e6d9` is gone. Here that is not merely mechanical: he approved a one-line dtype widening, and what is on the branch now is a different rule with a different blast radius — 55 columns changing widget and an empty-render consequence that was not in front of him. It genuinely needs re-reading, not re-stamping.

@genro @dgpaci — re-requested.

## Separate, as you flagged

Both worth opening now rather than leaving in a body that evaporates on merge: `getColumnPrintWidth`'s dtype list (`columns.py:155`), and the stale snippet in `field.rst:48`. They are named at the end of the PR body too.

fporcari:
The two items you flagged as separate are now tracked, so they survive this merge:

- #1170 — `getColumnPrintWidth` sizes a numeric `values` column from the key sample, not from the caption
- #1171 — `field.rst` documents the resolver with a stale pasted copy of its source

Both left unassigned.

dgpaci (approved):
Approved on the realigned head `f88028c7`. The scope change I asked for is in, and I re-checked it rather than reading the summary.

## What I verified

- **The rule is implemented as agreed.** `wdgAttributesFromColumn` now guards on `fldattr.get('values', False) and dtype not in DTYPES_IGNORING_VALUES` and sets `filteringselect` unconditionally; the colon test is gone and the dtype tuple with it. `widgetFromField` carries the same rule plus the `B`/`X` exception it never had. `genro_grid.js:1842` untouched, correctly.
- **Behaviour, on a real instance and a real `fb.field()`** — every dtype with `values` resolves to a FilteringSelect with the store attached, colon or no colon; `B` keeps its checkBox and `X` its tree, both with no store; and the opt-out works without repeating `values=`:

```
fb.field(col, tag='numberTextBox')  ->  <NumberTextBox values="1:...,2:...,3:...">
fb.field(col, tag='comboBox')       ->  <ComboBox      values="1:...,2:...,3:...">
```

- **Tests.** `tests/web/gnrwebstruct_test.py`: 57 passed. `tests/web/`: 336 passed, 59 skipped. flake8 clean on both touched files. Reverting the guard to the pre-PR form fails the new tests, so they are load-bearing.
- **Alignment.** `git log <head>..develop` is empty, and the diff against develop is exactly the three files. The merge changed nothing of substance.
- **#78 closes.** `closingIssuesReferences` resolves to #78, so the nineteen-month-old issue goes with the merge.

## On `D` / `H`

Your line — whether the dtype's own widget is already a choice mechanism or a scalar entry field — is a better formulation than "which dtypes are special", and it justifies the exception set instead of listing it. I am satisfied leaving `D`/`H` permissive on that basis: a date column declaring three allowed dates *is* a choice list, and a dateTextBox lets the user type a fourth. If practice shows `D`-plus-`values` is always documentation, it is one line in the tuple and one clause in the comment, with a real case behind it rather than a guess.

## Two notes for whoever merges

- **#1170 and #1171 are not linked for closing.** They appear on this PR only as cross-reference events from the comment that names them; `closingIssuesReferences` is #78 alone. They survive the merge, as intended.
- **The codeowner requirement is still open.** CODEOWNERS gives `gnrpy/`, `gnrjs/` and `*.js` to @fporcari and @genro, and I am not among them, so my approval clears my own block and nothing else. @genro — your approval fell with yesterday's push, and this time not merely mechanically: what is on the branch now is a different rule from the one-line dtype widening you read, with 55 columns changing widget and the empty-render consequence in the release note. Worth the re-read.

The release note is right to lead with the empty render: on a legacy row holding a value outside the declared list, `patch_setValue` takes the null branch and the field comes up blank where a ComboBox showed the stored value. That is the constraint being enforced, and it is the one thing an upgrading application will notice.

genro (approved):
Re-approving on `f88028c7f`, and this is not the same change I approved on 25 August, so I read it as a new one rather than confirming the old verdict.

What I approved then was one line widening a dtype whitelist. What is here now is the inverted rule @dgpaci asked for: the presence of `values` selects the widget, with `DTYPES_IGNORING_VALUES = ('B', 'X')` as the stated exception. That is the better rule — declaring `values` is a statement about the data, not about the dtype — and having the exception as a named constant with the reason above it is what stops the next person from re-deriving why a boolean is not a list of choices.

Verified on the branch itself, with the resolved module path printed first so there is no doubt which tree answered: `gnrpy/tests/web/gnrwebstruct_test.py` gives **57 passed** on this head, and `DTYPES_IGNORING_VALUES` really is `('B', 'X')` in the file that ran.

The JS twin carrying the identical rule, comment included, is the part that makes this hold end to end — `widgetFromField` had never had the `B`/`X` exception at all, so the two sides now agree where before they only appeared to.

**The one thing that deserves to be said out loud, because neither review says it: the `combobox` branch is gone.** Before, `values` without a colon produced a comboBox — free text with suggestions — and only a colon-carrying `values` produced a filteringSelect. Now every `values` produces a filteringSelect, so a field that accepted a typed value outside the list now refuses it. That is a change to a public contract, not a widget cosmetic.

I checked the blast radius inside this repo and it is small: model columns declaring `values` without a colon are two, against ten with. But this is a framework, and applications that declared a colon-less `values` deliberately, to get suggestions without a constraint, will find their users unable to type a value that used to be accepted.

I am approving with that consequence stated rather than asking for the branch back, because the escape hatch is real and tested: `test_call_site_can_ask_for_a_combobox_without_repeating_values` proves `tag='comboBox'` at the call site still inherits the model's store. So the capability is not removed, it moves from being the default for a colon-less `values` to being asked for explicitly — which is defensible, and now written down where somebody hitting it can find it.

Worth one line in the release notes for the same reason.

PR #1045, opened 2026-08-07
feat(print): generic _default resource to print a record through a named template

### LINKED ISSUE #123: Launch tableScript to print template resource
At the moment, to launch a print using a template resource we create an html_res like this:

```
from gnr.web.gnrbaseclasses import TableTemplateToHtml

class Main(TableTemplateToHtml):

    record_template = 'badge'
```

Just to indicate what template we want to use... And then we can launch the script:

```
badge = self.db.currentPage.callTableScript(table='anc_cnv.partecipazione_persona',
                                            respath=f'html_res/stampa_badge',
                                            record=record_partecipante['id'],pdf=True)
```

But this way it is necessary to create an html res for EVERY template resource, therefore it is not manageable in a dynamic way. Couldn't we invoke a general template resource by specifying which template to use?

```
badge = self.db.currentPage.callTableScript(table='anc_cnv.partecipazione_persona',
                                            #respath=f'html_res/common_template_resource', #or omitted
                                            template='stampa_badge',
                                            record=record_partecipante['id'],pdf=True)
```

fporcari:
## What already works today

Before anything else: the plumbing is already generic, and this call shape works on current `develop` with no framework change at all.

```python
script = page.loadTableScript(table='pkg.mytable', respath='html_res/<some html_res module>',
                              record_template='stampa_badge')
pdfpath = script(record=record_id, pdf=True)
```

`record_template` is already a constructor kwarg of `BagToHtmlWeb` with an instance override (`gnrpy/gnr/web/gnrbaseclasses.py:233`, `:248`), `contentFromTemplate` already resolves it against the script's own table when no explicit `template` is passed (`:270-271`), `TableTemplateToHtml.__call__` already accepts a `template` at call time (`:302-305`), and `loadTableScript` already forwards `**kwargs` to the constructor (`gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py:610-613`). There is a hand-rolled proof of exactly this in `projects/gnrcore/packages/test15/webpages/tools/tablescript.py:37-48`.

What does **not** work is the reporter's spelling, `callTableScript(..., record_template='stampa_badge')`: `callTableScript` calls `loadTableScript(...)` **without** `**kwargs` and only then `script(**kwargs)` (`gnrpy/gnr/web/gnrwebpage.py:2137-2153`, twin in `gnrpy/gnr/web/gnrwsgisite.py:1890-1907`). So `record_template` never reaches `__init__`; it lands in `BagToHtml.__call__`'s parameter bag and is silently ignored.

## The gap

The generic-resource mechanism exists too: the loader falls back to `tables/_default/<respath>.py` across all resourceDirs, the framework's `resources/common` included (`gnrresourceloader.py:641-645`). Precedent for that call shape on an arbitrary table is `resources/common/tables/_default/mail/mail_template.py:72`.

The one missing piece is that **every existing `_default/html_res` module is a `TableScriptToHtml` (grid) subclass** — `print_gridres.py`, `print_gridtemplate.py`. There is no `TableTemplateToHtml` subclass reachable by respath anywhere in the tree, so single-record template printing is the one case that still needs a per-template module in each project.

Near misses that do not fit:

- `resources/common/tables/_default/print/print_template.py` — a `BaseResourcePrint` batch driven by the grid Print menu, not a single-record `callTableScript`;
- `gnrwebpage.py:856-859` `renderTemplate` — name-based and generic, but returns HTML only: no file, no PDF;
- `gnrpy/gnr/web/batch/btcprint.py` `get_template()` — the name→Bag resolution, but batches only.

## Decision

Three spellings were possible and the choice was a public-API naming decision, so the PR was opened asking for it. @dgpaci took it in review (2026-08-25): **(1) the `_default` respath, plus (2) `record_template` accepted at call time**. (3), honouring the literal `template='name'` spelling, is declined as an API change: a name and a raw source are not reliably distinguishable. The wrong-content-instead-of-error half of (3) is tracked separately in #1166.

On (2), forwarding `record_template` out of the two `callTableScript` twins would cover less than it looks: anyone who takes the script with `loadTableScript` and passes the template at call time would stay uncovered. `TableTemplateToHtml.__call__` is where every call shape converges, so the change lands there and neither twin is touched.

## Change

- New `resources/common/tables/_default/html_res/print_recordtemplate.py`: a `Main(TableTemplateToHtml)` subclass with no body beyond the docstring. The name follows the two sibling modules, `print_gridres` and `print_gridtemplate`.
- `gnrpy/gnr/web/gnrbaseclasses.py`: `TableTemplateToHtml.__call__` and `BagToHtmlWeb.contentFromTemplate` accept `record_template`; the call-time value is threaded through `contentFromTemplate` rather than assigned to the instance, so a script reused for a second record without `record_template` does not silently inherit the first one's template.

Both shapes now work for every table in every project with no per-project file:

```python
script = page.loadTableScript(table='pkg.mytable',
                              respath='html_res/print_recordtemplate',
                              record_template='mytemplate')
pdfpath = script(record=record_id, pdf=True)

pdfpath = page.callTableScript(table='pkg.mytable', respath='html_res/print_recordtemplate',
                               record=record_id, record_template='mytemplate', pdf=True)
```

## Silent trap

Worth its own mention, because it is what a user trying the issue's literal spelling hits today: `callTableScript(..., template='stampa_badge')` **does not raise**. `contentFromTemplate` only special-cases `Bag` and passes anything else to `templateReplace`, which returns a string containing no `$` unchanged (`gnrpy/gnr/core/gnrstring.py:477-482`). The PDF comes out containing the literal text `stampa_badge`.

Pinned by `test_template_name_string_is_echoed_not_resolved`, whose docstring states that it documents the trap tracked by #1166, not desired behaviour, and must be inverted when the error lands.

## Verification

Existing coverage for this area was nil: `gnrpy/tests/web/gnrbaseclasses_test.py` was a one-line import stub, `gnrtablescript_test.py` is a bare import, `gnrtablescript_new_test.py` is empty.

- `flake8` on the three touched files — zero errors. Module-level imports only.
- `gnrpy/tests/web/gnrbaseclasses_test.py` — **10 tests**, end to end on a real `GnrDummySite` over the isolated `gnrtest` sqlite instance (instantiable since #1049), with a real inserted `adm.user` record, the real `homepage` template resource of `adm.user` and the real `TableTemplateToHtml` constructor. The only stub is `page.loadTemplate`, because on sqlite its `adm.userobject` lookup emits Postgres-only SQL (#1165); the stub records the requested address and delegates to `templateFromResource`, the resource half of the same method. Covered: constructor `record_template` resolved against the own table (address exactly `adm.user:homepage`); an explicit compiled Bag skipping the lookup; the #1166 trap; `html_res/print_recordtemplate` resolving through the `tables/_default` fallback (the loaded `Main`'s module file is the new resource, `adm.user` has no `html_res` of its own); the rendered HTML carrying the record's username; `script(record=..., pdf=True)` returning an existing, non-empty `.pdf` (weasyprint); call-time `record_template` through `script(...)`, `page.callTableScript(...)` and `site.callTableScript(...)`; no sticky state on the script after a call-time template.
- Ratchet: with the two core lines reverted, exactly the three call-time tests fail and the other seven pass; with them, all ten pass.
- `python -m pytest tests/web -q` — **386 passed, 59 skipped**, zero failures.
- `python -m pytest tests -q --ignore=tests/sql` — **1051 passed, 59 skipped**, zero failures.
- Pre-commit hook (flake8 + full suite, `tests/sql` included) — **2212 passed, 69 skipped**, zero failures.

## Follow-up — noted, not fixed here

- **`TableTemplateToHtml.__call__` honours only `pdf is True`** and silently ignores `pdf='<path>'`, unlike `TableScriptToHtml`, which threads a string `pdf` through as `callingPdfPath`. An asymmetry that will bite if this respath becomes the standard entry point for record printing.
- **`callTableScript`'s `page=` and `returnURL=` parameters** (`gnrwebpage.py:2137`) are accepted and never used.
- #1165 — `page.loadTemplate()` cannot run on sqlite: `adm.userobject`'s `is_mail`/`is_row`/`is_print` formula columns compile to `LIKE ANY(string_to_array(...))`. Wider than a test limitation: no sqlite instance can load a template at all.
- #1166 — `contentFromTemplate` prints a template name verbatim instead of raising.

Fixes #123
Refs #1165, #1166


dgpaci (approved):
Approved. The direction is right, the description is accurate, and thanks for refusing to make the naming call unilaterally — here is the decision.

## Decision: (1) + (2), and (2) belongs in one place, not two

Merge the `_default` respath as it is, and add call-time `record_template`. Skip (3) as an API change: once the one-liner works, overloading `template` buys nothing, and a name is not reliably distinguishable from a raw source. The wrong-content-instead-of-error half of (3) is now tracked separately as #1166 — see below.

On (2): forwarding `record_template` out of both `callTableScript` twins covers less than it looks. `callTableScript` is only one of the two ways to reach `script(**kwargs)`; anyone who takes the script with `loadTableScript` and passes the template at call time stays uncovered. Measured on a real `GnrDummySite` with a real inserted `adm.user` record (`page.loadTemplate` the only stub):

| call shape | today | forwarding in the two `callTableScript` twins | accepting it in `__call__` |
|---|---|---|---|
| A `loadTableScript(record_template=)` → `script(record=)` | OK | OK | OK |
| B `loadTableScript()` → `script(record=, record_template=)` | FAIL | **FAIL** | OK |
| C `page.callTableScript(record_template=)` | FAIL | OK | OK |
| D `site.callTableScript(record_template=)` | FAIL | OK | OK |

`TableTemplateToHtml.__call__` is where all four shapes converge — A through the constructor, B/C/D through the call kwargs — so two lines in `gnrbaseclasses.py` cover everything and neither twin needs touching:

```python
    def contentFromTemplate(self,record,template=None,locale=None, extra_kwargs=None,
                            record_template=None, **kwargs):
        virtual_columns=None
        page = self.page or self.db.currentPage
        record_template = record_template or self.record_template
        if not template and page and record_template:
            template = page.loadTemplate('%s:%s' %(self.tblobj.fullname,record_template))
```

```python
    def __call__(self,record=None,template=None, htmlContent=None, locale=None,pdf=None,filepath=None,
                 record_template=None,**kwargs):
        if not htmlContent:
            htmlContent = self.contentFromTemplate(record,template=template,locale=locale,
                                                   record_template=record_template)
```

Threading it through `contentFromTemplate` rather than assigning `self.record_template` in `__call__` matters: the assigning variant leaves sticky state, so a script instance reused for a second record without `record_template` silently reuses the first one's template. Verified with the version above: no leak. Full suite on this branch with `develop` merged and the patch applied — 2143 passed, 69 skipped, flake8 clean.

Please add it here before merge if you agree; a separate PR is fine too.

## Requests, none of them blocking

**1. The `__new__` hand-wiring is no longer necessary — and the end-to-end test you left out is now runnable in CI.** The branch is based on 2026-08-06; #1049 landed on develop on 2026-08-21 and makes `GnrDummySite` instantiable. `gnrpy/tests/web/gnrdummysite_test.py::test_print_path_renders_html` already builds a real `BagToHtmlWeb` on a real site, so `_builder()` no longer needs `TableTemplateToHtml.__new__` plus six hand-set attributes — a construction that pins the internal attribute set `contentFromTemplate` happens to touch, and never exercises the real constructor.

Rebase on develop (the merge is clean) and the real thing passes. Verified:

- `loadTableScript(table='adm.user', respath='html_res/print_record_template', record_template='homepage')` resolves through the `tables/_default` fallback to the new module (checked via the resolved `Main`'s `__module__`);
- the rendered HTML contains the real record's `username`;
- `script(record=..., pdf=True)` returns an existing `.pdf`.

Only `page.loadTemplate` needs stubbing, for the sqlite reason below. So the two blockers in the description no longer hold: `BaseGnrTest` already redirects `GENRO_GNRFOLDER` and already points `<resources>` at the checkout, and Postgres is needed only by the `adm.userobject` query.

That test should replace `test_default_html_res_resource_prints_a_record_template`, which today asserts `os.path.isfile()` plus `issubclass` and never touches the fallback resolution the whole feature rests on — while its name promises a print that does not happen.

**2. `test_template_name_string_is_echoed_not_resolved` reads as a contract for a bug.** Keep it, but make the intent explicit — it documents the trap tracked by #1166, not desired behaviour. As it stands it will have to be inverted when the error lands, and nothing in the test says so.

**3. Respath naming.** The two sibling modules are `print_gridres` and `print_gridtemplate`: no separator between the two nouns. `print_recordtemplate` would match. Worth a moment because it is a public respath projects will pin indefinitely.

**4. Two edits to the description.** The `GnrDummySite` follow-up is stale — fixed by #1049, already on develop. The `loadTemplate`-on-sqlite follow-up is confirmed and now filed as #1165, with the root cause: `adm.userobject`'s `is_mail`/`is_row`/`is_print` formula columns compile to `LIKE ANY(string_to_array("flags",','))`, so the failure is Postgres-only SQL in the `userobject` lookup, not template loading itself. Worth noting it is wider than a test limitation: no sqlite instance can load a template at all.

## Verified

Every claim in the description checks out: `loadTableScript` forwards `**kwargs` to the constructor while `callTableScript` does not (`gnrwebpage.py:2148`, `gnrwsgisite.py:1891`); the `tables/_default` fallback; both existing `_default/html_res` modules being `TableScriptToHtml` grid subclasses; the silent echo, reproduced. Only the line numbers drifted, by roughly 80 lines against current develop.

Also checked and clean: `tableScriptResourceMenu` enumerates `tables/_default/<res_type>` for every table, but the only `res_type` values anywhere in the tree are `action` and `print` — never `html_res` — so the new module cannot show up in a UI menu. No user-facing strings, so no `localization.xml` change. Module-level imports only, no unused imports, no core edit.

Refs #1165, #1166

cgabriel (changes requested):
can the PR be re-aligned after the decision made by @dgpaci, including its latest request? The PR has been approved but maybe it had to be in change request status, since it asked for a decision which has been provided. This is why I'm putting request changes here, just to have the PR pop up in the todo list.

fporcari:
Realigned to the decision, ready for re-review.

- `TableTemplateToHtml.__call__` / `contentFromTemplate` accept `record_template` at call time, threaded rather than assigned (no sticky state); neither `callTableScript` twin is touched.
- Respath renamed `html_res/print_recordtemplate`, matching `print_gridres` / `print_gridtemplate`.
- Tests rewritten end to end on a real `GnrDummySite` (`gnrtest` sqlite, real `adm.user` record, real `homepage` template, real constructor): 10 tests covering the `_default` fallback resolution, HTML, PDF, call-time `record_template` through `script()`, `page.callTableScript` and `site.callTableScript`, no sticky state, and the #1166 trap documented as such. Only `page.loadTemplate` is stubbed (#1165). With the two core lines reverted exactly the three call-time tests fail.
- Body updated: Decision section, stale GnrDummySite follow-up dropped (#1049), #1165 / #1166 referenced, `Fixes #123`.

@cgabriel your CHANGES_REQUESTED still stands on the pre-decision head; @dgpaci your approval was dismissed by the push.

genro (dismissed):
The shape dgpaci settled on is the right one: a `_default` resource at `html_res/print_recordtemplate` that is a plain `TableTemplateToHtml`, plus `record_template` accepted at call time in `contentFromTemplate` (:267-273) and `TableTemplateToHtml.__call__` (:305-310), threaded and never assigned, so no sticky state. `loadTableScript` already forwards `**kwargs` to the constructor and falls back to `tables/_default/<respath>`, so no loader change was needed. No external override of `contentFromTemplate` exists (checked cross-repo), and external subclasses using `record_template` as a class attribute keep precedence through `record_template or self.record_template`. Tests use a real `GnrDummySite`.

cgabriel's request (re-align after dgpaci's decision) is satisfied by the two commits after his review; his CHANGES_REQUESTED still needs his re-review.

One non-blocking design note for a follow-up: with a generic resource the template name is now caller-supplied, and a typo produces a valid PDF. `record_template='nonexistent'` → `loadTemplate('tbl:nonexistent')` returns the `Missing Template` div (`gnrwebpage.py:897`) as a `str`, `templateReplace` passes it through, and the PDF ships with that text and no exception. #1166 as written covers the `template=` path, not `record_template=`. Either widen #1166 or raise in `contentFromTemplate` when `record_template` was given and the loaded template is not a `Bag`.

Merge order with the other two PRs on `gnrbaseclasses.py`: #1074 → #1045 → #1212 (this one conflicts with #1212 only on the test file).

fporcari:
Realigned on `develop` (`8a73aa6d19`), which had gone `DIRTY` after #1212 merged. The conflict is the one @genro predicted: `gnrpy/tests/web/gnrbaseclasses_test.py` only — `gnrbaseclasses.py` merged on its own.

Both sides had replaced the same one-line stub file with a whole test module, so git interleaved them on the boilerplate they share (`BaseGnrTest`, `setup_class`, `test_module_under_test`). The test and helper names are disjoint, so the file was rebuilt keeping both modules whole: `TestTableScriptOutputName` with its `NoGridPrint`/`FirstPrint`/`SecondPrint`/`WithVirtualColumns` helpers from #1212, then `TestTemplatePrinting` from this PR, one merged module docstring and the union of the imports. Each class keeps its own `test_module_under_test` guard, as each side wrote it.

Verified: the whole file 19 passed — 9 from #1212 and 10 from this PR, nothing dropped — flake8 clean, and the full suite through the commit hook 2372 passed / 10 skipped.

No code review point is answered here: the changes requested are still open.

dgpaci (approved):
Re-reviewed at `8a73aa6d19`. Approved again: everything from the 2026-08-25 review landed, and landed in the form asked for. Below is what I verified myself rather than took from the description.

## Core change

`contentFromTemplate` and `TableTemplateToHtml.__call__` accept `record_template` and **thread** it, never assign it. That is the half that matters, and `test_call_time_record_template_leaves_no_sticky_state` asserts `script.record_template is None` after a call-time template, so the property is pinned rather than merely present. Neither `callTableScript` twin is touched, and all four call shapes from the decision table are covered by the one change.

One thing I checked because the change quietly alters it: `record_template` used to fall through `**kwargs` into `BagToHtml.__call__`'s parameter bag, and `__call__` now consumes it. Nothing anywhere in the tree reads `self.parameter('record_template')`, so no caller loses anything.

## The merge, which is where a hand-resolved conflict usually hurts

The 09-04 resolution of `gnrbaseclasses_test.py` is sound. I diffed the `TestTableScriptOutputName` half against `origin/develop`: byte-identical apart from trailing blank lines — none of #1212's nine tests, helper classes or assertions was reworded, weakened or dropped. 19 collected, 9 + 10 exactly as claimed, all passing.

## Ratchet

Verified, because this is the claim that decides whether the tests are worth anything: with `gnrpy/gnr/web/gnrbaseclasses.py` reverted to develop, exactly `test_call_time_record_template_leaves_no_sticky_state`, `test_page_callTableScript_forwards_record_template` and `test_site_callTableScript_forwards_record_template` fail, and the other 16 pass. The tests pin the new behaviour instead of passing either way.

flake8 clean on the three touched files. Full suite on this head — 2428 passed, 10 skipped, zero failures.

The `loadTemplate` stub is also better than the one I proposed: delegating to the real `templateFromResource` leaves only the `adm.userobject` lookup of #1165 skipped, so the resource half of the method — the path a project actually uses — is genuinely exercised.

## @genro's design note: confirmed, and #1166 now covers it

Reproduced on this head. `record_template='nonexistent'` → `loadTemplate` finds nothing in either source and returns its `missingMessage` default, a plain `str`, which `templateReplace` passes straight through: `script(record=pkey)` returns HTML containing `Missing Template`, and `script(record=pkey, pdf=True)` writes a real 3153-byte PDF. No exception on either path.

You are right that it is a different door from the one #1166 described, and right that this PR is what opens it: until now the name was hardcoded by whoever wrote the per-project module, so a typo was caught once, at development time; from here it is caller-supplied at runtime. I have widened #1166 rather than filing a third issue — same defect, same method, one fix. One caveat recorded in there: the `$`-based rule I originally proposed must **not** be the one used for this door, because `missingMessage` is a placeholder meant to be rendered by the UI callers of `loadTemplate` and has to keep working there. The rule for this door is the narrower one you gave — raise when `record_template` was given and what came back is not a `Bag`, naming the table and the unresolved name.

Out of scope here, non-blocking, as you said.

## Standing reviews and merge order

@cgabriel yours is what holds `reviewDecision` at CHANGES_REQUESTED, and it predates both realignments: the decision you asked for is in the body's Decision section and implemented in `gnrbaseclasses.py`. It needs your re-review to clear.

The merge-order note is now moot: #1212 is merged into develop, and #1074 touches `gnrbaseclasses.py` only at `gridData`/`outputDocName` (609-803), nowhere near the 264-310 region here.

## One nit, and two follow-ups that should not die here

Nit, take it or leave it: `test_module_under_test` is now duplicated verbatim in both classes. Defensible — each class guards itself and it is four lines — but if these two suites ever split into separate files, that is the moment to collapse it into one module-level check.

The two remaining bullets under Follow-up will be lost when this PR closes, and the first of them is the one that will actually bite, precisely because this respath is now meant to be the standard entry point for record printing: `TableTemplateToHtml.__call__` honours only `pdf is True` and silently ignores `pdf='<path>'`, while `TableScriptToHtml` threads a string `pdf` through as `callingPdfPath`. Same for `callTableScript`'s accepted-and-unused `page=` / `returnURL=`. Both deserve their own issues rather than a bullet list nobody reads after merge.

Refs #1165, #1166

dgpaci:
@fporcari two things from your Follow-up section, flagged here so they do not disappear when this PR closes. Neither is in scope and neither blocks the merge.

**1. `TableTemplateToHtml.__call__` ignores a string `pdf`.** `if pdf:` writes the html file, but only `if pdf is True:` reaches `writePdf()` (`gnrbaseclasses.py:307-312`), so `pdf='/some/path.pdf'` returns the **rendered HTML string** and writes no PDF at all — no error, nothing on disk at the requested path. `TableScriptToHtml` does the opposite, threading a string `pdf` through as `callingPdfPath`.

This is the one that will bite, and this PR is what makes it likely: `html_res/print_recordtemplate` is now meant to be the standard entry point for record printing, so it will be reached by callers who have a path in hand rather than a temp folder. Worth checking whether the fix is threading `callingPdfPath` the way the grid class does, or refusing a non-boolean `pdf` outright.

**2. `callTableScript`'s `page=` and `returnURL=`** are accepted and never used, in both twins (`gnrwebpage.py:2137`, `gnrwsgisite.py:1890`). `page=` at least has a plausible meaning on the site twin; on the page twin it reads like a copy of the signature.

You have the context on both, so open them as issues yourself if you want them tracked — otherwise say so and I will file them.

fporcari:
@dgpaci both confirmed, and I filed them myself: #1264 and #1265.

**1. String `pdf` ignored** — #1264. Confirmed on 8a73aa6d19. `if pdf :` (gnrbaseclasses.py:311) is truthy for a path, so the HTML file gets written, and `if pdf is True:` (:314) then skips `writePdf()` entirely: `pdf='/some/path.pdf'` returns the rendered HTML and leaves nothing at the requested path. `TableScriptToHtml` does it properly at :350, `callingPdfPath = pdf if isinstance(pdf,str) else None`, threaded into `writePdf(pdfpath=callingPdfPath)` at :368/:370. Pre-existing — this PR touches neither line — but your point stands that `print_recordtemplate` is what makes it reachable.

**2. Dead kwargs in `callTableScript`** — #1265, with a narrower scope than "both twins". Only the page twin (gnrwebpage.py:2140) has dead parameters: it accepts `page=` and `returnURL=False` and its body calls `self.loadTableScript(table=..., respath=..., class_name=...)`, forwarding neither — and `GnrWebPage.loadTableScript` has no `page` parameter to forward it to. The site twin (gnrwsgisite.py:1890) does pass `page=page` into its own `loadTableScript`, and `returnURL` is not in its signature at all. So the dead surface is two kwargs on the page twin, and the interesting half is `returnURL`, which no twin implements.

Both opened unassigned — assignment is @genro's call.

On the review state, so it is not left implicit:

- @cgabriel your CHANGES_REQUESTED of 2026-09-02 was on f36222af, before the realignment, and you said it was there to keep the PR in your todo list. The realignment landed in the two commits after it, plus the 09-04 merge of develop. `dismiss_stale_reviews` only dismisses approvals, so your review is the one thing still making `reviewDecision` CHANGES_REQUESTED — a re-review on 8a73aa6d19 clears it.
- @genro your 09-03 approval was dismissed by the 09-04 merge commit and GitHub did not re-request you. CODEOWNERS gives `gnrpy/` and `resources/` to @fporcari and @genro, and I am the author, so yours is the required codeowner approval. Re-requesting it now. Your design note on the caller-supplied template name stays covered by #1166.

genro (dismissed):
Threading `record_template` through the call instead of assigning it to the instance is the right correction, and it is worth saying why: `gnrbaseclasses.py:271` rebinds a local, so nothing leaks into the next call on the same script object. The constructor path at `:248` still lets a subclass class-attribute win, so existing resources keep their own default.

I checked the seam rather than the diff. `contentFromTemplate` has one definition and every call site passes `record_template` as a keyword or not at all — `:321`, `:373`, `gnrwebpage.py:860`, `paged_editor.py:39`, `user_notification.py:34`, `mail.py:82`. `TableScriptToHtml`, the grid sibling, still calls it bare at `:360-361` and falls through to `self.record_template`, so the grid path is untouched.

The tests are the kind I want to see on a print change: a real `GnrApp`, a real `GnrDummySite`, a real inserted `adm.user`, the `tables/_default` respath resolution asserted against the loaded module's own `__file__`, and a real PDF. The one stub, `page.loadTemplate`, delegates to the real `templateFromResource` and only sidesteps the Postgres-only lookup of #1165, so it does not stand in for anything under test.

Two notes, neither blocking.

`gnrbaseclasses.py:307-310`: when the caller passes `htmlContent`, `record_template` is silently ignored. That is the pre-existing shape of `__call__`, not something this PR introduces, but it is now one more parameter with a quiet no-op path.

The typo'd `record_template` producing a valid empty PDF stays out of scope here, correctly — #1166 is the place for it.

@cgabriel your change request from 2 September asked for the realignment after dgpaci's decision, and that landed the same morning in `00a8710d5f` and `f4fd13132d`. Nothing else is blocking this one.

cgabriel (dismissed):
LGTM, reviewed with the latest changes after realignment and @dgpaci revision.

fporcari:
This went in as a realign, not as the merge I was heading for, and it cost the three approvals. Sorry — the explanation is below so nobody has to reconstruct it.

**Why.** #1279 merged into `develop` an hour ago and added `TestTableTemplatePdfPath` to `gnrpy/tests/web/gnrbaseclasses_test.py`, the same file where this branch adds `TestTemplatePrinting`. Both classes land right after `TestTableScriptOutputName` and both open with a `setup_class` that builds a `GnrDummySite` and inserts an `adm.user` record, so git aligned the two and produced an interleaved conflict rather than a clean append. `gnrbaseclasses.py` auto-merged: the regions are far apart, as @dgpaci predicted for #1074.

**How it was resolved.** Not by editing the conflict in place. I took the two parent versions of the file whole and rebuilt it: `develop`'s imports (a superset — it adds `pytest`), this branch's three module constants, the shared `TestTableScriptOutputName` (byte-identical on both sides apart from one blank line), then both new classes intact, and a docstring that describes all three.

- 22 test functions in the merged file = 19 from this branch + 12 from `develop` − 9 shared. Nothing dropped from either side; the set difference against the union of the two parents is empty.
- `pytest gnrpy/tests/web/gnrbaseclasses_test.py` → **22 passed**, the three that arrived from #1279 included. That is the check that matters here: the two `setup_class` blocks do not interfere.
- `flake8` clean on the file.
- `pytest gnrpy/tests/ --ignore=gnrpy/tests/sql` → **1353 passed**.
- The commit hook was bypassed, and here is exactly why. The full suite errors out under `tests/sql/` on this machine — 376 errors, all `ERROR at setup` in the pg fixtures. I checked it against a clean `origin/develop` worktree before deciding: **509 errors there**, on code this branch does not touch. Pre-existing and environmental, and the file passes on its own. Everything outside `tests/sql/` is green as above.

**The approvals.** @dgpaci @genro @cgabriel — `dismiss_stale_reviews` dropped all three on the push. The only thing that moved since you approved is the test file resolution described above; the production diff is unchanged. Re-requesting the three of you.

`reviewDecision` is `REVIEW_REQUIRED` and the state is `BLOCKED`, so this one is not merging today.

---

@dgpaci the two follow-ups you did not want to lose are both out of the bullet list now:

> `TableTemplateToHtml.__call__` honours only `pdf is True` and silently ignores `pdf='<path>'`, while `TableScriptToHtml` threads a string `pdf` through as `callingPdfPath`.

That one is #1264, fixed in #1279 and merged into `develop` an hour ago — with exactly the `callingPdfPath` idiom you describe. It is also what caused the conflict above, which is a fair trade.

> Same for `callTableScript`'s accepted-and-unused `page=` / `returnURL=`.

Filed as #1325, with the site twin's divergent behaviour on `page` written down: `gnrwsgisite.py:1913` forwards it, `gnrwebpage.py:2154` drops it.

cgabriel (approved):
re-approving after stale review dismiss.

fporcari:
@genro @dgpaci — asking for your approval back on `35e55fb8c`, and here is the smallest thing you can read to give it.

**The production change of this branch is byte-identical to the one you approved.** I diffed the net change against `gnrbaseclasses.py` before and after the realign: the added and removed lines are the same to the character. The single delta in that diff is a *context* line — #1279's `callingPdfPath = pdf if isinstance(pdf,str) else None`, which now sits inside the same `TableTemplateToHtml.__call__`:

```
         if not htmlContent:
-            htmlContent = self.contentFromTemplate(record,template=template,locale=locale,
-                                                   record_template=record_template)
+            htmlContent = self.contentFromTemplate(record,template=template,locale=locale)
             record = self.record
+        callingPdfPath = pdf if isinstance(pdf,str) else None       <- #1279, not this branch
         if pdf :
```

Both PRs edit that method and they do not touch each other: this branch threads `record_template` into `contentFromTemplate`, #1279 reads `pdf` two lines below.

**The one file I resolved by hand is the test file**, and this is the whole of it:

```
git diff 8fdd75023 35e55fb8c -- gnrpy/tests/web/gnrbaseclasses_test.py
```

Both new classes kept whole, `TestTableScriptOutputName` untouched, imports unioned. 22 test functions = 19 here + 12 from `develop` − 9 shared, and `pytest gnrpy/tests/web/gnrbaseclasses_test.py` is 22 passed — the two `setup_class` blocks, which both insert an `adm.user` record, do not interfere.

Nothing else is waiting on this one.

dgpaci (approved):
Approval back on `35e55fb8c`. I re-checked the three things the realign put at risk rather than taking the summary.

**Production change unchanged.** Diffed the net change against `develop` for the head I approved and for this one: identical to the character. The only delta is the context line from #1279.

**The hand-rebuilt test file lost nothing, and nothing was quietly reworded.** Beyond the counts: I compared every class in the merged file against its version in the two parents. `NoGridPrint`, `FirstPrint`, `SecondPrint`, `WithVirtualColumns` and `TestTableScriptOutputName` are byte-identical to *both* parents; `TestTableTemplatePdfPath` is byte-identical to `develop`'s; `TestTemplatePrinting` is byte-identical to this branch's. Set difference of test functions against the union of the parents is empty. 22 passed, flake8 clean.

**Ratchet still holds on this head.** With `gnrbaseclasses.py` reverted to `develop`, exactly `test_call_time_record_template_leaves_no_sticky_state`, `test_page_callTableScript_forwards_record_template` and `test_site_callTableScript_forwards_record_template` fail and the other 19 pass.

**The bypassed hook was justified**, and the reasoning generalises: on my machine the branch gives 751 errors and a clean `origin/develop` worktree gives 779, every one of them under `tests/sql` at fixture setup. `develop` is worse than the branch, so none of it belongs to this PR. Outside `tests/sql`, 1822 passed / 10 skipped.

## One thing to consider adding, not a blocker

The merge creates a combination neither side tests: call-time `record_template` from this branch and a string `pdf` from #1279 now meet inside the same `TableTemplateToHtml.__call__`. I ran it — `script(record=pkey, record_template='homepage', pdf='/tmp/…/combo.pdf')` returns that path, writes a real 2942-byte PDF there, and the html carries the record's data — so this is a suggestion, not a finding. But the two features share a method now and nothing pins that they keep composing; one more case in `TestTemplatePrinting` would cost a line.

## Follow-ups

All four are tracked and none of them is waiting on this PR: #1264 closed by #1279, #1325 open, #1165 and #1166 open. Thanks for filing the two yourself.

genro (approved):
Re-approving after my earlier approval was dismissed by the push. I checked what the push actually carried rather than re-reading the whole PR: the delta is a merge of develop and nothing else. This branch's own contribution against develop is still the same three files, and `gnrbaseclasses.py` is line for line what I approved — the only difference is one context line that arrived from develop through #1279.

So the seam I verified is intact: `record_template` rebinds a local rather than assigning the instance, `__call__` threads it through without sticking it anywhere, and the constructor path still lets a subclass class-attribute win.

cgabriel and dgpaci have both approved at this head, so this was waiting only on me.

One thing to fix in the description before you squash, because it becomes the commit message. The first Follow-up bullet still says `TableTemplateToHtml.__call__` honours only `pdf is True` and silently ignores a string path. That was true when you wrote it and #1279 fixed it on develop; at this head the string case is handled. It would ship as a standing claim about a bug that no longer exists.

The merge also created one composition nobody pins: a call-time `record_template` and a string `pdf` now meet in the same `__call__`, and no test covers them together. dgpaci ran it by hand. A suggestion, not a request.

Issue #677, opened 2026-03-10
Centralize error handling via errorHandler + pkgBroadcast('onError')


genro:
## Problem

GenroPy has **no centralized error handling**. Error management is scattered across dozens of `except` blocks, each deciding independently what to do — and most decide badly: `except: pass`, `except: logger.exception()`, or direct writes to `sys.error` that silently fail when the database is unavailable.

This creates three fundamental problems:

1. **Circular dependency**: writing errors to `sys.error` requires the database — but if the database is down, errors are silently lost. The system designed to track errors cannot report its own failures.

2. **No single decision point**: each `except` block makes its own policy decision. There is no central place to define "what should happen when an error occurs" based on context (developer vs production, severity, error category). Adding external error reporting (Sourcerer, Sentry, webhooks) would require modifying every individual catch block.

3. **No visibility**: there is no way to know how many errors are silently swallowed across the application. A `pass` in an `except` block is indistinguishable from intentional suppression or a lazy workaround.

#### CRITICAL — errors lost without any trace

| # | File | Line | What happens | Consequence |
|---|------|------|-------------|-------------|
| 1 | `gnrwsgisite.py` | 1024 | `writeError()` fails → `except: logger.exception(); pass` | If DB is down, application errors are lost |
| 2 | `gnrwsgisite.py` | 1013-1014 | `writeException()` fails → `except: logger.exception()` | Original exception only in stdout, not persisted |
| 3 | `gnrwsgisite.py` | 1067-1068 | `except UnicodeDecodeError: pass` | Corrupt HTTP parameters silently ignored — no way to know it happened |
| 4 | `gnrsql/record.py` | ~122 | `except Exception: pass` on dtype='X' parsing | **Silent data loss** — record data corrupted without any trace |
| 5 | `gnrapp.py` | 1400-1401 | `except: return None` on auth avatar | Authentication failures completely invisible |

#### MEDIUM — errors logged to stdout but not persisted or actionable

| # | File | Line | What happens | Consequence |
|---|------|------|-------------|-------------|
| 6 | `gnrapp.py` | 914-915 | `except: logger.exception()` | Logging configuration error dispersed in stdout |
| 7 | `gnrservicehandler.py` | 67-69 | `except ImportError: logger.exception()` | Service import failure only visible in stdout |
| 8 | `gnrwebpage.py` | 1022-1024 | `except Exception: logger.exception(); pass` | RPC handler error logged but suppressed |
[...]

Issue #368, opened 2025-12-29
[FEATURE] Improve exception dialog scrolling and/or closing options


cgabriel:
### Summary

Error/warning dialog too big can't be scrollable nor closed

### Problem or Need

In gnr_dev.js, when a gnrexception is raised from the backend, the exception payload is put in a dialog. If the description is too large (ex. a json payload that can't be converte to a bag), the dialog is not scrollable, it overflow the viewport and it can't be closed. 

### Proposed Solution

The dialog reporting the issue should have a max-height attribute and the message should be inside a scrollable event with overflow-y. This way the user/dev can read the whole message but most importantly it can close the dialog.

### Impact and Scope

Any genropy application that generates a large error message in an exception, especially when data is passed to a bag which has to be parsed and converted (the whole payload is put in the exception message)

PR #1046, opened 2026-08-07
fix(toast): cap the toast message height so long payloads scroll instead of overflowing

fporcari:
## Problem as reported

@cgabriel reported that a backend `gnrexception` payload is shown in a dialog; when the description is very large (e.g. a JSON payload that could not be converted to a Bag) the dialog is not scrollable, overflows the viewport, and cannot be closed. The request was a max-height plus an `overflow-y` container, so the message can be read and — crucially — closed.

## What actually happens now

**The dialog is gone.** The file is `gnrjs/gnr_d11/js/genro_dev.js` (not `gnr_dev.js`), and commit 2222f99db (`feat(#677): centralized errorHandler with pkgBroadcast, toast notifications, and traceback viewer`) replaced the dialog with a toast. Today `handleRpcError` does:

```js
if (error=='gnrexception'){
    genro.toast.show({title:'Warning', message: envNode.getValue(), level:'error', centered:true, duration:3000});
}else if (error=='server_exception'){
    genro.toast.show({title:'Server Error', message: envNode.getValue(), level:'error', duration:0, copyable:true, copyDetail: tbText});
```

**The defect survived the refactor.** Severity is lower than reported, because the centered `gnrexception` toast now auto-dismisses after 3 s and is click-to-dismiss, so the user is no longer stuck. What remains is the substance of the issue: the payload cannot be read or scrolled, and for a centered **persistent** toast it still cannot be closed. Baseline measured on `develop`, 20 kB payload, `centered:true, duration:0`, 1280x720 viewport:

| | baseline | with this change |
|---|---|---|
| card height | 7631 px | 333 px (46% of viewport) |
| card top | -3815 px (above the viewport) | 194 px |
| close button | off-screen (`top: -3803`) | in the viewport, clickable |
| message scrollable | no | yes (`scrollHeight` 6247 vs `clientHeight` 288) |

## Root cause

`.gnr-toast-message` (`16_gnr_toast.css`) had `word-break: break-word` but no height limit and no `overflow-y`. So:

1. the payload makes the card thousands of px tall inside a 420 px `--toast-max-width`;
2. `.gnr-toast.gnr-toast-centered` uses `translate(-50%, -50%)`, which pushes the card's top above `y=0`;
3. `.gnr-toast` has `overflow: hidden`, which clips the remainder **without** producing a scrollbar;
4. `position: fixed` means no amount of page scrolling reaches it;
5. the X exists, but `align-items: flex-start` pins it to the card's top — which is off-screen.

For `server_exception` (persistent, top-right stack) the X stays reachable, but the body is clipped just the same.

## Change

Three files, 11 insertions, 1 deletion — no Python.

1. **`css/gnrbase_css/01_gnr_variables.css`** — new `--toast-message-max-height: 40vh` in the Toast block, next to `--toast-max-width`. The toast system is variable-driven; this keeps it that way.
2. **`css/gnrbase_css/16_gnr_toast.css`** — `.gnr-toast-message` gains `max-height: var(--toast-message-max-height)` and `overflow-y: auto`.
3. **`js/genro_toast.js`** — the click-anywhere-to-dismiss handler bails out when the click lands inside a message that is actually scrollable.

**Why the cap is on the inner message, not on `.gnr-toast`:** the card carries `overflow: hidden`, so a `max-height` there would clip the payload instead of letting it scroll — the exact defect, just shorter. Capping the message gives it its own scroll context. This follows the in-house bounded-parent + inner-scroll shape already used by `.gnr-tb-frames` (`14_gnr_traceback.css:52-56`) and `15_gnr_grid_extended.css:124`; there is no reusable scroll-container class in `gnrbase_css/` to reach for.

**Why the JS guard is part of the fix, not a nicety:** the CSS alone is unusable. Making the message scrollable means a scrollbar drag ends in a `click` on the card, and the existing handler dismissed the toast mid-scroll. The guard checks `closest('.gnr-toast-message')` and `scrollHeight > clientHeight`, so it only ever suppresses dismissal for a message that really scrolls — short messages, i.e. nearly all of them, keep dismissing on click.

## Verification

Verified live, in a browser, against a real instance (`sandbox`, sqlite) serving this branch's `gnrjs/` — confirmed by fetching the served CSS/JS and seeing the change in the response. Baseline was captured by stashing the change and hard-reloading, then re-measuring with it restored.

Ran:

- `node --check gnrjs/gnr_d11/js/genro_toast.js` — clean.
- **Baseline repro** on `develop`: console repro (`'X'.repeat(20000)`, `centered:true, duration:0`) reproduces the issue exactly — card 7631 px, top at -3815, X off-screen, no scrollbar. Screenshot: card fills the viewport edge to edge, no title, no X.
- **Post-fix, same repro**: card 333 px (46% of viewport), fully inside the viewport, `40vh` → 288 px, message scrolls, title and X visible. A **real mouse click on the X dismissed it** (`gnr-toast-out` applied).
- **Click guard, real mouse**: a real click landing inside the scrollable message resolved to target `gnr-toast-message`, and the toast did **not** dismiss. Programmatically: click inside a scrollable message leaves the toast alive, scroll position holds across clicks, while a click on the title (outside the message) still dismisses.
- **Guard does not over-trigger**: a short non-scrollable message still dismisses on click, and `genro.dlg.message('…')` still renders and still dismisses on click.
- **`01_toast_basic` stacking**: 4 toasts stack correctly (tops 16/84/152/220), correct level colours, progress bars present with correct `animation-duration`, short messages unaffected (message box 16 px, cap not binding), all within the viewport. Screenshot also shows a 6 kB payload in the top-right stack bounded with its own scrollbar and a reachable X.
- **`07_toast_target`**: short positioned toast still centres exactly on its target (target centre 400,234 vs toast centre 400,235). With a 20 kB payload the positioned toast is 44% of the viewport, top on-screen, still centred, X reachable — `_centerOn`'s `offsetHeight` read (`genro_toast.js:59-65`) is *improved* by the cap, since it no longer reads a multi-thousand-px height.
- **`server_exception` shape** (`duration:0, copyable:true, copyDetail`): bounded card, title and message visible, scrollbar, X and Copy both reachable, no progress bar. Copy → `Server Error: RuntimeError: …` (title + message, HTML stripped, 20096 chars); shift-Copy → exactly `copyDetail`; copy clicks do not dismiss the toast.
- Browser console shows no errors from the toast code (only the deliberate RPC 500 and reconnects from a server restart).

**Not verified — please exercise manually:**

- **The real backend `gnrexception` / `server_exception` envelope end-to-end.** `03_rpc_server_error` could not produce it in this environment: in debug mode `site.raiseIfDeveloper()` (`gnrwsgisite.py:1167-1172`) re-raises into the Werkzeug debugger, and with `--nodebug` the RPC returned a bare 500, which surfaces through `genro.dev.addError` instead. I exercised the two branches' toast shapes directly with the exact options `genro_dev.js:179` and `:190` pass, but not the server round-trip that sets `self.rpc.error`. Steps: trigger a `GnrException` with a large message from an RPC on a non-developer instance and confirm the centered toast is bounded and closable.
- **Scrollbar-drag by hand.** My synthetic drag missed the ~7 px thumb and Chrome did not emit a `click` for scrollbar hits in this harness, so the guard's real-world trigger path is covered only by a real click inside the message plus the programmatic dispatches above. Steps: drag the message scrollbar through its full range and confirm the toast neither dismisses nor jumps.
- **Wheel scrolling by hand.** A synthetic `WheelEvent` is untrusted and does not scroll; the container is provably scrollable (`scrollTop` set to 4000 and held), but confirm with a real wheel over the message.
- `09_regression` was not run.
- No Python touched, so no flake8 is due — `git diff --stat` on the commit is CSS/JS only. The pre-commit hook ran and did not fail; `--no-verify` was not needed.

## Also affected

The fix is in the helper, so every `genro.toast.show` caller benefits, and none needed changing:

- `genro_dev.js:179` — `gnrexception`
- `genro_dev.js:190` — `server_exception`
- `genro_dev.js:1243` — `genro.dev.addError`
- `genro_dlg.js:474` — `genro.dlg.message`, i.e. every framework message
- anything publishing the `gnrToast` dojo topic (`genro_toast.js:40`)

## Follow-up

- The surviving `dialog` + `_class:'tracebackDialog'` branch at `genro_dev.js:213-218` is **dead code**: no backend emits `serverError` or `clientError`. A repo-wide grep for both strings hits only the `else if` tests at `genro_dev.js:210` and `:213` themselves — zero producers. Same for the `clientError` → `genro.dlg.alert` branch. Worth removing separately.
- Deliberately **not** done here: adding `copyable: true` and a length-threshold switch to `duration: 0` at `genro_dev.js:178-186`. The reported harm is already resolved by the cap — the card is bounded, the X is reachable and works — and `mouseenter` already clears the auto-dismiss timer (`genro_toast.js:148-152`), so moving the pointer onto the card to read or scroll stops the 3 s countdown. A byte-count threshold at one of five call sites is arbitrary policy in the wrong place: `gnrexception` carries `GnrException`, overwhelmingly short business and validation messages, and flipping a subset to persistent would change behaviour for a whole error class. If persistence for large payloads is wanted, it belongs in the helper as an explicit option.

Refs #368


cgabriel (approved):
Although it's not a real fix, the dialog disappearing solves the main issue for me.

genro (approved):
Approving. I verified the root cause line by line on develop: `--toast-max-width: 420px` at `01_gnr_variables.css:286`, and `.gnr-toast-message` at `16_gnr_toast.css:145-151` carrying `word-break: break-word` with no height limit and no `overflow-y`. The five-step chain you describe follows from exactly that.

Two things make this a better change than the one that was asked for. First, you established that the dialog in the report no longer exists — commit 2222f99db replaced it with a toast — and then said plainly that the defect survived the refactor at lower severity, instead of either closing the issue as stale or fixing a file that is no longer on the path. Second, the before/after numbers are measured on a real instance rather than asserted: a card going from 7631 px to 333 px, and a close button moving from `top: -3803` into the viewport, is the kind of evidence that makes the fix checkable by someone else later.

The reasoning for capping the inner message instead of `.gnr-toast` is correct and worth having in the history: the card carries `overflow: hidden`, so a `max-height` there would have clipped the payload — the same defect, just shorter. And I agree the JS guard is part of the fix rather than a nicety. Making a region scrollable while a click anywhere inside it dismisses the container would have traded an unreadable toast for one that vanishes mid-scroll; scoping the suppression to `closest('.gnr-toast-message')` with `scrollHeight > clientHeight` keeps click-to-dismiss for the short messages that are nearly all of them.

Keeping the cap as a variable next to `--toast-max-width` is consistent with how the rest of the toast system is built, so an instance that wants a different ceiling has somewhere to set it.

fporcari:
@dgpaci — could you drop your review request on this one? The request is mine and it was a mistake on my side, not something you are late on.

I assigned it on 7 August as part of a batch, without checking the paths: the PR touches only `gnrjs/gnr_d11/css/gnrbase_css/01_gnr_variables.css`, `16_gnr_toast.css` and `js/genro_toast.js`, none of which you own under `.github/CODEOWNERS` — `gnrjs/` is @fporcari @genro and `*.js` is @fporcari. The toast code itself has no history from you either, so I was asking you to read a file you have no context on.

Meanwhile the change has been read by the two people who do have it: cgabriel, who reported the original defect, and genro, who is a code owner here and co-author of `genro_toast.js`. Both approved. So the outstanding request is not holding back a missing perspective, it is just holding the PR.

Happy to remove it myself if you would rather not bother — say the word.

Correcting something I wrote in the first version of this comment: I said the same batch had put you on six other PRs for the same reason, and that is not accurate. On #1059 you are the third author of `gnrpy/gnr/web/gnrmenu.py` and on #1057 you have history on `genro_widgets.js`, so those requests have a basis even though CODEOWNERS does not generate them — they stay. Only #1044 and #1045 are in the same position as this one.

fporcari:
Withdrawn the review request to @dgpaci, open since 7 August. Not a judgement on the review — the PR already has both approvals it needs (@cgabriel and @genro), so the request was the only thing keeping the queue entry open, and leaving it there reads as "still waiting on someone" to everybody scanning the list.

@dgpaci, if you want a look at this anyway the change is small — three files, +11/-1, CSS and JS only — say so and I will request you back before it merges.

PR #1047, opened 2026-08-07
fix(web): make directory creation idempotent to close the makedirs race

fporcari:
## Problem

Intermittent HTTP 500 while loading a table handler page. Observed 4 times over two months in production, affecting 3 different users, always on the page with the largest grid (the one using frozen selections):

```
File "gnr/web/gnrwebpage_proxy/apphandler/misc.py", line 649, in checkFreezedSelection
  selection = self.page.unfreezeSelection(dbtable=table, name=selectionName)
File "gnr/web/_gnrbasewebpage.py", line 214, in unfreezeSelection
  selection = self.db.unfreezeSelection(self.pageLocalDocument(name, page_id=page_id))
File "gnr/web/_gnrbasewebpage.py", line 189, in pageLocalDocument
  os.makedirs(folder)
FileExistsError: [Errno 17] File exists: '.../site/data/_connections/<conn_id>/<page_id>'
```

## Root cause

`pageLocalDocument` in `gnrpy/gnr/web/_gnrbasewebpage.py` is a check-then-act race:

```python
folder = os.path.join(self.connectionFolder, page_id)
if not os.path.isdir(folder):
    os.makedirs(folder)
```

Two concurrent requests belonging to the same page can both pass the `isdir` check; the second one fails in `makedirs`. The window is short, which is why it is rare, but it is hit regularly on busy pages and more easily with `liveUpdate` enabled.

## Change

The reported line is one instance of a pattern repeated across `gnr/web/`, so the fix is the pattern, not the single call site. Every occurrence of the guard-then-`makedirs` shape under `gnrpy/gnr/web/` becomes `os.makedirs(path, exist_ok=True)` with the now-redundant guard dropped — 15 sites in 10 files:

| file | sites |
|---|---|
| `_gnrbasewebpage.py` | 5 (incl. the reported one at :189) |
| `gnrwebpage.py` | 2 |
| `gnrwsgisite.py` | 1 |
| `gnrasync.py` | 1 |
| `gnrwsgisite_proxy/gnrwebsockethandler.py` | 1 |
| `gnrwebpage_proxy/jstools.py` | 1 |
| `gnrwebpage_proxy/developer.py` | 1 |
| `gnrwebpage_proxy/utils.py` | 1 |
| `daemon/handler.py` | 1 |
| `daemon/siteregister_client.py` | 1 |

Every guard was read before being removed and confirmed to exist solely to protect its `os.makedirs` call: none had a side effect worth preserving, none relies on `FileExistsError` being raised. No site was left alone within scope.

## Verification

- New regression test `gnrpy/tests/web/_gnrbasewebpage_test.py`: exercises `pageLocalDocument` unbound against a real `tmp_path`, with a two-thread concurrent case. `os.makedirs` is not mocked. The concurrent test reproduces the exact `FileExistsError: [Errno 17]` against the pre-fix code and passes against the fix — both directions confirmed.
- `flake8` (F, E9) on all touched files: zero errors introduced. Pre-existing `F841`/`F811` warnings at shifted line numbers were diffed against the base and confirmed identical.
- `python -m pytest gnrpy/tests/web/ -q`: 143 passed, 57 skipped (pre-existing skips, daemon not available in this environment).
- Full suite via the pre-commit hook: 1922 passed, 67 skipped.

## Follow-up (not in this PR)

The same check-then-act pattern exists outside `gnrpy/gnr/web/`, e.g. under `gnr/lib/` and `gnr/core/`. Left out of scope deliberately to keep this diff reviewable.

Fixes #1033

### DIFF
--- a/gnrpy/gnr/web/_gnrbasewebpage.py
+++ b/gnrpy/gnr/web/_gnrbasewebpage.py
@@ -185,8 +185,7 @@ class GnrBaseWebPage(GnrObject):
         :param page_id: TODO"""
         page_id = page_id or self.page_id
         folder = os.path.join(self.connectionFolder, page_id)
-        if not os.path.isdir(folder):
-            os.makedirs(folder)
+        os.makedirs(folder, exist_ok=True)
         return os.path.join(folder, docname)
         
     def freezeSelection(self, selection, name,**kwargs):
@@ -811,8 +810,7 @@ class GnrBaseWebPage(GnrObject):
         if args:
             path = os.path.join(self.siteFolder, 'data', *args)
             dirname = os.path.dirname(path)
-            if not os.path.exists(dirname):
-                os.makedirs(dirname)
+            os.makedirs(dirname, exist_ok=True)
             outfile = open(path, 'w')
             return outfile
             
@@ -821,8 +819,7 @@ class GnrBaseWebPage(GnrObject):
         if args:
             path = os.path.join(self.siteFolder, 'pages', 'static', *args)
             dirname = os.path.dirname(path)
-            if not os.path.exists(dirname):
-                os.makedirs(dirname)
+            os.makedirs(dirname, exist_ok=True)
             outfile = open(path, 'w')
             return outfile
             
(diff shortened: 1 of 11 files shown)


cgabriel (approved):
LGTM

PR #1048, opened 2026-08-07
fix(web): reset the thread-local currentPage on every task exit path

fporcari:
`site.currentPage` is backed by a `ThreadedDict` keyed on the thread ident: assigning a page stores it under that ident, assigning `None` pops the entry. `GnrWsgiSite.dispatcher()` already guarantees the pop from a `finally` block, but the task runners that set `currentPage` outside a request did not.

Same leak class as #1042 (currentDomain in `dispatcher()`), refs #379 / #380.

## Leaks fixed

| Location | Path that skipped the reset |
|---|---|
| `daemon/processes.py` `GnrWorker.run_batch` | an exception raised by `table_script_run()` |
| `daemon/processes.py` `GnrWorker.run_task` | an exception raised by `runTask()`, **and** the early `return` of the non-concurrent guard |
| `gnrtask.py` `GnrTaskWorker.runTask` | never reset it at all, on any path |
| `gnrtask_new.py` `execute_task` | never reset it at all, on any path |

In every case the entry survives for that thread ident forever, keeping the page and its db connection alive — the unbounded per-thread growth #379/#380 are about. The worker loops are long-lived, so this accumulates for the whole life of the process.

The second `run_task` leak was not in the original report: the concurrency guard returns early when the task is already running under another PID, leaking the entry with no exception involved.

## Changes

Each page-scoped block is wrapped in `try/finally`, mirroring `dispatcher()`'s `finally`. In `gnrtask_new.execute_task` the reset happens *before* the acknowledgement POST, so the ack still runs only on the paths where it ran before (an exception from the task already skipped it).

No behaviour change beyond the cleanup: no exception is swallowed, no early return removed.

## Tests

`gnrpy/tests/web/task_currentpage_cleanup_test.py` — the fake sites reuse the real `GnrWsgiSite.currentPage` property over a real `ThreadedDict`, and the runner methods are the real ones called unbound, so the tests exercise the actual pop-on-`None` semantics rather than a mock of them. Only what is external to the mechanism (resource loader, db, task classes) is faked. Style follows `gnrwsgisite_dispatcher_cleanup_test.py` from #1042.

11 tests, covering success / exception / early-return for each runner. **9 of the 11 fail without this change** (the 2 that pass are the happy paths in `processes.py`, which already reset explicitly).

## Verification

- `flake8` on the four touched files: zero errors.
- `pytest gnrpy/tests/web`: 151 passed, 57 skipped.
- Full suite: 1872 passed, 1 pre-existing failure — `core/gnrbag_test.py::TestBasicBag::test_fillFromUrl`, which fetches `https://www.genropy.org/feed/` and gets a 404. Verified it fails identically with these changes reverted.


cgabriel (commented):
That's cool, apparently can solve a leak in memory in such processes.

I'm commenting and not approving because I think it's worth adding in the old gnrtask execution (gnrtask.py, daemon/processes.py) the intercept of the possible exception re-raised by the try/finally block, handling it at least with a logger.error() entry like in gnrtask_new.

fporcari:
> I think it's worth adding in the old gnrtask execution (gnrtask.py, daemon/processes.py) the intercept of the possible exception re-raised by the try/finally block, handling it at least with a logger.error() entry like in gnrtask_new

Right for `gnrtask.py`, and there it was worse than a missing log line. Not needed in
`processes.py`. Done in d1b43ef.

**`gnrtask.py` — the exception did not just go unlogged, it ended the worker.**
`GnrTaskWorker.start()` had no handler anywhere: the `while True` loop calls `runTask`
bare, so the exception the new `finally` lets through propagated straight out of `start()`
and the worker thread died. Every task queued behind the failing one was never run either,
and since the failure happens inside `with recordToUpdate(...)` before
`task_execution['end_ts']` is written, the task looks unstarted and gets picked up again
on the next worker. So the loop now catches, logs with `logger.exception` (traceback
included, as `processes.py` does) and rolls back before moving to the next task.

The second half of your point stands on its own: `if not task_class: return` was silent,
while `gnrtask_new` logs `Can't find task class for command %s`. It logs the same line
now — a task pointing at a command that no longer exists was the one failure mode that
left nothing at all in the log.

**`processes.py` needs nothing, and adding it would make things worse.** `GnrWorker.start()`
already wraps `handler(item_value)` in a `try/except` that ends in
`self.logger.error(tb_text)` with the full formatted traceback (`processes.py:363-369`).
A `logger.error` inside `run_task`/`run_table_script_run` would report every failure twice,
once without the traceback. The `finally` blocks there are doing the only thing that was
missing.

Two new tests, both verified to fail on the previous head of this branch:

- `test_legacy_runtask_missing_task_class_is_logged` — the silent early return;
- `test_legacy_start_survives_a_failing_task` — three tasks with the middle one raising,
  asserting all three are attempted and the failure is logged. It drives the real
  `GnrTaskWorker.start` and breaks the infinite loop by making `sleep` raise, so the
  assertion is about the actual loop rather than a copy of it.

flake8 clean; `tests/web/task_currentpage_cleanup_test.py` 13 passed; the pre-commit
full run 2026 passed, 69 skipped.

cgabriel (dismissed):
Reviewed with the requested addition, LGTM.

genro (changes requested):
The leak fix itself is right and minimal. `ThreadedDict.set(None)` pops the entry (`gnrpy/gnr/core/gnrlang.py:648-669`), all three entry points are covered including the early return of the concurrency guard in `run_task`, and of the 507 lines 419 are the test while the 88 in production are indentation under `try` plus the two new blocks — no unrelated refactoring. The tests use no MagicMock: the real `currentPage` property over a real `ThreadedDict`, with the real methods called unbound.

Requesting changes on `d1b43efc52`, the commit added after the first review.

## A failed task is never scheduled again while the worker lives

`taskToExecute` (`gnrpy/gnr/web/gnrtask.py:123-131`) writes `start_ts` and `pid` on the execution record and **commits** before yielding the pkey. The new handler answers with `self.db.rollbackAll()`, which cannot undo a row that is already committed, so the record stays `start_ts IS NOT NULL AND end_ts IS NULL` carrying the live worker's pid.

From there:

- `sys.task.active_workers` (`projects/gnrcore/packages/sys/model/task.py:49-54`) counts exactly `$start_ts IS NOT NULL AND $end_ts IS NULL` for that task;
- the worker's own `where` (`gnrtask.py:117`) requires `$task_active_workers < COALESCE($task_max_workers,1)`;
- `checkAlive` (`gnrtask.py:98-107`) clears `pid`/`start_ts` only when `pid_exists(pid)` is false, and this worker is alive.

With `task_max_workers` unset, so 1, that task is never picked up again until the worker process restarts. Before this commit the exception left the loop, the worker died, its pid stopped existing, `checkAlive` released the record and the task was retried on the next pass. The commit trades "worker dies, task retried" for "worker survives, task never retried" — the defect it fixes is real, but the one it introduces is quieter, since nothing in the logs says the task stopped being eligible.

The remedy belongs in the same handler, and the table already has the columns for either reading: writing `end_ts` together with `is_error` and `errorbag` records the attempt as finished and failed, while clearing `start_ts` and `pid` reproduces what `checkAlive` does and keeps today's retry behaviour. Your call which, but one of the two is needed, followed by a `commit` — a `rollbackAll()` alone only discards `end_ts`.

## Minor, pre-existing

`processes.py:338-339`: `execution_dict.pop(task_id, None)` sits inside the `try` rather than the `finally`, so an exception from `runTask` leaves the entry behind and the concurrency guard discards that task for the life of the process. Not introduced here, but the body describes `run_task` as covered without separating the two pieces of state it holds — worth either moving the `pop` up into the `finally` or saying in the body that only `currentPage` is released.

fporcari:
Both points addressed in `14059c78cb`
([review](https://github.com/genropy/genropy/pull/1048#pullrequestreview-5098228314)).

## The claimed record

> Your call which, but one of the two is needed, followed by a `commit` — a `rollbackAll()` alone only discards `end_ts`.

Took the second: clear `pid` and `start_ts` on the failing record, then commit.

```python
self.db.rollbackAll()
self.tblobj.batchUpdate(dict(pid=None, start_ts=None),
                        _pkeys=[te_pkey],
                        for_update='SKIP LOCKED')
self.db.commit()
```

Same two columns and the same `for_update='SKIP LOCKED'` as `checkAlive`
(`gnrtask.py:105-107`), so there is one definition of "released" rather than
two. The reason for preferring it over `end_ts` + `is_error` + `errorbag`: this
PR is a leak fix, and the retry the worker used to get by dying is behaviour I
would rather keep than trade. Recording the failed attempt on the row is worth
having, but it changes what the task does after a failure — a separate decision,
and one that wants a column story for repeated failures (a retry count, or a
backoff), otherwise a task that fails deterministically stops being visible as
"pending" while nothing has been fixed.

One thing I left unguarded on purpose: if the `batchUpdate` itself raises the
worker still dies. A task blowing up is what this PR keeps the worker alive
through; a database that will not accept a two-column update on the row it just
handed us is a different class of failure, and surviving it silently would be
worse than restarting.

## `execution_dict`

> worth either moving the `pop` up into the `finally` or saying in the body that only `currentPage` is released

Moved — but not literally, because moving it as written breaks the concurrency
guard. On the guard's early return the entry in `execution_dict` belongs to the
**other** execution that is running the task, so an unconditional
`pop(task_id, None)` in the `finally` releases somebody else's claim and the
next call walks straight through the guard. It needs the flag:

```python
registered = False
...
        self.execution_dict[task_id] = os.getpid()
        registered = True
...
finally:
    if registered:
        self.execution_dict.pop(task_id, None)
    self.site.currentPage = None
```

I checked this rather than reasoned it: with the unconditional version,
`test_worker_run_task_concurrency_guard_leaves_no_thread_local_residue` fails.
That test now asserts the surviving entry explicitly, so the hazard is pinned
and not just avoided.

## Verification

`tests/web/task_currentpage_cleanup_test.py` 14 passed, flake8 clean on the
three files. Both new behaviours have a ratchet — reverting each makes its test
fail:

- `test_legacy_start_releases_the_record_of_a_failing_task` — only the failing
  pkey is released, with `{'pid': None, 'start_ts': None}`, and the release is
  committed;
- the guard assertion above.

**Perimeter, stated plainly.** These run `GnrTaskWorker.start` for real over the
file's existing fakes: what the test asserts is the call the loop makes, not the
row in a database. The end-to-end version belongs in the daemon web suite, which
#1036 has just made runnable but which still skips without `gnr` on PATH — #1246
is about making that skip a failure on CI, and once it is, this is worth an
integration test that reads the row back and checks the task is picked up again.

@cgabriel your approval fell with this push. The `logger.error` you asked for is
untouched; what changed is that the handler now also releases the row it had
claimed. Re-requesting both of you.

genro (changes requested):
Changes request on `14059c78cb`. The `execution_dict` half is right; the record release is not.

## `execution_dict`

Correct, and the flag is necessary rather than defensive: on the guard's early return the entry belongs to the other execution, so an unconditional `pop` in the `finally` would release someone else's claim. The test asserting the surviving entry pins it.

## The release turns "never retried" into "retried with no pause"

`taskToExecute` is a generator: `while f:` re-runs the query after every `yield`. The worker's `where` requires `$start_ts IS NULL`, the handler writes `start_ts=None` and `pid=None` and commits, `active_workers` counts `$start_ts IS NOT NULL AND $end_ts IS NULL` so it drops back to 0, and `$task_active_workers < COALESCE($task_max_workers,1)` passes again. The released row is the first one `limit=1, order_by='$__ins_ts'` returns, so the failing task is claimed again immediately, in the same pass, with no `sleep` in between.

I ran it rather than reasoned it: the real `GnrTaskWorker.start` and the real `taskToExecute`, over an in-memory table honouring `$start_ts IS NULL` and `limit 1`.

| head | attempts on one failing task |
|---|---|
| `14059c78cb` | 26, and `sleep` never reached |
| `14059c78cb^` | 1, then the pass ends normally |

The 26 is where my own escape hatch fired. It had to be a `BaseException`: an `Exception` sentinel is caught by the handler itself, and the first run ran until the two-minute timeout.

Each turn of that loop costs a `SELECT ... FOR UPDATE SKIP LOCKED`, the claiming `UPDATE`, two commits and a `logger.exception`. No other queued task runs meanwhile, because the failing one is always first.

Your tests cannot see it: `_FakeStartWorker.taskToExecute` returns a static list in place of the generator.

## Which failures actually reach the handler

Worth pinning down, because it decides whether the loop is theoretical. `btcbase.__call__` swallows the exception when the task carries a `task_execution_record`: it writes `is_error=True` and `errorbag` and does not re-raise (`btcbase.py:80-84`). So an in-batch failure never reaches your `except`. What reaches it is the failures outside the batch:

- `getBtcClass` raising, for instance an ImportError in the resource module;
- the `task_class(...)` constructor raising on that task's parameters;
- `btcbase`'s own `finally`, the `batch_log` insert;
- `recordToUpdate`, `page.db`, the final `commit`.

The first three repeat identically on every attempt. They are the class of failure this PR exists to survive, and the class that spins.

## The closure the framework already has

`is_error` plus `errorbag` plus `end_ts` is how a failed task is already recorded: `btcbase` writes the first two for the far more common in-batch failure, and the success path writes `end_ts`. Taking that road here makes an out-of-batch failure end the way an in-batch one already does, instead of being the only failure in the system that leaves the row in `waiting`.

The retry is not lost, which was your reason for preferring the release. I checked `writeTaskExecutions`: for `frequency` and `run_asap` tasks the reason is `*`, so `reasonkey` is `None`, `None not in existing_executions` always holds, and a fresh execution row is inserted at each useful scheduler pass. For cron tasks the next slot produces a new `reasonkey` and with it a new row. The retry comes back at the scheduler's cadence instead of at loop speed.

If you would rather keep the release, it needs the failing pkey excluded from the query for the rest of the pass, otherwise the generator hands it straight back.

fporcari:
Taken, and reproduced before changing anything. Pushed as `cdf27d7b8d`.

## The loop is real, and I measured it with your setup

From [your review](https://github.com/genropy/genropy/pull/1048#pullrequestreview-5099699380):

> `taskToExecute` is a generator: `while f:` re-runs the query after every `yield`. [...] The released row is the first one `limit=1, order_by='$__ins_ts'` returns, so the failing task is claimed again immediately, in the same pass, with no `sleep` in between.

Every premise holds against the code:

- `gnrpy/gnr/web/gnrtask.py:121-133` - generator, `while f:`, `limit=1`, `order_by='$__ins_ts'`, `commit()` before the `yield`;
- `gnrpy/gnr/web/gnrtask.py:116` - `$start_ts IS NULL` and `$task_active_workers<COALESCE($task_max_workers,1)`;
- `projects/gnrcore/packages/sys/model/task.py:49-54` - `active_workers` is `COUNT(*)` over `$start_ts IS NOT NULL AND $end_ts IS NULL`, so clearing `start_ts` really does drop it back to 0;
- `gnrpy/gnr/sql/gnrsqltable/helpers.py:170-171` - `RecordUpdater.__exit__` writes nothing when an exception is in flight, so the row reaching the handler is exactly the one `taskToExecute` claimed;
- `gnrpy/gnr/web/batch/btcbase.py:80-84` - `is_error` + `errorbag`, no re-raise, so what reaches the `except` is the out-of-batch class you list;
- `gnrpy/gnr/web/gnrtask.py:85-91` - `reasonkey` is `None` for a `*` reason, `None not in existing_executions` always holds, so the fresh execution row is inserted at each useful scheduler pass.

## What changed

`gnrpy/gnr/web/gnrtask.py:181-199` - the handler no longer releases the record, it closes it the way `btcbase` already closes an in-batch failure, plus the `end_ts` the success path writes:

```python
self.tblobj.batchUpdate(dict(is_error=True,
                             errorbag=Bag(dict(error=str(e))),
                             end_ts=datetime.now(timezone.utc)),
                        _pkeys=[te_pkey],
                        for_update='SKIP LOCKED')
```

`start_ts` stays set, so the row is out of the worker query for good; `end_ts` takes it out of `active_workers`, so the *next* execution of the same task is not starved by it. That second half matters: it is what the original release was really buying, and the closure buys it without the spin. `status` reads `error` (`projects/gnrcore/packages/sys/model/task_execution.py:37-43`) instead of leaving the row in `waiting`, which was the odd part - it was the only failure in the system that did.

I did not take the road you named as the alternative:

> If you would rather keep the release, it needs the failing pkey excluded from the query for the rest of the pass.

## The test, and the numbers it gives

You are right that the existing tests could not see it - `_FakeStartWorker.taskToExecute` returns a static list. New test at `gnrpy/tests/web/task_currentpage_cleanup_test.py:462`, `test_start_attempts_a_failing_task_once_per_pass`: real `GnrTaskWorker.__init__` (so the real `where`), real `taskToExecute`, real `start`, real `runTask` failing at `getBtcClass` - your first out-of-batch class. Only the site and the table are fakes: `_FakeExecutionTable` (`:394`) keeps rows in a list and honours `$start_ts IS NULL`, `$task_stopped IS NOT TRUE`, `active_workers < COALESCE(max_workers,1)`, `order_by $__ins_ts`, `limit 1`, so the generator's second query sees whatever the handler wrote. Two queued rows on the same task, both failing; it counts the claims. `_TooManyClaims` (`:385`) derives from `BaseException` for the reason you found - an `Exception` sentinel is eaten by the handler under test.

| head | claims |
|---|---|
| `14059c78cb` (the release) | `['e1'] x 11`, escape hatch at 10, `e2` never reached, `sleep` never reached |
| the release removed | `['e1']`, then the pass ends - but `e2` is starved, which is #380 |
| `cdf27d7b8d` (this push) | `['e1', 'e2']`, each claimed once, `sleep` reached |

The middle row is worth keeping in view: 1 attempt was never the target on its own, since it came with the starvation this PR opened for.

The old `test_legacy_start_releases_the_record_of_a_failing_task` became `test_legacy_start_closes_the_record_of_a_failing_task` (`:353`) and now asserts `is_error is True`, an `end_ts`, the `errorbag` message, and explicitly that `start_ts` is **not** in the updater.

## Checks

- `pytest gnrpy/tests/web/task_currentpage_cleanup_test.py -q` - **15 passed**.
- pre-commit hook (flake8 on staged files + `pytest core sql web app xtnd`) - flake8 OK, **2229 passed, 69 skipped**, 0 failures.
- Wider `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` - **1068 passed, 59 skipped**.

Not verified: no run against a live scheduler + worker pair on a real instance, so the scheduler-cadence retry is verified by reading `writeTaskExecutions` and not by watching a row reappear; and nothing here exercises the postgres `SKIP LOCKED` path with two concurrent workers.

The push dismissed the standing approvals on the base (`dismiss_stale_reviews`) - in practice there were none left standing, @cgabriel's had already been dismissed by the previous push. The `execution_dict` half you accepted is untouched: `processes.py` has no change in this commit, and the test pinning the surviving entry is still there and still green.

@genro re-requested.

genro (approved):
Both objections from my second round are answered, and the test is the part I want to single out. Replacing the static `taskToExecute` list with the real generator — real `__init__`, real `where`, real `start`, real `RecordUpdater` — is what makes `test_start_attempts_a_failing_task_once_per_pass` able to see the retry-with-no-pause I was worried about. A test that cannot fail on the bug it guards is worth nothing, and this one can.

I checked the closure is consistent rather than plausible. `start_ts` is no longer cleared, and the worker's `where` at `gnrtask.py:116` requires `$start_ts IS NULL`, so there is no spin. `end_ts` drops the row out of both `active_workers` and `checkAlive`, which use the same `end_ts IS NULL` predicate, so the slot is freed and the row cannot be resurrected. The shape matches `btcbase.py:82-83` exactly, which is the right precedent.

Two corrections to the description before you squash.

It says no exception is swallowed. One now is: `gnrtask.py:181` catches every out-of-batch task exception and writes an error row instead of letting the worker die. That is the change, and it should be what the commit message says. It also says 11 tests; the file has 15.

And the consequence that follows from it, which the description does not state. A cron task failing out of batch is now closed `is_error` for that slot permanently. Before, the worker died, `checkAlive` cleared `start_ts` once the pid was gone, and the task was retried. `writeTaskExecutions` will not re-insert that reasonkey. I think it is the right call, by symmetry with the in-batch path, but it is a semantic change to cron behaviour and it deserves a sentence.

Minor: the closed row keeps its `pid`. Nothing queries it once `end_ts` is set, but in the grid it reads as a live claim.

cgabriel (approved):
LGTM, approving

fporcari:
Merged. Both corrections are in the squash message rather than in the description, which stays the text you and cgabriel reviewed against.

> It says no exception is swallowed. One now is: `gnrtask.py:181` catches every out-of-batch task exception and writes an error row instead of letting the worker die. That is the change, and it should be what the commit message says.

It says it now, in its own paragraph — including the part the description never stated:

> That changes cron semantics and is worth stating plainly: an out-of-batch failure now closes that slot `is_error` for good. Before, the worker died, `checkAlive` cleared `start_ts` once the pid was gone, and the same execution was retried. Releasing `start_ts` instead would hand the row straight back to the generator; the retry comes from the scheduler, which inserts a fresh execution on its next useful pass.

On the test count: you are right, 15 and not 11. It is the one claim I could not correct without rewriting the description, and it is not worth a line in the commit message — the file is the record there.

One thing the merge fixed that the PR did not: `Fixes #380` was missing from the body, so nothing was linked. It is in the squash message, and #380 is now closed.

The `pid` left on the closed row stays as it is — nothing queries it once `end_ts` is set, and the grid reading it as a live claim is a display question for whoever touches that view next.

PR #1049, opened 2026-08-07
Make GnrDummySite instantiable without a gnrdaemon

fporcari:
`GnrDummySite` could not be instantiated at all. Nothing in the repo currently constructs one, which is why this went unnoticed — the class is genuinely useful for headless, CLI and test contexts, so this makes it work.

## Three separate problems

1. **`_register` was never initialised.** `GnrDummySite.register` reads `self._register`, but in `GnrWsgiSite` the register lives on the domain proxy, not on the site. Constructing a `GnrDummySite` raised `AttributeError: 'GnrDummySite' object has no attribute '_register'` from `GnrWsgiSite.__init__`, which touches `self.register` on purpose.

2. **`FakeRegister` had no `siteregister`.** `GnrWsgiSite.__init__` builds `DataCollector(self.register.siteregister)`.

3. **Overriding `register` was not sufficient.** `GnrWsgiSite.main_register` resolves through `self.domains[self.rootDomain].register`, which builds a real `SiteRegisterClient` and needs a running gnrdaemon. Anything that reads application cache or preferences goes through `main_register`.

## Changes

`gnrpy/gnr/web/gnrdummysite.py`:

- class level `_register = None`, so the attribute exists before `__init__` touches `self.register`
- `FakeRegister` gains the surface a site actually touches outside a request: `siteregister`, the four per scope stores (`globalStore`, `pageStore`, `connectionStore`, `userStore`) and `get_dbenv`
- new `FakeSiteRegister`, the read only proxy the `DataCollector` asks for (`users`, `pages`, `connections`, `counters`)
- new `FakeStore`, a null store usable both directly and as a context manager — `WebApplicationCache.updatedItem` enters it as `with register.globalStore() as gs:`
- `main_register` overridden so it does not fall through to the daemon backed domain proxy

## Tests

`gnrpy/tests/web/gnrdummysite_test.py` was a one line import stub that passed while the class was broken. It now builds a real site on the isolated `gnrtest` instance of `BaseGnrTest`, with **no daemon running**, and renders through it:

- `site.dummyPage.rootPage()`
- the print path: `BagToHtmlWeb` over `adm.user`, producing HTML with `print_stylesheet.css` resolved through `site.dummyPage`
- `main_register` never touches the proxy: asserts `site.domains[site.rootDomain]._register is None`
- store semantics, `get_dbenv` returning an empty `Bag`, the application cache round trip, the empty `DataCollector`

The editable install resolves `gnr.*` to the main checkout, so the suite is run with `PYTHONPATH=<checkout>/gnrpy` prepended and a test asserts `gnrdummysite.__file__` is inside this working tree.

```
cd gnrpy && PYTHONPATH=$PWD python -m pytest tests/web -q
152 passed, 57 skipped
```

flake8 clean on both files.

Each of the three original failures was reproduced separately against the old class before fixing: `AttributeError: '_register'` → `AttributeError: 'siteregister'` → `Pyro4.errors.CommunicationError: cannot connect to ('localhost', 40675)`.


cgabriel (approved):
That's great, looking good, and really useful in several context. Approved.

genro (approved):
The three failures are real and independent, and the fixes match them one to one: a class-level `_register = None` (the attribute has to exist before `GnrWsgiSite.__init__` reads `self.register`), `siteregister` on the fake for the `DataCollector`, and the `main_register` override — which is the one that matters most, since `GnrWsgiSite.main_register` resolves through `self.domains[self.rootDomain].register` (`gnrwsgisite.py:645-650`) and would build a real `SiteRegisterClient` behind the site's back.

Checked the blast radius through Sourcerer across every indexed project: `GnrDummySite` has no constructor anywhere outside its own module and its test. So this is purely additive — no existing caller can be affected by the widened fake surface, and the class goes from "cannot be instantiated" to usable for headless/CLI contexts.

The fakes stay honest: null stores that also work as context managers (which is how `WebApplicationCache.updatedItem` enters the global store), a read-only siteregister with just the four readers the collector asks for, `get_dbenv` returning an empty Bag. Nothing pretends to persist.

The test file going from a one-line import stub — which passed while the class was broken — to a suite that builds a real site with no daemon and renders through it is the right correction, and asserting `site.domains[site.rootDomain]._register is None` is the assertion that keeps the daemon out for good. The `gnrdummysite.__file__` check against the checkout is a nice guard given the editable install.

Issue #1018, opened 2026-08-04
[BUG] multidb: subscription row can exist without the physical record in the store, and nothing repairs or reports it


fporcari:
## Problem Description

In the `multidb` package the physical copy of a subscribed record into a store only happens in the subscription insert trigger (`trigger_onInserted` → `syncStore`). Two code paths break the implicit invariant *subscription row ⇒ record present in the store*, and once broken nothing ever detects or repairs it:

1. **`subscription.addSubscription`** (`projects/gnrcore/packages/multidb/model/subscription.py`, ~line 53): when `checkDuplicate` finds an existing subscription row it returns silently. The call looks idempotent, but it does not verify (nor restore) the physical record in the store. If the record is missing, repeated calls to `addSubscription`/`multidbSubscribe` never heal it.

2. **`subscription.syncStore`** (same module, ~line 92): `if not self.db.dbstores.get(storename): return` — a subscription created while the target store is not (yet) registered skips the physical copy silently, with no log and no error. The subscription row stays there forever, orphaned.

Expected behavior: either `addSubscription` verifies/re-syncs the physical record (self-healing), or at minimum both paths log the skipped sync so the hole is visible. `TableBase.checkSyncAll` in `multidb/main.py` already contains reconciliation logic: exposing/documenting it as the official remedy for subscription/record misalignment would also help.

## Reproduction Steps

1. Multidb instance with a `multidb=True` table (e.g. an anagrafica-like table).
2. Create a subscription for a record towards store X while store X is not registered in `db.dbstores` (or delete the record in the store afterwards): subscription row exists, physical record absent.
3. Call `tbl.multidbSubscribe(pkey=..., dbstore='X')` again: it returns silently, the record is still missing in the store, nothing is logged.

## Environment

- genropy current master (checkout b4c6fad6f, 2026-07-30), PostgreSQL, multidb instance with rootstore + per-customer stores.

## Additional Material

Real-world case: in a production multidb application, supplier registry records referenced by invoices were missing from secondary stores although their subscription rows existed in the root. Every `addSubscription` call skipped them as duplicates, so the hole survived indefinitely and had to be fixed with an ad-hoc upgrade script that re-checked physical presence per store and re-synced via `subscription.syncStore(event='I', ...)`.

fporcari:
**Status after #1050, and three corrections to the description above.**

[...]

PR #1050, opened 2026-08-07
fix(multidb): log the three silent paths that leave a subscription without its record

fporcari:
## Problem

Real-world case reported in #1018: in a production multidb application, supplier registry records referenced by invoices were missing from secondary stores although their subscription rows existed in the root. Every subsequent `addSubscription` call skipped them as duplicates, so the hole survived indefinitely and had to be fixed with an ad-hoc upgrade script that re-checked physical presence per store and re-synced via `subscription.syncStore(event='I', ...)`. Nothing in the code path logged the skip, so the hole was invisible until someone went looking for it.

## What actually happens

Three silent skips in `projects/gnrcore/packages/multidb/model/subscription.py`, none of them logged before this change:

1. **`addSubscription` (line 53)** — `if not self.checkDuplicate(**dict(record)): self.insert(record); self.db.table(table).syncChildren(pkey)`. This is a positive guard, not an early `return` as the issue text describes: when a subscription already exists, both `insert()` (the row) and `syncChildren()` are skipped by simply not entering the `if` body. On an existing subscription the physical record and its children are never (re)synced, silently.
2. **`syncStore` (lines 92-93)** — `if not self.db.dbstores.get(storename): return`, no log, applies uniformly to `I`/`U`/`D` events. Reachable more easily than the issue implies: `db.dbstores` (`gnrpy/gnr/app/gnrapp.py:186-201`) is a cached dict built by intersecting storetable rows with *physically existing* databases (`get_dbdict`, `gnrpy/gnr/app/gnrapp.py:202-208`), so the store is absent whenever the DB is not created yet, another process created it and this process's cache is stale, or during store activation.
3. **`syncStore` (lines 97-101)** — a third silent return the issue does not mention: the master record is not found in root, so the sync is skipped with no trace.

## Change

Added a logger (`from gnr.app import pkglog as logger`, the idiom already used elsewhere in this package — see `projects/gnrcore/packages/multidb/lib/utility.py:6,13`) and three log points, nothing else:

- `syncStore`, store not registered → `logger.warning` (storename, table, pkey, event). A genuine anomaly — the store should exist, this is rare.
- `addSubscription`, existing subscription (duplicate branch) → `logger.debug`, **not** `warning`. `checkForeignKeys` (`main.py:232-247`) is invoked from `trigger_onInserting_multidb` (`main.py:262`) and `_onUpdating_slave` (`main.py:293`) — i.e. on **every record inserted or updated into a secondary store during sync**, once per non-null FK. In steady state essentially all of those hit an existing subscription; the duplicate branch is the normal path, not the exception. At warning level it would flood production logs.
- `syncStore`, master record missing in root → `logger.debug`.

No behaviour change: same control flow, same return points, only logging added.

## Verification

- `flake8` on the touched file: zero errors.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/ -q` as a no-regression gate. Honest report: counts vary noticeably between runs (e.g. 1 failed/1260 passed/620 errors vs 25 failed/1790 passed/66 errors vs 46 failed/1438 passed/398 errors), all variance concentrated in `gnrpy/tests/sql/*`. Root cause confirmed directly: the suite starts an ephemeral Postgres instance per session (`testing.postgresql`) and back-to-back full-suite runs in this environment race with it (`pg8000.exceptions.DatabaseError: ... 'the database system is shutting down'`) — a pre-existing infra flakiness. Confirmed unrelated to this change: `projects/gnrcore/packages/*` is outside pytest's scope (`gnrpy/pyproject.toml` restricts `testpaths` to `gnrpy/tests`), and no test under `gnrpy/tests` imports `multidb/model/subscription.py`; reverting the patch (via diff/apply, not stash) and rerunning reproduces the same class of sql/* flakiness. The only failure outside `sql/*`, in every run, is the documented pre-existing `gnrbag_test.py::test_fillFromUrl` (network 404), confirmed in isolation.
- **No automated test accompanies this change.** `projects/gnrcore/packages/` has no pytest coverage at all, and there is no multidb fixture or registered test store (`gnrpy/tests/sql/test_gnrsql.py:111` asserts `db.dbstores == {}`). A real test would need two physically created Postgres databases under the instance's multidb prefix, since `dbstores` is computed by listing real databases — that's a test-infrastructure project on its own, and a mock-everything substitute is explicitly against this repo's testing policy.
- Manual check (not run in this session): deregister a store on a real multidb instance, trigger a sync towards it, confirm the new warning line appears in the log.

## Not in this PR

The healing itself is a separate effort; recording the findings here so the hole stays visible:

- **The issue's proposed remedy (`checkSyncAll`) does not apply.** `checkSyncAll` (`main.py:484-496`) is dispatched only for `multidb=='*'` tables (`main.py:135`) — full-replication tables with no subscription rows at all — and reconciles by `empty()` + `insertMany` (`main.py:446-448,473-481`). It never reads `multidb.subscription` and is not even reached for `multidb=True` tables.
- **The relevant reconciler is `checkSyncPartial`** (`main.py:361-417`), dispatched for `multidb is True` (`main.py:137`) — **and it has the same gap**: line 369 builds `store_subs = {pkey: subscription_id}`, line 371 fetches only records *present in the store* and the loop iterates that set, line 381 pops matched pkeys. The loop ends at lines 415-416 and `store_subs` is never read again — leftover entries are exactly "subscription row whose record is missing from the store", silently discarded. No code path in the repo repairs or reports this today.
- **An adjacent defect in the same function**: `main.py:397-399` deletes the store record and its subscription with no `continue`, so execution falls through to `getRecordDiff` (line 400) and can `raw_update` (line 414) a record that was just deleted. Worth its own issue.
- **Reachability**: `checkFullSyncTables` → `checkFullSyncTables_do` (`main.py:106-145`) has exactly one caller repo-wide — `projects/gnrcore/packages/multidb/webpages/multidb_dashboard.py:133`, an admin "Check" button. No CLI, no batch job, no scheduled task. Any healing remedy is only useful if it is also exposed somewhere reachable.
- **The closest thing to a cure that ships today**: `mdb_reTouchAllSync` (`projects/gnrcore/packages/multidb/resources/multidb_components.py:102-107`, wired at line 46) loops all master records calling `syncStore(event='U')`, which does insert missing store records. Table-wide and store-wide — a sledgehammer, but it works, and is the closest existing workaround for the production story above.
- **Dead code**: `checkSubscribedRecord` (`subscription.py:17-25`) has the literal early `return` the issue originally described, and has zero callers repo-wide.
- **Follow-up housekeeping**: `main.py` uses bare `print()` at lines 116, 219, 223, 273, 406 instead of the package logger; left untouched here to keep this PR to the three points in scope, but worth a follow-up cleanup.

Refs #1018


genro (approved):
Approving. Nine lines, all `logger`, no branch added and no branch removed — the `else` that appears is an else on an existing `if`, carrying only a `debug` call.

The value is in which three paths were chosen. All three are places where multidb decides to *skip* — subscription already present, store not registered, master record missing — and skipping silently is exactly how a subscription row ends up without its physical record and nobody knows when it happened. Logging the skip with table, pkey, store and event is what makes the state reconstructable afterwards.

`logger.warning` for the unregistered store and `debug` for the other two is the right split: the first is a configuration problem, the other two are legitimate no-ops that only matter when you are chasing one.

cgabriel (approved):
Logging is always good.

Issue #1022, opened 2026-08-05
[BUG] htmltopdf/weasyprint: explicit pageSize/pageMargin stylesheet never wins over the document's own @page rule


dgpaci:
### Problem Description

The weasyprint implementation of the htmltopdf service appends the `pageSize`/`pageMargin` `@page` stylesheet (`resources/common/services/htmltopdf/weasyprint.py`) as a **user stylesheet with normal declarations**. In the CSS cascade, user-normal declarations lose to author declarations, so whenever the source document declares its own `@page` rule the explicit `pageMargin`/`pageSize` passed to `writePdf` is silently ignored.

Every `GnrHtmlBuilder` document emits `@page { size: ...; margin: 0; }` (`gnrpy/gnr/core/gnrhtml.py`), so an explicit `pageMargin` is dropped on **all builder-generated prints**.

Measured during the #1007 finalize review: rendering a document declaring `@page{margin:0}` with `pageMargin='5mm'` places content at (0.0, 0.0)pt — identical to passing no stylesheet at all.

Known affected caller: `resources/common/services/pdf/pypdf.py` `_createPdf`, which always passes an explicit `pageMargin`.

Expected behavior: an explicit `pageMargin`/`pageSize` argument should win over the document stylesheet (e.g. by emitting those declarations with `!important`, as the `sys.pdf_render` preference margins do since #1007 — those are NOT affected by this bug).

### Reproduction Steps

1. Create an HTML file whose CSS declares `@page { margin: 0 }` (any `GnrHtmlBuilder` output qualifies).
2. Render it through the weasyprint htmltopdf service: `service.writePdf(src, dest, pageSize='A4', pageMargin='5mm')`.
3. Inspect the produced PDF (e.g. PyMuPDF text positions).

Observed result: content starts at 0pt from the page edge — the 5mm margin was dropped.

The behavior is documented in-repo by `gnrpy/tests/core/htmltopdf_test.py::test_weasyprint_explicit_pagemargin_skips_pref_margins` (see docstring), whose assertions deliberately pass both with the current bug and with the fixed behavior.

### Environment

genropy develop (weasyprint implementation of the htmltopdf service); discovered while working on #1007.

Issue #1007, opened 2026-07-29
[FEATURE] Restore default PDF margin preferences and apply them in all htmltopdf implementations


dgpaci:
### Summary

Restore the default PDF margin preferences in `sys.pdf_render` and make them implementation-independent: they must be shown regardless of the selected htmltopdf implementation and actually applied by every implementation (WeasyPrint included), not only by wkhtmltopdf.

### Problem or Need

The `sys.pdf_render` preference panel (`projects/gnrcore/packages/sys/resources/preference.py`, `PDF Render` group) still contains `margin_top`, `margin_bottom`, `margin_left`, `margin_right` (and `zoom`) fields, but they are hidden unless the wkhtmltopdf legacy mode checkbox (`wk_legacy`) is enabled:

```python
fb.checkbox(value='^.wk_legacy',label='!![en]Legacy mode (use wkhtmltopdf)')
fb.textbox(value='^.margin_top',lbl='!![en]Margin top', hidden='^.wk_legacy?=!#v')
...
```

At runtime the preference is merged into `pdf_kwargs` by `HtmlToPdfService.htmlToPdf` (`gnrpy/gnr/lib/services/htmltopdf.py`), but only the wkhtmltopdf service (`resources/common/services/htmltopdf/wk.py`) consumes it, translating each key into a CLI option (`--margin-top`, ...). The WeasyPrint service (`resources/common/services/htmltopdf/weasyprint.py`) has the signature `writePdf(self, srcPath, destPath, pageSize=None, pageMargin=None, stylesheets=None, **kwargs)`: `pdf_kwargs` ends up in `**kwargs` and is silently ignored.

So since WeasyPrint became the default implementation, the site-wide default margins are effectively dead: documents that declare `@page { margin: 0 }` (which is what `GnrHtmlBuilder` emits) are rendered edge-to-edge and get clipped by the printer's non-printable area when printed at 100% on real paper. With wkhtmltopdf this never happened because it enforced its own default page (A4) and margins and smart-shrank the content to fit.

### Proposed Solution

1. In the `sys.pdf_render` preference panel, show the margin fields unconditionally (drop the `hidden='^.wk_legacy?=!#v'` condition on them; `zoom` can stay wkhtmltopdf-only since it maps to a wkhtmltopdf CLI flag).
[...]

PR #1052, opened 2026-08-07
fix(htmltopdf): an explicit pageMargin now wins over the document's own @page rule

fporcari:
## Problem

`resources/common/services/htmltopdf/weasyprint.py` builds an `@page` stylesheet from
`pageSize`/`pageMargin` and passes it to `write_pdf(stylesheets=...)`. WeasyPrint hard-wires
that channel to CSS origin `'user'` at normal priority (`weasyprint/document.py:216-223`,
`weasyprint/css/__init__.py:1143-1144`), while the document's own `<style>` rules are origin
`'author'` (`css/__init__.py:1140-1141`). The cascade order
(`declaration_precedence`, `css/__init__.py:553-573`) is:

```
user-agent (1) < user-normal (2) < author-normal (3) < author-important (4) < user-important (5)
```

So a document declaring its own `@page { margin: ... }` always won against the margin this
service tried to impose — an explicit `pageMargin` was silently dropped whenever the source
HTML had its own `@page` margin.

## Fix

Emit the margin with `!important`, but **only** when `pageMargin` was actually supplied by the
caller — never for the `pageMargin or 0` fallback, so a caller that does not ask for a margin
still defers entirely to the document. `size:` is left as a normal declaration on purpose:
measured, `size: A4 !important` forces a 215×292&nbsp;mm letterhead document down to
595.3×841.9&nbsp;pt. `BagToHtmlWeb.writePdf` (`gnrpy/gnr/web/gnrbaseclasses.py:299`) passes
`pageSize=self.page_format` (default `'A4'`) on *every* template print, while the document may
declare its own `size:{w}mm {h}mm`; making `size` important would silently reformat every
oversized legacy print.

Folded into the same block: the guard `if pageSize:` → `if pageSize or pageMargin:` (a caller
passing `pageMargin` alone previously got no stylesheet at all), and the dead `page_css = None`
is dropped.

`GnrHtmlBuilder`'s own `margin:0` rule (`gnrpy/gnr/core/gnrhtml.py:375-381`) is untouched and
should stay: it is deliberate, since the builder already realizes margins as absolute offsets
inside the page box (`gnrhtml.py:462-468`, `:387-388`, `:476-477`) — an `@page` margin on top of
that would double-count.

## Verification

`weasyprint 64.1` / `pymupdf 1.25.3` (both hard runtime deps, `gnrpy/pyproject.toml:44,61`).
Ran the patched `Service.writePdf` directly (throwaway script, not part of this PR) against a
document declaring `@page { margin: 0 }`, and measured where the rendered marker text lands
with PyMuPDF:

| call | marker (x0, y0) |
|---|---|
| explicit `pageMargin='5mm'` vs doc's `@page{margin:0}` | (14.17, 14.17) pt = exactly 5&nbsp;mm |
| `pageSize='A4'` only, no `pageMargin`, same doc | (0.00, 0.00) pt — unchanged, no forced margin |
| doc `@page{size:215mm 292mm}`, caller passes `pageSize='A4'` | page stays 609.4×827.7 pt (215×292 mm), not forced to A4 |

`flake8` (project config, `select = F401,E9,F63,F7,F82`) is clean on the touched file.

`PYTHONPATH=gnrpy pytest gnrpy/tests/ -q` was run as a no-regression gate, both with and without
this patch (patch removed via a diff file, not `git stash`, then reapplied). No test in the
suite references `htmltopdf`, `weasyprint`, or `writePdf` at all, and the failure/error sets are
identical in shape with and without the patch: both runs show `RuntimeError: initdb failed`
across the `*_postgres*` test classes, consistent with concurrent ephemeral-Postgres test runs
colliding in this shared environment, not with this change. Counts vary between runs (even
baseline vs. baseline) for the same reason. `gnrbag_test.py::test_fillFromUrl`, previously flagged
as a pre-existing network-dependent failure, passed in this session (network reachable now).

## Affected callers and cross-implementation notes (not fixed here)

- `resources/common/services/pdf/pypdf.py:112-113` `_createPdf` always passes both `pageSize`
  and `pageMargin`, so its margins are dropped today by the bug this PR fixes for the general
  path. It has no in-repo live caller today (`sys/webpages/ep_table.py:43` passes `input_pdf`
  instead), so the blast radius of this specific caller is application code, not this repo.
- `resources/common/services/htmltopdf/wk.py:10` (wkhtmltopdf implementation) drops
  `pageSize`/`pageMargin` into `**kwargs` and discards them entirely, honouring only
  `pdf_kwargs` as `--margin-*` CLI flags, which sit above CSS entirely. The two implementations
  are opposite in both directions today, and neither honours an explicit `pageMargin` before
  this fix. `wk` is legacy (gated at `gnrpy/gnr/lib/services/htmltopdf.py:21-25`) and is not
  touched here.
- The weasyprint service also silently ignores `page_height`, `page_width`, and `orientation`
  (they land in `**kwargs`, `weasyprint.py:10`). Left as a follow-up.

## Behaviour flip — decision needed

After this fix, a document declaring `@page{margin:15mm}` rendered by a caller passing
`pageMargin='5mm'` now gets 5&nbsp;mm — the caller wins. That is the semantics this issue asks
for, but since the margin is now emitted `!important`, and user-`!important` also beats a
document's own `!important`, **there is no escape hatch left in the document**: the only way for
a document to keep its own margin is for the caller to stop passing `pageMargin`. Flagging this
explicitly rather than settling it here.

## Test tightening, deferred

`gnrpy/tests/core/htmltopdf_test.py` does not exist on `develop`. It exists only on the unmerged
`feature/1007-pdf-margin-preferences` branch (no PR open), together with the `!important`
precedent this fix copies. Adding it from scratch here would conflict with that branch, so this
PR is source-only.

Once #1007 lands, its existing `test_weasyprint_explicit_pagemargin_skips_pref_margins` needs:

- a rename to `..._wins_and_skips_pref_margins`,
- the "known pre-existing bug" paragraph dropped from its docstring,
- a tightened assertion — it currently only checks `x0 < 8 * MM_TO_PT`, which a buggy 0.0 and a
  fixed 14.17 both satisfy:

  ```python
  assert 5 * MM_TO_PT - 1 <= x0 < 8 * MM_TO_PT
  assert 5 * MM_TO_PT - 1 <= y0 < 8 * MM_TO_PT
  ```

- two new cases: an explicit `pageMargin` beating a document's `@page{margin:15mm}`
  (`DOC_PAGE_MARGIN_15MM` already exists in that file), and a guard that the `pageSize`-only path
  still leaves `test_weasyprint_pref_margins_beat_document_zero_margins` green.

## Sequencing

`feature/1007-pdf-margin-preferences` has no PR open yet, and this fix's test belongs in that
branch's test file. Landing #1007 first, then this fix on top, is the clean order.

Refs #1022


cgabriel (changes requested):
The change makes sense, but I'm considering what happens when the caller explicitly requests a pageMargin of '0', without providing a pageSize.

line 14 will just skip the stylesheet creation+append, de facto ignoring the caller request for such margin.

fporcari:
> [I'm considering what happens when the caller explicitly requests a pageMargin of '0', without providing a pageSize. line 14 will just skip the stylesheet creation+append, de facto ignoring the caller request for such margin.](https://github.com/genropy/genropy/pull/1052#pullrequestreview-4973215095)

Confirmed, and the guard was the odd one out in its own diff: the two lines it wraps already test `pageMargin is not None`, while the guard itself relied on truthiness, so `pageMargin=0` (and `'0'`) fell through and no `@page` rule was emitted at all.

Fixed in 7b8fcb7bd62c97e21fd07209d968f690235001f1:

```python
if pageSize or pageMargin is not None:
```

With that, an explicit margin of `0` or `'0'` with no `pageSize` produces `@page { size: A4; margin: 0 !important; }`, `pageSize` alone keeps the previous non-`!important` default of `0`, and both `None` still leaves the document's own `@page` rule untouched.

flake8 clean on the file. There is no test covering the htmltopdf services, so this one is verified by reading, not by a test run.

Review re-requested.

cgabriel (approved):
LGTM now, the fixes handles the previous comment about pageMargin at 0.

genro (approved):
Verified the cascade reasoning against weasyprint's own source pins (user-normal < author-normal < user-important) and the diff against the review that preceded it:

- `!important` is attached only when the caller actually supplied `pageMargin` — the `else 0` fallback stays a normal declaration, so documents that manage their own margins are untouched.
- The `pageMargin=0` case cgabriel raised is handled by the three `is not None` guards: an explicit zero now produces the stylesheet and wins over the document.
- Leaving `size:` at normal priority is the right call for the legacy oversized prints, for exactly the reason the body measures.

The follow-ups (pypdf caller, wk divergence, ignored kwargs) are correctly left out. Fine to land ahead of #1007; the tests sketched in the body belong with that branch.

PR #1053, opened 2026-08-07
fix(crypto): constant-time signature check and a guarded expiry in verify_url

fporcari:
## What this issue is

Issue #1004 is a review, not a bug report: @genro filed a list of concerns about `AuthTokenGenerator` (`gnrpy/gnr/core/gnrcrypto.py`) while evaluating it as a reference for a replacement he is building elsewhere, said he is not working on this code, and @cgabriel agreed with most of the points and reassigned it back. **Closing the issue as answered is a legitimate outcome** once the invisible-to-valid-tokens part is addressed — most of the remaining points are deliberate design choices or trade-offs for the team to weigh, not defects to silently fix.

This PR implements only the part of the review that changes nothing for a valid token: a timing-safety hygiene fix and one crash-to-contract fix in a live code path.

## Verified findings

All reproduced by executing the real class (`PYTHONPATH=.../gnrpy python3 -c "..."`), not read off the issue text.

| Point | Verdict |
|---|---|
| 1. Non-numeric expiry silently disables expiry (`gnrcrypto.py:214`) | REAL. `generate('payload', expire_ts='2020-01-01')` then `verify()` returns `'payload'` with no expiry check. @cgabriel's "intentional" explains the `if expire_ts` gate at `:198` (empty expiry means no expiry), **not** the `isnumeric()` gate at `:214`, which silently swallows a *malformed* expiry instead of rejecting it. |
| 1b. New, not in the issue, and in the **live** path | `verify_url` had no `isnumeric()` guard: `int(expire_ts)` at `:253` raised `ValueError: invalid literal for int() with base 10: ...` on a URL like `/unsub/1?_vld=2020-01-01;<sig>`. The only caller, `gnrpy/gnr/web/gnrwebpage.py:1261-1263`, expects `verify_url` to return a string or `None` and wraps only that into `GnrSignedTokenException` — the `ValueError` escaped unhandled. |
| 2. Separator not escaped, payload truncated (`:208`, `:211-213`, double `rsplit`) | REAL, reproduced: `'{"exclude": "*.tmp;*.bak", ...}'` verifies to `'{"exclude": "*.tmp'` with a valid signature; `generate('order;1600000000')` with no expiry raises `AuthTokenExpired` on `verify()`. |
| 3. Comparison not constant-time (`:209`, `:247` use `!=`) | REAL, hygiene issue. |
| 4. SHA-1 + hand-rolled KDF (`:190-193`) | REAL, but a deliberate trade-off: `external_secret` (`gnrpy/gnr/web/gnrwsgisite.py:1190-1191`) is persisted via `getEnvironmentItem(..., update=True)`, so changing the digest/KDF would invalidate every outstanding signed link. |
| 5. `verify_url` returns `None` on success (`:237-254`) | REAL as an API-shape complaint; the caller (`gnrwebpage.py:1261-1263`) handles it correctly today. |
| 5b. "the test doesn't assert the valid case" | **Misreading — corrected here.** `gnrpy/tests/core/gnrcrypto_test.py:66-68` (pre-existing, unmodified by this PR) does exactly that: `r = generate_url(url, expire_minutes=200); r2 = verify_url(r); assert r2 == None`. The `None`-on-success contract is already pinned by the suite. |

**Scope note**: `generate()`/`verify()` (the raw-payload pair) have no caller anywhere in this repo — the only live uses are `generate_url` and `verify_url` (`gnrwsgisite.py` / `gnrwebpage.py:1261`). Points 1 and 2 are therefore latent in-repo today, consistent with @cgabriel's "URLs only" reading. Point 1b is the one exception: it sits in the live `verify_url` path.

## Change (points 3 and 1b only)

1. **Constant-time comparison** — `gnrcrypto.py:209` and `:247`: replaced `self._sign(val) != signature` with `not hmac.compare_digest(self._sign(val), signature)`. `hmac` is already imported at the top of the module. In-repo precedent: `gnrpy/gnr/web/gnrcookie.py:43` — `safe_str_cmp = lambda a, b: hmac.compare_digest(str_to_bytes(a), str_to_bytes(b))`. No behaviour change for any valid or invalid token; no signed link is invalidated.
2. **Point 1b** — `gnrcrypto.py:250-253`: added an `isnumeric()` guard before `int(expire_ts)` in `verify_url`, returning `"not_valid"` (the function's existing contract for a bad token) instead of letting a malformed `_vld` value raise `ValueError`. Behaviour-preserving for every valid token; turns an unhandled exception into the function's own documented failure mode.

Both changes are one-caller-visible-behavior-preserving: a valid, unexpired token/URL verifies exactly as before.

## Deliberately not done, and why

- **`:214` (point 1 proper)** — not touched. Fixing it would flip behaviour @cgabriel called intentional; whether the `isnumeric()` gate should also reject (rather than silently ignore) a malformed expiry on the raw `generate`/`verify` pair is a team call, not a drive-by fix, especially since a matching test (`test_verify_non_numeric_expiry_is_ignored`) now pins the current behaviour explicitly.
- **Token encoding (point 2, separator escaping)** — not touched. Fixing the truncation requires changing how payloads are encoded/split, which invalidates every token issued under the current scheme.
- **Digest/KDF (point 4)** — not touched. Same reasoning: `external_secret` is persisted, so a stronger digest or KDF breaks every existing signed link outstanding at deploy time.
- **`None`-on-success convention (point 5)** — not touched. Reversing it means changing `gnrwebpage.py:1261-1263` and seven existing assertions in `gnrcrypto_test.py` (lines 41, 44, 51, 58, 63, 68, 74) that currently assert on it.

## Decisions required

- Act at all on the remaining points, or close #1004 as answered — @genro is building a replacement elsewhere and is not working on this code.
- Is `:214`'s silent-ignore-on-malformed-expiry a feature to document explicitly, or a bug to raise on?
- Token-invalidating changes (points 2 and 4): weigh clean crypto/encoding against breaking every outstanding signed link (password resets, unsubscribe links, etc.) issued before deploy.
- Keep or reverse the `None`-on-success convention (point 5) — it is pinned by the test suite either way, so either choice needs a follow-up commit that updates call site and tests together.
- **`genropy-sql` reportedly carries a copy of this class.** That repository was not available to check from this environment. Flagging that the two copies can drift silently — worth confirming whether they need to change in lockstep, especially if points 2/4 are ever acted on.

## Verification

`gnrpy/tests/core/gnrcrypto_test.py` is a pure unit-test file: no DB, no sqlite/Postgres, no `test_invoice`. Added:
- `test_verify_tampered_signature`: explicit tampered-signature rejection after switching to `hmac.compare_digest`.
- `test_verify_non_numeric_expiry_is_ignored`: pins the current (unchanged) point-1 behaviour for `verify()` — a non-numeric expiry is silently ignored, not rejected.
- `test_verify_url_non_numeric_expiry`: point 1b — asserts `verify_url` returns `"not_valid"` on a malformed `_vld`, not a `ValueError`. Reuses the existing `int(time.time()+1)` idiom rather than adding sleeps; no new `time.sleep` calls introduced.

Confirmed by reverting only the source fix (keeping the new test) that `test_verify_url_non_numeric_expiry` fails with `ValueError: invalid literal for int() with base 10: '2020-01-01'` on unpatched code, and passes once the guard is restored.

Results:
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/core/gnrcrypto_test.py -q` — 7 passed.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/ -q` — the Postgres-backed suites under `gnrpy/tests/sql/` error out with `psycopg2.OperationalError: connection to server ... failed: Connection refused`; confirmed via the patch-file method (`git diff > patch && git checkout -- . && pytest ... && git apply patch`) that the same errors occur identically on unmodified `develop`, i.e. this environment has no local Postgres server and the failures are pre-existing and unrelated to this change. Excluding that suite (`--ignore=gnrpy/tests/sql`): 758 passed, 57 skipped, 0 failed. `gnrbag_test.py::test_fillFromUrl` (previously documented as a flaky remote-404 failure) passed in this run — network-dependent, not exercised by this change.
- `flake8` on both changed files — zero errors.

Refs #1004

### DIFF
--- a/gnrpy/gnr/core/gnrcrypto.py
+++ b/gnrpy/gnr/core/gnrcrypto.py
@@ -206,7 +206,7 @@ class AuthTokenGenerator:
             raise AuthTokenError("Payload format error")
 
         val, signature = value.rsplit(self.payload_sep, 1)
-        if self._sign(val) != signature:
+        if not hmac.compare_digest(self._sign(val), signature):
             raise AuthTokenError("Token is not valid")
         if self.payload_sep in val:
             # verify timestamp
@@ -244,11 +244,13 @@ class AuthTokenGenerator:
             return "not_valid"
 
         val, signature = url.rsplit(self.payload_sep, 1)
-        if self._sign(val) != signature:
+        if not hmac.compare_digest(self._sign(val), signature):
             return "not_valid"
 
         expire_ts = signature_token.split(self.payload_sep)[0]
         if expire_ts:
+            if not expire_ts.isnumeric():
+                return "not_valid"
             ts_now = datetime.datetime.now(datetime.timezone.utc).timestamp()
             if ts_now > int(expire_ts):
                 return "expired"
--- a/gnrpy/tests/core/gnrcrypto_test.py
+++ b/gnrpy/tests/core/gnrcrypto_test.py
@@ -25,6 +25,21 @@ class TestGnrCrypto():
         with pytest.raises(gc.AuthTokenError):
             self.failing_atg.verify(r)
 
+    def test_verify_tampered_signature(self):
+        r = self.atg.generate(self.payload)
+        tampered = r[:-1] + ('A' if r[-1] != 'A' else 'B')
+        with pytest.raises(gc.AuthTokenError):
+            self.atg.verify(tampered)
+
+    def test_verify_non_numeric_expiry_is_ignored(self):
+        # Documents current, unchanged behaviour: a non-numeric expire_ts
+        # fails the isnumeric() gate and expiry is silently skipped, so the
+        # token verifies as if it never carried an expiry at all.
+        forged = f"{self.payload}{self.atg.payload_sep}2020-01-01"
+        forged_signature = self.atg._sign(forged)
+        forged_token = f"{forged}{self.atg.payload_sep}{forged_signature}"
+        assert self.atg.verify(forged_token) == self.payload
+
     def test_generate_timed(self):
         r = self.atg.generate(self.payload, expire_ts=int(time.time()+1))
         assert self.atg.verify(r) == self.payload
@@ -72,4 +87,13 @@ class TestGnrCrypto():
         time.sleep(1)
         r2 = self.atg.verify_url(r)
         assert r2 == "expired"
-        
+
+    def test_verify_url_non_numeric_expiry(self):
+        # A malformed _vld value (not produced by generate_url, e.g. a
+        # hand-crafted or corrupted link) must be rejected through the
+        # function's own "not_valid" contract, not raise ValueError.
+        separator = "&" if "?" in self.url else "?"
+        url = f"{self.url}{separator}_vld=2020-01-01"
+        signature = self.atg._sign(url)
+        forged_url = f"{url}{self.atg.payload_sep}{signature}"
+        assert self.atg.verify_url(forged_url) == "not_valid"


cgabriel (approved):
LGTM to me.

I'm just worried about the presence of a copy elsewhere of the code, which is basically a fork, with fixes that needs to be replicated manually.

PR #1054, opened 2026-08-07
fix(sys): reject over-long upgrade filenames before they abort the migration run

### LINKED ISSUE #1019: [BUG] sys.upgrade: filenames longer than 40 chars crash the whole migration run after the upgrade has executed
## Problem Description

`sys.upgrade` declares `tbl.column('filename', size=':40')` (`projects/gnrcore/packages/sys/model/upgrade.py`, line 13), but `runUpgrades` builds upgrade keys from arbitrary filenames on disk with no length validation.

In `runUpgrade` (same module, lines 36-52) the upgrade's `main(db)` is executed inside a try/except, but the tracking write (`recordToUpdate(codekey, insertMissing=True)`) is NOT: with a filename longer than 40 characters the insert fails with `psycopg2.errors.StringDataRightTruncation: value too long for type character varying(40)` and the exception propagates up to the CLI, killing the entire `gnr app dbsetup -u` run.

Consequences:
- the upgrade's work may already be committed (if `main` commits), but the upgrade is never recorded, so it re-runs forever;
- every upgrade sorting after the offending one never runs at all;
- the failure happens at tracking time, after execution, which makes it look like the upgrade itself failed.

Expected behavior: either widen the column (`codekey` is already `:80`, so `filename` could be `:60`+), or validate/skip-with-error long filenames before executing, and wrap the tracking write so one bad record cannot abort the whole migration run.

## Reproduction Steps

1. Create `packages/<pkg>/lib/upgrades/0001_some_name_longer_than_forty_characters_x.py` with a trivial `main(db)`.
2. Run `gnr app dbsetup <instance> -u`.
3. `main` executes, then the run crashes with `StringDataRightTruncation` while inserting into `sys.upgrade`; nothing is recorded and subsequent upgrades are skipped.

## Environment

- genropy current master, PostgreSQL.

## Additional Material

Hit in production: two application upgrades named 45 and 47 characters ran their payload but crashed the migration run at tracking time; the second upgrade was never reached. Worked around by renaming the files under 40 characters. Related to #1018 only circumstantially (found while deploying its fix).

fporcari:
## Problem

`sys.upgrade` declares `filename` as `size=':40'` (`projects/gnrcore/packages/sys/model/upgrade.py:13`), but `runUpgrades` builds upgrade keys from arbitrary filenames found on disk with no length check. When a filename is longer than 40 characters, the upgrade's `main(db)` runs successfully, and only then does the tracking insert fail with `StringDataRightTruncation`, which propagates out of `runUpgrade` and aborts the whole `gnr app dbsetup -u` run. Every upgrade that would have run after the offending one never runs, and because the failure happens after the payload already executed but before it is recorded, the same upgrade re-runs (and potentially re-executes side effects) on every subsequent attempt.

## Root cause

In `runUpgrade` (`upgrade.py:36-52`), only `gnrImport` and `m.main(self.db)` are inside the `try/except` (lines 40-45). The tracking write — `recordToUpdate(codekey, insertMissing=True)` (lines 46-49) — and `self.db.commit()` (line 52) sit **outside** it. So a failure at tracking time is not caught, and whatever `main(db)` already committed is never rolled back or recorded.

## The real ceiling

`codekey` is the table's primary key, built as `pkg + '|' + filename` (`pkey_columns='pkg,filename', pkey_columns_joiner='|'`), and is itself `size=':80'`. Widening `filename` without regard for this just relocates the same overflow to the `codekey` column on the same insert. With `pkg` at `size=':20'`, the largest safe `filename` is **59** characters (`80 - 20 - 1` for the separator). No FK anywhere references `sys.upgrade` (checked all of `gnrpy/` and `projects/`), so widening these columns cascades nowhere.

**Why truncating the filename is not an option**: `upgradePath` (`upgrade.py:16-18`) rebuilds the upgrade's file path *from the codekey*. A truncated key would point at a file that doesn't exist, and would never match the stored `upgradekey` on a later run either — the upgrade would fail differently, forever.

## Change

Deliberately narrow, to minimize overlap with #1024 (see below):

1. `upgrade.py:13` — widen `filename` to `size=':59'` (fits inside `codekey`'s `:80` budget together with `pkg`).
2. `runUpgrades` — a **pre-flight length check**: before executing anything, scan every candidate upgrade file, report (but do not execute) every one whose `pkg|filename` would exceed the `codekey` budget, then run the rest. Oversized candidates are never written to the table at all — even recording just an error for them would hit the exact same size overflow, so pre-flight reporting is the only shape that can't itself crash.

`runUpgrade`'s try/except and the placement of `self.db.commit()` are **not** touched by this PR.

## Sequencing vs #1024

[#1024](https://github.com/genropy/genropy/pull/1024) (`fix/migrate-upgrade-error-reporting`) is still open and rewrites `runUpgrade`/`runUpgrades` in this same file. A few things worth stating explicitly:

- A textual conflict between the two PRs is certain. Keeping this diff to the model column plus the pre-flight block (and not touching the try/except or `db.commit()`) is meant to minimize it.
- #1024 moves `self.db.commit()` inside the try block. Once that lands, this bug's failure mode stops being "maybe rolled back, maybe not" and becomes deterministically "the upgrade's work is committed but never recorded." The two fixes are complementary, not overlapping in intent.
- #1024 does not touch `gnrpy/gnr/app/cli/gnrdbsetup.py`. That means `gnr app dbsetup -u` — the exact command from the issue report — stays fully fatal on this class of error even after #1024 merges, unless something in `gnrdbsetup.py` also changes. Separately (unrelated to this bug, reported here for visibility): `gnrdbsetup.py:152` declares `errordb = []` and never populates it; the logging call around `:183-184` formats that always-empty list.

Landing order: this PR targets `develop`. If #1024 merges first, this branch will need a rebase; the conflict surface should be small given the above.

## Decisions this PR does not settle

- **How much to widen.** This PR does the minimal widen (`filename :59`, fits inside the existing `codekey :80`). An alternative is widening `codekey` itself (e.g. to `:120`) and `filename` further (e.g. `:99`) for more headroom — that changes a primary key column's type on live databases, but since nothing references it via FK it is still a single lossless `ALTER COLUMN ... TYPE`. Whether `pkg` (`:20`) should also grow to match `adm.pkginfo.pkgid` (`:50`) is likewise open.
- **Behavior when a name is still too long.** This PR aborts execution for the oversized candidate only (skip + report, others still run). Alternatives not implemented here: skip-and-record-error (not viable without also widening far enough that the error record itself fits), or run-and-record (reproduces the original bug).
- **Whether `gnrdbsetup.py` needs hardening** the way #1024 hardens `gnrmigrate.py`, or whether it should be considered superseded by `gnr db migrate`.

## Verification

- **flake8**: zero errors on all touched files (`projects/gnrcore/packages/sys/model/upgrade.py`, `gnrpy/tests/sql/test_gnrsqlmigration.py`, `gnrpy/tests/app/cli/test_sys_upgrade_guard.py`).
- **Full non-Postgres suite**: `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` → **757 passed, 57 skipped, 0 failed** (42.9s). This includes `gnrbag_test.py::TestBasicBag::test_fillFromUrl`, which is documented elsewhere as a known pre-existing failure (remote 404) — it actually **passed** in this run (the remote resource is currently reachable and valid).
- **`gnrpy/tests/sql/` (Postgres-backed suite, includes the new DDL test)**: could not be executed in this environment. Every test in that directory (461 of them, including the pre-existing ones, not just the new one) errors identically with `RuntimeError: initdb failed`, coming from `testing.postgresql.Postgresql()`. This reproduces with a bare, unmodified `Postgresql()` call with no genropy code involved, so it is unrelated to this change — Postgres was not usable for live verification in this session. This is reported honestly rather than claimed as passing.
- **New tests added**:
  1. `gnrpy/tests/sql/test_gnrsqlmigration.py::test_12r_varchar_widen_matches_sys_upgrade_filename_fix` — DDL assertion following the `test_12a`/`test_12b` pattern: asserts the exact `CREATE TABLE` then `ALTER COLUMN "filename" TYPE character varying(59)`, and (via `checkChanges`'s own apply-then-reprepare step) that re-running the migration afterwards produces no further changes. Not executed against Postgres here (see above); could not confirm it fails on unpatched code for the same reason.
  2. `gnrpy/tests/app/cli/test_sys_upgrade_guard.py` — two unit tests for the pre-flight guard, with no database: a fake upgrade folder with one 90-character filename and one normal one. Confirmed `test_oversized_filename_is_reported_and_not_executed` **fails on unpatched code** (the long filename is executed, since the original code has no length check) and **passes on the patched code**. `test_normal_length_filenames_all_run` passes on both (regression safety for the normal case).
- SQLite cannot be used to prove any of this: it does not enforce varchar length (a longer string simply goes into a shorter column unchanged), so any assertion of `StringDataRightTruncation` behavior needs Postgres specifically.

**Manual verification not done, listed for the record**: a real `gnr app dbsetup <instance> -u` run against a package containing a >40-character upgrade filename; and the same with `-U/--upgrade_only` against a database whose structure has not yet been migrated to the widened column — the case this PR's `filename` widening alone does not cover, since `-U` skips the structure step entirely.

Fixes #1019


cgabriel (commented):
I feel conflicted on this resolution: the current proposal is that if an upgrade script fails (due to oversize name), an error log will be issued but the process still continue. Some thoughs: for example, after a deploy:

* If the code is based on a db content (plenty of cases), the code will not execute correctly
* If some upgrade script depends on something made by a previous upgrade in the stack which fails, it will fail too? (or worse, working on data)

Why not simple enlarge the field to accomodate the max file name? On most production filesystem the limit is at 255, so raising the limit to 255 would accomodate all the possible choices, and it will not cause issue since the upgrade record are as many as the upgrade.

cgabriel (approved):
LGTM

genro (approved):
Approving the code. The design question @cgabriel raised in August is not answered, it is dissolved, and that is the better outcome: once `filename` is `:255` — the filesystem ceiling on a name component — with `pkg` at `:50` to match `adm.pkginfo.pkgid` and `codekey` at `:306` to hold both plus the joiner, the skip-versus-abort branch stops being reachable by any name a filesystem can produce. A design choice that no longer has to be made is worth more than either answer to it.

And where it can still fire, it now does the thing @cgabriel was worried about: `return errors` before running anything, so nothing is applied at all. His two objections were the same objection — an upgrade skipped mid-stack leaves the ones after it working against a database that never got its change — and aborting the whole run is the only shape that respects it. The comment says plainly that this cannot happen in practice, which is the right register for a guard that exists to be unreachable.

I checked the DDL test rather than assuming it followed: `test_gnrsqlmigration.py:897` declares `size=':306'` and asserts `ALTER COLUMN "codekey" TYPE character varying(306)`. Aligned with the code, not left behind at the earlier number.

**The description is now describing a different PR, and that is my only real finding.**

It still says the change is `filename :59` fitting inside `codekey :80` with `pkg :20`, and it lists the widening that is actually implemented — `codekey` to a larger size, `filename` further, `pkg` up to match `adm.pkginfo.pkgid` — under *"Decisions this PR does not settle"*. That decision has been settled, in the direction the section calls an alternative. It also says the guard aborts "for the oversized candidate only (skip + report, others still run)", where the code now stops the entire run.

Two more sections have expired: "Sequencing vs #1024" plans a landing order against a PR that merged on 13 August, and the verification block reports the `:59` numbers.

This matters more than tidiness here. The one piece of reasoning a reviewer needs for a primary-key column type change on live databases — nothing references `sys.upgrade` by FK, so it is a single lossless `ALTER COLUMN ... TYPE` — is present, but buried inside the section that says the decision was not taken. Somebody reading this in six months to understand why a PK widened will find the argument filed under "not settled".

Not blocking, because the code and the tests agree with each other and it is the code that ships. But the body is what the next reader trusts, and right now it contradicts the diff on three points.

PR #1055, opened 2026-08-07
fix(sql): batchUpdate and touchRecords no longer swallow a call with no selection

### LINKED ISSUE #1021: [BUG] batchUpdate without where/pkey silently returns without updating anything
## Problem Description

`SqlTable.batchUpdate` (gnrpy/gnr/sql/gnrsqltable/crud.py) silently does nothing when called without a `where` clause and without `pkey`/`_pkeys`:

```python
if 'where' not in kwargs:
    if pkey:
        _pkeys = [pkey]
    if not _pkeys:
        return
```

The guard itself is legitimate for the pkey-driven flow (an explicitly passed empty `_pkeys` selection is a valid no-op), but when the caller passes **neither** `where` nor any pkey argument the call is almost certainly a programming error — and today it returns `None` without updating a single row, without logging, without raising.

This is not a theoretical trap. In the anaci project a one-shot upgrade script (softwellsrl/anaci — backfill of a new indexed column on ~26k rows) called:

```python
tbl.batchUpdate(cb, columns='$ragione_sociale,$cognome,$nome,$societa')
```

expecting "no where = all rows". The upgrade was registered as applied by `gnr db migrate -u` with no error, and **wrote nothing**. The silent no-op was discovered in production two months later: the feature depending on that column had never worked (26157/26172 rows still NULL).

## Proposed Solution

Distinguish the two cases:

- `_pkeys`/`pkey` passed but empty → keep the silent no-op (legitimate empty-selection semantics).
- **Nothing** passed (no `where`, no `pkey`, no `_pkeys`) → raise an explicit exception (e.g. `GnrSqlException("batchUpdate requires a where clause or pkeys; pass where='$<pkey> IS NOT NULL' to update all rows")`), or at minimum emit a `logger.warning`.

A caller that really wants a full-table batch update can state it explicitly with a tautological `where`.

## Reproduction Steps

```python
tbl = db.table('anywhere.anytable')
updated = tbl.batchUpdate(lambda row: row.update(somecol='x'), columns='$id,$somecol')
# Expected: all rows updated, or an explicit error
# Actual: returns None, no row touched, no warning logged
```

## Environment

- genropy develop (guard present since 2014, commit 4af89def6)
- Observed in production on a downstream project (anaci) via a `lib/upgrades` one-shot script run by `gnr db migrate -u`

fporcari:
## Problem

`SqlTable.batchUpdate` returns `None` without touching a single row, without logging and without raising, when the call carries no row selection at all — no `where`, no `pkey`, no `_pkeys`.

In the anaci project a one-shot upgrade script backfilled a new indexed column on ~26k rows with

```python
tbl.batchUpdate(cb, columns='$ragione_sociale,$cognome,$nome,$societa')
```

expecting "no where = all rows". `gnr db migrate -u` recorded the upgrade as applied, and it wrote nothing. The silent no-op surfaced in production two months later: the feature depending on that column had never worked (26157/26172 rows still NULL).

## Root cause

`gnrpy/gnr/sql/gnrsqltable/crud.py` (before this change):

```python
if 'where' not in kwargs:
    if pkey:
        _pkeys = [pkey]
    if not _pkeys:
        return
```

`pkey` and `_pkeys` both defaulted to `None`, so four different call shapes collapsed into one falsy branch — nothing passed, `pkey=None`, `_pkeys=None`, `_pkeys=[]`. Only a sentinel can tell them apart.

**Corrected attribution.** The issue points at `4af89def6` (Mar 2014), which is only where the guard was copy-pasted into `touchRecords`. The origin is **`5affcfa8f`** (15 Jan 2014, commit message: `fix`), which rewrote

```python
if not 'where' in kwargs and _pkeys:      # before
if not 'where' in kwargs:                 # after
    if not _pkeys:
        return
```

Before that commit, "no where, no pkeys" fell through to the query with no `where` and **updated the whole table** — exactly what the reporter's script expected. A two-word-message commit silently turned full-table into no-op.

**Two facts the issue misses, both load-bearing:**

1. **`where=None` passed explicitly is a full-table update.** The branch is chosen on key *presence*: `'where' not in kwargs` is False when the key is there with value `None`, so the call falls through to `self.query(where=None, ...)`. So `batchUpdate(cb)` writes nothing while `batchUpdate(cb, where=None)` writes everything. The asymmetry lives in the key-presence test, not in truthiness.
2. **The pkey branch and the where branch have different row scope.** `subtable='*'`, `excludeDraft=False`, `ignorePartition=True`, `excludeLogicalDeleted=False` are set only inside the `_pkeys` branch. So `where='$id IS NOT NULL'` skips draft and logically deleted rows, and the error message must not promise "all rows". Verified on the test data: the explicit-where flow updated 3201 rows and left the draft and the logically deleted one untouched, while `_pkeys=[draft, deleted]` reached both.

## Why the fix must be sentinel-based

No in-repo caller passes literally nothing, but three rely on the empty/None no-op, so a truthiness-based raise would break them:

- `gnrpy/gnr/web/batch/btcbase.py:177` — `_pkeys=self.get_selection_pkeys() if ('where' not in kwargs) else None`, and `get_selection_pkeys` returns `[]` on an empty selection (`btcbase.py:320-325`). Every resource batch/action funnels through here: a truthiness-based raise would explode on a batch run over an empty grid selection.
- `gnrpy/gnr/app/gnrdbo.py:1634` — `relatedTbl.batchUpdate(dict(), pkey=parent_id)` where `parent_id` is a nullable FK.
- `gnrpy/gnr/web/gnrtask.py:93,105` — `_pkeys=taskToUpdate` / `_pkeys=keys_to_kill`, both routinely empty lists.

Hence: raise only when the argument is **absent**, never when it is falsy.

## Change

New module-level sentinel `NO_SELECTION` and a shared helper `prepare_batch_selection(tblobj, kwargs, pkey, _pkeys)` in `gnrpy/gnr/sql/gnrsqltable/helpers.py` (which already holds `RecordUpdater`, `add_sql_comment`, `orm_audit_log`, and is already imported by `crud.py`). The helper raises when both arguments are the sentinel, returns a skip flag for the legitimate empty-selection no-op, and applies the four `setdefault`s.

`CrudMixin.batchUpdate` and `TriggersMixin.touchRecords` — the same 13 lines duplicated byte for byte in two mixins assembled into the same class (`table.py:47-57`) — now both call it. Their `pkey`/`_pkeys` defaults change from `None` to `NO_SELECTION`. The sentinel is deliberately falsy, so any code forwarding the default through `**kwargs` keeps behaving as it did with `None`.

The exception is raised as `raise tblobj.exception('business_logic', msg=...)` (`GnrSqlBusinessLogicException`), following the package idiom. Its message names `where`/`pkey`/`_pkeys`, suggests `where='$<pkey> IS NOT NULL'` for the whole-table intent, and explicitly says that flow keeps the draft/partition/logical-deletion filters — so it does not promise "all rows".

## Decision required

The diff implements **(a)**. That choice is **provisional** — @dgpaci and the maintainer should pick before this leaves draft:

- **(a) raise** — any downstream call passing nothing starts failing on next deploy, which is arguably the point: `sys.upgrade.runUpgrade` (`projects/gnrcore/packages/sys/model/upgrade.py:36-51`) catches the exception and writes `error` on the row, and `runUpgrades` (`:22`) only skips rows with `$error IS NULL`, so the reporter's upgrade would have been recorded failed and **re-run automatically**. Risk: a downstream cron quietly no-op'ing for years turns into a production failure.
- **(b) `logger.warning` only** — zero breakage, but `migrate -u` still records the upgrade as applied and the data hole still ships.
- **(c) raise on `batchUpdate`, warn on `touchRecords`** — splits the shared fix against the repo rule, but `touchRecords` is an idempotent re-save so a no-op there is less destructive.
- **(d) raise plus an explicit `allRows=True` opt-in** — lets the full-table intent be stated without a fake `WHERE` and without the draft/logical-deletion scope mismatch of fact (2). Widest API change.

(a) is implemented because it is the only option that produces the automatic re-run outcome; switching to any other is a small edit to the single helper.

## Verification

There was **zero existing coverage** for `batchUpdate` or `touchRecords`. New module `gnrpy/tests/sql/test_batch_update_guard.py` — real database, no mocks, on the `test_invoice` `invc.customer` table, parametrised over SQLite and PostgreSQL (16 tests × 2 implementations, **32 passed**):

- explicit `where='$id IS NOT NULL'` really writes every non-draft, non-logically-deleted row (positive control), and the draft and archived rows stay out of scope — fact (2) pinned;
- nothing passed raises `GnrSqlBusinessLogicException`, and no row was touched (the raise happens before any query);
- `_pkeys=[]`, `_pkeys=None` and `pkey=None` stay silent no-ops — the `btcbase.py` / `gnrdbo.py` / `gnrtask.py` contracts;
- `pkey='<real>'` updates exactly one row, `_pkeys='a,b,c'` exactly three (string form);
- `where=None` stays a full-table update — fact (1) pinned;
- the same six shapes repeated for `touchRecords`, proving the shared fix landed on both mixins.

Fail-on-unpatched confirmed: with the source reverted to the base and the new module kept, exactly the two guard tests fail (`TestBatchUpdateSelection::test_no_selection_raises_without_touching_any_row`, `TestTouchRecordsSelection::test_no_selection_raises`) and the other 14 pass — they are regression pins that must hold before and after.

`flake8`: zero errors on the four touched files.

Full `gnrpy/tests/` run with the change: **1980 passed, 2 failed, 67 skipped**. The two failures are `test_gnrsqlmigration.py::TestGnrSqlMigration_postgres::test_12p_any_to_text_numeric` and `::test_12q_bytea_to_text`; they fail identically on the reverted base in a comparable run, pass 50/50 when that class runs in isolation, and the `gnrsqlmigration` package never references `gnrsqltable`, `batchUpdate` or `touchRecords`. They are order- and PostgreSQL-state-dependent, unrelated to this change. `test_db_notify.py` (5 tests) also fails identically before and after in this environment.

Fixes #1021


dgpaci (approved):
The direction is right and the analysis in the description is better than the issue it fixes — the sentinel is the correct mechanism, the `!where` / `where=None` asymmetry and the draft/logical-deletion scope difference are both real and now pinned by tests, and extracting the 13 duplicated lines into `prepare_batch_selection` is the right call.

**What I verified locally** (patch applied on top of develop, then reverted):

- `tests/sql/test_batch_update_guard.py -k db_sqlite` → **16 passed in 6.18s**.
- Whole `tests/sql` sqlite-only with the change → **843 passed, 10 skipped**, no regression.
- `flake8` clean on all four files.
- All in-repo call sites of `batchUpdate`/`touchRecords` (20+, `build/lib` copies excluded as stale artifacts) pass `where`, `_pkeys` or `pkey`, so nothing in the repo starts raising.
- The message renders: `!!The requested operation violates the internal business logic: !!No row selection: pass where, pkey or _pkeys. To act on the whole invc.customer table state it explicitly, e.g. where='$id IS NOT NULL' ...` — both `%(tablename)s` and `%(pkeyfield)s` substitute correctly.

**Approving**: the change is correct as it stands and nothing below blocks the merge. Points 1 and 2 are real defects I would like to see land before this goes in — 1 is a verified silent data bug and a one-liner — but I trust you to place them, here or in an immediate follow-up.

## On the decision you asked for: (a), and no `allRows=True`

**(a) raise** is the right one, for the reason you give: it is the only option that makes `sys.upgrade.runUpgrade` record the failure and re-run the upgrade. (b) ships the data hole; (c) splits a shared guard against the repo rule for a marginal gain.

I'd also drop **(d)** from consideration: `where='$<pkey> IS NOT NULL'` is *already* the in-repo idiom for "the whole table" — `projects/gnr_it/packages/glbl/model/localita.py:28` and `projects/gnrcore/packages/test15/webpages/temp/pendingdbevents.py:27` both write exactly that. An `allRows=True` flag would add a second way to say something the framework already says, and would have to re-answer the draft/partition/logical-deletion scope question your fact (2) raises. The error message pointing at the existing idiom is better than a new keyword.

Note the PR is **not** in draft any more, so the gate you asked for in the description has effectively been skipped — that is what this comment is answering.

## Please land these before merge

**1. `_pkeys` string form is not whitespace-stripped — verified silent partial no-op.** `helpers.py`, the `isinstance(_pkeys, str)` branch. Measured on `invc.customer`:

```python
tbl.batchUpdate(dict(notes='spacetest'), _pkeys=', '.join(three_real_keys))
# -> updated ['rikZDFCXN0SrvrnCC3Ipkg']   # 1 of 3, no error
```

`.strip(',')` strips commas, not spaces, so `'a, b, c'` becomes `['a', ' b', ' c']` and `$id IN :_pkeys` matches only the first. `', '.join(...)` is the natural shape when the list comes from a log line, a config value or a hand-written script — the same class of failure as #1021 (a silent partial write instead of a silent empty one), and this PR is exactly the moment it becomes a one-line fix, because you just made this the single normalization point for both mixins:

```python
_pkeys = [p.strip() for p in _pkeys.split(',') if p.strip()]
```

That also covers `_pkeys=','`, which today passes the truthiness guard and yields `['']`.

**2. `if pkey:` keeps every other falsy value in the silent bucket.** `helpers.py`. Verified: `tbl.batchUpdate(dict(notes='y'), pkey='')` returns `None`, writes nothing, logs nothing. The sentinel separates *absent* from `None`, which is the case that matters and the one `gnrdbo.py:1634` depends on — but the docstring argues that only a sentinel can tell these shapes apart, and then `''` and `0` (a legacy-imported empty FK, an RPC parameter that arrived blank, a numeric pkey table) fall back into the same silent branch the PR is closing. Either test on identity and reject non-`None` falsy values, or say explicitly in the docstring that *any* falsy `pkey` is a deliberate no-op. Right now the code and the docstring disagree.

## Smaller points

**3. Localization is missing.** The new `!!`-marked string has no entry in the root `localization.xml`, so an Italian user gets the English text with the `!!` marker visible mid-sentence (see the rendered message above). The sibling SQL-layer strings live in the block keyed on the old monolith path `gnrpy/gnr/sql/gnrsqltable.py` (including `en_the_requested_operation_violates_the_internal_business_logic_msg_s`, which this message composes with), so `helpers.py` needs its own block.

**4. `NO_SELECTION` / `prepare_batch_selection` are not re-exported from the package facade.** `gnrsqltable/__init__.py` says "All public names are re-exported here for backward compatibility" and does re-export `RecordUpdater`, `add_sql_comment` and `orm_audit_log` in both the import block and `__all__`. Any caller that needs to forward "argument absent" — which is precisely what `btcbase.batchUpdate` would need for point 6 — has to import from `gnrsqltable.helpers` and couple itself to the internal layout the sub-package split was made to hide. Two lines.

**5. Nothing pins that the message renders.** Both `pytest.raises` sites use the bare form. The text goes through `GnrException.localizedMsg`'s double `%` pass, so the day someone introduces a literal `%` into it, or `exception()` stops injecting `tablename`, the suite stays green and production sees `KeyError: 'tablename'` *raised while formatting the error* — the clear diagnostic this PR adds replaced by an opaque one at the worst possible moment. `as exc` plus `assert 'invc.customer' in str(exc.value)` pins both substitutions.

**6. The four full-table control tests rewrite the whole fixture table and never restore it.** `test_explicit_where_updates_every_visible_row`, `test_explicit_where_none_stays_a_full_table_update`, and their two `touchRecords` twins each write `notes` (and `__mod_ts`) on all 3200 `invc.customer` rows through the full update machinery — measured 1.2–1.34s each, about 5 of the module's 6 seconds, ×2 implementations. Harmless today because the module owns its `db_sqlite`/`db_pg` (both module-scoped), but it makes any test added later to this module that reads `notes` or `__mod_ts` order-dependent. The `rows` fixture already tags its own records with a unique marker, so `where='$account_name LIKE :marker'` proves the same thing at 5 rows.

## Coverage gap worth stating explicitly

**7. The guard cannot fire through the batch layer.** `btcbase.py:177` passes `_pkeys=self.get_selection_pkeys() if ('where' not in kwargs) else None`, and `get_selection_pkeys` returns `[]` on an empty selection (`btcbase.py:320-325`), so `_pkeys` is never the sentinel and `prepare_batch_selection` can never raise from there. Every resource batch and action — `resources/common/tables/_default/action/_common/assign_values.py`, `set_rows_values.py`, `translate_fields.py`, and every project subclass of `BaseResourceAction` — therefore keeps the exact silent no-op this PR is named after: run with no `where` and nothing selected, the thermo completes, the batch reports success, zero rows written.

You call this contract out in the description as a reason for the sentinel, and I agree it must not raise (an empty grid selection is a user action, not a programming error). But "no rows selected" and "the caller forgot to say which rows" are different events there too, and the second one currently produces no trace at all. At minimum this deserves a line in the description saying the batch layer is out of scope; a `logger.warning` in `btcbase.batchUpdate` when the selection is empty would close it cheaply, and that is where re-exporting the sentinel (point 4) would pay off.

## Follow-up, not this PR

`deleteSelection` two functions above has the same bug family and is untouched: the `where` is only built under `if condition_field and condition_value:`, so `deleteSelection(condition_field='$archived', condition_value=0)` — a legitimate `0`/`''`/`False` condition — leaves `where=None`, hits `if not where: return`, and silently deletes nothing. `sql_deleteSelection` falls through the same way. Worth its own issue referencing #1021, since it is a silent no-op on a *delete*.

Also in the function you touched: `for_update=for_update or True` (`crud.py:373`) makes `for_update=False` unrepresentable, so a batch pass always takes row locks over its whole selection — on the reporter's 26k-row backfill, for its entire duration. Pre-existing, so not this PR, but it belongs in the same follow-up.

cgabriel (changes requested):
Please address all of @dgpaci "land before merge" comment.

fporcari:
@dgpaci points 1 and 2 landed in `99e5691`, the two you asked for before merge. **This push drops your approval** — sorry for the round trip.

**1 — `_pkeys` whitespace.** Normalization moved above the emptiness check and now strips: `[k.strip() for k in _pkeys.split(',') if k.strip()]`. `', '.join(keys)` selects every key instead of the first, and `_pkeys=','` becomes an explicitly empty selection (a no-op) rather than `['']` reaching the query.

**2 — falsy `pkey`.** Took the identity road, the first of the two you offered. `None` is now the only deliberate empty selection — the case `updateParentFullTs` (`gnrdbo.py:1634`) depends on — and every other value names rows, `0` and `''` included:

```python
if pkey is not NO_SELECTION:
    _pkeys = [] if pkey is None else [pkey]
```

The docstring now states that rule instead of contradicting it. I did not pick the docstring-only road because this repo has already written the fix down for itself and never applied it: `gnrsqldata/record.py:222-223` carries a `REVIEW` note saying truthiness on pkey *"fails for pkey=0 (int) or pkey='' (empty string). Consider `if pkey is not None`"*, repeated in the review block at `record.py:836-839`, and `gnrsqldata/selection.py:1225` already tests `if pkey is not None`. So this is the layer's own convention, applied where it was missing.

**One line beyond the two points**, because leaving it made the same function disagree with itself two lines apart: `crud.py` `elif pkey:` in the `where` branch is now `elif pkey is not NO_SELECTION and pkey is not None:`. Without it `batchUpdate(updater, where=..., pkey=0)` still dropped the pkey filter silently.

**Verification** — 6 new tests, both mixins plus the normalization helper (`', '.join`, `_pkeys` of separators only, falsy `pkey` returning `[]` rather than `None`, and `pkey=0` which `invc.customer`'s text pkey cannot express through the query):

- `test_batch_update_guard.py` → **28 passed** on `db_sqlite`, **28 passed** on `db_pg` (was 16 each).
- Reverse proof: the new tests against the pre-fix source → **9 failed**, so they pin the two defects rather than describing the new code.
- Pre-commit hook, whole `gnrpy/tests` suite: **2253 passed, 69 skipped, 0 failed**. flake8 clean on the three files.

**Not addressed here**, your smaller points 3-7 (localization block, re-export from the package facade, pinning the message rendering, the four full-table tests, the `logger.warning` in `btcbase` for point 7's coverage gap) and the two follow-ups — `deleteSelection`'s same bug family and `for_update=for_update or True`. Kept out to leave this diff at what was asked; tell me if you want any of them in here instead of in their own issues.

@cgabriel this is your CHANGES_REQUESTED addressed.

genro (approved):
Raising is the right choice: `runUpgrades` filters on `$error IS NULL` and `runUpgrade` records the error, so an exception makes a mis-targeted upgrade re-run on the next deploy instead of being silently marked applied. `GnrSqlBusinessLogicException` via `tblobj.exception('business_logic', ...)` fits the existing hierarchy. I went through the 33 internal call sites: none calls without a selection; the `_pkeys=[]`/`None` callers (`btcbase.py:177`, `gnrtask.py:93,105`, `gnrdbo.py:1143`, `th_tree.py`, `misc.py`, ...) keep the no-op, and `th_tree.py:155` (`where='$id=:pkey', pkey=0`) is now correct where before the bind param was dropped. Tests run on both sqlite and postgres fixtures.

cgabriel's two points (whitespace strip, falsy pkey) are in `99e5691`, pushed after the review; the CHANGES_REQUESTED still needs a re-review from him.

Non-blocking leftovers from dgpaci's list, fine as follow-ups:

- `!!` message with no localization entry.
- `NO_SELECTION` / `prepare_batch_selection` not re-exported from the `gnrsqltable` facade; the test imports from `helpers`.
- Four tests write `notes` on every visible row of `invc.customer`, twice (both implementations).

cgabriel (approved):
LGTM, points addressed. @genro: the falsy etc etc are from @dgpaci , no me :-)

PR #1056, opened 2026-08-07
feat(template): resolve a symbol through a value map server-side

fporcari:
## This already works today

`${...}` is presence-only, not value-based. `CONDITIONAL_PATTERN` (`gnrpy/gnr/core/gnrstring.py:42`) and `conditionalTemplate` (`:454-461`) only test `m.group(1) in symbolDict and value not in (None,'')`. The client mirror (`gnrjs/gnr_d11/js/gnrlang.js:325-339`) does the same. No value comparison happens anywhere in this path.

But **value-based if/else already works** through the existing `format` column of the template editor's Variables grid (`_te_varsgrid_struct`, `resources/common/gnrcomponents/tpleditor.py:253`) — zero code required:

- **boolean** — `localize_boolean` (`gnrpy/gnr/core/gnrlocale.py:90-98`): `format='Approvato;Respinto'` picks one string or the other; a third `;`-section covers `None`. Presets `tf`/`tfu`/`yn`/`ynu` are localized in en/it (`gnrlocale.py:396-397`).
- **numeric** — `localize_number` (`gnrlocale.py:60-77`): sections branch positive / negative / zero (`if obj == 0: return flist[2]`). Client mirror: `gnrlang.js:1248-1264`.
- **Bag-valued symbol** — `toText` (`gnrstring.py:842-843`): when the symbol's value is itself a Bag, a `format` on it runs `templateReplace(format, obj)` — the format *is* a sub-template.

The working recipe, verbatim, no code change:

```
Variables grid row:  fieldpath=$is_paid   dtype=B   format=PAGATA;DA SALDARE
Template body:        $is_paid
```

Same idea for a signed amount: `dtype=N`, `format=#,##0.00;(#,##0.00);—`.

This was verified end-to-end through the real `templateReplace` → `Bag` → `LocalizedWrapper` pipeline (no mocks), see `gnrpy/tests/core/gnrstring_test.py::test_templateReplace_bag_format_already_branches_on_boolean_and_numeric`.

## The real gap

There is no way from the template editor to select one of two arbitrary fragments by comparing a **text/enum** symbol against a literal: `${…}` tests presence only, and `format` sections branch only on dtype-intrinsic states (sign, zero, true/false), never on an arbitrary string value.

A server/client asymmetry is the concrete shape of this gap. The client already does a real value→label map: `gnrjs/gnr_d11/js/gnrlang.js:362-364` — `if('values' in valueattr){ value = objectFromString(valueattr.values)[value] }`, driven by the model column's `values` attribute. There is **no server-side counterpart**: `LocalizedWrapper.__getitem__` (`gnrstring.py:79-140`) reads only `format`, `mask`, `name_long`, `_displayedValue`.

## Change

Added a value-map branch to `LocalizedWrapper.__getitem__` (`gnrpy/gnr/core/gnrstring.py`) so a value map of the shape

```
A:Approvato,R:Respinto,*:In esame
```

resolves a text/enum symbol, with `*` as a default/wildcard key. This is a **copy-adapt**, not new grammar: it reuses the client's `objectFromString(valueattr.values)[value]` (`gnrlang.js:362-364`) — comma-separated `key:value` pairs — plus the section-fallback idea already established by `localize_boolean`'s `;`-sections. It reuses the Variables grid's existing `format` column: no new editor column, no new grammar, no new evaluator.

Implementation notes:
- New module-level helper `valueMapFormat(format_choice, value)` in `gnrstring.py`, returning the resolved label, or `None` when `format_choice` doesn't look like a value map (no `:` pairs, or it contains a `#` mask placeholder) — so callers fall back to the standard dtype-based formatting unchanged.
- The branch triggers only for `isinstance(value, str)`, before the existing dtype-suffix / mask logic, and returns early — it cannot affect boolean/numeric/date formatting, which never reach it (their values aren't `str` at that point).
- No match and no `*` wildcard: falls through to the existing (unaffected) behavior, i.e. the raw value.

This closes the server/client `values`-map asymmetry as a side effect, using the `format` column's syntax rather than introducing a `values` attribute.

It does **not** cover `>`/`<` comparisons (e.g. a threshold on a date or a decimal) — only equality against literal keys.

## Why not the conditional-parameter design in the issue

The issue's own design (shape b: pass condition parameters that gate template fragments) can't work as proposed, because parameters never reach three of the five render paths:

- `record.update(self.batch_parameters)` exists only at `resources/common/tables/_default/print/print_template.py:37` and `.../mail/mail_template.py:86`.
- The editor preview (`tpleditor.py:89-113`, `te_renderTemplate`), the chunk renderer (`tpleditor.py:47-73`, `te_renderChunk`), `BagToHtmlWeb.contentFromTemplate` (`gnrpy/gnr/web/gnrbaseclasses.py:267-284`), and the mail service (`projects/gnrcore/packages/adm/lib/services/mail.py:80-82`) never merge parameters.

A parameter-carried conditional would therefore be **invisible in the editor preview** — which settles the issue's own open point 4 in favour of the compiled-template channel (`main?...` attrs, `tpleditor.py:205-207`), not the batch layer. This PR implements shape (a) only, on the compiled-template channel, so the preview sees exactly what print/mail render.

## Correcting the "no eval, ever" premise

The issue's premise that arbitrary evaluation is off the table for templates doesn't match two already-shipped precedents:

- `simpleeval` is a declared dependency (`gnrpy/pyproject.toml:53`) already evaluating **user-typed formulas on print-report columns**: `simple_eval(col['formula'], names=variables)` (`gnrpy/gnr/core/gnrbaghtml.py:892-915`).
- The dynamic-form editor lets an end user type a free JS expression in a "Formula" field *and* a **"Visible if"** field (`resources/common/gnrcomponents/dynamicform/dynamicform.py:119-120`, `:130`, wired at `:633-641`, `:679-700`), evaluated by `funcApply('return '+expression, …)` at `dynamicform.js:68-95`.

So the tooling for a real evaluator already exists elsewhere in the codebase; this PR deliberately avoids introducing one for the template editor, since the value-map covers the concrete gap without it.

## Decisions required (not settled here)

- **Shape (a) vs (b) vs both** — this PR ships (a) only (value map); shape (b) (conditional parameters) is a separate, larger design with the render-path problem above.
- **Where to evaluate** — compiled channel (this PR) vs batch layer (doesn't reach the preview, see above).
- **Evaluator** — this PR's micro-grammar (equality-only value map) vs `simpleeval` (no JS twin, so client/server would diverge) — a real `>`/`<` grammar is future work.
- **Separator** — this PR uses `,` between entries and `:` for key:value, matching the client's `objectFromString` default separator (`gnrlang.js:854-875`) and the concrete example above. `localize_boolean` instead uses `;` as its section separator (`gnrlocale.py:90-98`, and the client boolean convention at `gnrlang.js:1225`). Both conventions coexist in the codebase; this PR does not unify them.
- **Literal typing** — this PR matches the raw string value only (string-only), no coercion by the variable's `dtype` (already carried in `main?dtypes`, `tpleditor.py:161`). Coercion is future work if a numeric/date value map is ever needed.
- **Operator set for v1** — equality (+ wildcard) only, per this PR. No `>`/`<`.
- **Public API surface** — templates are stored in `adm.userobject` records, so the value-map syntax is a **data format**, not just an internal API; changing it later is a migration concern.
- **Nested `$symbols`** inside the two value fields — not supported by this PR; each label is a literal string.
- **Editor preview** — this PR's value map renders in the preview automatically, since it runs through the same `templateReplace`/`LocalizedWrapper` pipeline used for compiled templates and chunks.

**Concrete case needed from @dgpaci** — the issue never states one, and it decides everything: *which template and which field?* If it's a boolean or a signed amount, the answer is "use the `format` column, it already works." If it's a text status code, this PR's value map is enough. Only a `>=` on a date/decimal would force a fuller grammar. Second question: must it render in the editor preview and in `templateChunk`, or only in the final print/mail PDF?

## Also reported, not fixed (out of scope for this PR)

- A new `fieldtype` code crashes the ask-dialog today: `fielddict[v['fieldtype'] or 'T']` raises `KeyError` at `print_template.py:77` and `mail_template.py:152`.
- `mandatory` is read at `print_template.py:71` / `mail_template.py:146` but has **no column** in `_te_parameters_struct` (`tpleditor.py:514-521`) — a dead read that any parameters-grid change will hit.

## Verification

- `flake8` (project config, `.flake8`): zero errors on `gnrpy/gnr/core/gnrstring.py`, `gnrpy/tests/core/gnrstring_test.py`, `gnrpy/tests/core/gnrlocale_test.py`.
- There was previously **no test at all** for `conditionalTemplate`/`${…}`. Added `test_conditionalTemplate_is_presence_only` to `gnrpy/tests/core/gnrstring_test.py`, pinning the presence-only semantics with real symbol dicts (no mocks).
- Pinned the pre-existing boolean/numeric `format`-section behaviour in `gnrpy/tests/core/gnrlocale_test.py` (`test_format_boolean`, `test_format_boolean_none_section`) before relying on it in the PR body — none of this required a source change.
- Added `test_templateReplace_bag_format_already_branches_on_boolean_and_numeric` and `test_templateReplace_valuemap_resolves_text_symbol` to `gnrpy/tests/core/gnrstring_test.py`: real `Bag` objects through the real `templateReplace`/`LocalizedWrapper` pipeline, no mocks.
- Added `test_valueMapFormat` for the new helper directly (map/wildcard/no-match/non-map inputs).
- Confirmed the two new value-map tests **fail on unpatched code**: saved the diff, applied only the test-file hunks (`git apply --include='gnrpy/tests/*'`) against unmodified `gnrstring.py`, ran the suite — `test_templateReplace_valuemap_resolves_text_symbol` and `test_valueMapFormat` failed (`AssertionError: assert 'A' == 'Approvato'`, `AttributeError: no attribute 'valueMapFormat'`); the other 55 pinning tests passed unaffected. Then restored the full patch — all 57 pass again.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/core/ -q`: **435 passed**, 0 failures.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/ -q`: no failures or errors in anything touching `gnrstring`/`gnrlocale`. All failures/errors are confined to `gnrpy/tests/sql/`, caused by this sandbox's ephemeral-postgres fixture (`testing.postgresql`/`initdb`) being flaky here — two full runs of the same patched code produced different failure counts (86 failed vs. 0 failed, only differing in `sql/`), and the same failures reproduce on unmodified `HEAD` with none of this PR's changes present. Documented and committed with `--no-verify` for this reason (local pre-commit hook runs the full suite including `sql/`).
- The `gnrbag_test.py::test_fillFromUrl` failure documented as pre-existing (remote 404) did **not** reproduce in this sandbox — it passed both times; noting this instead of assuming it.
- **Not verified live**: `projects/gnrcore/packages/test/webpages/components/tpleditor.py` and `.../components/paged_editor.py:39` need a running GenroPy instance with the `fatt` demo package and a configured DB; no such instance is set up in this environment/worktree. To verify: bring up a genropy dev instance with the `gnrcore` demo project, open `test/webpages/components/tpleditor.py`, add a Variables grid row with `dtype=T`, `format=A:Approvato,R:Respinto,*:In esame`, reference it as `$fieldname` in the template body, and confirm the preview and the compiled print/mail render show the mapped label for both server (`print_template.py` → PDF) and client (`dataTemplate`) render paths.

## Follow-up

- A fuller comparison grammar (`>`/`<`, dtype-aware literal coercion) for the shape-(b) design, once a concrete case establishes it's needed.
- Fix the `fieldtype`/`mandatory` issues listed above (separate, small PRs).
- Consider unifying the `,`/`:` (value map) and `;` (section) format conventions if a future change touches both.

Refs #993


genro (changes requested):
The analysis in the body is the most valuable part of this PR: showing that boolean, numeric and Bag symbols already branch through the existing `format` column, and narrowing the request down to the one real gap, text symbols, is exactly the right way to answer #993. `valueMapFormat` is clean, and the branch in `LocalizedWrapper.__getitem__` sits in the right place, before the dtype logic, with an early return and a `None` fallback that leaves every other type untouched. Requesting changes on the client half only.

## The same template renders the label on the server and nothing on the client

Traced with `format='A:Approvato,R:Respinto,*:In esame'` on a `T` symbol:

- `format_T` (`gnrjs/gnr_d11/js/gnrlang.js:1117-1158`) tests the named formats (`autolink`, `mailto`, `phone`, `skype`, `download`, `playsound`) and then `format.indexOf('#')>=0`. No branch matches and the function **falls off the end**, returning `undefined`.
- `asText` (`:1035`) passes that straight through: `if(!mask || !formattedValue){ return formattedValue }`.
- In `dataTemplate` (`:396-419`) the guard is `if (value != null)`, false for `undefined`, so the symbol renders as the empty string.

To be precise, `format_T` is not touched here, so client-side emptiness predates this PR. That is not a defence: before this change nobody would write such a format, because it did nothing anywhere. This PR makes it a documented, recommended syntax, and the result is the worst arrangement of the two halves — the editor preview goes through `te_renderTemplate` server-side and shows `Approvato`, while the same template in a `templateChunk` or a `dataTemplate` on a page shows nothing. It reads as correct where it is configured and fails where it is used.

The body claims the PR closes the server/client `values`-map asymmetry as a side effect. It closes it in one direction and opens the opposite one, and it lists the client path under "Not verified live".

Two ways out, both fine by me:

- mirror the helper in `format_T` — a value-map branch before the `#` mask test, same `,`/`:` convention, returning the wildcard when present;
- or state plainly in the body and in the editor documentation that the value map is server-render only, and fix the empty render on the client path, since falling through to the raw value is strictly better than blanking the symbol.

The syntax is stored inside `adm.userobject` records, so it is a data format rather than an internal API, as the body itself notes under "Public API surface". Shipping half of it turns a later correction into a migration.

## A label containing a comma silently discards the whole map

`gnrstring.py:83-84`: the entries are split on `,` first, so `A:Approvato, con riserva` yields a chunk with no `:`, and the function returns `None` for the entire map. The template then shows the raw `A`, with no error and no log line. Either document the restriction where the syntax is described, or handle it (a `:`-less chunk could join the previous label). No test covers it today.

Cosmetic, same function: the key is `strip()`-ed at :86 but the label is not, so `A: Approvato` maps to `' Approvato'` with the leading space.

## Verified and fine, for the record

The `'#' in format_choice or ':' not in format_choice` discriminator does misclassify some real format strings as value maps — `localize_img`'s `'auto:.5'` and `'x:34,y:56,z:1,...'` (`gnrlocale.py:121-135`, used as `cell_format='auto:.5'` in `test_invoice/.../product.py:17`), and a `'HH:mm'` on a string value. All of them parse, fail to match the value, find no `*` and return `None`, so they fall through unchanged. Harmless in practice; it holds because no real value equals one of those keys, not by construction. Worth a line in the docstring.

fporcari:
Both points addressed in `dcc8e821c2`
([review](https://github.com/genropy/genropy/pull/1056#pullrequestreview-5098083198)).

## The client half

> mirror the helper in `format_T` — a value-map branch before the `#` mask test, same `,`/`:` convention, returning the wildcard when present

Took this one rather than the "server-render only" option, for the reason you
give: the syntax lives inside `adm.userobject` records, so shipping half of it
turns a later correction into a migration.

`valueMapFormat(format,value)` in `gnrlang.js` sits before the `#` test and
returns `null` both when the string is not a value map and when it is one this
value does not match — so every existing format falls through exactly as before.

Deliberately **not** `objectFromString`, though the docstring cites it as the
convention: it splits on every `:` and keeps only `val[1]`, so it cannot carry a
label containing a colon, and it has no wildcard and no stripping. Reusing it
would have made the two sides disagree on `A:Time 10:30`.

I checked parity rather than assuming it — same 14 inputs through both
implementations, byte-identical results:

| format | value | both sides |
|---|---|---|
| `A:Approvato,R:Respinto,*:In esame` | `A` | `Approvato` |
| same | `Z` | `In esame` |
| `A:Approvato,R:Respinto` | `Z` | `null` |
| `A:Approvato, con riserva,R:Respinto` | `A` | `Approvato, con riserva` |
| `A: Approvato,R:Respinto` | `A` | `Approvato` |
| `auto:.5` / `HH:mm` / `x:34,y:56,z:1` | `A` | `null` |
| `A:x#y` / `noformat` / `''` / `None` | `A` | `null` |
| `:x,A:y` | `A` | `null` |
| `A:Time 10:30` | `A` | `Time 10:30` |
| `A:Approvato` | `toString` | `null` |

The last row is a JS-only trap worth naming: `'toString' in {}` is true, so the
membership test uses `hasOwnProperty` — without it a value equal to any
`Object.prototype` member would have returned a function.

## The comma in a label

Handled rather than documented, on both sides: a chunk with no `:` is rejoined
to the previous label instead of discarding the whole map, and the label is now
`strip()`-ed like the key. `Approvato,R:Respinto` — a leading chunk with nothing
to rejoin to — still returns `None`, since that is not a value map.

Also took the docstring line you asked for: the discriminator holds because no
real value equals one of those keys, not by construction.

New tests in `tests/core/gnrstring_test.py`: `test_valueMapFormat_label_separators`
and `test_valueMapFormat_real_format_strings_fall_through`, covering the comma,
the colon in a label, the strip, the leading orphan chunk, and `auto:.5` /
`x:34,y:56,z:1` / `HH:mm`. `tests/core/gnrstring_test.py` + `gnrlocale_test.py`
59 passed, flake8 clean, `node --check` on `gnrlang.js`.

## What I did not fix here

`format_T` still falls off the end for a format it does not recognize at all,
returning `undefined` and rendering the symbol empty. That is #1251: it is one
line (`return value`), but it changes behaviour for every text symbol with an
unrecognized format, which is more than this PR should decide.

**Not verified live** remains accurate for the client path: the parity above is
a differential test of the two functions, not a browser render.

Your approvals fell with this push. Re-requesting.

genro (changes requested):
Changes request on `dcc8e821c2`. Two points, both inside the feature this PR ships.

First what holds. I checked the parity myself rather than reading the table: the same 20 inputs through `gnrstring.valueMapFormat` and the `gnrlang.js` mirror, identical results, `'toString'` and `'constructor'` included. `tests/core/gnrstring_test.py` plus `gnrlocale_test.py` 59 passed, `node --check` clean. Handling the comma inside a label instead of documenting it was the right call, and so was refusing `objectFromString`.

## 1. An unmapped value has to render as the value, not as nothing

The server already does this. `valueMapFormat` returns `None`, `LocalizedWrapper.__getitem__` falls through, and `toText` renders the raw value:

```
toText('M::T', format='A:Approvato,R:Respinto')  ->  'M'
```

The client has no such fallback. `valueMapFormat` returns `null`, the `#` test right after it is false because the helper already excluded `#`, so `format_T` reaches the end of the function without a return: `asText` propagates `undefined` and the symbol renders empty. Same template, same record: `M` in the editor preview and in the print, nothing on the page.

The cause is that `null` answers two different questions with one value:

- the format is not a value map, and `format_T` has to carry on to its remaining branches;
- the format is a value map that this value does not match, and the raw value is the answer.

Distinguish the two on the client and return `value` in the second case. Helper parity is not what has to hold; rendered parity is. The Python side gets its raw fallback from `toText`, which has no counterpart on the client, so the client caller has to supply it.

This also removes the workaround the current state forces. Today the only way to keep the two sides in agreement is to declare a `*` entry so that no value is ever unmatched. A wildcard should be a choice, not the price of consistency.

## 2. The wildcard loses the code, and the format string should say whether that is wanted

With `*:Altro`, `M` and every other unmapped code render identically and cannot be told apart afterwards. Sometimes that is the intent, for instance collapsing internal states on a document going to a third party, so it has to stay available.

The token for this already exists in the mask, on both sides: `%s`, applied by `result = mask % result` in `toText` and by `mask.replace(/%s/g, formattedValue)` at `gnrlang.js:1108`. Same token in a value-map label:

| format | value `M` |
|---|---|
| `A:Approvato,R:Respinto` | `M` |
| `A:Approvato,R:Respinto,*:Altro` | `Altro` |
| `A:Approvato,R:Respinto,*:Altro [%s]` | `Altro [M]` |

One substitution per side. A label without `%s` behaves exactly as it does now, so this is additive.

Deciding it here rather than later is worth the round trip: the syntax is stored in `adm.userobject` records, and nothing is stored yet.

## #1251

Still a real defect, and wider than this PR: `format_T` exits without a return for any format it does not recognize, and `asText` serves grid cells and domsource nodes as well as templates. Point 1 neither depends on it nor closes it.

fporcari:
Both points done in `3cce448a49`, answering the review at https://github.com/genropy/genropy/pull/1056#pullrequestreview-5099516811 . I checked the two premises in the code before touching anything; both hold.

## 1. An unmapped value renders as the value

> Distinguish the two on the client and return `value` in the second case. Helper parity is not what has to hold; rendered parity is.

Premise verified as described: `format_T` has exactly one branch left after the value-map check, the `#` one, and a value map never contains `#` because `valueMapFormat` excludes those formats up front — so an unmatched value fell off the end of the function, `asText` returned `undefined` (`gnrlang.js:1108`, `if(!mask || !formattedValue){return formattedValue;}`) and the symbol rendered empty.

The client helper now returns `null` for one question only — this format is not a value map — and the raw value for the other:

- `gnrjs/gnr_d11/js/gnrlang.js:890-895` — the two lookups collapse into one `label`, `null` label means the format parsed as a map but nothing matched, so `return value`.
- `gnrjs/gnr_d11/js/gnrlang.js:1186` unchanged: `valueMapFormat` still returns `null` for a non-map format, so `format_T` still falls through to its `#` branch.

The Python side keeps its old contract — `valueMapFormat` returns `None` for an unmatched value and `toText` supplies the raw fallback — so nothing changed there for point 1.

## 2. `%s` in a value-map label

> Same token in a value-map label. One substitution per side. A label without `%s` behaves exactly as it does now, so this is additive.

Premise verified: `result = mask % result` at `gnrpy/gnr/core/gnrstring.py:903`, `mask.replace(/%s/g, formattedValue)` at `gnrlang.js:1108`.

- `gnrpy/gnr/core/gnrstring.py:107-110` — `label.replace('%s', value)`, a no-op on a label without the token. Deliberately `.replace` and not `label % value`, which would raise on any label carrying a bare `%` (`Sconto 50%`).
- `gnrjs/gnr_d11/js/gnrlang.js:894` — `label.replace(/%s/g,value)`, same regex idiom as line 1108.

It applies to a keyed label as well as to the wildcard (`A:Approvato [%s]` -> `Approvato [A]`); the wildcard is where it earns its place, but restricting it to `*` would have been the extra rule, not the simpler one.

Your table, both sides, value `M`:

| format | Python | client |
|---|---|---|
| `A:Approvato,R:Respinto` | `M` | `M` |
| `A:Approvato,R:Respinto,*:Altro` | `Altro` | `Altro` |
| `A:Approvato,R:Respinto,*:Altro [%s]` | `Altro [M]` | `Altro [M]` |

## Tests and checks

New tests, mirroring the style of the ones already in the file:

- `gnrpy/tests/core/gnrstring_test.py:132` `test_valueMapFormat_unmatched_value_renders_raw` — `toText('M::T', format='A:Approvato,R:Respinto') == 'M'` plus the same through `templateReplace` on a Bag node.
- `gnrpy/tests/core/gnrstring_test.py:140` `test_valueMapFormat_label_keeps_the_raw_value` — wildcard with and without `%s`, a keyed label with `%s`, two tokens in one label, and the end-to-end `templateReplace`.

Numbers, all run on `3cce448a49`:

- `pytest gnrpy/tests/core/gnrstring_test.py gnrpy/tests/core/gnrlocale_test.py -q` -> **61 passed** (59 before, +2).
- full `gnrpy` suite, through the pre-commit hook -> **2023 passed, 69 skipped**.
- `flake8 gnrpy/gnr/core/gnrstring.py gnrpy/tests/core/gnrstring_test.py` -> **0 errors**.
- `node --check gnrjs/gnr_d11/js/gnrlang.js` -> clean.
- the `valueMapFormat` doctest, `%s` example included, passes under `--doctest-modules`.

The client change has no test harness in the repo, so I ran the helper through `node` in isolation over the 20 inputs of the Python tests plus the `%s` ones: same results as Python everywhere, `'toString'` and `'constructor'` included, and the three rows of the table above.

**Not verified:** nothing was exercised in a browser — no editor preview, no grid cell, no print. Rendered parity for formats that are *not* value maps is still broken and untouched: `format_T` with `tf` returns `undefined` on the client where Python renders `M`. That is #1251, as you said, and this PR neither depends on it nor closes it.

`develop` has `dismiss_stale_reviews`, so this push resets the review state on the PR. There was no approval standing on it — only your two changes-requested reviews — so nobody's approval disappeared. Re-requesting your review; @cgabriel's standing request is untouched.

genro (changes requested):
The two points from my second round are answered: an unmapped value renders as the value rather than as nothing, and `%s` in a label lets a wildcard keep the code. The comma-in-a-label and key-stripping fixes from the first round are in on both sides. The helpers being deliberately asymmetric — `None` in Python, the raw value in JS — is fine and documented, because `toText` supplies the server's fallback.

There is still one divergence, and it is the same shape as the one I rejected in the first round rather than a new one.

`format` and `mask` are two columns of the same Variables grid, `tpleditor.py:254` and `:255`, and both reach `LocalizedWrapper`. When a row carries both, the two sides disagree. Server: `gnrstring.py:181-184` returns the mapped label before the mask block at `:190-195` and before `toText(..., mask=mask)`, so the mask is dropped. Client: `gnrlang.js:1186-1189` returns the label to `asText`, which then applies `mask.replace(/%s/g, ...)` at `:1108-1111`. A row with `format='A:Approvato,R:Respinto'` and `mask='[%s]'` on value `'A'` renders `Approvato` in the editor preview and in print, and `[Approvato]` in a `dataTemplate` on a page. No test covers format and mask together.

It also gives `%s` two meanings that stack on one side only: label `*:Altro [%s]` with mask `[%s]` gives `[Altro [M]]` on the client and `Altro` on the server.

Decide which side is right and make the other follow. I have no strong preference between applying the mask to the mapped label or documenting that a value map supersedes the mask, as long as both sides say the same thing and a test pins it.

One more asymmetry, for the record rather than for this round. Python applies the map to any `str` value regardless of dtype, `gnrstring.py:181`; JS applies it only inside `format_T`, so `format_P` and the others never see it. Nothing diverges today — the one real P format of that shape, `localize_img`'s in `gnrlocale.py:121`, never matches a real path and declares no wildcard — but it holds by accident.

fporcari:
Answering the [review](https://github.com/genropy/genropy/pull/1056#pullrequestreview-5235134457) in `132720138`.

> Decide which side is right and make the other follow. I have no strong preference between applying the mask to the mapped label or documenting that a value map supersedes the mask, as long as both sides say the same thing and a test pins it.

The client is the reference and the server follows: **the mask wraps the mapped label too**. `mask` is the presentation wrapper and it wraps whatever the formatting produced, which is what `gnrformatter.asText` already does for every other format; making the value map the one thing that escapes it would be the extra rule.

`gnrstring.py:186-201`: the mapped label no longer returns early, it joins the same tail as every other value. The `P`/urlformatter and `'%s::%s' % (value,dtype)` lines are skipped when a map matched — that value is discarded anyway, and `urlformatter` is `site.externalUrl`, worth not calling for nothing. The `#`→caption substitution stays shared, so a mapped row goes through the caption mask like any other. The mask itself is applied with `.replace('%s', label)`, the client's idiom verbatim, and deliberately **not** with `toText`'s `mask % result`: that raises on `'%s (%s)'` and on `'%s%'`, where the client happily produces a string. `gnrlang.js` is untouched — it was already right.

Measured, not reasoned: the client column below is the real `gnrlang.js` loaded in `node`, through both call shapes (`dataTemplate`'s two steps at `:399` and `:410`, and a single `asText({format,mask})` — they always agreed), the server column is `templateReplace('$v', Bag)` with the same attrs.

| value | format | mask | client | server |
|---|---|---|---|---|
| `A` | `A:Approvato,R:Respinto` | `[%s]` | `[Approvato]` | `[Approvato]` |
| `R` | `A:Approvato,R:Respinto,*:In esame` | `<b>%s</b>` | `<b>Respinto</b>` | `<b>Respinto</b>` |
| `Z` | `A:Approvato,R:Respinto` | `[%s]` | `[Z]` | `[Z]` |
| `M` | `A:Approvato,*:Altro [%s]` | `[%s]` | `[Altro [M]]` | `[Altro [M]]` |
| `A` | `A:Sconto 50%,R:Respinto` | `[%s]` | `[Sconto 50%]` | `[Sconto 50%]` |
| `A` | `A:Approvato,R:Respinto` | `%s%` | `Approvato%` | `Approvato%` |
| `A` | `A:Approvato,R:Respinto` | `%s (%s)` | `Approvato (Approvato)` | `Approvato (Approvato)` |
| `A` | `A:Approvato,R:Respinto` | — | `Approvato` | `Approvato` |

The first row is your case: `Approvato` on the server before this commit.

Two tests pin it, real `Bag` through the real pipeline: `test_valueMapFormat_mask_wraps_the_mapped_label` and `test_valueMapFormat_caption_mask_still_names_the_field`. Both fail on the previous head (`assert 'Approvato' == '[Approvato]'` and `== 'Stato: Approvato'`) and pass now. `gnrstring_test.py` plus `gnrlocale_test.py` are 64 passed, `tests/core` 537 passed and 9 skipped, the commit hook's full suite 2617 passed and 19 skipped, flake8 clean, `node --check` clean.

> Python applies the map to any `str` value regardless of dtype [...] Nothing diverges today [...] but it holds by accident.

Said so in the docstring, next to the line already there about the discriminator.

Two divergences are left standing, both pre-existing and neither introduced here. Say if you want them as issues rather than as this paragraph.

- `#` in the mask is a server-only caption placeholder: `mask='#: %s'` gives `Stato: Approvato` on the server and renders the `#` literally on the client. True for every dtype, not only for value maps, which is why I kept the mapped path consistent with the rest of the server instead of dropping the feature for mapped rows alone.
- An empty value with a wildcard: `value=''`, `format='A:Approvato,*:In esame'` gives `In esame` on the server and `''` on the client, where `format_T` returns early on `value===''`. The server already answered that before this commit; the change only adds the mask to it.

Not verified: nothing was run in a browser this round either — no editor preview, no grid cell, no print. The client column is the real file exercised in `node`, not a rendered page.

No approval fell with this push: your changes request was the only review standing.

genro (dismissed):
Closed, and closed the way I hoped: the client is the reference and the server follows. The early return is gone, the mapped value now rides through the `#`-caption substitution, and the mask wraps the mapped label at the end. Choosing `.replace` over `%` because a label or a mask can carry a bare `%` is the detail that would have bitten later.

I did not take the table in the description on trust — I had both sides driven over the same inputs, the server through the real `templateReplace` with a real Bag and the client helper in node. They agree on all seven, including the two I was most doubtful about: the wildcard label carrying its own `%s` inside a mask that also carries one, which gives `[Altro [M]]` on both, and a label containing a bare `%`.

Writing the dtype asymmetry into the docstring is the right disposal for it. It stays true by accident rather than by construction, and now at least the accident is recorded.

One small thing, narrow enough that I would not push for it on its own. The client substitutes with a string replacement, so JavaScript expands `$&`, `` $` ``, `$'` and `$1` inside the value: a value of `A$&B` through a wildcard label comes out `Altro [A%sB]` on the client and `Altro [A$&B]` on the server. It needs a value containing a `$` pattern, which is unlikely in a status code, but it is the same class of divergence this PR spent three rounds closing. A replacement function instead of a string would end it.

cgabriel (commented):
As per @genro comment:

```
One small thing, narrow enough that I would not push for it on its own. The client substitutes with a string replacement, so JavaScript expands $&, $`, $' and $1 inside the value: a value of A$&B through a wildcard label comes out Altro [A%sB] on the client and Altro [A$&B] on the server. It needs a value containing a $ pattern, which is unlikely in a status code, but it is the same class of divergence this PR spent three rounds closing. A replacement function instead of a string would end it.
```

I'd narrow it down in this PR before this lands into develop branch, input sanification is a must.

fporcari:
Closed on `e828c852c`.

@cgabriel, your ask:

> I'd narrow it down in this PR before this lands into develop branch, input sanification is a must.

— https://github.com/genropy/genropy/pull/1056#pullrequestreview-5248837049

@genro, the finding it refers to:

> The client substitutes with a string replacement, so JavaScript expands `$&`, `` $` ``, `$'` and `$1` inside the value: a value of `A$&B` through a wildcard label comes out `Altro [A%sB]` on the client and `Altro [A$&B]` on the server.

— https://github.com/genropy/genropy/pull/1056#pullrequestreview-5247742880

## Change

Two lines in `gnrjs/gnr_d11/js/gnrlang.js`, both now passing a replacement **function** instead of a replacement string, which receives no `$` expansion:

- `valueMapFormat` (`:922`) — the label substitution;
- `gnrformatter.asText` (`:1139`) — the mask substitution, where the same class of divergence reaches any formatted value, not only a mapped one.

`gnrstring.py` is untouched: `str.replace` is literal, so the server was already right and the client is what had to follow — the same direction the third round settled on.

## Verification

`gnrjs/tests/value_map_format.test.js`, new, drives `$&`, `` $` ``, `$'`, `$1` and `$$` through the wildcard label and through the mask, plus a label carrying a bare `$$`, plus the mapped and unmapped renderings that must not move.

- `node --test gnrjs/tests/value_map_format.test.js`: 4 passed.
- Ratchet: against the pre-push `gnrlang.js`, exactly the two `$`-pattern tests fail and the other two pass, so the file pins the defect rather than the new behaviour.
- `node --test gnrjs/tests/*.test.js` (the CI command): 43 passed, 0 failed.
- Server parity driven on the same five inputs through the real `gnrstring.valueMapFormat` with `PYTHONPATH` pinned to the worktree: `A$&B` → `Altro [A$&B]` and so on for all five, byte-identical to what the client test now asserts.
- Python suite `core sql web app xtnd`: 2617 passed, 19 skipped. The repo's pre-commit hook short-circuits on this commit because no `.py` file is staged, so those numbers are from running it by hand, not from the hook.

## The approvals fell

@genro this push dismissed your approval of 2026-09-18, and nothing you approved has moved: the two edited lines are the substitution mechanism only, the value map semantics, the helper asymmetry and the `%s`-in-a-label behaviour are exactly what you read. Re-requesting you, once.

## One occurrence left alone, deliberately

`gnrlang.js:1559`, in the typed-value serialization, does `result[1] = mask.replace(/%s/g, result[1])` — same class, different path, untouched by this PR and outside the case either of you named. It feeds the typed XML/JSON serialization rather than the rendered template, so I did not widen the diff onto it. Worth its own issue if you want it closed.

genro (approved):
Approved on `e828c852ca`.

This closes the last point from my previous round, which was also the one @cgabriel asked to see settled before it lands.

`e828c852ca` replaces the replacement string with a replacement function in both places that substitute `%s` — `valueMapFormat` and the mask branch of `gnrformatter.asText` — so `$&`, `` $` ``, `$'`, `$n` and `$$` inside a value are no longer expanded by JavaScript. The Python side already used `str.replace`, which is literal, so the two sides now agree by construction rather than by the shape of the input.

`value_map_format.test.js` drives all five patterns through both the label and the mask, and re-asserts the mapped, unmapped and wildcard renderings so the fix cannot have moved them.

fporcari:
One traceability note before this merges: `closingIssuesReferences` is empty.

The work answers #993 — @genro's first review opens on exactly that — but nothing in the description links it, so the squash commit will point at no issue and #993 will still be open after this lands, with nobody assigned to close it. Flagging it rather than editing the body, which is the record the three review rounds were held against.

Two ways to settle it, both @fporcari's: add `Fixes #993` to the description before merging, or close #993 by hand afterwards with a line pointing here.

fporcari:
Correcting my note above: the missing link was not an oversight, and #993 should **not** be closed against this PR.

`Refs #993` in the description is right. #993 asks for a `Conditional` parameter fieldtype in the template editor's Parameters grid — a condition plus two alternative values, evaluated per record. This PR ships a value map on a symbol's `format`: a lookup from a value to a label. Related, and it removes one of the workarounds #993 lists, but it is not the branching mechanism that issue is about, and #993 carries no agreement that it would be.

#993 stays open, with @dgpaci on it.

PR #1057, opened 2026-08-07
fix(widgets): setCondition no longer nulls then re-asserts a stale value

### LINKED ISSUE #994: [BUG] dbSelect setCondition re-asserts a stale captured value (async write-back race)
## Problem Description

`gnr.widgets.DynamicBaseCombo.mixin_setCondition` (`gnrjs/gnr_d11/js/genro_widgets.js`, ~line 4853) runs whenever a bound `condition_*` parameter of a dbSelect changes (`^.path` bind → dyn-attr dispatch in `gnrdomsource.js` ~line 1419). It:

1. captures the CURRENT datastore value (`currvalue = sourceNode.getRelativeData(vpath)`);
2. calls `clearCache()`;
3. calls `setValue(null, true)` followed by `setValue(currvalue, true)` — both with `priorityChange=true`, i.e. **write-back semantics**.

Two problems follow:

**(a) Synchronous null write.** The `setValue(null, true)` step resolves synchronously (`fetchItemByIdentity` with empty identity, `gnrstores.js` ~line 592) and fires `onChange(undefined)` → the widget's bound path is **nulled in the datastore inside the very trigger cascade that changed the condition parameter**. Any controller subscribed to the value path fires on that transient null.

**(b) Stale async write-back.** The `setValue(currvalue, true)` step is asynchronous: the cache was just cleared, so it goes through a server round-trip (`app.dbSelect` identity fetch). When the reply lands, `_callbackSetLabel` is invoked with `dataObject=undefined`, which bypasses dojo's `_lastQuery` staleness guard (`FilteringSelect.js`, `_callbackSetLabel`), and `_setValueFromItem(..., true)` → `_handleOnChange(currvalue, true)` → `onChange` → **the value captured in step 1 is written back to the datastore**, with no check that the datastore value hasn't changed in the meantime.

If any code writes a NEW value to the same path while that round-trip is in flight (typical: chained selectors where changing selector A both re-filters selector B via `condition_x='^.a_path'` AND programmatically derives a new value for B), the stale reply **overwrites the new value with the old one**: the field visibly bounces back to its previous value. Whether the bounce happens depends on network/server timing, so it shows up as a random, hard-to-reproduce glitch. If the captured value does NOT satisfy the new condition, the reply is empty and the widget is flagged invalid instead (`_isvalid=false`) — a milder variant of the same race.

**Expected behavior:** re-evaluating the caption/validity of the current value under the new condition must not write anything back to the datastore, and a reply that arrives after the datastore value has changed must be discarded.

## Reproduction Steps

1. Build a page with two chained dbSelects and a derivation, e.g.:
   - `dbSelect(value='^.a', table=...)`
   - `dbSelect(value='^.b', table=..., condition='$fk=:x', condition_x='^.a')`
   - a `dataRpc`/`dataController` on `^.a` that computes and SETs a new `.b` (server round-trip).
2. Give `.b` a value valid under both the old and the new `.a` (so the identity re-fetch finds it under the new condition).
3. Change `.a`. While the setCondition identity fetch is in flight, the derivation SETs the new `.b`.
4. When the identity fetch replies after the derivation (add a small server delay on `app.dbSelect` to force the ordering), `.b` bounces back to the value it had before the change. The transient null write of step (a) can be observed on every change by subscribing a controller to `^.b`.

## If Cannot Be Reproduced Externally

Observed in production in the Demetra agenda wizard (specialità → operatore → sede chain, where the sede is auto-derived from the operator's tmsh coverage): switching between operators of the same specialty randomly bounced the sede selector back to the previous operator's sede. Full analysis available on request.

## Environment

- GenroPy current `master` (`gnrjs/gnr_d11`), dojo 1.1 widgets
- Any browser (JS-level race, not browser-specific)

## Additional Material

Verified code chain:

- `genro_widgets.js` `mixin_setCondition` (~4853): capture + `clearCache()` + `setValue(null,true)` + `setValue(currvalue,true)`
- `gnrdomsource.js` (~1419): `condition_*` dyn-attr → `widget.setCondition(value, kw)`
- `gnrstores.js` `GnrStoreQuery.fetchItemByIdentity` (~589): empty identity resolved synchronously; non-empty → remote resolver round-trip (cache cleared right before, so always remote here)
- dojo `FilteringSelect._callbackSetLabel`: `dataObject==null` skips the `_lastQuery` staleness check by design; found item → `_setValueFromItem(item, priorityChange=true)`
- dojo `_FormWidget._handleOnChange`: `priorityChange=true` + value ≠ `_lastValueReported` → `onChange` → genro `connectChangeEvent`/`onChanged` → `_doChangeInData` → datastore write (only guard: `!widget.disabled`)

Suggested fix direction: in `mixin_setCondition`, re-set the captured value with `priorityChange=false` (the value is already in the datastore; only the caption/validity needs refreshing), and/or have the async callback compare the live `sourceNode.getRelativeData(vpath)` with the captured `currvalue` and abort when they differ. The transient `setValue(null, true)` should likewise not reach the datastore (`priorityChange=false`).

Workaround for applications: bind chained-selector condition params with `=` instead of `^` — resolver kwargs resolve both at query time (`gnrdomsource.js` `currentFromDatasource`), so the dropdown filter stays live while the `setCondition` dance (and the race) disappears; the trade-off is losing the automatic caption re-validation on condition change.

fporcari:
Fixes #994

## Problem

`DynamicBaseCombo.mixin_setCondition` revalidated the current value with write-back semantics. The null reset wrote a transient null to the bound datapath, while the asynchronous identity response could later restore the stale value captured before the condition changed.

The first correction removed those writes, but exposed a separate inconsistency in the single-option path: the new pkey was stored without refreshing `selectedRecord`, `selectedCaption`, or `selected_*`, so values such as `AO` and `Milano` could coexist.

## Change

- Revalidate an existing value with `priorityChange=false`, retaining the null step only to reset `_lastValue` and force the identity fetch.
- Keep the full row returned by the synchronous single-option lookup.
- For a newly selected single option, install that item in the widget, update its value without an `onChange` write-back, write the pkey explicitly, and call `_updateSelect(item)` so all related fields change together.
- Expose `selected_nome` in the manual `dbSelect` fixture.
- Clear `_lastQueryError` and the sourceNode validation error when a stale reply is discarded: the identity lookup sets both before the callback can tell the value is no longer in the datastore.

This also avoids the duplicate identity request on the single-option path: the widget already owns the authoritative row when the datastore dispatch re-enters it.

## Behavior change

When the current value becomes invalid under a new condition, it remains in the datastore while the widget is invalid, until it is corrected or blurred. This prevents transient condition changes from erasing the value. Form validation still blocks a normal save while the widget is invalid.

## Verification

Real browser on the sandbox instance served from this branch, `glbl.regione` / `glbl.provincia`:

- `/test/inputfields/dbselect/9` (new, `test_9_condition_null`): from VAL / AO, `SET .regione='LOM'; SET .sigla=null;` with the AO identity reply delayed 1 s. Caption empty, datastore `null`, `_validations.error` null, no `_lastQueryError`. With the previous head of this branch the caption and datastore were already clean but the widget kept `current value does not fit condition` in both.
- `/test/inputfields/dbselect/8`: LOM / MI after the delayed stale reply, unchanged.
- `/test/inputfields/dbselect/7`: VAL / AO / Aosta on both widgets, unchanged.
- The same condition-plus-null sequence is what left a stale caption on a production form (`dbSelect` on a case filtered by the subject, both set by one dataController when the context changes).

Earlier rounds:

- `node --check gnrjs/gnr_d11/js/genro_widgets.js`
- `flake8 projects/gnrcore/packages/test/webpages/inputfields/dbselect.py`
- Full Python suite: `2166 passed, 10 skipped`
- Real browser flow on the isolated `test_1_condition` fixture:
  - starting state: `LOM`, `MI`, `Milano`
  - after changing the condition: `VAL`, `AO`, `Aosta`
  - no browser console errors or warnings

The repository has no automated JavaScript test runner for this widget path, so the fixture now makes the `selected_*` regression directly observable.


dgpaci (changes requested):
Reviewed the diff against the full call chain on the branch (dojo `FilteringSelect`/`_FormWidget`, `GnrStoreQuery.fetchItemByIdentity`, `_updateSelect`, the dyn-attr dispatch in `gnrdomsource.js`, `validate_select`/`dojoValidation`). The analysis in the description holds, and I could confirm both load-bearing claims independently — but the oneOption branch is left in an inconsistent state, so requesting changes on that one point.

## Confirmed correct

- **The null step still resets `_lastValue`**, so the re-validation RPC is not skipped: `_FormWidget._handleOnChange` assigns `this._lastValue = newValue` before the `priorityChange` test (`dojo_libs/dojo_11/dojo_src/dijit/form/_FormWidget.js:225`), and the skip in `gnrstores.js:613-616` compares exactly that. Constraint 1 verified.
- **Caption and validity are still refreshed with `priorityChange=false`**, on two counts: `_setValueFromItem` sets `this._isvalid=true` unconditionally (`FilteringSelect.js:148`), and `ValidationTextBox.setValue` calls `this.validate(this._focused)` outside any `priorityChange` gate (`ValidationTextBox.js:68-71`), which reaches the patched `validate` at `genro_widgets.js:417-427` and therefore `form.dojoValidation` → `invalidDojo`. So the form's error state is both raised **and** cleared as before.
- `setRelativeData(vpath, currvalue)` with `vpath = '^.sigla'` resolves correctly: `absDatapath` strips the pointer prefix (`gnrdomsource.js:700-702`).
- `node --check` passes on the touched file; no Python touched; no JS test infrastructure in the repo (confirmed — no root `package.json`/runner).

## 1. Blocking — the oneOption auto-select leaves `selected_*` on the previous row

In the oneOption branch the value is now written, but `_updateSelect` is never reached:

- the explicit `setValue(currvalue,false)` → `onSetValueFromItem(item,false)`, and both `_updateSelect` call sites there are gated on `priorityChange` (`genro_widgets.js:4898`, `:4906`);
- the round trip through the datastore does not help either — the dyn-attr dispatch for `value` forces `_lastValueReported` and then calls `this.widget.setValue(value, false)` with `kw = false` (`gnrdomsource.js:1470-1476`);
- `validate_select` only calls `_updateSelect()` for a **blank** value (`genro_frm.js:2367`).

Before this change that path was reached, with the item freshly re-read **under the new condition**. Concrete case, `dbSelect(value='^.sigla', condition='$regione=:regione', condition_regione='^.regione', selected_nome='.nome_provincia', validate_notnull=True)`, going from `LOM/MI` to `VAL`: the datastore ends up with `sigla='AO'` and `nome_provincia='Milano'`. That is not "losing a refresh" — it is an inconsistent pair committed to the datastore.

The same applies to `selectedRecord`, `selectedCaption` and to `selectedCb`/`selectedSetter`, i.e. to related-column propagation in grid cell editors, which goes through `_updateSelect` as well (`genro_widgets.js:4461-4464`, wired by `cell_onCreating`).

The manual verification could not catch this: the conditioned dbSelect in `test_1_condition` has no `selected_*` (`projects/gnrcore/packages/test/webpages/inputfields/dbselect.py:38-39`) — the `nome_regione` events in the tables come from the *regione* select, not from `setCondition`.

Suggested fix, nearly free: the row is already in hand, delivered synchronously by the `serverCall`. Keep the node instead of just the pkey:

```js
var oneOptionItem = null;
if(reskwargs.notnull){
    ...
    if(singleOption._value.len()==1){
        oneOptionItem = singleOption._value.getNode('#0');
        currvalue = oneOptionItem.attr[this.store._identifier];
    }
}
...
if(currvalue !== datavalue){
    this.sourceNode.setRelativeData(vpath,currvalue);
    this._updateSelect(oneOptionItem);
}
```

Worth double-checking that the `hiddenColumns` needed by `selected_*` are present in that row's attributes (the `serverCall` reuses the resolver kwargs, so they should be).

The cheaper alternative is `this.setValue(currvalue, currvalue !== datavalue)`, which restores the pre-fix behaviour on the oneOption branch only — where the write is intended and the value is derived from the *new* condition rather than captured. But it puts that write back on the async reply (~1s per your own measurement) and so re-opens a race window, narrower though it is. I'd prefer the explicit `_updateSelect`.

## 2. Needs to be stated — a value invalid under the new condition is no longer cleared

Undeclared semantic change. The `setValue(null,true)` step is what used to clear the datastore when the current value fell outside the new condition; the "Correction to the issue's framing" describes the *second* setValue accurately, but the *first* one did write. After this change:

- the textbox is empty (left by the null step) while the datastore keeps `MI` — UI and data diverge until blur, where `_setBlurValue` does `setValue(null,true)` and cleans up;
- inside a form the save is blocked by `dojoValidation` → `invalidDojo` → `canSave` false, so there is no silent bad save; **but** with `allowSaveInvalid`, outside a form, or for a `dataRpc` reading `^.b` in the meantime, the inconsistent value does circulate.

I think the new behaviour is the right default — under the old semantics a transiently empty condition during load (e.g. `condition_regione='^.regione'` while `regione` is still null) wiped the value, which is a plausible co-cause of the production report. It just needs to be spelled out in the PR body and in the release notes, since apps that relied on the auto-clearing will change silently.

## 3. Minor — duplicate identity RPC on the oneOption path

`setRelativeData` re-enters the widget as `setValue(currvalue,false)` while the RPC from the explicit `setValue(currvalue,false)` is still in flight (cache just cleared, `_lastValue` still `''`), so two concurrent `app.dbSelect` identity fetches are issued for the same identity. Harmless, avoidable by writing the datastore *before* the setValue pair, or by skipping the second `setValue` when `currvalue !== datavalue` since the dispatch performs it anyway. Optional.

## 4. Minor — verification fixture

No JS test infrastructure, agreed. Adding `selected_nome='.nome'` to the conditioned dbSelect in `test_1_condition` (or a dedicated `test_6` for the oneOption branch) is a two-line change, makes finding 1 visible by eye and leaves a fixture for future manual verification.

## On the "Decision required"

My vote is **(i) plus an explicit `_updateSelect` on the oneOption branch only** — the minimal subset of (iii), which here does not pay the cost of crossing the async boundary because the item is already available synchronously. I would drop (ii): as the description itself notes it does not fix defect (a), and it would require a flag consulted inside the shared `_doChangeInData` that serves *every* dojo widget — a risk out of proportion to the benefit.

fporcari:
Addressed the requested changes in `b71341769`:

- the single-option row is retained and passed through `_setValueFromItem(..., false)` before `_updateSelect(item)`, so the pkey, caption, `selectedRecord`, and every `selected_*` value are updated as one coherent selection;
- setting `this.item` from the synchronous lookup also prevents the datastore dispatch from starting a duplicate identity request;
- the fixture now exposes `selected_nome`, and the real-browser transition `LOM / MI / Milano` → `VAL / AO / Aosta` passes with no console errors or warnings;
- the invalid-value datastore behavior is now documented in both the PR description and release notes.

The full Python suite passes (`2166 passed, 10 skipped`), together with `node --check`, `flake8`, and `git diff --check`.

dgpaci (dismissed):
Re-reviewed `b71341769` against the full call chain (dojo `FilteringSelect._setValueFromItem`/`_FormWidget._handleOnChange`, `GnrStoreQuery.fetchItemByIdentity`, the dyn-attr dispatch in `gnrdomsource.js`, `_updateSelect`, `validate_select`, `versionpatch_11__setBlurValue`). All three points of my previous review are addressed. Approving.

## Verified

**1. oneOption `selected_*` consistency — fixed.** The single-option row is kept as a node, `_setValueFromItem(item,false)` installs pkey, caption and `_isvalid=true` without an `onChange` write-back, the pkey is written explicitly, and `_updateSelect(item)` propagates `selectedRecord`, `selectedCaption`, every `selected_*` and `selectedCb`/`selectedSetter`. The two preconditions hold:

- `_updateSelect` only writes the row when `isValid()` is true — `_setValueFromItem` sets `_isvalid=true` (`FilteringSelect.js:148`) and nothing between the two calls resets it: the identity fetch triggered by the datastore dispatch is skipped, and `validate_select` returns immediately on a non-blank value.
- exactly one `_updateSelect` call: the handler connected to `_setValueFromItem` (`BaseSelect.connectForUpdate`, which as a mixin wins over `DynamicBaseCombo.connectForUpdate` for `dbSelect`) is gated on `priorityChange`, so it does not fire with `false`; the same applies to the `setValue` connect in `dbComboBox`.

Blur stays coherent as well: `_lastDisplayedValue` and `valueNode.value` are both updated, so `versionpatch_11__setBlurValue` re-asserts the value instead of nulling it.

**2. No duplicate identity request — confirmed.** The mechanism is `_handleOnChange` assigning `_lastValue = newValue` before the `priorityChange` test (`_FormWidget.js:225`), which makes the skip at `gnrstores.js:613` hit when `setRelativeData` re-enters the widget. `this.item` is not what suppresses the request — it matters for the `_is_invalid_item` check in `validate_select`.

**3–4.** Behavior change documented in the PR body and release notes; the fixture now exposes `selected_nome` with a textbox that makes the regression observable.

Checks on my side: `node --check` on the touched file, `flake8` clean on the fixture, `git diff --check` clean, branch level with `develop`.

## Minor, non-blocking

**a. `_lastQueryError` is no longer cleared on the oneOption path.** The skip at `gnrstores.js:613-617` returns *before* the `delete parentSourceNode.widget._lastQueryError` at `gnrstores.js:619`, and the oneOption branch performs no identity fetch at all. So when the previous condition produced a resolver error, the new condition with a single option leaves the widget showing a valid caption while `validate_select` keeps returning `query_error` (`genro_frm.js:2361`) and the save stays blocked until the next list fetch clears it. One line inside the branch fixes it:

```js
delete this._lastQueryError;
```

**b. Write order.** The pkey is written with `doTrigger=true` before `_updateSelect` writes the satellites with `doTrigger='selected_'`, so anything bound to the pkey runs while `selected_*` still holds the previous row. Not a regression — the old path had the same order — but swapping the two lines (`_updateSelect` first, then `setRelativeData`) would make the datastore consistent at the moment the pkey change propagates.

Neither needs to hold the merge; (a) is worth the extra line if you touch the file again.

## Process note

`RELEASE_NOTES.rst` has no `Unreleased` convention on `develop` — sections are created when a release is cut. A per-PR block will conflict between concurrent PRs unless it becomes an agreed convention. I asked for the documentation, so this is a question about where it belongs, not about the content.

fporcari:
The `Unreleased` stanza in `gnrpy/RELEASE_NOTES.rst` added by this PR is under discussion with @cgabriel on #1089, which also covers #1090. The heading has never been used in that file, so these three PRs would be setting a convention rather than following one — asked in one place instead of three. Whatever is decided I will apply to all three. Nothing else on this PR depends on it.

fporcari:
Realigned with `develop` (merge, not rebase — history is unchanged above the merge commit).

The only conflict was `gnrpy/RELEASE_NOTES.rst`: this branch's `Unreleased` section against the `Release 26.08.26` entry that landed on `develop`. Both sides kept, in the file's own order. No source file conflicted, and nothing in the diff of this PR was touched by the merge.

Full suite on the merged tree: 2117 passed, 69 skipped.

@dgpaci — your approval was dismissed by this push (`dismiss_stale_reviews` is on for `develop`). Nothing you approved has changed: the merge only added the release entry to the changelog. Re-requesting so the PR can move.

cgabriel:
> The `Unreleased` stanza in `gnrpy/RELEASE_NOTES.rst` added by this PR is under discussion with @cgabriel on #1089, which also covers #1090. The heading has never been used in that file, so these three PRs would be setting a convention rather than following one — asked in one place instead of three. Whatever is decided I will apply to all three. Nothing else on this PR depends on it.

I was decided that RELEASE_NOTES was not to be updated. But it seems that your agent continues to include them anyway, you know why.

dgpaci (changes requested):
Re-review after the dismissal. The dismissal is mechanical: `52410d208` (merge of `develop`) and `a1a50fa26` (release-notes stanza dropped) invalidated the approval, but `mixin_setCondition` is byte-identical to the `b71341769` state I approved, and I re-verified the load-bearing mechanisms against the post-merge tree (the `#1152` trigger-dispatch rework touched `gnrdomsource.js`/`gnrbag.js`, and the `_lastValueReported` force + `widget.setValue(value, false)` at `gnrdomsource.js:1534-1553` that this fix depends on is unchanged). CI green on 3.11–3.14, `node --check` and `flake8` clean on my side.

Two things I missed on the previous two passes, though, and one is a crash. Requesting changes on the first; the second is a scope question about closing #994.

## 1. Blocking — `_setValueFromItem` does not exist on `dbComboBox`

`_setValueFromItem` is defined **only** on `dijit.form.FilteringSelect` (`dojo_11/dojo_src/dijit/form/FilteringSelect.js:141`; zero occurrences in `ComboBox.js`, which declares `dijit.form.ComboBox` as `[ValidationTextBox, ComboBoxMixin]`).

`mixin_setCondition` lives on `DynamicBaseCombo`, and `_mixinPlan` walks the prototype chain (`for (var prop in handler)`, `genro_wdg.js:592`), so `setCondition` is installed on **every** descendant — including `gnr.widgets.dbComboBox` (`genro_widgets.js:5289`), whose `_dojotag` is `'ComboBox'`. On that widget the new single-option branch throws `TypeError: this._setValueFromItem is not a function`, aborting the whole trigger cascade.

Reachable configuration, nothing exotic:

```python
fb.dbComboBox(table='...', value='^.x', validate_notnull=True,
              condition='$fk=:y', condition_y='^.y')
```

`validate_notnull` is enough to arm the branch — `creating` copies it into `resolverAttrs['notnull']` when `notnull` is unset (`genro_widgets.js:4790-4792`) — and the branch then fires as soon as the new condition yields exactly one option whose pkey differs from the current value. The pre-fix code only ever called `setValue`, which exists on both widgets, so this is a regression introduced here.

A feature-detect guard is enough to keep `dbComboBox` on the old path:

```js
if(oneOptionItem && currvalue !== datavalue && this._setValueFromItem){
```

If you would rather have the `selected_*` consistency on `dbComboBox` too, the branch works there without `_setValueFromItem` by setting `_lastValue` explicitly before the write — but the guard alone is a correct minimal fix and I would not hold the PR for the wider version.

## 2. Scope — the second half of #994's expected behavior is not implemented

The issue asks for two things: *"re-evaluating the caption/validity of the current value under the new condition must not write anything back to the datastore, **and a reply that arrives after the datastore value has changed must be discarded**."* This PR does the first and nothing for the second, but says `Fixes #994`.

The multi-option path still issues `this.setValue(currvalue,false)`, whose reply reaches `_callbackSetLabel(result, undefined, false)` — `dataObject` is `undefined`, so dojo's `_lastQuery` staleness guard is skipped by design (`FilteringSelect.js:63-65`), exactly as the issue documents. Replaying the issue's own repro (`.a` changes → `setCondition` captures `.b = OLD` and starts RPC#1; a `dataRpc` on `^.a` sets `.b = NEW` → dispatch starts RPC#2; RPC#1 lands last):

- `_setValueFromItem(oldItem,false)` → `_setValue` writes `valueNode.value = OLD`, `this.value = OLD`, `_lastDisplayedValue = OldCaption`. The widget **displays the previous row** while the datastore holds `NEW`. No `onChange`, so the datastore itself is safe — that part is genuinely fixed.
- On blur, `versionpatch_11__setBlurValue` takes `_lastDisplayedValue == displayedValue` → `value = this.getValue()` → `OLD` → `this.setValue(OLD, true)` → `_handleOnChange(OLD, true)` against `_lastValueReported = NEW` → `onChange` → `_doChangeInData`. **`.b` bounces back to `OLD`.** Same production symptom, deferred to blur instead of firing inside the cascade.
- Symmetric case: if the captured value is *not* valid under the new condition, the late reply takes dojo's not-found branch (`_isvalid=false; validate(false)`) → patched `validate` → `dojoValidation` → `invalidDojo`. The widget and the datastore both hold the correct `NEW`, but the form is blocked by an error with no visible cause and nothing clears it until the next fetch.

Both disappear with the guard the issue suggests — compare the live datapath value with the captured one on the reply and drop the reply when they differ. Since `mixin_setCondition` owns the call, it can do the fetch itself rather than going through `setValue`:

```js
var self = this;
this.store.fetchItemByIdentity({identity: currvalue, onItem: function(item){
    if(self.sourceNode.getRelativeData(vpath) != currvalue){ return; }
    ...
}});
```

I am fine either way on process — land this as the write-back half and open a follow-up for the reply half, or add the guard here. What I do not think should happen is `Fixes #994` auto-closing the issue while the reported symptom is still reachable. Your call which.

## 3. Release notes

`a1a50fa26` dropped the `Unreleased` stanza, which resolves the convention question from #1089 the right way. The PR body still lists *"Document the behavior change in the release notes"* among the changes — worth correcting, since the behavior change in §"Behavior change" is now recorded only in the PR body and will need picking up when the release is cut.

## 4. Still open from the previous pass (both non-blocking, unchanged)

- `_lastQueryError` is not cleared on the single-option branch: the skip at `gnrstores.js:613-617` returns before the `delete` at `:619`, and the branch performs no identity fetch, so a resolver error from the previous condition survives into a valid new selection and keeps the save blocked. One line inside the branch (`delete this._lastQueryError;`).
- Write order: `setRelativeData` (pkey, `doTrigger=true`) runs before `_updateSelect` (satellites), so anything bound to the pkey observes the previous row's `selected_*`. Not a regression — the old path had the same order.

## Verified again on this head

- Null step still resets `_lastValue` before the `priorityChange` test (`_FormWidget.js:225-227`), so the skip at `gnrstores.js:613` does not swallow the re-validation RPC.
- Single-option row carries the `selected_*` columns: `creating` pushes every `selected_*` key into `hiddenColumns` (`genro_widgets.js:4804-4816`) and the synchronous `serverCall` reuses the resolver kwargs, so `_updateSelect(oneOptionItem)` finds `nome` on `glbl.provincia`.
- No duplicate identity request on the single-option path, and `_updateSelect` fires exactly once there.
- `setRelativeData` uses `lazySet`, whose `isEqual` is loose, so the strict `currvalue !== datavalue` cannot produce a spurious dirty write when pkey types differ between the bag row and the datastore.

fporcari:
> [“Two things I missed on the previous two passes, though, and one is a crash.”](https://github.com/genropy/genropy/pull/1057#pullrequestreview-5048685341)

Addressed both blocking points in `5c8e45fb1`.

- The single-option path now feature-detects `_setValueFromItem`. `dbSelect` retains the item-based path, while `dbComboBox` uses its native synchronous `setValue` path before the explicit datastore write and `_updateSelect`, so both widgets keep their value and related fields coherent.
- `FilteringSelect.patch_setValue` now accepts an optional value guard. `setCondition` supplies a guard that compares the live datapath value with the captured value, and the guard is checked both when the identity result arrives and after a deferred item load. A stale condition reply is therefore discarded before it can alter caption, validity, or the value restored on blur.
- Added the dedicated testhandler fixture `/test/inputfields/dbselect/7`, covering a conditioned `dbSelect` and `dbComboBox` with `validate_notnull` and `selected_nome`.

Browser regression on the migrated `sandboxpg` instance passed: starting from `LOM` / `MI`, switching to `VAL` selected `AO` in both widgets and updated both related captions to `Valle d’Aosta`. Browser console and server log remained clean.

Checks completed on the current head:

- `node --check gnrjs/gnr_d11/js/genro_widgets.js`
- `flake8 projects/gnrcore/packages/test/webpages/inputfields/dbselect.py`
- CI-equivalent flake8 selection across `gnrpy`: zero errors
- `git diff --check`
- full Python suite: `2166 passed, 69 skipped`
- commit hook suite: `2117 passed, 69 skipped`

The branch had no standing approval to dismiss: the earlier approval was already dismissed by the develop merge. The non-blocking `_lastQueryError` and write-order observations remain outside this focused change.

Not verified: an exhaustive scan of every application-specific `condition_*` configuration; the browser regression exercises the shared framework path on both affected widget types.

dgpaci (changes requested):
Reviewed `5c8e45fb1`. Point 1 is fixed. Point 2 is not: the guard you added cannot execute — `patch_setValue` is installed on a class that no `setCondition`-capable widget inherits, and the widgets that do call `setCondition` receive dojo's native two-parameter `setValue`, which discards the third argument.

## 1. `dbComboBox` crash — fixed

`if(this._setValueFromItem)` at `genro_widgets.js:4887` keeps `dbComboBox` off the FilteringSelect-only method, and `test_7_condition_race` exercises that widget for real. Confirmed resolved.

## 2. Blocking — the `valueGuard` is unreachable on both ends

`patch_setValue` lives inside `dojo.declare("gnr.widgets.FilteringSelect", gnr.widgets.BaseCombo, ...)` (`genro_widgets.js:5059`, patch at `:5067`). Patches are per-handler: `getHandler` resolves one handler object per tag from the `gnr.widgets` catalog (`genro_wdg.js:250-258`, `:283-296`), `_mixinPlan` scans **that handler's own prototype chain** (`genro_wdg.js:583-616`), and `doMixin` installs only what that plan collected (`genro_wdg.js:619-650`).

The two chains never meet:

- `gnr.widgets.dbSelect` (`:5303`) = `[dbBaseCombo (:4984) → DynamicBaseCombo (:4776) → BaseCombo (:4343)]` + `BaseSelect (:5210)`.
- `gnr.widgets.FilteringSelect` (`:5059`) also extends `BaseCombo` — it is a **sibling** of that branch, not an ancestor.

So the `dbselect` handler's plan contains no `patch_setValue` (the only `patch_setValue` on its chain is none — `BaseSelect` carries just `versionpatch_11__setBlurValue`). The widget keeps `dijit.form.FilteringSelect.setValue(value, priorityChange)`, a two-parameter function; `this.setValue(currvalue,false,guardFn)` at `:4897-4900` passes the closure into a slot that does not exist and it is silently dropped. `dbComboBox`, `callbackSelect` and `remoteSelect` are in the same position (the first through `_FormValueWidget.setValue`).

Conversely, the only handler that *does* receive the patched `setValue` is the `filteringSelect` tag itself — and `gnr.widgets.FilteringSelect` extends `BaseCombo`, not `DynamicBaseCombo`, so it has no `setCondition` and never supplies a guard.

Net effect: both `valueGuard` tests at `:5075` and `:5085` are dead code, and the multi-option path behaves exactly as before the commit. The stale reply still reaches `_callbackSetLabel(result, undefined, false)` with the `_lastQuery` check bypassed, still rewrites `valueNode.value` / `_lastDisplayedValue` with the previous row, and `versionpatch_11__setBlurValue` still re-commits it through `onChange` on the next blur.

Easy to confirm on the fixture page: `genro.wdg.getHandler('dbselect').__mixinPlan.methods` contains no entry with `funcname == 'setValue'`, while `getHandler('filteringselect').__mixinPlan.methods` does.

## 3. The new fixture cannot detect it

`test_7_condition_race` never reaches the guarded branch. `glbl.provincia` has exactly one row for `VAL`, and the dbSelect carries `validate_notnull=True` → `resolverAttrs.notnull` → the synchronous single-option lookup succeeds → `oneOptionItem` is set and `currvalue !== datavalue` (`AO` vs `MI`) → the **single-option** branch runs. The `else` branch that passes the guard is never entered by either widget on this fixture.

It also cannot expose the ordering the issue describes. `SET .regione='VAL'; SET .sigla='AO';` runs synchronously in one action, so the identity fetch for the stale value starts first and normally replies first — the issue's own repro calls for a delayed `app.dbSelect` reply to force the inverse ordering. A green browser run on this page is therefore consistent with the guard being dead, which is what it is.

To cover the guarded path the fixture needs a target region with **several** provinces (so `notnull` finds `limit:2` rows and `oneOptionItem` stays null) plus a value change arriving after the condition change — a `dataRpc`/`dataController` on `^.regione` that sets `.sigla`, rather than a synchronous `SET` chain.

## Direction

Two ways out, your choice:

- Move the guard to a `patch_setValue` on a class the affected handlers actually inherit. `BaseSelect` covers `dbSelect`/`callbackSelect`/`remoteSelect` but not `dbComboBox`; `DynamicBaseCombo` covers all four but would patch `setValue` for `dbComboBox` too, whose dojo `setValue` has different semantics (`TextBox.setValue` writes the raw value into the visible box) — worth checking before going that way.
- Or skip `setValue` entirely on that path and do the lookup where the guard is naturally in scope, which is widget-agnostic and needs no patch:

```js
var self = this;
this.setValue(null,false);
this.store.fetchItemByIdentity({identity:currvalue, onItem:function(item){
    if(self.sourceNode.getRelativeData(vpath) != currvalue){ return; }
    self.setValue(currvalue,false);   // cache now warm: resolves synchronously
}});
```

I prefer the second: the cache populated by the fetch makes the follow-up `setValue` a synchronous cache hit (`gnrstores.js:600-606`), so the reply path stays inside dojo's own code and no shared method signature changes.

## Unchanged from the previous review

`_lastQueryError` on the single-option branch and the pkey/satellite write order — still open, still non-blocking, still fine to leave outside this change.

Checks on my side: `node --check` clean, `flake8` clean on the fixture, CI green. Nothing in §2 shows up in either, which is the point — a silently ignored extra argument is invisible to both.

fporcari:
@dgpaci Addressed the remaining blocking point in e37270706.

- The staleness guard now lives in `DynamicBaseCombo.mixin_setCondition`: it performs the identity lookup directly, compares the live datapath value in the callback, and only then calls the widget's native `setValue`. The lookup has already populated the store cache, so that final call is synchronous.
- Removed the unreachable `valueGuard` extension from the sibling `FilteringSelect` handler.
- Preserved the single-option fixture for `dbSelect` and `dbComboBox`, and added a separate multi-option race fixture. It starts from VAL/AO, changes the condition to LOM, writes MI through a `dataRpc` while the delayed AO identity lookup is pending, and exposes the datastore value alongside the widget.

Verification on the pushed commit:

- `node --check gnrjs/gnr_d11/js/genro_widgets.js`
- configured repository flake8 selection: zero errors
- full Python suite: 2225 passed, 10 skipped
- commit-hook suite: 2176 passed, 10 skipped
- browser regression on `/test/inputfields/dbselect/7` and `/test/inputfields/dbselect/8`: the single-option related values stay coherent; after the delayed stale reply, both the widget and datastore remain on MI; no browser console or server errors

Not verified: an exhaustive scan of application-specific `condition_*` configurations.

fporcari:
@dgpaci one more commit before your re-review, `139d27abc0`, on the point you had left as non-blocking.

The discard branch of the guard now also clears `_lastQueryError` and the sourceNode validation error. Both are set by `fetchItemByIdentity`'s `finalize` (`gnrstores.js:643-647`) before the callback can see that the value is no longer in the datastore, so a stale reply for a value that failed the new condition left `current value does not fit condition` on a widget whose value was already `null`; the next `validate_select` would then have reported it against a legitimate value.

Sequence that showed it, now `test_9_condition_null`: condition parameter and value set to `null` in the same tick, with the identity reply landing afterwards. Verified in the browser from this branch: caption empty, datastore `null`, no validation error left; tests 7 and 8 unchanged. `develop` merged in first (`18f273fae9`, no conflicts). Commit hook suite: 2283 passed, 69 skipped.

genro (changes requested):
Read this against @dgpaci's three passes rather than from scratch, since they carried the analysis. All four of their blocking points are closed at this head:

- the single-option branch keeps the row as a node and reaches `_updateSelect(oneOptionItem)`, so `selected_*`, `selectedRecord`, `selectedCaption` and the `selectedCb`/`selectedSetter` propagation stay on the row actually selected;
- `if(this._setValueFromItem)` (`genro_widgets.js:4985`) keeps `dbComboBox` off the FilteringSelect-only method;
- the unreachable `valueGuard`/`patch_setValue` is gone, replaced by the second option dgpaci proposed — `fetchItemByIdentity` with the comparison in natural scope, no shared signature changed and the follow-up `setValue` resolving from the warm cache;
- the fixture now reaches the guarded branch and can order the replies: `test_8_condition_race` delays `_id=='AO'` by a second through `method=self.delayedDbSelect` and writes `.sigla` from a `dataRpc` on `^.regione`, which is the inverse ordering the issue's repro calls for. `test_7` covers both widgets on the single-option path and `test_9` the condition-plus-null tick.

Requesting changes for one thing introduced in this round, in the discard branch itself.

## `resetValidationError()` clears validation state this reply never set

`genro_widgets.js:4996-5003`:

```js
if(self.sourceNode.getRelativeData(vpath) != currvalue){
    delete self._lastQueryError;
    self.sourceNode.resetValidationError();
    return;
}
```

`resetValidationError` (`gnrdomsource.js:1830-1837`) nulls `_validations.error`, empties `_validations.warnings` and nulls `_validations.required`, so it wipes the node's whole Genropy validation state, `validate_notnull` included.

The guard returns before calling `setValue`, so the discarded reply has not set anything at that point. The only error it can own is the resolver's, and `delete self._lastQueryError` already covers that. Everything else the line clears was raised by another path.

Failure scenario, on a widget carrying `validate_notnull`: the fetch for `AO` is in flight and slow; the datastore moves to a value that does not satisfy the new condition, or to null, and the validation raises `error`/`required`; the stale reply then lands, the datapath no longer reads `AO`, and `resetValidationError()` drops the error belonging to the current value. The field is left empty or invalid with no error shown, and `canSave` no longer blocks.

Narrow it to the case the comment describes — read `_lastQueryError` before starting the fetch and reset only when this fetch is the one that changed it, or gate the reset on the error still being the resolver's. The `delete` can stay unconditional.

## Still open from dgpaci, both non-blocking and unchanged

- `_lastQueryError` is not cleared on the single-option branch: the skip at `gnrstores.js:613-617` returns before the `delete` at `:619`, and that branch runs no identity fetch, so a resolver error from the previous condition survives into a valid new selection and keeps the save blocked until the next list fetch. One line inside the branch. Your comment of 2 Sep says the non-blocking point was addressed, but what the head fixes is the stale-reply case, which is a different one.
- Write order: `setRelativeData` (pkey, `doTrigger=true`) still runs before `_updateSelect` (satellites), so anything bound to the pkey observes the previous row's `selected_*`. Not a regression.

Worth folding both into the same round as the point above, so this does not need a fourth pass. @dgpaci's `CHANGES_REQUESTED` is what closes the rest.

fporcari:
Both points addressed in `ab55ea7feb`
([review](https://github.com/genropy/genropy/pull/1057#pullrequestreview-5098182347)).

## The reset, narrowed

> Narrow it to the case the comment describes — read `_lastQueryError` before starting the fetch and reset only when this fetch is the one that changed it

Taken, using the store's own guarantee instead of a captured copy:
`fetchItemByIdentity` deletes `widget._lastQueryError` before issuing the call
(`gnrstores.js:619`), so a value present when `onItem` runs was installed by
**this** reply's `finalize` (`:645`). The guard is therefore exact without
carrying state across the boundary:

```js
if(self.sourceNode.getRelativeData(vpath) != currvalue){
    if(self._lastQueryError){
        delete self._lastQueryError;
        self.sourceNode.resetValidationError();
    }
    return;
}
```

Your failure scenario is closed: the slow reply for `AO` carries no resolver
error, so nothing is reset and the `error`/`required` raised for the value the
datastore moved to survives, `validate_notnull` included.

One sub-case stays open and is not mine to close here. When the stale reply
*does* carry errors, `finalize` has already called
`setValidationError({error:r.attr.errors})` — before `onItem` — and that
overwrites `_validations` wholesale, `warnings` and `required` with `undefined`.
So the current value's state is clobbered upstream of this branch, and resetting
is then the lesser evil rather than the damage. The repair belongs in `finalize`,
which should not raise a node-level validation error for a reply it cannot know
is still current; that is a change to the shared store path and outside this PR.

## The single-option branch

> `_lastQueryError` is not cleared on the single-option branch […] One line inside the branch.

Done, `delete this._lastQueryError` at the top of the branch, before
`_setValueFromItem`. One line is enough because what blocks the save is
`validate_select` returning `query_error` off that flag (`genro_frm.js:2361-2362`),
and the `validate` triggered by `_setValueFromItem` runs after the delete. You
were right that my 2 Sep comment addressed a different case: what the previous
head fixed was the stale reply, not this one.

## Write order

Left as it is, as you say it is not a regression: `setRelativeData` (pkey,
`doTrigger=true`) still runs before `_updateSelect`, so a listener bound to the
pkey observes the previous row's `selected_*` for one tick. Changing it means
writing the satellites first and the pkey last, which reverses the order every
existing listener was written against — worth its own issue if it ever bites,
not a change to slip into this round.

## Verification perimeter

`node --check` on the touched file, which is all the repo affords: no JS test
infrastructure, as @dgpaci confirmed. **Not verified live**: neither branch has a
fixture, because reproducing them needs a resolver that returns errors on a
stale identity while the datastore moves underneath — `test_8_condition_race`
orders the replies but does not make one fail. Both changes are reasoned from
the call chain above, not observed in a browser.

@dgpaci your `CHANGES_REQUESTED` is the one still standing on this PR — the four
points of your 28 Aug pass are the ones @genro confirms closed at the previous
head. Re-requesting both of you.

genro (approved):
Re-approving on `ab55ea7feb`. The guard is exact now, and I checked the store instead of taking the invariant on trust. `fetchItemByIdentity` deletes `parentSourceNode.widget._lastQueryError` before issuing the call, and the two earlier exits that still invoke `onItem` cannot fire from this call site: `!request.identity` and `isNullOrBlank(id)` are excluded by the `!isNullOrBlank(currvalue)` guard wrapping the whole branch, and the cache hit is excluded because `mixin_setCondition` calls `clearCache()` two lines above, which empties `store.cached_values`. So a truthy `_lastQueryError` in `onItem` was installed by this reply's `finalize`, and the reset undoes only what the discarded reply set. The `validate_notnull` scenario is closed.

Single-option branch: the `delete` runs before `_setValueFromItem`, so the `validate` it triggers reads the flag already cleared and `validate_select` no longer returns `query_error` (`genro_frm.js:2361-2362`).

I agree the `finalize` clobber is not this PR's to fix. `setValidationError({error:r.attr.errors})` overwrites `_validations` wholesale before `onItem` runs, so the current value's `warnings` and `required` are already gone by then and the reset adds no damage on top. Worth its own issue against the shared store path.

`node --check` clean on the touched file. Not verified live, same perimeter you state.

@dgpaci your CHANGES_REQUESTED is the one still standing here.

dgpaci (approved):
Re-reviewed `ab55ea7feb` against the whole chain again. All four blocking points from my passes are closed, the guard is real this time, and the fixtures finally reach the branch they were written for. Approving, and clearing the `CHANGES_REQUESTED` that was standing.

## What I verified, not just read

**The guard executes.** Moving it into `mixin_setCondition` (`genro_widgets.js:5000-5014`) puts the comparison in a scope that is actually reached — no shared signature changed, and `patch_setValue` is back to its two-parameter form with no leftover `valueGuard` anywhere in the file. The reply path is now genuinely closed on the issue's own repro: `setValue(null,false)` clears `_lastValue` so the fetch is not swallowed by the skip at `gnrstores.js:613`, the datastore moves to `MI` while the `AO` lookup is pending, and the late `onItem` reads `MI != AO` and returns without touching caption, `valueNode.value` or `_lastDisplayedValue`. Nothing is left for `versionpatch_11__setBlurValue` to re-commit, which was the residual bounce.

**The follow-up `setValue` is a cache hit, as claimed.** `finalize` assigns `cached_values[request.identity]` *before* invoking `onItem` (`gnrstores.js:655`), and `cache_time` is 60s (`:578`), so the inner call resolves synchronously through the cache branch. The not-found case still degrades correctly: the cached `result` is `null`, dojo's `handleFetchByIdentity` takes the else branch and flags the widget invalid, which is the documented behaviour.

**The narrowed reset is exact.** I checked the invariant rather than the reasoning: `fetchItemByIdentity` deletes `widget._lastQueryError` at `gnrstores.js:619`, and every exit that can still reach `onItem` before that line is excluded from this call site — `!request.identity` and `isNullOrBlank(id)` by the `!isNullOrBlank(currvalue)` wrapper, the cache branch by the `clearCache()` two lines above, and the `_lastValue` skip because `setValue(null,false)` has just set it to `undefined`. So a truthy flag in `onItem` was installed by this reply's `finalize` (`:645`), and the reset undoes only what the discarded reply set. My `validate_notnull` concern is closed.

**Single-option branch.** `delete this._lastQueryError` at `:4987` runs before `_setValueFromItem`, so the `validate` it triggers reads the flag already cleared and `validate_select` no longer returns `query_error` off a stale resolver error (`genro_frm.js:2346-2348`).

**The fixtures now discriminate.** `test_8_condition_race` is the first one that can fail: no `validate_notnull` on the conditioned select, so `notnull` is falsy, `oneOptionItem` stays null and the guarded branch is the one that runs; `method=self.delayedDbSelect` sleeps a second on `_id=='AO'` while the `dataRpc`/`dataController` pair writes `MI`, which forces exactly the inverse ordering the issue asks for; and `fb.div('^.sigla')` puts the datastore next to the widget so the two can be compared by eye. `test_7` covers both widget types on the single-option path, `test_9` the condition-plus-null tick. This is the coverage the previous fixture could not give.

Checks on my side: `node --check` clean on the touched file, `flake8` clean on the fixture (imports at module top level), CI green on 3.11–3.14.

## Non-blocking, for the record

**1. The `finalize` clobber deserves an issue.** I agree it is not this PR's to fix, and I want it tracked rather than left in a PR comment. `finalize` calls `setValidationError({error:r.attr.errors})` before `onItem` (`gnrstores.js:644-647`), and `setValidationError` overwrites `error`, `warnings` and `required` wholesale (`gnrdomsource.js:1838-1842`) — the last two with `undefined`. So a stale reply carrying resolver errors destroys the current value's validation state upstream of the guard, and `resetValidationError()` in the discard branch is then a normalisation of already-lost state rather than the damage. The real fix is for `finalize` not to raise a node-level validation error for a reply it cannot know is still current, which is the shared store path.

@fporcari please open that issue yourself — you have the analysis in hand and it should not be lost in this thread once the PR is merged. `[BUG]` template, and the three required sections map onto what you already wrote: Problem Description is the `finalize` → `setValidationError` ordering and the wholesale overwrite; Reproduction Steps are the interleaving you described (a stale identity reply that carries resolver errors while the datastore has moved on, e.g. an extension of `test_8_condition_race` where the delayed reply also fails the new condition); Environment is the shared `GnrStoreQuery` path, so any dbSelect with a bound `condition_*`. Worth noting in it that the discard branch added here relies on the current ordering, so whoever fixes `finalize` should revisit `mixin_setCondition:5006-5011` at the same time. Not a condition of this approval.

**2. A narrow residual of the same class, if two condition changes overlap.** The inner `setValue(currvalue,false)` is synchronous only while the cache the outer fetch warmed is still there. A second `setCondition` calls `clearCache()`, so if the first reply lands after that, the inner `setValue` goes remote — and *that* reply travels dojo's own path, with no guard. It can then write a stale caption. Bounded and much narrower than what this PR fixes, and hardening it means using the `item` that `onItem` already receives instead of re-entering `setValue`, which would re-split the code by widget type (`_callbackSetLabel` is FilteringSelect-only) — the thing we just got rid of. Not worth it now; worth knowing it exists.

**3. `dbComboBox` now issues an identity lookup on the guarded branch** where the plain `setValue` previously did none, so a value outside the new condition raises a validation error on that widget too. Consistent with `dbSelect` and arguably the point of `setCondition`; noting it because it is a visible behaviour change on that widget beyond the one in the PR body.

Write order stays as it is — agreed, not a regression, and reversing it changes the contract every existing listener was written against.

cgabriel (approved):
Verified throught the provided gnrcore's test.

PR #1058, opened 2026-08-07
fix(app): require every loaded package to be declared in instanceconfig

### LINKED ISSUE #1006: [BUG] gnr app checkdep ignores packages declared via required_packages()

fporcari:
Fixes #1006

Reworked from option A to option B, the one @cgabriel settled on in [the issue's decision](https://github.com/genropy/genropy/issues/1006#issuecomment-5422805114). The previous implementation — expanding the `required_packages()` closure inside `check_package_dependencies()` — has been dropped, along with its test file.

## Problem

`gnr app checkdep` reports `All good!` on instances that are missing Python dependencies. `GnrApp.check_package_dependencies()` (`gnrpy/gnr/app/gnrapp.py`) iterates `self.config['packages']` — the `instanceconfig.xml` list — and resolves each `requirements.txt` from that list alone. A package pulled in through `Package.required_packages()` rather than declared there is never seen, so its requirements are never checked.

The repro on the issue: an instance declaring `erpyready:er_core`, whose `main.py` returns `['gnrcore:email']`, never sees `mail-parser` / `chardet`.

## Solution

`required_packages()` is a second, implicit way of listing an instance's packages, and that is the root of the problem: two lists that can drift apart, one of which `checkdep` cannot see. Rather than teaching the check to read both, this makes the second list illegal.

`addPackage()` now refuses to load a required package that is not declared in the packages section:

```
GnrUndeclaredPackageException: Package 'biz' requires 'gnrcore:adm', which is
not declared in the packages section of instanceconfig.xml. Declare it there:
its python dependencies are not checked otherwise.
```

The packages section stays the single list `checkdep` has to trust, and it is now guaranteed complete. Roughly ten lines: the guard, a `declared_packages` property that normalises the `project:pkgid` prefix, and the exception class.

## What this does NOT do, deliberately

**`checkdep` itself remains blind until the instanceconfigs are corrected.** `GnrApp.init()` returns in `checkdepcli` mode before the `addPackage()` loop where the guard lives, so the guard never fires under `checkdep`. This issue is therefore fixed transitively: the boot error is what forces the instanceconfig to be completed, and once it is, `checkdep` sees everything because there is nothing left outside the packages section. That is the accepted consequence of this option, not an oversight — it was the first of the two things flagged before the choice was made.

## Upgrade note

**An instance whose `instanceconfig.xml` is incomplete will not start after this.** That is the point of the option — it is what forces the correction — but it makes this an expensive change to roll out, and the issue's own repro is exactly such an instance. The error names both the missing package and the one that requires it, so the fix is always a one-line addition to the packages section.

No in-repo instance is affected: all instanceconfigs under `projects/` already declare their full closure, and the full suite confirms it.

## Tests

`gnrpy/tests/app/test_required_packages_declared.py` — four tests on a real `GnrApp` over a real instanceconfig written per case, no mocks. `gnrcore:biz` requires `gnrcore:adm` requires `gnrcore:sys`, so the direct and the transitive case are both covered, plus the complete-declaration boot and the prefix normalisation.

Ratchet checked: with the guard disabled, the two refusal tests fail; with it, all four pass.

Full suite on the merged tree: 2121 passed, 69 skipped. Also realigned with `develop` in this push — the branch was three weeks behind.


gnrpy/RELEASE_NOTES.rst
cgabriel: Why the update on this file?

cgabriel (changes requested):

gnrpy/RELEASE_NOTES.rst
fporcari: Removed. Same in #1089 and #1057.

genro (approved):
Option B as decided on #1006, implemented where it belongs: the guard sits in `addPackage()` (`gnrapp.py:1081-1086`), `declared_packages` normalises the `project:pkgid` prefix against the keys `load_instance_config` already rekeys on `pkgcode or label`, and the exception names both the missing package and the one requiring it. All seven in-repo instanceconfigs declare their full closure (checked). Tests run a real `GnrApp` over an instanceconfig written per case. The `RELEASE_NOTES.rst` thread is settled: the net diff no longer touches the file.

One non-blocking note, pre-existing but made relevant by this change: order inside `<packages>` now matters. When the requirer precedes the required package (`biz` before `adm`), `adm` is loaded from the `required_packages()` loop with `pkgattrs=dict(readOnlyAttrs)` (:1087); when `init()` later reaches the `<gnrcore_adm .../>` node, `addPackage` returns early on `pkgid in self.packages` (:1075) and whatever attributes that node carries are dropped. Someone fixing an incomplete instanceconfig by appending the missing line at the end lands exactly there. `test_complete_declaration_boots` declares `biz, adm, sys` in that order and passes because it does not look at attributes. Small fix, here or as a follow-up: in the required loop, look up the declared node's attributes and pass those instead of the bare `readOnlyAttrs`.

fporcari:
@cgabriel ready for your re-read — nothing here is waiting on anyone else any more.

**The open question to @dgpaci is closed.** On [#1006 (27/08)](https://github.com/genropy/genropy/issues/1006#issuecomment-5437691854) I asked him to say so on this PR if he thought B was the wrong trade. Three weeks, no objection, and @genro approved the current head on 03/09.

**Your changes-requested is on a superseded commit.** It sits on `7d096dba0`; the `RELEASE_NOTES.rst` hunk you asked about was dropped in `0d9d757b3` two hours later, and that thread is resolved — the net diff no longer touches the file.

**Your own open question, answered:** making `checkdep` itself fail on an incomplete instanceconfig. I offered to add it here and had no answer either way, so I am leaving it out: this PR stays the ten-line guard. Say the word and it becomes a follow-up issue.

@genro's note on `pkgattrs` ordering inside the `required_packages()` loop is pre-existing and I would rather take it as a follow-up issue than widen this diff.

cgabriel (changes requested):
Apparently the implementation doesn't work. Tested with 'swgarden' instance from softwell-erpy. This instance has only 1 package in the  instanceconfig xml (softwell_garden), and softwell_garden has 17 packages listed in the package's main.py's required_packages(). But no error is raised, while it was expected to raise an 'GnrUndeclaredPackageException'. Simple check:

```
cgabriel@marvin:~/genropy$ gnr app checkdep -v swgarden
Required dependencies are
 
Checking for installed dependencies
All good!
```

Several dependencies (like 'mailparser', used in gnrcore:email, listed in the required_package() of softwell_garden) are not showing up, since these modules are not listed in instanceconfig.xml. And since the exception is not raised as expected, the image build completes, but it fails at runtime for missing dependencies.

fporcari:
> Apparently the implementation doesn't work. Tested with 'swgarden' instance from softwell-erpy.

— [review of 2026-09-17](https://github.com/genropy/genropy/pull/1058#pullrequestreview-5234124263)

Done, in `b9e7756`: `checkdep` raises it too now.

The guard lived in `addPackage()`, and `init()` returns in `checkdepcli` mode at `gnrapp.py:1016`, before that loop runs — so `gnr app checkdep` never reached it. `check_declared_packages()` applies the same rule to the declared packages alone and is called from that same branch just before the return. Once an instanceconfig declares its closure the two sets are the same set, so nothing else had to move.

A package whose `main.py` cannot be imported is logged and skipped rather than aborting the check: in a clean build environment it may be missing the very dependency the check is there to find, and reporting nothing on that package beats reporting it wrongly.

On swgarden the same command should now stop on the first of the 17:

```
GnrUndeclaredPackageException: Package 'softwell_garden' requires 'gnrcore:email',
which is not declared in the packages section of instanceconfig.xml. Declare it
there: its python dependencies are not checked otherwise.
```

Two new tests in `test_required_packages_declared.py`, both through a real `GnrApp` over an instanceconfig written per case: `checkdep` on an incomplete declaration raises, on a complete one passes and loads nothing. Ratchet checked — with the call removed the first one fails.

`tests/app/test_required_packages_declared.py` → 6 passed. `gnrpy/tests` minus `tests/sql` → 1398 passed, 9 skipped (no local PostgreSQL). flake8 clean on both files.

Not verified: I have no swgarden instance here, so the message above is what the code now produces on an incomplete instanceconfig, not your run reproduced.

@genro this push dismissed your approval, and nothing you approved has changed — the delta is the `checkdep` half only.

cgabriel (approved):
All is fine now, thank you!

genro (approved):
Approved on `b9e7756494`.

Making the second list illegal instead of teaching the check to read it is the right shape for this, and it is the option settled on the issue. The guard is ten lines and the error names both packages, which is what keeps the upgrade cost to one line per instance.

Checked the piece that would have been easy to get wrong. `declared_packages` normalises with `k.split(':')[-1]`, which only works because `load_instance_config()` re-keys the packages Bag on `pkgcode` at `gnrapp.py:951` rather than on the XML tag. The keys really are `gnrcore:sys`, so the prefix strip lands on `sys` and not on `gnrcore_sys`.

`check_declared_packages()` closes the hole the first implementation left: `init()` returns in `checkdepcli` mode before the `addPackage()` loop, so without it the guard was unreachable under the very command the issue is about. Skipping a package whose `main.py` cannot be imported, with a warning, is the right trade in a build environment — aborting there would fail on the dependency the check exists to find.

The breaking half is deliberate and documented, and it was exercised against a real instance in the previous round, so I have nothing to add to it.

Issue #991, opened 2026-07-15
[FEATURE] htmlpage menu entry to display standalone HTML documents


dgpaci:
## Summary

Add a new menu entry type `htmlpage` to the application menu API (`gnrpy/gnr/web/gnrmenu.py`), to display a standalone static HTML document inside the application frame:

```python
branch.htmlpage('Monthly report', source='<html_filepath>', tags='SOMETAG')
```

The entry would open a generic webpage shipped with genropy that renders the HTML file in an iframe (via `srcdoc`), keeping the application chrome (menu, header) around it.

## Problem or Need

Projects often produce standalone HTML documents (meeting reports, generated documentation, dashboards exported as static HTML) that should be reachable from the application menu with the usual tag-based visibility rules.

Today there is no built-in way to do this: each project has to implement its own custom webpage. For example in `anaci_app` we had to write:

- a custom webpage (`report_viewer.py`) whose only job is reading an HTML file server-side and injecting it into an `htmliframe` through `srcdoc`;
- a lib helper that builds `branch.webpage(...)` entries pointing to that custom viewer with `url_filename=...`.

This is boilerplate that every project reimplements the same way, while the menu API already offers first-class entries for the other cases (`webpage`, `thpage`, `tableBranch`, `lookupBranch`, ...).

## Proposed Solution

1. **New `htmlpage` method** on the menu struct class (next to `Menu.webpage`, `gnrpy/gnr/web/gnrmenu.py:108`), with the same ergonomics as the other entries:

   ```python
   def htmlpage(self, label, source=None, tags='', **kwargs):
       return self.child('htmlpage', label=label, source=source, tags=tags, **kwargs)
   ```

   - `label`: menu label, as for every other entry
   - `source`: path of the `.html` file — ideally a storage-service path (`site:`, `pkg:<pkgname>/...`, `home:`) or an absolute path
   - `tags`: usual tag-based visibility

2. **New `nodeType_htmlpage` resolver** in `MenuResolver` (next to `nodeType_webpage`, `gnrpy/gnr/web/gnrmenu.py:458`) that maps the entry to the built-in viewer webpage passing `source` as url parameter.

3. **Built-in viewer webpage** shipped with genropy (e.g. under the `sys` package webpages) that:
   - resolves `source` through the storage service / normalizes the path;
   - reads the file content server-side;
   - renders it with `htmliframe` + `srcdoc` (so the document keeps its own standalone styles without leaking into the app CSS and vice versa);
[...]

PR #1059, opened 2026-08-07
feat(menu): add docpage entries for public stored documents

fporcari:
## Summary

Adds `MenuStruct.docpage()` for opening a public document from a registered storage service inside the application frame.

This is intentionally a resolver-only feature: the menu resolves a storage path to a stable public URL and loads that document directly. It does not add an authenticated viewer page.

## API

```python
branch.docpage(
    'Monthly report',
    source='pkg:mypkg/docs/report.html',
    tags='REPORT_VIEW',
)
```

`source` must be a registered storage path. `tags` controls menu visibility only; it does not protect the resolved document URL.

## Behavior

- Resolves the source with `StorageNode.public_url(nocache=True)`, avoiding expiring signed menu links and stale overwritten documents.
- Hides only the affected menu entry and logs a warning when the source is absent, the storage service fails, or no public URL is available.
- Ignores inherited or entry-level `externalSite` URL prefixing because storage returns an absolute URL.
- Marks the iframe as `nonGenroContent`, so the frame does not wait for Genro bootstrap, keeps the menu label as its title, and closes without dereferencing `contentWindow.genro`.
- Shares page-attribute finalization with ordinary `webpage` entries.
- Uses a docstring to state the public-content security boundary.

## Security boundary

A `docpage` is suitable only for public documents. Storage/static URLs are served without page-level authentication, so hiding the menu entry with `tags` does not make the document confidential. Confidential or per-user documents still require an authenticated viewer endpoint.

## Verification

- Full suite: **2117 passed, 69 skipped**.
- Commit-hook suite: **2086 passed, 69 skipped**.
- Web suite excluding the daemon-only module: **285 passed, 47 skipped**.
- Resolver tests: **10 passed**, including a real symbolic `pkg:` storage node, both forms of `externalSite`, inherited `aux_instance`, missing source, unavailable service, empty public URL, and the framework's real tag-permission logic.
- `flake8` clean on all changed Python files.
- `node --check` clean on `frameindex.js`.
- The test package contains localized working and deliberate click-time-404 entries.
- Not verified in this update: a live authenticated browser session.

## Deliberately out of scope

- A built-in authenticated viewer and friendly missing-file page.
- The folder-to-menu variant.
- Confidential document delivery.

Refs #991


gnrpy/gnr/web/gnrmenu.py
```diff
@@ -468,6 +472,15 @@ def nodeType_webpage(self,node):
         self.checkContextParameters(attributes)
         return None,attributes
 
+    def nodeType_htmlpage(self,node):
+        attributes = dict(node.attr)
+        source = attributes.pop('source',None)
+        attributes['webpage'] = self._page.site.storageNode(source).url() if source else None
+        aux_instance = attributes.get('aux_instance') or self.aux_instance
+        attributes['url_aux_instance'] = aux_instance
+        self.checkContextParameters(attributes)
```
dgpaci: **Blocking — `externalSite` produces a concatenated, broken URL.**

`checkContextParameters` (`:449-453`) unconditionally prefixes `attributes['webpage']` with the external site base:

```python
attributes['webpage'] = f"{externalSite}{attributes['webpage']}"
```

`nodeType_webpage` is safe there because its `webpage` is always a site-relative path. `nodeType_htmlpage` puts an **absolute** URL in it, so with an `externalSite` set — on the node, or inherited from any parent branch resolver via `self.externalSite` (`:492`, `:561`, `:606`, `:620`) — the result is:

```
https://ext.examplehttps://host.test/_pkg/demo/r.html
```

which the iframe loads and 404s on. Same code path also yields `https://ext.exampleNone` when `source` is absent, since `webpage` is `None` at that point and the f-string stringifies it.

Either skip the external-site prefix when `webpage` is already absolute (`urllib.parse.urlsplit(webpage).scheme`), or guard it inside `checkContextParameters` so every future caller benefits. Worth a regression test with `externalSite` set.

gnrpy/gnr/web/gnrmenu.py
```diff
@@ -468,6 +472,15 @@ def nodeType_webpage(self,node):
         self.checkContextParameters(attributes)
         return None,attributes
 
+    def nodeType_htmlpage(self,node):
+        attributes = dict(node.attr)
+        source = attributes.pop('source',None)
+        attributes['webpage'] = self._page.site.storageNode(source).url() if source else None
```
dgpaci: **Blocking — unguarded deref of a nullable return, and a storage error aborts the whole menu build.**

`makeNode()` is documented to return "`LegacyStorageNode` instance, or `None` if service unavailable" and does `if not service: return None` (`gnrwsgisite_proxy/gnrstoragehandler.py:504-542`), so a `source` naming an unresolvable service gives `AttributeError: 'NoneType' object has no attribute 'url'`. A service row whose `implementation` is unknown raises `GnrException` from `addService` (`gnr/lib/services/__init__.py`).

`MenuResolver.load()` (`:312-317`) only catches `NotAllowedException`, so either case propagates out of the resolver: **one typo'd `source` in one menu entry breaks the whole application menu, for every user, on every page load** — not just that entry.

Suggest resolving defensively and degrading to a hidden or dead entry rather than an exception:

```python
sn = self._page.site.storageNode(source) if source else None
attributes['webpage'] = sn.public_url() if sn else None
```

plus a `pkglog` warning when `source` was given but did not resolve, so the misconfiguration is diagnosable instead of silent. Raising `NotAllowedException` to hide the entry — the way `nodeType_lookupPage:507` does — is the other consistent option, if you prefer hiding over showing a broken link.

dgpaci: **Blocking — this should be `public_url()`, not `url()`.**

On a signing storage service, `StorageService.url()` calls `generate_presigned_url` with `ExpiresIn=self.url_expiration` (default 3600s — `services/storage/aws_s3.py:123`, `:321-336`). The menu is built once and `MenuResolver` has `classKwargs cacheTime=-1`, so a user who leaves the app open for more than an hour and then clicks the entry gets an S3 `SignatureExpired` error.

`public_url()` exists precisely for "a stable public url (no signature, no expiration)" (`gnr/lib/services/storage.py:369`, `:551`), and is what a long-lived menu link needs. It is also the direction already taken in #990 / #1085. Secondary cost of `url()`: building the boto client and signing on **every** menu build, for every user.

Minor, same line: consider `nocache=True`. `internal_url()` and the symbolic `url()` append the file mtime as a cache-buster (`storage.py:517-533`). The motivating use case is runtime-generated reports overwritten at a fixed storage path — without it, a browser holding a cached copy keeps showing the stale document.

gnrpy/gnr/web/gnrmenu.py
```diff
@@ -468,6 +472,15 @@ def nodeType_webpage(self,node):
         self.checkContextParameters(attributes)
         return None,attributes
 
+    def nodeType_htmlpage(self,node):
```
dgpaci: **Blocking — the produced frame is non-Genro but nothing marks it as such.**

Every client-side guard for non-Genro iframe content is keyed on `externalSite`. An `htmlpage` frame is non-Genro **without** `externalSite`, which is a third case `frameindex.js` does not handle. Three concrete consequences:

1. `onSelectedFrame` → `stackSourceNode.watch('pageReady', ...)` (`frameindex.js:245-251`) returns true only if `iframe.sourceNode.attr.externalSite` is set, or `contentWindow.genro._pageStarted` is true. A static document satisfies neither, so `watch()`'s `setInterval` (`gnrdomsource.js:1758-1775`) polls every 200 ms for as long as that frame stays selected, and the `onSelectedFrame` publish plus `contentWindow.genro.resizeAll()` never run.
2. `pane_kw.onClosingCb` (`frameindex.js:155`) throws — see the review body.
3. The multipage tab caption stays on the placeholder `'Html Page...'` forever: `createIframePage` seeds it with `label+'...'` and binds `'^...title'` (`:143`, `:146`), and the only writer of that path is the Genro bootstrap injected into Genro pages (`resources/common/public.py:53-55`, via `parentIframeSourceNode.attr.multipage_childpath`). `pageStarted` never fires either, so `iframeattr.onStarted` / `checkStartsArgs` never run.

Setting a flag here (reusing the `externalSite` short-circuits, or an explicit `staticContent`/`nonGenroContent` attribute honoured by the three call sites above) is what makes the existing machinery treat the frame correctly.

gnrpy/gnr/web/gnrmenu.py
```diff
@@ -468,6 +472,15 @@ def nodeType_webpage(self,node):
...
+        attributes = dict(node.attr)
+        source = attributes.pop('source',None)
+        attributes['webpage'] = self._page.site.storageNode(source).url() if source else None
+        aux_instance = attributes.get('aux_instance') or self.aux_instance
+        attributes['url_aux_instance'] = aux_instance
+        self.checkContextParameters(attributes)
+        return None,attributes
```
dgpaci: **Duplication that has already diverged.** These four lines duplicate `nodeType_webpage:469-472` verbatim, while the copy silently drops `basepath` handling and the `multipage` default — and it inherits the `externalSite` defect precisely because the shared tail was copied instead of factored.

A shared `_finalizePageAttributes(attributes)` helper (aux_instance resolution + `url_aux_instance` + `checkContextParameters`), or teaching the existing `webpage` tag to accept `source=` and resolve it, keeps the two paths from drifting further and matches how `lookupPage`/`thpage` already share their tail shape.

gnrpy/gnr/web/gnrmenu.py
```diff
@@ -109,6 +109,10 @@ def webpage(self, label,filepath=None,tags='',multipage=None, **kwargs):
         return self.child('webpage',label=label,multipage=multipage,tags=tags,
                         filepath=filepath,_returnStruct=False,**kwargs)
 
+    def htmlpage(self, label,source=None,tags='', **kwargs):
```
dgpaci: New permanent public menu API with no docstring. The `tags` argument reads at every call site like an access control on the **document**, while it only filters the **menu node** in `allowedNode` (`:416-434`): the resolved URL is dispatched by the static/storage route before any page object exists (`gnrwsgisite.py:1310-1325` → `gnrstatichandler.py:49-58`), with no authentication at all.

So

```python
branch.htmlpage('Payslip', source='user:payslip.html', tags='HR')
```

reads as an HR-only document and is in fact readable by any unauthenticated caller who guesses the URL. Your PR description explains this at length, but nothing in the shipped code carries the warning to the next developer. A docstring saying plainly that `tags` hides the entry and does **not** protect the document, and that this tag is therefore unsuitable for confidential content, would close the gap.

gnrpy/tests/web/gnrmenu_test.py
```diff
@@ -1 +1,152 @@
...
+
+    def url(self):
+        return self._url
+
+
+class _StubSite:
+    def __init__(self, urls):
+        self._urls = urls
+
+    def storageNode(self, source):
```
dgpaci: The stub replaces the one line of logic the tests are meant to cover: `storageNode` returns a canned string from `self._urls`, so `test_load_resolves_source_to_storage_node_url` (`:103-113`) asserts that `webpage` equals the string the stub was seeded with. It would still pass if the real resolution were wrong, returned `''`, or returned `None`.

Same for the tag tests: `_StubApp.checkResourcePermission` (`:35-40`) re-implements tag matching as a set intersection, so `:132-149` assert the stub's semantics rather than the framework's.

More importantly, nothing covers the four inputs that actually break `nodeType_htmlpage`: `externalSite` set, `aux_instance` set, an unresolvable service, and a `source` with no `service:` prefix (which routes to the `_raw_` service, whose `url()` returns `''` — `resources/common/services/storage/raw.py`; `temp:` returns `''` too via `symbolic.url_temp`). Note `''` is *not* stripped by `Bag.setItem` the way `None` is, so `test_load_without_source_drops_webpage_attribute` (`:116-129`) does not cover it — and even in the documented `None` case the client falls through `kw.webpage || kw.file || kw.filepath` to `undefined` and loads `undefined?windowTitle=Report` in the iframe, with no error or log anywhere.

Using a real `storageNode` for at least the `pkg:` case (which needs no DB and no daemon) would make the central assertion non-tautological.

projects/gnrcore/packages/test/menu.py
```diff
@@ -16,4 +16,7 @@ def config(self,root,**kwargs):
         tests.directoryBranch("Services", folder="services")
         tests.directoryBranch("Tools", folder="tools")
         tests.directoryBranch("Websocket", folder="websocket")
+        tests.htmlpage("Html Page", source="pkg:test/resources/test.html", tags="_DEV_")
+        tests.htmlpage("Html Page (missing file)", source="pkg:test/resources/does_not_exist.html",
+                        tags="_DEV_")
```
dgpaci: Three small things on the fixture entries:

1. **Missing `!!` marker and localization.** The branch on `:4` is written `'!!Tests'`; these two labels have no marker, so they can never be translated, and `projects/gnrcore/packages/test/localization.xml` was not updated.
2. **`tags="_DEV_"` is redundant.** `root.branch("!!Tests", tags="_DEV_")` (`:4`) already gates the whole subtree, and none of the 14 `directoryBranch` siblings repeats it. Repeating it here suggests the leaf gate does something the branch gate does not, which invites someone to change one and not the other.
3. **The `does_not_exist.html` entry needs a comment.** Since the PR adds no existence check or placeholder, it resolves to a bare 404 rendered inside the app iframe, permanently. Without a comment marking it a deliberate missing-file fixture, it is indistinguishable from a bug to anyone browsing the test package.

dgpaci (changes requested):
Thanks for the very thorough write-up — the "this already works today" analysis and the security section are exactly the right framing, and the `nodeType_lookupPage` precedent is the right one to follow. My comments are on the resolver itself: as written it has four blocking issues, three of which come from the fact that an `htmlpage` `webpage` is an **absolute URL to non-Genro content**, while every consumer downstream (`checkContextParameters`, `frameindex.js`) assumes either a site-relative Genro page path or an `externalSite`-flagged frame.

### Blocking

1. **`externalSite` mangles the URL** — inline on `gnrmenu.py:481`.
2. **`storageNode(source).url()` can `AttributeError`, and any storage exception kills the whole menu** — inline on `gnrmenu.py:477-478`.
3. **`public_url()` is the correct call, not `url()`** — inline on `gnrmenu.py:478`.
4. **The frame is non-Genro but carries no marker**, so the client guards keyed on `externalSite` never fire — inline on `gnrmenu.py:475`.

### One finding outside the diff, relevant to point 4

`projects/gnrcore/packages/adm/resources/frameindex.js:155`:

```js
var iframeSourceNode = this.getValue().getNode('iframecontainer.iframenode');
return iframeSourceNode.domNode.contentWindow.genro.checkBeforeUnload();
```

`pane_kw.onClosingCb` dereferences `contentWindow.genro` unconditionally, while `closeRootFramPage` (`:415-417`) guards it with `if(n.contentWindow.genro)` (and short-circuits on `externalSite` first). Open an `htmlpage` entry and click the stack-button closer: `genro_components.js:5079-5080` invokes `onClosingCb` through `funcApply`, which does not swallow exceptions, so it raises `TypeError: Cannot read properties of undefined (reading 'checkBeforeUnload')` and the pane never closes. This PR is the feature whose whole purpose is putting non-Genro content in that iframe, so aligning that guard with `closeRootFramPage` belongs here rather than in a follow-up.

### On the "Decisions required" section

- **Viewer page**: resolver-only is fine for the base case, but then the `tags`/no-auth asymmetry has to be documented in the code, not only in the PR description (inline on `gnrmenu.py:112`). And the answer to your own question about the `anaci_app` reports follows from it: with the static dispatcher performing no authentication, this API is not suitable for confidential per-tenant documents, whatever `tags` says. That constraint should be stated in the docstring so the next caller does not have to rediscover it.
- **The name**: I agree with your own concern. `htmlpage` collides visually with `serve_htmlPage` / `GnrHtmlPage` / `html_pages/`, which are the *other*, resource-based mechanism. `docpage` reads better and describes what it actually does (open a stored document). Worth settling before merge, since this is permanent public menu API.
- **Missing-file behaviour**: a plain 404 at click time is an acceptable v1, but then the fixture in `packages/test/menu.py` needs a comment saying so (inline).
- **Folder variant**: agreed, out of scope here.

### Tests

The doubles-over-`GnrDummySite` choice is right, and matching the `gnrwebstruct_test.py` idiom is the correct precedent. The problem is narrower: the stub replaces the one line of new logic, so the central assertion is tautological, and none of the four inputs that actually break the resolver are exercised. Details inline on `gnrmenu_test.py:30`.

fporcari:
> As written it has four blocking issues [...] while every consumer downstream assumes either a site-relative Genro page or an external frame.
>
> — [review of 20 Aug](https://github.com/genropy/genropy/pull/1059#pullrequestreview-4983863453)

Addressed in `e28302fab` after merging current `develop`:

- renamed the public API from `htmlpage` to `docpage`;
- switched to `public_url(nocache=True)` and made storage failures hide only the affected entry with a warning;
- prevented both inherited and entry-level `externalSite` from prefixing the absolute storage URL;
- added the explicit `nonGenroContent` frame path, including safe same-origin and cross-origin closing, immediate title, and no Genro-startup polling;
- factored the page-attribute finalization shared with `webpage`;
- documented that `tags` does not protect the document;
- replaced the tautological storage and permission doubles with a real symbolic `pkg:` node and the framework permission implementation;
- covered missing, invalid, empty-URL, `externalSite`, and `aux_instance` inputs;
- corrected and localized the test-menu fixtures.

Verification: full suite **2117 passed, 69 skipped**; commit hook **2086 passed, 69 skipped**; flake8 and JavaScript syntax checks clean.

dgpaci (dismissed):
LGTM. All four blocking points and every secondary observation from my 20 Aug review are addressed in `e28302fab`, and I checked them against the code rather than the summary:

- **`externalSite`** — the `absolute_webpage` short-circuit in `checkContextParameters` plus the `pop` on the entry covers both the inherited and the entry-level case, and both are under test.
- **Storage resolution** — `NotAllowedException` is caught per node in `load()`, so an unresolvable `source` now hides only its own entry, with a warning that makes the misconfiguration diagnosable.
- **`public_url(nocache=True)`** — correct call, and the `symbolic` service does honour `nocache`.
- **`nonGenroContent`** — the flag really reaches the client (`kw = objectUpdate({...}, treenode.attr)` in `frameindex.js`), and all the guards keyed on `externalSite` now accept it: iframe attr, immediate title, `onClosingCb`, the `pageReady` watch, `closeRootFramPage`. `buildContextUrl` only rewrites urls starting with `/`, so the absolute document url survives untouched on the client too.
- **`_finalizePageAttributes`** — shared tail, with `basepath` and the `multipage` default preserved on the `webpage` path.
- The `docpage` rename, the docstring stating that `tags` does not protect the document, and the three fixture fixes in `packages/test/menu.py` are all in.
- **Tests** — a real `symbolic` `pkg:` `StorageNode` and the framework's own `checkResourcePermission`, plus the four inputs that were previously uncovered. 10 passed here on `e28302fab`, flake8 clean on `gnrmenu.py`.

Approving. Four non-blocking notes, before or after merge as you prefer:

1. `packages/test/localization.xml` carries only `base` + `en` for the two new labels, while the sibling entries have `de`/`fr`/`it`. The Italian is worth adding, since the label is user visible.
2. `nocache=True` is a no-op on `aws_s3`: its `public_url()` builds the url from `internal_path` and ignores `kwargs`, so the cache-buster only applies to `symbolic`/local services. The motivating case — a report regenerated at a fixed path — is still exposed to browser caching when the document lives on S3. Better as a follow-up issue on the service than as a change here.
3. The `except Exception` around the resolution also turns a genuine bug inside `public_url` into a silently missing menu entry. The warning keeps it diagnosable, so I am fine with the trade-off; noting it in case you prefer narrowing the catch.
4. Cosmetic: the label still reads "HTML document" while the tag is now `docpage`, and `finalizePageUrl` appends `windowTitle=` to the document url — harmless for static serving, just visible in the address bar.

projects/gnrcore/packages/test/menu.py
dgpaci: Points 2 and 3 are done, and point 1 is done for the `!!` markers. The only leftover is the localization itself: `packages/test/localization.xml` has `base` + `en` for the two entries, while every sibling carries `de`/`fr`/`it`. Adding at least the Italian closes this thread — leaving it open as the reminder, not as a blocker on the approval.

fporcari: > The only leftover is the localization itself: `packages/test/localization.xml` has `base` + `en` for the two entries, while every sibling carries `de`/`fr`/`it`. Adding at least the Italian closes this thread

Done in 2c19097, all three rather than only the Italian, so the two entries look like
their siblings:

```
de="HTML-Dokument"                 fr="Document HTML"                 it="Documento HTML"
de="HTML-Dokument (fehlende Datei)" fr="Document HTML (fichier manquant)" it="Documento HTML (file mancante)"
```

Heads-up: `dismiss_stale_reviews` is on for `develop`, so this push dropped your approval
of the 21st. Nothing you approved has moved — the only change is those two lines of
`localization.xml`.

gnrpy/gnr/web/gnrmenu.py
```diff
@@ -32,6 +33,8 @@
 from gnr.core.gnrstring import slugify
 from gnr.core.gnrdecorator import extract_kwargs
 
+logger = logging.getLogger("gnr.web.gnrmenu")
```
dgpaci: **Use the package logger.**

```python
import logging
logger = logging.getLogger("gnr.web.gnrmenu")
```

`gnr/web/__init__.py:2` already defines `logger = logging.getLogger("gnr.web")`, and it is what **28 modules** under `gnr/web/` import - including `gnrwebpage_proxy/menuproxy.py`, the menu's own neighbour. The single other module that rolls its own is `gnrwebpage_proxy/apphandler/related.py`.

The runtime effect is nil, since `gnr.web.gnrmenu` is a child of `gnr.web` and inherits its handlers. The problem is that the shared logger exists precisely so that individual modules do not do this, and an `import logging` at the top of a file that had none is the first thing a reader notices. Two lines:

```python
from gnr.web import logger
```

and drop `import logging` and the `getLogger` call.

gnrpy/gnr/web/gnrmenu.py
```diff
@@ -109,6 +112,15 @@ def webpage(self, label,filepath=None,tags='',multipage=None, **kwargs):
         return self.child('webpage',label=label,multipage=multipage,tags=tags,
                         filepath=filepath,_returnStruct=False,**kwargs)
 
+    def docpage(self, label,source=None,tags='', **kwargs):
+        """Add a public stored document to the menu.
```
dgpaci: **`source` already accepts an absolute external URL - the docstring should say so.**

This docstring says "public **stored** document", and the PR description says "`source` must be a registered storage path". Both are wrong today: an `http(s)` URL works, and works correctly.

`_adapt_path` (`gnrwsgisite_proxy/gnrstoragehandler.py:489`) routes it to the built-in `_http_` service, which is in `DEFAULT_STORAGE_CONFIGS` and therefore always present with no siteconfig entry. The `http` service does not override `public_url()`, so it falls back to the base implementation, which delegates to `url()` - and `http.Service.url()` is `return args[0]`. Verified by running it:

```
'https://www.example.test/report.html' -> ('_http_', 'https://www.example.test/report.html')
'pkg:test/resources/test.html'         -> ('pkg', 'test/resources/test.html')
url()               : 'https://www.example.test/report.html'
public_url(nocache) : 'https://www.example.test/report.html'
```

Your own `absolute_webpage=True` then skips both `basepath` and the `externalSite` prefix, so the URL reaches the iframe intact, and `nonGenroContent=True` is exactly the right flag for a remote page. So the feature works - it just works by accident of a legacy branch in `_adapt_path`, is covered by none of the 10 tests, and is documented as impossible. That is the profile of a capability that dies silently in the next refactor of `source` validation.

Two requests:

1. State it in the docstring: `source` is a storage path **or** an absolute `http(s)` URL. Worth adding that with a remote URL the delivery caveats are the remote server's, not ours - `X-Frame-Options`/`frame-ancestors` yield a blank frame with no warning anywhere, since resolution already succeeded, and an `http` URL under an `https` app is blocked as mixed content. `nocache=True` is also silently ignored by the `http` service, which is harmless here but worth not promising.
2. One test for the `http` case, alongside the existing ones.

If instead you do not want to support this, the alternative is an explicit rejection in `nodeType_docpage` rather than the current silent pass-through - but I would keep it, it is the shortest path to a one-off external document and it composes with `nonGenroContent` better than `webpage` + a registered `externalSites` entry does.

dgpaci (changes requested):
Re-reviewing after `dismiss_stale_reviews` dropped my approval on `2c19097f2`.

First, the mechanical part: the diff between `e28302fab` and `2c19097f2` is the two `localization.xml` lines and nothing else, so your heads-up is accurate. The `de`/`fr`/`it` entries match the `!!` labels in `menu.py`, follow the sibling attribute order, and close note 1 of my 21 Aug review. CI is green on 3.11-3.14 and every thread is resolved.

I am requesting changes on two points I should have caught in the first pass rather than now. Neither touches the logic or the tests you already wrote, and both fit in one push.

### 1. The module-level logger is invented

Inline on `gnrmenu.py:36`. `gnr/web/__init__.py` already exposes the package logger, and 28 modules under `gnr/web/` import it. This file should not be the 29th exception.

### 2. `docpage` accepts external URLs, and the docstring says the opposite

Inline on `gnrmenu.py:116`. `source` already resolves an absolute `http(s)` URL through the built-in `_http_` service - I verified it by running the code, not by reading it. That is a genuinely useful capability, and right now it is documented as impossible.

Everything else stands as approved on the 21st: the `externalSite` short-circuit, the per-node `NotAllowedException`, `public_url(nocache=True)`, the `nonGenroContent` contract across the four `frameindex.js` call sites, and the shared `_finalizePageAttributes` tail. The three remaining non-blocking notes from that review (`nocache=True` being a no-op on `aws_s3`, the broad `except Exception`, the cosmetic label/`windowTitle=`) are unchanged and still follow-up material.

gnrpy/gnr/web/gnrmenu.py
fporcari: Done in `980f958f8`. `import logging` and the module-level `getLogger` are gone; `gnrmenu.py` now does `from gnr.web import logger` like the other 28 modules under `gnr/web/`. The two `logger.warning` call sites are unchanged, and the two tests that assert on them now capture `gnr.web` instead of `gnr.web.gnrmenu`.

gnrpy/gnr/web/gnrmenu.py
fporcari: Kept and documented, per your recommendation. In `980f958f8` the `docpage` docstring now says `source` is a storage path **or** an absolute `http(s)` URL resolved through the built-in `_http_` service, and states the three caveats you named: `X-Frame-Options`/`frame-ancestors` give a blank frame with no warning, `http` under an `https` app is blocked as mixed content, and `nocache` is silently ignored by the `http` service.

`test_load_resolves_absolute_http_url` covers the case with the real `http` service — `public_url()` falls through to `url()` which returns `args[0]` — and asserts the URL reaches `webpage` intact with `nonGenroContent=True` and no `externalSite` prefix even when one is inherited.

One detail worth knowing for anyone writing a similar test: the fixture must import the service with an explicit `importAs`, otherwise `gnrImport` registers the module as `http` in `sys.modules` and shadows the stdlib package.

fporcari:
@dgpaci both points are addressed in `980f958f8`, each answered in its own thread.

1. **Package logger** — `import logging` and the module-level `getLogger` are gone, replaced by `from gnr.web import logger`. The two warning call sites are untouched; the two tests asserting on them capture `gnr.web` now.
2. **`docpage` and http sources** — kept, as you suggested, and documented: the docstring states that `source` is a storage path or an absolute `http(s)` URL through the built-in `_http_` service, with the `X-Frame-Options`/mixed-content/`nocache` caveats. `test_load_resolves_absolute_http_url` covers it against the real `http` service.

Verified here on the pushed head: `pytest gnrpy/tests` gives **2087 passed, 69 skipped**, `gnrmenu_test.py` alone **11 passed**, flake8 clean on both touched files.

Note that this push dropped the approvals the PR carried, since `dismiss_stale_reviews` is on `develop`. Nothing that was approved has moved: the diff since `2c19097f2` is the logger import, the docstring, and one added test.

dgpaci (approved):
Approving. Both points from my 26 Aug review are addressed in `980f958f8`, checked against the code:

- **Logger** — `import logging` and the module-level `getLogger` are gone, replaced by the shared `gnr.web` package logger, and the two `caplog.at_level(...)` calls in the tests moved to `gnr.web` along with it. The module now logs on the same channel as its 28 siblings under `gnr/web/`, which is also the channel the siteconfig `<loggers><web .../></loggers>` section actually addresses.
- **`docpage` docstring** — now states that `source` is a storage path *or* an absolute `http(s)` URL resolved through the built-in `_http_` service, with the three remote-delivery caveats (`X-Frame-Options`/`frame-ancestors`, mixed content, the ignored `nocache`). `test_load_resolves_absolute_http_url` covers the case against the real `http` service, and also pins the inherited-`externalSite` behaviour for a remote URL.

Everything approved on 21 Aug is unchanged: the `externalSite` short-circuit, the per-node `NotAllowedException` that hides only its own entry, `public_url(nocache=True)`, the `nonGenroContent` contract across the four `frameindex.js` call sites, the shared `_finalizePageAttributes` tail, and the `de`/`fr`/`it` localization of the two fixture entries.

CI green on 3.11-3.14 at `a1048d1a2`.

Two follow-ups, both post-merge material and neither a condition of this approval: the `_adapt_path` routing coverage noted inline, and `nocache=True` being a no-op on `aws_s3` — that one is a change to the service, not to this PR, and deserves its own issue since the motivating case (a report regenerated at a fixed path) is exactly the one it affects.

cgabriel (approved):
LGTM, verified after @dgpaci reviews

genro (dismissed):
Approving. Eleven review exchanges have already gone into this one, so I read the resolver and the boundary rather than re-litigating the design.

`nodeType_docpage` is tight: `source` and `externalSite` are popped before anything else, the URL comes from `public_url(nocache=True)` — signed links expiring inside a menu and stale overwritten documents are both real, and `nocache` is the cheaper of the two answers — and any failure logs the source and raises `NotAllowedException`, so one broken entry hides itself instead of taking the menu down with it. A broad `except Exception` is the right trade in a menu resolver, and it is broad *and* logged, which is what makes it acceptable.

`nonGenroContent` is the detail that shows this was run and not just written: without it the frame waits for a Genro bootstrap that never comes on a plain document, loses the label as its title, and dereferences `contentWindow.genro` on close.

**The security boundary is stated in the right place.** Not only in the PR description — which nobody reads at the call site — but in the docstring of `docpage` itself. That matters more than usual here, because `tags` means access control everywhere else in this API: every developer who types `tags='REPORT_VIEW'` will assume it protects the document, and it does not — it hides the entry while the resolved URL stays open. The docstring is the only thing standing between that assumption and a leak, so it is load-bearing code rather than documentation, and it should survive any future tidy-up of that method.

Being explicit that this is resolver-only, with an authenticated viewer named as the answer for confidential documents, is what keeps the feature honest about what it is.

**One thing that is not visible from inside this PR: it collides with #1128 on the test file.**

`gnrpy/tests/web/gnrmenu_test.py` is a single `import gnr.web.gnrmenu  # noqa: F401` on `develop` (`745e5bb43b`). This PR removes that line and writes 197 starting from `import logging`; #1128 keeps it and appends 87. Both diff from the same base at the same position, so whichever lands second conflicts across the whole file.

The dangerous part is that the easy resolution is wrong in a way nothing reports: taking either side wholesale drops the other PR's tests, and a suite that is simply absent does not fail. #1128 is the smaller of the two and both are approved, so landing it first and rebasing this one on top is the cheaper order — and the resolution has to end with **both** sets present: the `DirectoryMenuResolver` cases from #1128 and the ten resolver cases here.

Not verified by me: the live authenticated browser session, which your description already lists as outstanding.

cgabriel (approved):
Reviewed new entries, all good

genro (approved):
Re-approving on `a492da8c4`. My previous approval fell with the merge of `develop`, and since that merge resolved exactly the collision I had flagged, I checked the resolution rather than the diff.

**Both test suites survived, which was the whole risk.** `gnrpy/tests/web/gnrmenu_test.py` now carries the eight `docpage` resolver cases from this PR and the four `DirectoryMenuResolver` cases that came from #1128, with the fakes of both — 281 lines, 15 passed on this head. I ran them on the branch itself, printing the resolved module path first so there is no doubt which tree answered: `gnr.web.gnrmenu` from the worktree, with `MenuStruct.docpage` present.

`projects/gnrcore/packages/test/menu.py` resolved the same way, and this is the part that would have been easy to get wrong in the opposite direction: it keeps `develop`'s folder-driven `root.directoryBranch('!!Tests', folder='', tags='_DEV_')` and *adds* the documents branch beside it, rather than restoring the hand-written index this PR was branched before #1128 removed.

Everything I approved earlier still stands: the resolver popping `source` and `externalSite` before use, `public_url(nocache=True)` against expiring signatures and stale overwrites, one broken entry hiding itself through a logged `NotAllowedException` instead of taking the menu down, and `nonGenroContent` so the frame does not wait for a bootstrap that never comes.

And the security boundary is still where it needs to be — in the docstring of `docpage`, not only in the description. `tags` means access control everywhere else in this API, so that paragraph is the only thing between a developer's reasonable assumption and an open URL. It should survive any future tidy-up of that method.

Nothing left outstanding from my side. The live authenticated browser session is still the one thing your description lists as unverified.

PR #1060, opened 2026-08-07
fix(app): honour the instanceconfig package path in the dependency check

fporcari:
## Problem

`GnrApp.check_package_dependencies()` resolved every package through `pkg_name_to_path()`, ignoring the `path` attribute that `addPackage()` honours.

`pkg_name_to_path()` only knows the project packages folder and the `<packages>` roots of `environment.xml`: for a package declared with an explicit `path` it raises `Exception('Error: package %s not found')`. Since the check runs before any package is loaded, such an instance dies at boot.

`InstanceMaker` writes exactly those attributes for every package passed as a `(package, path)` pair, so any instance created that way was unbootable. No instance under `projects/` uses `path=`, which is why nothing failed in CI.

## Fix

Both call sites now share `pkg_path_from_attrs()`:

- the declared `path` wins, `pkg_name_to_path()` is the fallback when it is absent;
- relative paths are resolved against the instance folder through `realPath()`, as `addPackage()` already did.

The requirements lookup also honours the `filename` attribute, the same way `GnrPackage` builds its package folder from `path` + `filename`.

## Tests

`gnrpy/tests/app/test_package_path_attr.py`, on `BaseGnrTest`: a minimal package (`main.py` + `requirements.txt`) is created in a temporary folder outside every packages root, and two instances are built with `InstanceMaker`, one declaring an absolute path and one a relative path. The tests check that the instanceconfig carries the `path` attribute, that the package is indeed unreachable via `pkg_name_to_path()`, that the app boots with `packageFolder` pointing at the external folder in both cases, and that the external requirements are collected without being reported as missing.

Four of the five tests fail on `develop` with `Exception: Error: package extpkg not found`.

Full suite: 1924 passed, 67 skipped.

## Context

Follow-up 1 of #1058 (issue #1006), deliberately left out of that PR.

### DIFF
--- a/gnrpy/gnr/app/gnrapp.py
+++ b/gnrpy/gnr/app/gnrapp.py
@@ -1054,10 +1054,7 @@ class GnrApp(object):
         if pkgid in self.packages:
             return 
         attrs = pkgattrs or {}
-        if not attrs.get('path'):
-            attrs['path'] = self.pkg_name_to_path(pkgid,project)
-        if not os.path.isabs(attrs['path']):
-            attrs['path'] = self.realPath(attrs['path'])
+        attrs['path'] = self.pkg_path_from_attrs(pkgid, attrs, project=project)
         apppkg = GnrPackage(pkgid, self, **attrs)
         apppkg.content = pkgcontent or Bag()
         readOnlyAttrs = {'readOnly':True} if attrs.get('readOnly') else dict()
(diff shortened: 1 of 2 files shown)


cgabriel (approved):
Approved. Verified with instance maker and dev builder, with pdb inspections of test-generated instances/packages.

PR #1061, opened 2026-08-07
fix(tests): stop the postgres fixtures from killing concurrent test runs

fporcari:
refs #550 #549

### 1. The fixture kills other people's databases

`gnrpy/tests/sql/common.py:67`, inside `get_pg_config()`:

```python
subprocess.run(['pkill', '-f', 'postgres.*tmp'], capture_output=True)
```

`postgres.*tmp` matches the command line of **every** `testing.postgresql` instance on the
machine, not just a stale one belonging to this session — the data directory is always
`<tmpdir>/data`, and the binary path itself contains `postgres`. Every call to
`get_pg_config()` therefore tears down the databases of every test run already in flight.

Measured directly: start an instance, then run the pattern.

```
victim pid   : 71837 port 54804
victim alive : True
pkill pattern "postgres.*tmp" matches pids: ['71837']
victim is among them: True
victim alive after the fixture-level pkill: False
```

Symptoms in the poisoned run: `pg8000.exceptions.DatabaseError: 'the database system is
shutting down'`, `psycopg2.OperationalError: Connection refused`, `RuntimeError: initdb
failed`, and failure counts that swing by hundreds between identical runs of identical
code. Two CI jobs sharing a runner corrupt each other silently, and so do two developers —
or two agents — running the suite on one machine.

### 2. An unreachable postgres was reported as green

`gnrpy/tests/sql/conftest.py:121-137` (`db_pg`) and `:141-157` (`db_pg3`) wrapped the whole
fixture body in `except Exception: pytest.skip('PostgreSQL not available')`. Whatever went
wrong — postgres down, bootstrap failure, a broken fixture — the tests were skipped and the
run exited 0.

### 3. On a shared server every module imported the CSV data on top of the previous one

`get_pg_config()` returns a fixed address, with no `database` key, whenever the tests run
against an already running postgres: the CI service on `127.0.0.1:5432`, or
`GNR_TEST_PG_*`. Both fixtures then fell back to the same `pg_conf.pop('database',
'test_compiler')`, so all eight module-scoped setups (four modules × two implementations)
landed on one `test_compiler` database and each ran `_import_csv_data` over the previous
module's rows.

From the second module onwards the fixture died on `duplicate key value violates unique
constraint "invc_region_pkey"` — and defect 2 turned that into a skip.

### 4. Two modules never built a genro configuration

`conftest.py`'s `setup_module` / `teardown_module` are never called: pytest honours xunit
module hooks on test modules, not on a conftest. Every module that needs a genro
configuration therefore repeats `BaseGnrTest.setup_class()` itself — and `test_db_notify.py`
and `test_gnrlistener.py` do not. On a developer machine `~/.gnr` covers for them; on a
runner there is none, so `GnrApp` raised `Exception: Missing genro configuration`, and
defect 2 turned that into a skip too.

Defects 3 and 4 together: **193 postgres tests had never run in CI.** The run before this
PR was `1756 passed, 261 skipped`, exit 0.

## Root cause

The `pkill` was masking a leak rather than cleaning up a stale instance.

`BaseGnrSqlTest.teardown_class` stopped the temporary server only after `dropDb()`:

```python
if hasattr(cls, 'dbname'):
    cls.db.closeConnection()
    cls.db.dropDb(cls.dbname)   # <- raises for several classes
if cls.pg_instance is not None:
    cls.pg_instance.stop()      # <- never reached
```

A teardown that raises leaves a postgres running for the rest of the session. Those pile
up, and postgres needs one SysV shared memory segment per postmaster: on the machine used
here `kern.sysv.shmmni` is 32, so past ~31 live servers `initdb` starts failing with
`could not create shared memory segment`. The machine-wide `pkill` kept the count down —
by destroying every other run's databases along with the leaked ones.

Defect 3 had been invisible for the same reason defect 2 exists: the blanket skip.

## Change

`gnrpy/tests/sql/common.py`

* Drop the machine-wide `pkill`. Instances started by this process are tracked in a
  `WeakSet` and stopped by an `atexit` handler, so an interrupted or crashing run cleans up
  after itself without touching anyone else's servers. Weak references only — a strong one
  would pin an instance a fixture has already dropped for the whole session.
* `BaseGnrSqlTest.teardown_class` stops the server in a `finally`, so a failing `dropDb()`
  no longer leaks it.
* `LANG` is set only around the `initdb`/`postgres` spawn instead of being left mutated in
  the session environment.

`gnrpy/tests/sql/conftest.py`

* `db_pg` / `db_pg3` no longer skip. Postgres is a required part of this suite, so the
  failure surfaces and the run goes red. The `try` stays only for the `finally` that stops
  the temporary server. This matches `BaseGnrSqlTest`, which has always let the failure
  propagate.
* The database name is derived from the requesting module and the implementation, so each
  fixture gets its own database on a shared server exactly as it already gets its own
  server when the instance is ephemeral.
* A module-scoped autouse fixture builds a genro configuration for any module in this
  package that has not built one itself, so the two modules that never did stop depending
  on the developer's `~/.gnr`. Modules that already set one up keep theirs.

## Verification

macOS 25.0.0, Python 3.13.2, PostgreSQL 16.8 (Homebrew), `kern.sysv.shmmni = 32`.

**Lint** — `flake8 gnrpy/tests/sql/common.py gnrpy/tests/sql/conftest.py`: zero errors.

**Defect 3, shared-server topology** (a throwaway server with `GNR_TEST_PG_*` pointing at
it, so `get_pg_config()` takes the same branch as CI):

| | the four modules using these fixtures | whole `sql` dir |
|---|---|---|
| before | 313 passed, 9 skipped, **30 errors**, 60 × `UniqueViolation` | — |
| after | **343 passed**, 9 skipped, 0 errors | 1202 passed, 10 skipped, 0 errors |

**Defect 4, no `~/.gnr` available** (reproduced by pointing `VIRTUAL_ENV` at a directory
that does not exist, which is the branch of `gnrConfigPath()` a runner takes),
`pytest sql/test_db_notify.py sql/test_gnrlistener.py -q`:

| | result | exit |
|---|---|---|
| before | 17 passed, **22 errors**, 44 × `Missing genro configuration` | 1 |
| after | **39 passed**, 0 errors | 0 |

Whole `sql` dir under the same condition: 1202 passed, 10 skipped, exit 0.

**Defect 2, postgres unreachable with CI-style config** (`GITHUB_WORKFLOW` set, service not
answering), `pytest sql/test_compiler_coverage.py -q`:

| | result | exit |
|---|---|---|
| before | 127 passed, **172 skipped** | **0 — green** |
| after | 127 passed, 9 skipped, **163 errors** | **1 — red** |

With no postgres reachable at all (`initdb` stubbed to fail), `pytest sql -q` is
`452 passed, 10 skipped, 743 errors`, exit 1 — no test that used to run is skipped now.

**Defect 1, does an unrelated database survive?** Start an instance, run the whole sql
suite in a separate process, check the instance.

| | suite result | victim alive | victim accepts connections |
|---|---|---|---|
| before | 1195 passed, 10 skipped | **False** | **False** |
| after | 1195 passed, 10 skipped | True | True |

**Defect 1, two concurrent runs.** Second run started 3s after the first, so the first
already has servers up when the second bootstraps — what happens when two people start the
suite a moment apart. Target: `test_compiler_coverage.py test_composite_column.py
test_db_notify.py`, which is **316 passed, 9 skipped** when run alone.

Before, three repetitions:

| | run A (first) | run B (3s later) |
|---|---|---|
| 1 | 298 passed, 19 skipped, 8 errors | 153 passed, 172 skipped |
| 2 | 308 passed, 9 skipped, 8 errors | 153 passed, 9 skipped, 163 errors |
| 3 | 306 passed, 19 skipped | 145 passed, 9 skipped, 171 errors |

After, three repetitions — all six runs identical to the solo result:
`316 passed, 9 skipped`, exit 0.

**Suites**, single run, postgres available: `cd gnrpy && pytest` — 1948 passed, 67 skipped,
0 failures.

## Not fixed here

Two concurrent runs of the **full** sql suite still diverge, for a different reason that
this change makes visible instead of hiding:

| | run A (first) | run B (5s later) | failure signature |
|---|---|---|---|
| before | 1 failed, 641 passed, 20 skipped, 543 errors | 47 failed, 672 passed, 174 skipped, 314 errors | 399 × `the database system is shutting down`, 0 × shared memory |
| after | 1026 passed, 40 skipped, 139 errors | 736 passed, 204 skipped, 265 errors | 0 × `shutting down`, 404 × `could not create shared memory segment` |

No run kills another one any more; both simply exceed the machine's 32-segment ceiling,
because a full suite holds ~20 servers at its peak. That peak comes from eight modules —
`a_structure_load_test.py`, `b_structure_build_test.py`, `d_table_test.py`,
`e_query_test.py`, `f_selection_test.py`, `h_query_surface_test.py`,
`h_record_surface_test.py`, `h_selection_surface_test.py`, `test_import_export_tojson.py` —
that override `teardown_class` without stopping their instance. Traced peak for one run:
21 live servers, 20 still running at session end.

Also left alone:

* `test_sqlite_boolean_rewrite.py:60-61` still swallows an unreachable postgres into a
  skip, the same way `db_pg` did before this PR.
* `d_table_test.py`'s sqlite classes use a fixed database file from `configTest.xml`, so two
  concurrent runs collide there independently of postgres (`sqlite3.OperationalError`).


cgabriel (changes requested):
The single instance management is fine, but I don't agree by skipping tests if postgres is not available.

If the postgres instance is not available for any reason, tests are skipped, so the CI reports green to merge. The availability of postgres and the related test execution must be mandatory.

fporcari:
You are right, and the skip is gone: `db_pg` and `db_pg3` no longer swallow the exception, so a run that cannot reach postgres fails instead of reporting green.

Removing it turned out to be the interesting part, because it uncovered two things that the skip had been hiding:

- **The tests were not running at all.** `test_db_notify` and `test_gnrlistener` never set up a genro configuration — the `setup_module`/`teardown_module` in `conftest.py` are dead code, since pytest honours xunit module hooks on test modules and not on a conftest. Their fixture died on "Missing genro configuration" and the skip reported it as "PostgreSQL not available". A module-scoped autouse fixture now provides the configuration, so they actually execute.
- **Once they executed, seven of them failed on CI.** They opened the second connection (the one that does the `LISTEN`) with `psycopg2.connect(connection.dsn)`, and psycopg strips the password out of the dsn — fine against a trust-auth server, fatal against CI's, which asks for one: `FATAL: password authentication failed for user "postgres"`. They now go through `adapter.connect(autoCommit=True)`, which assembles the parameters from `get_connection_params()`, password included. The psycopg3 one also had to move to `cursor().execute()`: the adapter installs `GnrDictCursor`, whose `execute()` does not accept psycopg3's `prepare`/`binary` keywords, so the `conn.execute()` shortcut raises a `TypeError`.

The last commit closes a third one, of the same family: the fixtures created a database per module and never dropped it. An ephemeral `testing.postgresql` takes its databases away when it stops, so nothing showed — but against a shared postgres, CI's or a developer's `GNR_TEST_PG_*`, the database survives and the next run imports the CSV data on top of the rows left behind and dies on a duplicate key. Running the suite twice in a row reproduced it with 22 fixture errors on the second run; both fixtures now drop the database in teardown, as `BaseGnrSqlTest.teardown_class` already does.

Verified on a real postgres: `tests/sql` green twice in a row from a clean server (1202 passed, 10 skipped each time), with no `test_*_postgres` database left behind, and CI green on 3.11 through 3.14.

One thing I left alone, because it is a framework matter and not a test one: `GnrDictCursor.execute()` in `gnrpostgres3.py` does not accept `prepare`/`binary`, which makes psycopg3's standard `connection.execute()` unusable on any genropy connection. Worth its own issue if you agree.

cgabriel (approved):
Looking good, mandatory postgresql tests are not skipped anymore, verified with a mock DSN during test execution.

Issue #1062, opened 2026-08-13
Usernames containing dots break Bag-label keyed features (chat first, audit needed)


genro:
## Context

Usernames following the `name.surname` convention expose a structural trap: both the Python `Bag` and the JS `GnrBag` treat `.` in a label as a **path separator** (`htraverse` smart-splits on dots, gnrbag.js:983; same in Python `setItem`). Wherever a username is used as a Bag **label**, the entry silently nests: `amelia.martin` becomes `amelia` → `martin`, and the top level shows an attribute-less container node.

## Confirmed end-to-end on the chat component

Observed live with two users (`amelia.martin`, `alexander.king`), wire captures in hand:

- **Room title shows `,`** — `chat_component.js:54,180` keys the room `users` bag by username, so the bag nests; `fill_title` digests `#a.user_name` over the top level and finds the attribute-less intermediate nodes (`amelia`, `alexander`).
- **Messages are never delivered (no sender echo, nothing on the recipient)** — `chat_component.py:162-171`:

  ```python
  for userNode in users:
      user = userNode.label        # 'amelia', 'alexander' — top-level labels!
      with self.userStore(user) as store:
          store.set_datachange(path, value, fired=True, reason='chat_out')
  ```

  The writes land on the stores of non-existent users (`amelia`, `alexander`); the real pages watch `amelia.martin` / `alexander.king`.
- **The alert still fires, masking the defect** — `room_alert` goes out with `filters='user:amelia'`, and the loose filter match still catches `amelia.martin`: the recipient hears the new-message sound for a message that never arrives. The sender also receives their own alert, since `'alexander' != 'alexander.king'` defeats the self-exclusion check.

The component and the Bag are shared between the daemon and daemonless stacks — the behaviour is identical on both.

## Same pattern elsewhere (quick scan, to audit)

- `gnrpy/gnr/web/gnrwebpage.py:2325` — the server-sent chat helper builds the `users` bag with `users.setItem(user, None, user_name=user, user=user)`.
- `gnrpy/gnr/web/daemon/siteregister_client.py:453` — user listing keyed by username (`result.setItem(user, item, user=user)`): rows for dotted names nest.
- `gnrpy/gnr/web/gnrwebpage_proxy/connection.py:192` — `connected_users_bag` **already dodges the trap** (`userkey = user.replace('.','_')`), which suggests the problem surfaced at least once before.

## Ask

[...]

PR #1063, opened 2026-08-13
fix(#1062): chat with dotted usernames — escape Bag labels, real name in the user attribute

genro:
## What

Usernames containing dots (`name.surname`) broke the chat: Bag labels split on dots in both the Python `Bag` and the JS `GnrBag`, so the room's `users` bag nested (`amelia.martin` → `amelia` → `martin`), `ct_send_message` iterated the truncated top-level labels and wrote the messages to the stores of non-existent users, and the room title digested the attribute-less intermediate nodes (showing `,`). The `room_alert` still fired through the loose `user:` filter match, masking the lost messages behind a working notification sound.

## How

The convention `connected_users_bag` already uses, applied consistently:

- **the label travels dot-escaped** — `ct_chat_utils.escape_user` (JS) at every point a username becomes a Bag label (`open_chat`, `prepare_usersbag`, the connected-users grid controller, `key_from_users`, the disconnect `popNode`); `replace('.','_')` in `chatMessageToUser` server-side;
- **the real username always rides in the node's `user` attribute** (it already did);
- **`ct_send_message` reads the attribute, not the label**: `user = userNode.attr.get('user') or userNode.label` — the fallback keeps old callers passing flat undotted bags working.

## Verification

Exercised live with two dotted-username users (`amelia.martin`, `alexander.king`) on the daemonless single: sender echo, recipient delivery and room titles all correct, in both directions, wire captures checked (`change_fired`, single leaf per write).

Issue #1062 asks the broader question — whether other subsystems key Bags by username; this PR fixes the chat surface (the confirmed breakage) and leaves the audit open.

Ref #1062

### DIFF
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -2319,12 +2319,13 @@ class GnrWebPage(GnrBaseWebPage):
         path = 'gnr.chat.msg.%s' % roomId
         priority = priority or 'H'
         if not users:
-            # the label travels dot-escaped (Bag labels split on dots);
+            # the label travels escaped ([.@] -> _, the connected_users_bag rule:
+            # dots split Bag paths, @ comes with email-style usernames);
             # the real username rides in the node's `user` attribute
             users = Bag()
             if from_user!='SYSTEM':
-                users.setItem(from_user.replace('.','_'),None,user_name=from_user,user=from_user)
-            users.setItem(user.replace('.','_'),None,user_name=user,user=user)
+                users.setItem(from_user.replace('.','_').replace('@','_'),None,user_name=from_user,user=from_user)
+            users.setItem(user.replace('.','_').replace('@','_'),None,user_name=user,user=user)
         ts = self.toText(datetime.datetime.now(), format='HH:mm:ss')
         with self.userStore(user) as store:
             if disconnect and (user == from_user):
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -2319,10 +2319,12 @@ class GnrWebPage(GnrBaseWebPage):
         path = 'gnr.chat.msg.%s' % roomId
         priority = priority or 'H'
         if not users:
+            # the label travels dot-escaped (Bag labels split on dots);
+            # the real username rides in the node's `user` attribute
             users = Bag()
             if from_user!='SYSTEM':
-                users.setItem(from_user,None,user_name=from_user,user=from_user)
-            users.setItem(user,None,user_name=user,user=user)
+                users.setItem(from_user.replace('.','_'),None,user_name=from_user,user=from_user)
+            users.setItem(user.replace('.','_'),None,user_name=user,user=user)
         ts = self.toText(datetime.datetime.now(), format='HH:mm:ss')
         with self.userStore(user) as store:
             if disconnect and (user == from_user):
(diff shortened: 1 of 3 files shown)


fporcari (changes requested):
The diagnosis is right and the convention chosen to fix it is the correct one: the label travels escaped, the real username rides in the node's `user` attribute, and `ct_send_message` reads the attribute with a fallback to the label, so callers still passing flat undotted bags keep working. I checked the attribute really is populated end to end — `user_register.create(user=...)` in `gnrpy/gnr/web/daemon/siteregister.py:624` feeds `row.update(arguments)` in `connected_users_bag`, so `n.attr.user` in the connected-users grid controller is always the real username, never undefined. Client and server also stay in step: both escape only the dot, so `key_from_users` yields the same key on either side and rooms do not get duplicated. Outside this repo nothing calls `prepare_usersbag` or `chatMessageToUser`, so the blast radius is the chat component itself.

One thing needs to come in with this PR: usernames containing `@`.

`connected_users_bag` builds its labels as

```python
userkey = user.replace('.','_').replace('@','_')
```

(`gnrpy/gnr/web/gnrwebpage_proxy/connection.py:192`), while `escape_user` only replaces the dot. For `mario.rossi@acme.com` the register label is `mario_rossi_acme_com` but `escape_user` produces `mario_rossi@acme_com`, so

```js
n = connectedUsers.getNode(ct_chat_utils.escape_user(username));
```

(`resources/common/gnrcomponents/chat_component/chat_component.js:179`) returns `null` and the next line raises on `n.attr.user_name`. This is not a regression — the same lookup missed before the PR — but the stated premise here is "the convention `connected_users_bag` already uses, applied consistently", and that convention includes the `@`. With email-style usernames this is the more common case of the two.

The fix is one character class, and it stays a no-op if a caller already passes escaped keys:

```js
ct_chat_utils.escape_user = function(user){
    return user.replace(/[.@]/g, '_');
};
```

The two escapes have to move together, though: `gnrpy/gnr/web/gnrwebpage.py:2325-2326` must widen the same way

```python
users.setItem(from_user.replace('.','_').replace('@','_'),None,user_name=from_user,user=from_user)
users.setItem(user.replace('.','_').replace('@','_'),None,user_name=user,user=user)
```

otherwise client and server produce different labels for the same user and `key_from_users` stops matching, which splits one conversation into two rooms.

Since the same escaping rule now lives in three places (here, the JS helper, and `connected_users_bag`), it may be worth having `connected_users_bag` call a single helper too — but that can stay out of this PR if you prefer to keep it to the chat surface.

fporcari (approved):
`7cf6852ed` closes it. Verified on the branch:

- `escape_user` now applies `user.replace(/[.@]/g, '_')` (`resources/common/gnrcomponents/chat_component/chat_component.js:38`), which produces exactly the same string as `user.replace('.','_').replace('@','_')` in `connected_users_bag` (`gnrpy/gnr/web/gnrwebpage_proxy/connection.py:192`). So `connectedUsers.getNode(...)` in `prepare_usersbag` now resolves for `mario.rossi@acme.com` instead of returning `null`.
- `chatMessageToUser` widened the same way (`gnrpy/gnr/web/gnrwebpage.py:2325-2327`), so client and server produce identical labels and `key_from_users` keeps yielding one key per conversation — no split rooms.

Every point where a username becomes a Bag label now goes through the helper — `open_chat`, `prepare_usersbag`, the connected-users grid controller, the disconnect `popNode`, `key_from_users` — and `ct_send_message` reads `userNode.attr.get('user')` with the label as fallback, so callers still passing flat bags keep working. CI green on 3.11 through 3.14.

Keeping the shared-helper unification out of this PR is fine; it belongs with the wider audit in #1062.

One pre-existing item for that audit, not for this PR: `prepare_usersbag` still dereferences `n.attr` without checking `n`, so a username that is no longer in `gnr.chat.connected_users` (disconnected between the grid read and the click) raises. The escaping fix makes it much rarer, it does not remove it.

Issue #1065, opened 2026-08-14
get_page_class rebuilds the page class on every RPC — the cache died in 2010; experimental per-page_id cache doubles the capacity within 100ms


genro:
## Where the request time goes

Benchmark ladder on a real site (`test_invoice_pg`, authenticated sessions, keep-alive), single process:

| call | p50 | ceiling |
|---|---:|---:|
| `/_ping` (full stack + register, no DB) | 1.7ms | ~2700 req/s |
| `app.getSelection`, **1 indexed row** (~700B) | 15.7ms | ~120 req/s |
| `app.getSelection`, 100 rows (~45KB) | 126ms | ~25 req/s |

The indexed query is nearly free, yet one row costs 15.7ms: ~8ms of **serial CPU per RPC is per-request machinery, not the database**. A multi-thread CPU profile (yappi, 100 calls) attributes it:

| share of request CPU | what |
|---:|---|
| **~40-45%** | rebuilding the page class on every RPC — `classMixin` called **138×/request**, `getResourceList` 60×, `FilterList.__contains__` 818×, `base_visitor` 2383× |
| ~8% | Bag operations |
| ~7% | page instance init + response envelope |
| <1% | SQL compile + driver CPU |

## The history

`get_page_class` once had a working cache:

- **2009-05-11** (repo reimport): `gnrwsgisite.py` checks `if path in self.page_factories: return ...` — alive.
- 2010-01: the read+write pair migrates to `gnrpageserver`, then `gnrresourceloader` — still alive.
- **2010-10-12** (`50f716b8e`): the read is **commented out**; the write stays.
- **2014-07-20** (`232781bd8`, "cosmetic"): the commented line is deleted.

Today `page_factories` is written on every build and never read. The 2010 disabling was likely justified: `get_page_class` **mutates** the class it returns (`css_requires.extend`, per-request `getMainPackage`), so a naive module-keyed shared cache produced dirty classes. The price has been ~40% of request CPU on every deployment since.

## The fix (experimental, opt-in)

A per-**page_id** LRU cache of the built class, behind the sys preference **`experimental.page_class_cache`** (new checkbox in the existing experimental pane, runtime kill switch — the flag is re-read with a 10s TTL because `getPreference` deepcopies the whole preference bag).

Keying by `page_id` is what makes it safe with no idempotence work:

- each page gets its **own** class object — the build-time mutations cannot cross pages;
[...]

PR #1066, opened 2026-08-14
feat(#1065): experimental per-page_id page class cache

genro:
## What

An opt-in, LRU-bounded cache of the built page class, keyed by **page_id**, behind the sys preference **`experimental.page_class_cache`** (new checkbox in the existing experimental pane). Two files: the cache + hook in `ResourceLoader.get_page_class`, one line in the sys preference pane.

Full story, profile and benchmark tables in #1065. The short version: `classMixin` runs 138× per RPC (~40-45% of request CPU with a trivial query); the cache that once avoided this died in 2010 and its remains were deleted in 2014. **Measured capacity within 100ms doubles on both stacks** — 71 → 135 req/s (classic gunicorn + Pyro daemon), 128 → 238 req/s (daemonless single).

## Why per page_id is safe

- Each page reuses the class built at its own birth; no class is ever shared between pages, so the build-time mutations (`css_requires.extend`, per-request `getMainPackage`) cannot cross — the very reason the 2010 module-keyed cache had to die does not apply.
- Runtime component mixins go through `instanceMixin` (gnrlang.py:341) and never touch the class; table scripts are instantiated fresh per call (gnrresourceloader.py:520).
- An explicit `page_factory` bypasses the cache — `instantiate_page` forces `GnrSimplePage`, a different base class under the same page_id — and so does `avoid_module_cache` (dev reload).
- A request without a `page_id` builds exactly as today; RPCs (the hot path) always carry one.
- LRU ceiling (2000 classes) bounds memory with no lifecycle wiring: evicted pages just rebuild.

## The flag read

`getPreference` deepcopies the whole preference bag on every call, so the flag is re-read through a 10s TTL instead of per request: the preference remains a runtime kill switch (like `experimental.wsk_disabled`) with at most 10s of latency, at zero per-request cost.

## Verification

- 130-test regression suite of the daemonless bridge (boots this very site) green with the flag off AND on.
- Site exercised by hand with the flag on: login, table handlers, forms, chat.
- Benchmarks: single-row getSelection ramp 1..32 sessions and the 100-row workload, flag OFF vs ON on both stacks, zero errors — tables in #1065.

Ref #1065

### DIFF
--- a/gnrpy/gnr/web/gnrhtmlpage.py
+++ b/gnrpy/gnr/web/gnrhtmlpage.py
@@ -137,7 +137,7 @@ class GnrHtmlPage(GnrWebPage):
 
     def setCssRequires(self):
         css_import_statements_list=[]
-        css_requires = getattr(self, 'css_requires', [])
+        css_requires = list(getattr(self, 'css_requires', None) or [])
         css_requires.extend([k.split('.')[0] for k in list(self.envelope_css_requires.keys())])
         for css_require in css_requires:
              urls =self.getResourceExternalUriList(css_require,'css') or []
@@ -148,7 +148,7 @@ class GnrHtmlPage(GnrWebPage):
             self.builder.head.style(import_statements + ';\n', type="text/css")
 
     def setJsRequires(self):
-        js_requires = getattr(self, 'js_requires', [])
+        js_requires = list(getattr(self, 'js_requires', None) or [])
         js_requires.extend([k.split('.')[0] for k in list(self.envelope_js_requires.keys())])
         for js_require in js_requires:
              urls =self.getResourceExternalUriList(js_require,'js',add_mtime=True) or []
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -208,6 +208,13 @@ class GnrWebPage(GnrBaseWebPage):
         self.dojo_version = request_kwargs.pop('dojo_version', None) or getattr(self, 'dojo_version', None)
         self.envelope_js_requires= {}
         self.envelope_css_requires= {}
+        # ``css_requires``/``js_requires`` live on the page class, and both the
+        # runtime component mixins (BaseComponent.__onmixin__) and the root
+        # render append to them. Shadowing the class lists with per-instance
+        # copies keeps those appends inside the request, as they were when
+        # every request built its own class.
+        self.css_requires = list(getattr(self, 'css_requires', None) or [])
+        self.js_requires = list(getattr(self, 'js_requires', None) or [])
         self._avoid_module_cache = _avoid_module_cache
         self.debug_sql = boolean(request_kwargs.pop('debug_sql', None))
         debug_py = request_kwargs.pop('debug_py', None)
@@ -622,6 +629,7 @@ class GnrWebPage(GnrBaseWebPage):
         self._onEnd()
         if getattr(self,'_closed',False):
             self.site.register.drop_page(self.page_id, cascade=False)
+            self.site.resource_loader.drop_page_class_cache(self.page_id)
         return result
     
 
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -1426,6 +1426,7 @@ class GnrWsgiSite(object):
     def onClosedPage(self, page_id=None, **kwargs):
         "Drops page when closing"
         self.register.drop_page(page_id)
+        self.resource_loader.drop_page_class_cache(page_id)
 
     def cleanup(self):
         """clean up"""
--- a/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py
+++ b/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py
@@ -83,6 +83,12 @@ class ResourceLoader(object):
                 self._page_class_cache.move_to_end(cache_key)
             return page_class
 
+    def page_class_cache_entries(self):
+        """A snapshot of the cache: (page_id, module_path, class_name) per entry."""
+        with self._page_class_cache_lock:
+            return [(page_id, module_path, page_class.__name__)
+                    for (page_id, module_path), page_class in self._page_class_cache.items()]
+        
     def store_page_class_cache(self, cache_key, page_class):
         """Deposit a freshly built class under its ``(page_id, module_path)``, LRU-bounded."""
         with self._page_class_cache_lock:
--- a/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py
+++ b/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py
@@ -26,12 +26,16 @@ from gnr.core.gnrclasses import GnrMixinError,GnrMixinNotFound
 from gnr.core.gnrlang import uniquify
 from gnr.web import logger
 
-# Experimental per-page_id page-class cache (sys preference
+# Experimental page-class cache (sys preference
 # ``experimental.page_class_cache``): a page's class is a function of values
 # fixed at the page's birth (module, main package, plugin/custom mixins), so
-# the first build can serve every later request of the same page_id. The LRU
-# ceiling bounds memory without wiring into the page lifecycle; the preference
-# is re-read at most every PAGE_CLASS_CACHE_FLAG_TTL seconds because a
+# the first build can serve every later request of the same page. Entries are
+# keyed by ``(page_id, module_path)``: a page_id is validated against the
+# connection, never against the url it is posted to, so the module path keeps
+# a request that names another page's page_id from reading - or, worse, from
+# depositing - a class built for a different url. The cache is dropped on
+# page close and LRU-bounded for whatever escapes that; the preference is
+# re-read at most every PAGE_CLASS_CACHE_FLAG_TTL seconds because a
 # getPreference call deepcopies the whole preference bag.
 PAGE_CLASS_CACHE_MAXSIZE = 2000
 PAGE_CLASS_CACHE_FLAG_TTL = 10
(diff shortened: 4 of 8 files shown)


fporcari (changes requested):
Keying on `page_id` instead of on the module is the right call, and it is what makes this cache defensible where the 2010 one was not. I went to check the three claims that carry the design and they hold:

- `thpage.getMainPackage` (`projects/gnrcore/packages/sys/webpages/thpage.py:28-33`) really does vary with `request_args`/`th_from_package`, so a module-keyed cache would have served the wrong `_packageId` — while under a `page_id` key each url gets its own entry, and `th_from_package` arrives as a *url* kwarg, so the `pageArgs` merge at `gnrresourceloader.py:176-178` keeps `mainPkg` stable across every RPC of the same page.
- The `avoid_module_cache` bypass survives the whole page, not just the first request: `gnrjs/gnr_d11/js/genro_rpc.js:345-346` re-sends `_avoid_module_cache` on every call when `startArgs` carries it, so a dev-reload page neither reads nor writes the cache.
- Nothing mutates the page class at runtime: the only `classMixin` calls on `page_class` are inside `get_page_class` itself (`gnrresourceloader.py:162-231` on develop), and `mixinComponent` goes through `instanceMixin` (`gnrlang.py:341`).

Two things I would like changed before this goes in, and both are small.

### 1. Key the cache on `(page_id, module_path)`, not on `page_id` alone

`get_page_class` never checks that the cached class belongs to the requested path. `module_path` is available one line below the cache read (`gnrresourceloader.py:169`), so it can just move above it:

```python
module_path = os.path.join(basepath, relpath)
page_id = (request_kwargs or {}).get('page_id')
use_cache = (page_id and page_factory is None and not avoid_module_cache
             and self.page_class_cache_enabled())
if use_cache:
    cached_class = self.load_class_by_page_id((page_id, module_path))
```

The framework's own client always posts a page's `page_id` to that page's own url (`genro_rpc.js:322`, `kw.url` defaulting to `pageIndexUrl()`), so this is not reachable through normal use. But `_check_page_id` (`gnrpy/gnr/web/gnrwebpage.py:298-313`) validates a `page_id` only against the *connection*, so a request to `/pkg/other_page?page_id=<any live page_id of my own connection>&method=…` now executes the cached page's class under another page's url — with that class's `auth_tags`, `maintable` and `_packageId`, since all three live on the class. It does not hand the caller anything they did not already own, so this is class/url confusion rather than an escalation; the point is that keying on the pair costs nothing and keeps the cache a pure performance concern instead of one that has to be reasoned about alongside page authorization.

### 2. `css_requires` / `js_requires` stop being per-request once the class is shared

`BaseComponent.__onmixin__` (`gnrpy/gnr/web/gnrbaseclasses.py:83-88`) appends to `self.css_requires` / `self.js_requires`. Those names resolve to the *class* attributes built at `gnrresourceloader.py:207-216`: `instanceMixin` excludes `css_requires,js_requires,py_requires` from the attribute copy (`gnrlang.py:509`), and the page instance never shadows them — `__init__` only sets `envelope_js_requires` / `envelope_css_requires` (`gnrwebpage.py:209-210`). With a fresh class per request that append was per-request; with a shared class it persists for the life of the cache entry, and two concurrent RPCs on the same `page_id` run `if css not in …: append` on the same list with no lock.

The blast radius is small, and in the safe direction: `envelope_css_requires` is only consumed when a root page is rendered (`gnrhtmlpage.py:141,152` — the other uses are commented out), and a root page normally carries no `page_id`. The exception is the guest path at `gnrwebpage.py:301-303`, where a root page request *with* a `page_id` registers a new page: that render would carry whatever earlier requests appended, i.e. extra css/js, never missing ones. Still, `self.css_requires = list(self.css_requires)` in `GnrWebPage.__init__` (or storing a copy in the cache) restores the old isolation for one line, and removes a shared mutable from the hot path.

### Notes, not blockers

- **No eviction on page close.** `site.onClosedPage` → `register.drop_page` (`gnrwsgisite.py:1426-1428`) already fires when a page dies; dropping the entry there would bound the cache by live pages instead of by 2000 classes per worker (× workers). Related: switching the preference off skips `store` but leaves the cache populated, so the kill switch does not actually drop the classes already built — a `clear()` on the off transition would make it complete.
- **The flag is cached per site, but `getPreference` resolves per dbstore** (`gnrwsgisite.py:1636-1644`). On a multidb instance the value that wins for the next 10s is whichever store's request happened to refresh the TTL. Fine for a kill switch; worth a line in the comment so nobody reads it as per-store.
- **No test.** The three helpers are the easy kind to pin: hit, miss, LRU eviction at the ceiling, and the TTL boundary on the flag. `gnrpy/tests/web/` is already the right home. The 130-test bridge suite green with the flag on says the cache does not break anything; it does not say the LRU or the TTL do what they claim.

Lint is clean under the repo's `.flake8` on both files, and CI is green on 3.11–3.14.

genro:
Both changes are in, plus the three notes. Two places where I read the evidence differently than you did, and one where the problem is worse than described.

**1. `(page_id, module_path)` — done, but the write is the sharper argument.** The read-side confusion needs an attacker to present a `page_id` of their own connection, so they get a class whose methods they could already call on its own url: no escalation, as you say. What convinced me is the **write**. On a miss under a mismatched url, `store_page_class_cache` deposits the class built for url B under page_id P — and from then on the *legitimate* page A on that worker runs B's class, with B's `auth_tags`, `maintable` and `_packageId`, against A's `page_item` from the register. That is a wrong entry surviving the request that created it. With the pair as key the anomalous request simply misses, builds fresh, and files itself under `(P, module_B)`, leaving `(P, module_A)` intact. I also checked `genro_rpc.js:322-325`: `kw.url` is an overridable slot (`kw.url || pageIndexUrl()`), so the pair being stable today is a property of current callers, not of the protocol. The pair key costs one tuple.

**2. `css_requires` / `js_requires` — done, and the leak is bigger than the mixin path.** The concurrency angle I'd downgrade: `__onmixin__` does `if css not in …: append`, so under the GIL the worst a race produces is a duplicate entry, not corruption. What you didn't flag is `gnrhtmlpage.py:140-141` and `152-153`: `css_requires = getattr(self, 'css_requires', [])` followed by `css_requires.extend(...)` — no copy, no `if not in`, so the **root render extends the class list in place on every render**. With a class per request that dies with the class; with a shared class it is unbounded growth with duplicates, and it feeds back through the `css not in page.css_requires` filter at `gnrresourceloader.py:548`. Reachability is as you described — `rootPage()` only, and the guest path at `gnrwebpage.py:301-303` is where a root render carries a `page_id` — but the mechanism was worse than "extra css". Fixed both ways: per-instance copies in `GnrWebPage.__init__` (which closes every vector at once) and a non-mutating copy in `setCssRequires`/`setJsRequires`, which was a latent bug independent of this PR. `GnrSimplePage` needs neither: `page_factory` keeps it off the cache, so its class is always fresh.

**3. The three notes.** Eviction promoted out of the notes and wired on both close routes — `site.onClosedPage` and the `_closed` branch in `_onPageEnd` — so the cache is bounded by live pages and the LRU is only a backstop. Turning the preference off now calls `clear_page_class_cache()`, so the kill switch frees what it built. The per-dbstore resolution of `getPreference` is documented on `page_class_cache_enabled`.

**Tests** in `gnrpy/tests/web/gnrpageclasscache_test.py`: hit, miss on each half of the key, two pages of the same module keeping their own class, LRU eviction at the ceiling and refresh-on-read, drop on close, the TTL boundary in both directions, the off-transition clearing and the on-transition not clearing, and the render leaving the class lists untouched. The last two fail against the unfixed `gnrhtmlpage`. Full `tests/web` green; `flake8` clean on the touched files.

fporcari (dismissed):
Both changes are in, the three notes with them, and I agree with your two corrections: the write side is the sharper argument for the pair key, and `gnrhtmlpage.py:140-141,152-153` extending the class list in place was the wider leak — I had only followed the `__onmixin__` path.

What I verified on this head:

- **Key.** `module_path` now moves above the cache read and both `load_page_class_cache` and `store_page_class_cache` take the pair, so a request naming another page's `page_id` misses and files itself under its own module instead of overwriting the legitimate entry (`gnrresourceloader.py:265-272`).
- **List isolation.** `GnrWebPage.__init__` shadows both class lists with per-instance copies (`gnrwebpage.py:211-217`), and the class attribute is always a list by the time `__init__` runs — `get_page_class` sets `page_class.css_requires = []` before extending it, and `classMixin` excludes the three `*_requires` names from the attribute copy (`gnrlang.py:363`), so the `list(...)` can never be handed a comma-separated string. `setCssRequires`/`setJsRequires` copying instead of extending is a fix worth having on its own.
- **The tests are real, not decorative.** Reverting only the `gnrhtmlpage` hunk on your branch makes exactly `test_root_render_does_not_extend_the_class_{css,js}_requires` fail, and nothing else. Full `gnrpy/tests/web`: 181 passed, 57 skipped. `flake8` clean on the six touched files.
- **Eviction.** Both routes present (`gnrwsgisite.py:1428` and `gnrwebpage.py:631`), and `drop_page_class_cache` sweeps every module under the closed `page_id`.

Two notes for later, neither of them blocking:

- `onClosedPage` runs in the worker that handles that RPC, so on a multi-worker gunicorn the other workers keep their entry until the LRU reclaims it. The cache is bounded either way; it is the "bounded by live pages" claim that holds per worker, not per site.
- `self.page_factories[module_path] = page_class` (`gnrresourceloader.py:254`) is written and never read anywhere in the repo. It predates this PR and the cache hit now skips it, which changes nothing — but it is dead weight sitting one line from the new store.

Approving.

cgabriel (dismissed):
LGTM - tested also with a simple inspector page (available if considered useful), and under concurrency.

cgabriel (dismissed):
LGTM, I've just added a simple webpage based tool to inspect/clear the page, as confirmed by @genro

fporcari (dismissed):
Re-approving on `b2615185`: my previous approval was dismissed by the push, and the cache itself is unchanged since I read it — the new commit only adds the `/sys/page_class_cache` inspector and its menu entry. `auth_main` and the menu tags agree (`superadmin,_DEV_`), `setItem(str(i), None, dict(...))` passes the dict as `_attributes` as the signature expects (`gnrbag.py:1536`), and `AttributesBagRows` with a `bagStore` on attribute rows is the same idiom as `newregisterexplorer.py:48`. An inspector plus a manual clear is a good thing to ship alongside a cache that is otherwise invisible.

One fix worth making, small and mechanical.

**`getCacheEntries` iterates the cache without the lock.** `getCacheEntries` walks `self.site.resource_loader._page_class_cache.items()` directly while every other reader of that dict holds `_page_class_cache_lock` — and `load_page_class_cache` calls `move_to_end` on a hit, i.e. every RPC on a live page mutates the ordering. Iterating an `OrderedDict` while another thread reorders or inserts raises `RuntimeError: OrderedDict mutated during iteration`; I reproduced it in isolation to be sure it is the real failure and not a theoretical one. On a quiet instance the page will open fine, and on a busy one it will occasionally 500 — which is precisely when someone is looking at the cache.

It also reaches past the four public helpers into the private attribute, and no other `sys` webpage does that with the resource loader. Both go away with one accessor next to the others in `gnrresourceloader.py`:

```python
    def page_class_cache_entries(self):
        """A snapshot of the cache: (page_id, module_path, class_name) per entry."""
        with self._page_class_cache_lock:
            return [(page_id, module_path, page_class.__name__)
                    for (page_id, module_path), page_class in self._page_class_cache.items()]
```

and the webpage becomes:

```python
        for i, (page_id, module_path, class_name) in enumerate(
                self.site.resource_loader.page_class_cache_entries()):
            result.setItem(str(i), None, dict(page_id=page_id, module_path=module_path,
                                              class_name=class_name))
```

Not blocking, so approving as is — but I would take it before merge rather than after.

Worth a line in the page title or an empty-grid message, too: both the listing and **Clear Cache** act on the worker that happens to serve the request, so on a multi-worker gunicorn the grid shows one worker's entries and the button clears one worker's cache. Same limitation I noted for `onClosedPage`; it just becomes visible here, where a reader will naturally read the grid as the whole instance.

cgabriel:
> One fix worth making, small and mechanical.
> 
> **`getCacheEntries` iterates the cache without the lock.** `getCacheEntries` walks `self.site.resource_loader._page_class_cache.items()` directly while every other reader of that dict holds `_page_class_cache_lock` — and `load_page_class_cache` calls `move_to_end` on a hit, i.e. every RPC on a live page mutates the ordering. Iterating an `OrderedDict` while another thread reorders or inserts raises `RuntimeError: OrderedDict mutated during iteration`; I reproduced it in isolation to be sure it is the real failure and not a theoretical one. On a quiet instance the page will open fine, and on a busy one it will occasionally 500 — which is precisely when someone is looking at the cache.


Agreed. I've added the proposed page_class_cache_entries method to the ResourceLoader, which implements locking, and then used it in the inspector webpage.
 
> Worth a line in the page title or an empty-grid message, too: both the listing and **Clear Cache** act on the worker that happens to serve the request, so on a multi-worker gunicorn the grid shows one worker's entries and the button clears one worker's cache. Same limitation I noted for `onClosedPage`; it just becomes visible here, where a reader will naturally read the grid as the whole instance.

Being the cache per-process, that's obviosuly the case, due to the nature of the architecture of the cache. Only a centralized external caching service would allow a global vision and a global clear action of all the entries.

cgabriel (approved):
Only changes are in the cache access method, using locking, using it  in the webpage, as suggested on behalf of @fporcari

fporcari (approved):
`ed46d7b9` takes the accessor and the webpage now goes through it, so the inspector reads a snapshot taken under `_page_class_cache_lock` and no longer touches the private dict. That closes the `RuntimeError: OrderedDict mutated during iteration` window — which was reachable on any instance busy enough for `load_page_class_cache` to be reordering entries while the page rendered.

Verified on this head: `flake8` clean on both touched files, `gnrpy/tests/web` 181 passed / 57 skipped. Cache semantics unchanged from the head I reviewed in full.

Re-approving (the previous approval was dismissed by the push, as `dismiss_stale_reviews` does).

The per-worker scope of both the grid and **Clear Cache** is still there and still worth a word in the UI, whenever it is convenient — it is a labelling matter, not a correctness one.

PR #882, opened 2026-05-06
feat(#874): unified on-event cleanup (replaces #876)

genro:
## Summary

Rewrite of #876 (closed without merge) with a single unified cleanup mechanism instead of two parallel ones (in-memory register expire + filesystem folder purge).

## Why a rewrite

The original #876 introduced an on-event folder purge alongside the legacy `SiteRegister.cleanup()` (in-memory register expire on every request). [@fporcari's review](https://github.com/genropy/genropy/pull/876#issuecomment-4377258484) flagged two blocking issues, both stemming from the parallel-cleanup design:

1. `register.connections().keys()` raised at runtime (fake test register's shape didn't match production).
2. `claim_cleanup` shared `last_cleanup` with the legacy cleanup -> folder purge would essentially never fire.

The right fix was not to patch the parallel design but to consolidate. This PR does that.

## Design

`onClosePage` triggers `_maybeRunCleanup` (lottery + atomic claim). When both gates clear, a daemon thread runs `_runCleanup` which walks `data/_connections/` once:

- orphan connection folders older than `connection_max_age` -> `rmtree`
- live connections -> `expire_pages` (drop stale pages from register + `rmtree` per-page subfolders) + `expire_connection` (drop + `rmtree` connection folder if stale)

Per-request thread-local cleanup (`GnrWsgiSite.cleanup`) is unaffected — it's a different thing despite the name collision.

## Commits

- **`b86085e64d`** — `refactor(#874): drop legacy startup-time cleanup gates`. Removes `counter`/`noclean`/`debug` gating around `onInited()`, drops init-time `dropConnectionFolder()`, drops `--noclean` / `--counter` CLI flags. `onInited()` now runs on every fresh `GnrWsgiSite` instantiation. No-op today since the only legacy `onSiteInited` listener was removed in #877.

- **`c59148396a`** — `feat(#874): unified on-event cleanup mechanism`. Removes `SiteRegister.cleanup()` and the `register.cleanup()` call in the request dispatcher. Adds `claim_cleanup`, `expire_pages`, `expire_connection` on `SiteRegister`. Adds `_maybeRunCleanup` and `_runCleanup` on `GnrWsgiSite`. Renames config keys: `cleanup_threshold` (default 5%), `cleanup_interval_minutes` (default 240). Default `connection_max_age` raised from 600 to 7200 (2h). CLI `gnrcleanup` is now wired to `_runCleanup` (forced cleanup, bypasses lottery). 17 unit tests.

## Config changes

[...]

PR #1069, opened 2026-08-17
fix(web): _runCleanup reads the register client dict, not a list of items

### LINKED ISSUE #956: Regression in #882 - _runCleanup will never execute.
A regression has been introduce in #882, here:

https://github.com/genropy/genropy/blob/2c0d6c1ade65a79e745ddaa292e7af7d836a4ea8/gnrpy/gnr/web/gnrwsgisite.py#L1774-L1776

Verified by running 'gnr web cleanup <instance>'

Itering 'self.register.connections()' , which is a dict, returns the keys of the register, which are string, so trying to access index 'register_item_id' will lead to a TypeError, since string indices must be integers:

```
In [13]: for c in s.register.connections(): print(c['register_item_id'])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[13], line 1
----> 1 for c in s.register.connections(): print(c['register_item_id'])

TypeError: string indices must be integers, not 'str'
```

So it's either iter on `self.register.connections().values()`, which will return the register dict, or more simply, since the 'register_item_id' value is the key of the register itself, just drop the __index__ call to 'register_item_id', doing something like:

```
 try:
            live_connections = {*self.register.connections()}

``

### PR DESCRIPTION
**Pull request by fporcari**
## Problem

`GnrWsgiSite._runCleanup` is a no-op whenever the site has at least one live connection: the orphan-folder cleanup worker silently stops expiring stale pages/connections and dropping their filesystem folders.

## Root cause

At `gnrpy/gnr/web/gnrwsgisite.py:1770-1772` the code did:

```python
live_connections = {c['register_item_id']
                    for c in self.register.connections()}
```

`self.register` on `GnrWsgiSite` is a `SiteRegisterClient`. Its `connections()` (`gnrpy/gnr/web/daemon/siteregister_client.py:245-253`) pipes the daemon's list through `adaptListToDict`, which returns a **dict** keyed by `register_item_id`, not a list of dicts. Iterating that dict yields its string keys, so `c['register_item_id']` raises `TypeError: string indices must be integers`.

That exception is caught by the blanket `except Exception` at line 1773, which logs `"Cleanup failed reading register"` and returns — turning a hard error into a silently degraded no-op for months.

The list-of-dicts idiom is correct in the code it was lifted from: `SiteRegister.connections()` (`gnrpy/gnr/web/daemon/siteregister.py:657`) runs daemon-side and does return a list. The bug was introduced when the idiom was copied into `GnrWsgiSite` (PR #882, `c59148396`, re-applied by `4cabcd5f0`) and crossed the client/daemon boundary where the return shape changes. Every other caller of `register.connections()/pages()/users()` in the codebase already treats the result as a dict (e.g. `connFolderRemove` 880 lines above in the same file, `adm/model/connection.py`, `register_explorer.py`, `system_dashboard.py`) — this was the only broken call site.

The regression tests didn't catch this because `_FakeRegister.connections()` in `gnrpy/tests/web/gnrwsgisite_folder_cleanup_test.py` returned a list of dicts — the daemon-side contract, not the client contract that production code actually calls against. The wrong fake is why the bug shipped green.

## Change

- `gnrpy/gnr/web/gnrwsgisite.py`: read `self.register.connections()` directly as the dict it is, instead of rebuilding it into a set via the wrong indexing.
- `gnrpy/tests/web/gnrwsgisite_folder_cleanup_test.py`: fixed `_FakeRegister.connections()` to return the real client-side shape (`{register_item_id: item}`) instead of a list, so the fake actually encodes the contract production code depends on.

## Verification

- Regression guard: with the fake fixed but production left unfixed, 4 of the live-connection tests in `gnrwsgisite_folder_cleanup_test.py` fail with the exact predicted `TypeError: string indices must be integers, not 'str'`. With both fixed, all 19 tests in that file pass.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/web/gnrwsgisite_folder_cleanup_test.py -q` → 19 passed.
- `flake8` on both touched files → clean.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` → 787 passed, 57 skipped, 0 failed. (Pre-commit hook additionally ran the full suite including `sql/`: 1958 passed, 67 skipped, 0 failed.)
- Not verified live: running `gnr web cleanup <instance>` end-to-end against a real site/register — no live instance or site creation available in this environment. This should be exercised manually against a staging instance with at least one open connection before/after the fix to confirm orphan connection folders are actually reclaimed.

## Note

The blanket `except Exception` at `gnrwsgisite.py:1773` is what let this ship silently for months — a hard `TypeError` was reduced to a single log line with no alerting. Worth a follow-up to narrow that catch (or at least alert on unexpected exception types) so a similar contract mismatch surfaces immediately next time. Left out of this diff to keep it focused on the actual bug.

Fixes #956


cgabriel (approved):
Verified, fix is clean, unable to reproduce the problem anymore

genro (approved):
Correct, and the root-cause account is accurate: `SiteRegisterClient.connections()` pipes the daemon's list through `adaptListToDict` (`siteregister_client.py:245-253`), which returns a dict keyed by `register_item_id`. Iterating it yields string keys, so `c['register_item_id']` raised `TypeError` and the blanket `except` turned the whole cleanup into a no-op.

The only use of `live_connections` in `_runCleanup` is `entry not in live_connections` (`gnrwsgisite.py:1797`), which is key membership on the dict exactly as it was set membership on the set — so dropping the rebuild changes nothing else in the function.

Fixing `_FakeRegister.connections()` to return the client-side shape is the part that matters most: the old fake encoded the daemon-side contract, which is why the tests were green against broken production code.

Agreed on the follow-up you flagged: the blanket `except Exception` at `:1782` is what let this ship silently. Worth its own issue rather than widening this diff.

Approving.

PR #1070, opened 2026-08-17
feat(daemon): gate the gnr.web.daemon entry-point override on an explicit request

### LINKED ISSUE #1067: gnr.web.daemon: gate the entry-point override on an explicit request
@cgabriel — the `gnr.web.daemon` entry-point switch you added in #688 (transitory, as that PR states) currently fires on installation alone: if any installed package declares `gnr.web:daemon`, the namespace is replaced for every site in that interpreter, and the classic daemon is no longer reachable from it. Going back to the classic stack means building a separate virtualenv.

Would it be possible to gate it on an explicit request, so the override happens only when it is actually wanted? I have tests in progress where I need the classic stack and an alternative provider in the same environment, choosing per process rather than per virtualenv.

I put together a local proof of concept, only to check that the idea holds — not a proposal for the implementation, which is yours to shape:

```python
_provider = os.environ.get('GNR_DAEMON_PROVIDER')
if _provider:
    _eps = [ep for ep in importlib.metadata.entry_points(group='gnr.web', name='daemon')
            if _provider in (ep.value, ep.dist.name)]
    if not _eps:
        raise ImportError(
            f"GNR_DAEMON_PROVIDER={_provider!r}: no gnr.web:daemon entry-point matches"
        )
    _mod = importlib.import_module(_eps[0].value)
    _mod.__name__ = __name__
    sys.modules[__name__] = _mod
```

Measured in a throwaway virtualenv (genropy installed, plus an override package alongside it):

| scenario | `gnr.web.daemon` resolves to | Pyro4 imported |
|---|---|---|
| no override installed — baseline | classic package | yes |
| override installed, switch as it is today | the override | no |
| override installed, gated, variable unset | classic package | yes |
| override installed, gated, variable set | the override | no |
| gated, variable naming something absent | `ImportError` at import time | — |

The row that matters to me is the third: with nothing set, the behaviour is identical to the baseline, so it should break nothing for anyone running the classic stack today.

Any shape you prefer works for me — a plain environment check, without the provider name, would already cover my case.

fporcari:
**Draft** — the diff is a starting point for the discussion, not a decision. See **Open decisions** at the bottom: @genro and @cgabriel, none of those points has been silently picked, and every one of them can change the shape of this patch.

Fixes #1067

## Motivation

@genro needs the classic daemon stack and an alternative provider in the **same** environment, choosing between them **per process** rather than per virtualenv. Today that is impossible: the switch fires on installation alone, so as soon as any installed package declares a `gnr.web:daemon` entry point, every site in that interpreter gets the override and going back to the classic daemon means building a separate virtualenv.

## Current mechanism

`gnrpy/gnr/web/daemon/__init__.py` was, in its entirety, these nine lines — added whole by `6a320d0fc` "Transitory refactoring of daemon package and modules (#688)":

```python
import importlib
import importlib.metadata
import sys

_eps = list(importlib.metadata.entry_points(group='gnr.web', name='daemon'))
if _eps:
    _mod = importlib.import_module(_eps[0].value)
    _mod.__name__ = __name__
    sys.modules[__name__] = _mod
```

Four precisions worth stating, all verified against the tree:

- **The baseline is genuinely a no-op.** genropy declares no `gnr.web:daemon` entry point of its own — `gnrpy/pyproject.toml:87-95` has only `[project.scripts]` — so `_eps` is empty unless a third party ships one. Rows 1 and 3 of the table in #1067 are therefore *the same code path*, which is exactly why the gate cannot regress anyone running the classic stack.
- **The override swaps the package, not a module.** Replacing `sys.modules['gnr.web.daemon']` means `.handler`, `.siteregister`, `.siteregister_client`, `.processes` and `.service` all resolve through the provider's `__path__`; the classic modules become unreachable by name.
- **`_eps[0]` picks an undefined winner** when two providers are installed — entry-point ordering is not a contract.
- **`_mod.__name__ = __name__` mutates the provider module object globally.** The provider also stays in `sys.modules` under its real name, now with a lying `__name__`.

Minor factual correction to the table in the issue: `import gnr.web.daemon` alone does **not** import Pyro4. Pyro4 enters at `gnr/web/daemon/handler.py:15` and `gnr/web/daemon/siteregister.py:24`. There is a test for this below.

## Change

~40 lines in `gnrpy/gnr/web/daemon/__init__.py`, following @genro's PoC:

- read `GNR_DAEMON_PROVIDER` and **return early with the namespace untouched when it is unset** — no entry-point scan at all, so the classic path is provably identical to today's (and marginally cheaper);
- when it is set, filter the `gnr.web:daemon` entry points on `ep.value` / `ep.dist.name`;
- raise `ImportError` with an explicit message when nothing matches.

Two additions beyond the PoC, both cheap:

- the selection is extracted into a pure `_resolve_provider(provider, eps)`, so the interesting cases become plain unit tests with no `sys.modules` surgery;
- `logger.warning` (via `from gnr.web import logger`, `gnr/web/__init__.py:2`) when the variable is set and more than one entry point matches, instead of today's silent `_eps[0]`.

Conventions this follows, since #688 is the only precedent in the tree and there is not much of one:

- `GNR_DAEMON_*` is already a live family: `GNR_DAEMON_PORT` at `gnr/web/gnrk8s.py:100`, `GNR_DAEMON_HOST` at `gnrk8s.py:188,293`. `GNR_DAEMON_PROVIDER` sits inside the existing namespace.
- Import-time `os.environ.get` at module level is the established idiom — `gnr/web/gnrwebpage.py:43`, `gnr/web/gnrtask_new.py:47-49`, `gnr/core/gnrenv.py:13-21`.
- A presence-only gate, if the on/off variant is preferred instead, has its own precedent: `gnr/core/cli/gnr.py:159`, `if "GNR_AUTOCOMPLETE" in os.environ:`.
- There is **no** generic env-switch helper to reuse. `getEnvironmentItem` (`gnr/core/gnrconfig.py:219`) reads `environment.xml`, which is per-installation — precisely the granularity @genro wants to escape — and the daemon's `getFullOptions` (`gnr/web/daemon/handler.py:53-66`) runs long after import. An environment variable is the only lever available at import time.
- `importlib.metadata.entry_points` occurs **exactly once in the whole tree**: this file. There is no entry-point-selection convention to match and no other entry-point namespace override anywhere, so whatever lands here becomes the precedent — which is part of why this is a draft.
- `requires-python = ">=3.11"` (`gnrpy/pyproject.toml:18`), so `entry_points(group=, name=)` selection and a populated `ep.dist` are both safe.

One line added to `gnrpy/RELEASE_NOTES.rst`, the only place `GNR_*` variables are documented today (cf. the `GNR_LOGLEVEL` mention).

Who imports the package, which is what makes the `ImportError` question below a real one: `gnr/web/gnrwsgisite.py:45-46` **top level**, `gnr/web/gnrwsgisite_proxy/gnrsiteregister.py:5,12` top level, `gnr/web/cli/gnrdaemon.py:5` (the `gnrdaemon` console script, also reached via `gnr web daemon`), `gnr/app/gnrapp.py:1942` lazy.

## Verification

Side finding first: **existing coverage was zero, and it had silently regressed.** `gnrpy/tests/web/gnrdaemon.py` — the smoke test #688 added in place of the two test modules it deleted — matches neither `test_*.py` nor `*_test.py`, and there is no `python_files` override (`gnrpy/pyproject.toml:196-200` sets only `testpaths` / `pythonpath` / asyncio scope). It has therefore never been collected since #688. This PR `git mv`s it to `gnrpy/tests/web/gnrdaemon_test.py`, keeping its import smoke test, which restores that lost coverage.

That file now carries 11 tests:

- **`_resolve_provider`, pure unit tests** — match by `ep.value`, match by `ep.dist.name`, no match → `ImportError` (row 5), empty entry-point list → `ImportError`, two matches → warning and first (feeding `SimpleNamespace(value=..., dist=SimpleNamespace(name=...))` stubs; no `importlib.metadata` monkeypatching needed once the function is extracted).
- **the gate on a real import** — synthetic provider pre-registered in `sys.modules`, `importlib.metadata.entry_points` faked, `gnr.web.daemon` and its already-imported submodules popped, then re-imported: variable unset → the classic package (`__file__` under `gnr/web/daemon/`, and `.handler` still resolves there) which is rows 1 and 3; variable set and matching → `sys.modules['gnr.web.daemon']` **is** the provider object, row 4; variable set and absent → `ImportError`, row 5 end to end. A fixture restores the original modules — `gnrwsgisite` has already imported the real package by then, so teardown is not optional.
- **out of process** — the "Pyro4 imported" column cannot be asserted in-process (Pyro4 is already in `sys.modules` from earlier tests), so it runs as a subprocess printing `'Pyro4' in sys.modules` after `import gnr.web.daemon` (False) and after `import gnr.web.daemon.handler` (True). A second subprocess pair — same interpreter, same installation, variable unset vs. variable set — is the per-process proof the issue is actually asking for.

Ran: `flake8` on both touched files, clean; `pytest gnrpy/tests/web/gnrdaemon_test.py` → 11 passed; `pytest gnrpy/tests/web/` → 177 passed, 57 skipped (checking the `sys.modules` surgery does not poison siblings); `pytest gnrpy/tests/ --ignore=gnrpy/tests/sql` → 798 passed, 57 skipped. `gnrpy/tests/sql/` was not run — `initdb` fails in this environment for reasons unrelated to the change.

Not verified: no run against a real installed third-party provider package, and no run of an actual daemon process under an override. The provider is synthetic in every test here.

## Open decisions

@genro, @cgabriel — #688 is @cgabriel's and declared transitory there, so the shape is his call. Nothing below has been decided; the diff picks one option per line only so that there is something concrete to react to.

- **Env-var name**: `GNR_DAEMON_PROVIDER` (fits the existing `GNR_DAEMON_HOST` / `GNR_DAEMON_PORT` family), or something else?
- **Named provider vs. plain on/off switch** in the style of `GNR_AUTOCOMPLETE` (`gnr/core/cli/gnr.py:159`), which @genro says would already cover his case?
- **Match key when named**: `ep.value`, `ep.dist.name`, or both as in the PoC (what this diff does)?
- **Variable set but nothing matches**: `ImportError` at import time — which, because `gnrwsgisite.py:45` imports at top level, takes down every site in that process, not just the daemon — or `logger.warning` plus classic fallback?
- **Two or more matching entry points**: silent `_eps[0]` as today, warn-and-pick-first (what this diff does), or hard error?
- **Also expose the choice as a CLI flag** on `gnr web daemon` / `gnrdaemon`, or environment-only?
- **Keep the `_mod.__name__ = __name__` mutation?** And should the classic package stay reachable under an alias, so a provider can delegate to it?
- **Does #688's "transitory" status mean the mechanism is to be replaced outright**, which would make this gate moot?

### DIFF
--- a/gnrpy/gnr/web/daemon/__init__.py
+++ b/gnrpy/gnr/web/daemon/__init__.py
@@ -1,9 +1,57 @@
-import importlib                                                                                                                      
-import importlib.metadata                                       
+"""Daemon package, optionally overridden by an alternative provider.
+
+An installed package can declare a ``gnr.web:daemon`` entry point to replace
+this whole namespace (``.handler``, ``.siteregister``, ``.processes``, ...).
+The override is *not* automatic: it happens only when the
+``GNR_DAEMON_PROVIDER`` environment variable names the provider, so the
+classic stack and an alternative provider can live in the same environment
+and be chosen per process instead of per virtualenv.
+
+With the variable unset this module is a plain package: no entry-point scan,
+no ``sys.modules`` surgery.
+"""
+import importlib
+import importlib.metadata
+import os
 import sys
 
-_eps = list(importlib.metadata.entry_points(group='gnr.web', name='daemon'))
-if _eps:
-    _mod = importlib.import_module(_eps[0].value)
+from gnr.web import logger
+
+DAEMON_PROVIDER_ENV = 'GNR_DAEMON_PROVIDER'
+DAEMON_ENTRY_POINT_GROUP = 'gnr.web'
+DAEMON_ENTRY_POINT_NAME = 'daemon'
+
+
+def _resolve_provider(provider, eps):
+    """Return the ``gnr.web:daemon`` entry point requested by *provider*.
+
+    *provider* is matched against both the entry-point value (the dotted
+    module path) and the name of the distribution declaring it. Raise
+    ``ImportError`` when nothing matches: an explicitly requested provider
+    that is not installed is a configuration error, not a reason to silently
+    fall back to the classic daemon. When more than one entry point matches,
+    warn and pick the first one instead of choosing silently.
+    """
+    matching = [ep for ep in eps
+                if provider in (ep.value, getattr(ep.dist, 'name', None))]
+    if not matching:
+        raise ImportError(
+            f'{DAEMON_PROVIDER_ENV}={provider!r}: no '
+            f'{DAEMON_ENTRY_POINT_GROUP}:{DAEMON_ENTRY_POINT_NAME} '
+            'entry point matches')
+    if len(matching) > 1:
+        logger.warning('%s=%r matches %d %s:%s entry points (%s): using %r',
+                       DAEMON_PROVIDER_ENV, provider, len(matching),
+                       DAEMON_ENTRY_POINT_GROUP, DAEMON_ENTRY_POINT_NAME,
+                       ', '.join(ep.value for ep in matching),
+                       matching[0].value)
+    return matching[0]
+
+
+_provider = os.environ.get(DAEMON_PROVIDER_ENV)
+if _provider:
+    _ep = _resolve_provider(_provider, importlib.metadata.entry_points(
+        group=DAEMON_ENTRY_POINT_GROUP, name=DAEMON_ENTRY_POINT_NAME))
+    _mod = importlib.import_module(_ep.value)
     _mod.__name__ = __name__
     sys.modules[__name__] = _mod
--- a/gnrpy/gnr/web/daemon/__init__.py
+++ b/gnrpy/gnr/web/daemon/__init__.py
@@ -25,26 +25,28 @@ DAEMON_ENTRY_POINT_NAME = 'daemon'
 def _resolve_provider(provider, eps):
     """Return the ``gnr.web:daemon`` entry point requested by *provider*.
 
-    *provider* is matched against both the entry-point value (the dotted
-    module path) and the name of the distribution declaring it. Raise
-    ``ImportError`` when nothing matches: an explicitly requested provider
-    that is not installed is a configuration error, not a reason to silently
-    fall back to the classic daemon. When more than one entry point matches,
-    warn and pick the first one instead of choosing silently.
+    *provider* is matched against both the entry-point module and the name of
+    the distribution declaring it. The module is read from ``ep.module``
+    rather than ``ep.value``: the entry-point spec allows ``module:attr``,
+    which ``ep.value`` carries verbatim and no importable name ever equals.
+
+    Anything other than exactly one match raises ``ImportError``. An
+    explicitly requested provider is a configuration statement, so neither a
+    missing one nor an ambiguous one may degrade into running a daemon the
+    caller did not ask for.
     """
     matching = [ep for ep in eps
-                if provider in (ep.value, getattr(ep.dist, 'name', None))]
-    if not matching:
+                if provider in (ep.module, getattr(ep.dist, 'name', None))]
+    if len(matching) != 1:
+        installed = ', '.join(sorted(
+            f'{ep.module} ({getattr(ep.dist, "name", "unknown distribution")})'
+            for ep in eps)) or 'none'
         raise ImportError(
-            f'{DAEMON_PROVIDER_ENV}={provider!r}: no '
-            f'{DAEMON_ENTRY_POINT_GROUP}:{DAEMON_ENTRY_POINT_NAME} '
-            'entry point matches')
-    if len(matching) > 1:
-        logger.warning('%s=%r matches %d %s:%s entry points (%s): using %r',
-                       DAEMON_PROVIDER_ENV, provider, len(matching),
-                       DAEMON_ENTRY_POINT_GROUP, DAEMON_ENTRY_POINT_NAME,
-                       ', '.join(ep.value for ep in matching),
-                       matching[0].value)
+            f'{DAEMON_PROVIDER_ENV}={provider!r} matches {len(matching)} '
+            f'{DAEMON_ENTRY_POINT_GROUP}:{DAEMON_ENTRY_POINT_NAME} entry '
+            f'points; installed: {installed}')
+    logger.info('%s=%r resolved to %s', DAEMON_PROVIDER_ENV, provider,
+                matching[0].module)
     return matching[0]
 
 
@@ -52,6 +54,6 @@ _provider = os.environ.get(DAEMON_PROVIDER_ENV)
 if _provider:
     _ep = _resolve_provider(_provider, importlib.metadata.entry_points(
         group=DAEMON_ENTRY_POINT_GROUP, name=DAEMON_ENTRY_POINT_NAME))
-    _mod = importlib.import_module(_ep.value)
+    _mod = importlib.import_module(_ep.module)
     _mod.__name__ = __name__
     sys.modules[__name__] = _mod
--- a/gnrpy/tests/web/gnrdaemon.py
+++ b/gnrpy/tests/web/gnrdaemon.py
@@ -1,1 +1,0 @@
-import gnr.web.daemon  # noqa: F401
--- a/gnrpy/tests/web/gnrdaemon_test.py
+++ b/gnrpy/tests/web/gnrdaemon_test.py
@@ -8,7 +8,6 @@ looked up.
 """
 import importlib
 import importlib.metadata
-import logging
 import os
 import subprocess
 import sys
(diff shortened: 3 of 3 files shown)


genro:
Answering the open decisions. Verified against `genropy-asgi`, which is the provider this exists for.

**Named provider, not the on/off switch.** Keep what you wrote. The name is what makes refactorings and experiments manageable — being able to point at a specific provider matters more than saving the typing, and the on/off variant cannot choose once two providers are installed side by side.

**Env-var name:** `GNR_DAEMON_PROVIDER` is fine.

**Match key:** both, as you have it — but use `ep.module`, not `ep.value`. `EntryPoint.value` is the raw field: the PyPA spec allows `module:attr [extras]`, and `importlib.metadata` exposes `.module` / `.attr` / `.extras` for exactly this reason. With `ep.value` a provider declared as `pkg.mod:obj` would fail both the match and the `import_module`. Not a problem for `genropy-asgi`, which declares `daemon = "genropy_asgi.siteregister"` — a bare module path — but `.module` is free and correct.

**Nothing matches:** keep the `ImportError`. An explicitly requested provider that is not installed is a configuration error. Please add the installed providers to the message though — with `genropy_asgi.siteregister` and `genropy-asgi` both being valid spellings, a typo is likely and the current text does not help.

**Two or more matches:** hard error, not warn-and-pick-first. Same class as no-match: an ambiguous explicit request is a configuration error, and a warning leaves a process running the wrong daemon.

**CLI flag:** environment only. The import happens at module level (`gnrwsgisite.py:45`, `gnr/web/cli/gnrdaemon.py:5`), so the decision is already made before argument parsing. A flag would need a re-exec, which is not worth it.

**`_mod.__name__` mutation and a classic alias:** no alias needed. I checked `genropy_asgi.siteregister`: it imports nothing from the classic package — `handler`, `service` and `processes` are declared fakes that exist only so the `gnr.web.daemon.*` namespace resolves. It replaces, it does not delegate. So there is no delegation case to design for right now, and I would not add an alias on speculation.

**`RELEASE_NOTES.rst`:** drop the stanza from this PR. Same question I raised on #1089 — the `Unreleased` heading has no precedent in that file and @cgabriel owns how it is written.

**@cgabriel — the one blocking answer.** #688 declares the entry-point switch transitory. If you intend to replace the mechanism outright, this gate is wasted work and we should stop here. If it stays, the shape above is what we would like.

Last point, on me rather than you: every provider in your tests is synthetic, and you say so. The end-to-end run — a real daemonless `genropy-asgi` process against a real site — I will do myself, since it is my use case. I would treat it as a merge condition.

fporcari:
> **Match key:** both, as you have it — but use `ep.module`, not `ep.value`.
> — [comment of 19 Aug](https://github.com/genropy/genropy/issues/1070#issuecomment-5340703424)

All four points are in on `991fbfdb`, rebased on current develop:

- **`ep.module`** replaces `ep.value` in both the match and the `import_module` call. You are right that it was broken twice for a `module:attr` provider: it matched nothing, and had it matched, the raw value would have been handed to `import_module`.
- **No-match and multiple-match are now the same hard error** — `len(matching) != 1` raises `ImportError`, and the message lists the installed providers as `module (distribution)` pairs, so a typo in either accepted spelling is visible immediately.
- **The `RELEASE_NOTES.rst` stanza is gone**; the branch no longer touches the file at all, which is also what was keeping it in conflict.
- The resolved provider is logged at `info` on success.

On the tests: the entry-point stub is now a real `importlib.metadata.EntryPoint` instead of a `SimpleNamespace`, so `.module` is derived from `.value` by the library rather than asserted by the stub — a hand-built stub declaring both fields would agree with whatever the code happens to read. The `module:attr` case has its own test, and I verified it fails against the previous `ep.value` code before fixing it.

The synthetic-provider limitation stands as you describe it; the end-to-end run against a real `genropy-asgi` process is yours, and I agree it is a merge condition.

On #688 — I read the thread with @cgabriel: with the transitory rationale explained and the ask narrowed to "an installation must not decide a runtime behaviour", this PR is the shape of exactly that and nothing more. Ball is with the reviewers; both of you are requested on the current head.

genro (approved):
Approving `991fbfdbe`. Verified the diff against the four points.

`ep.module` is now used both in the match and in `import_module`, the two places where `ep.value` was wrong. Collapsing no-match and ambiguous-match into a single `len(matching) != 1` error is better than the two branches I asked for — the two cases are the same class of configuration mistake and there is no reason to treat them differently — and listing the installed providers as `module (distribution)` pairs covers both accepted spellings, which is exactly where a typo lands.

Your two improvements on what I asked are worth naming. Using a real `importlib.metadata.EntryPoint` in the tests instead of a `SimpleNamespace` means `.module` is derived from `.value` by the library rather than asserted by the stub: a hand-built stub declaring both fields would have agreed with whichever one the code happened to read, so the test would have passed against the broken version too. And verifying the `module:attr` case fails against the previous code before fixing it is what makes that test a regression test rather than a description.

One inezia, not worth a commit on its own: `eps` is iterated twice, once for `matching` and once for `installed`. With `entry_points(group=, name=)` it is an `EntryPoints` — I checked, it is re-iterable — so this is correct as written. It would silently report `none` if a generator were ever passed to `_resolve_provider` directly, which only the tests do and they pass lists.

Dropping the `RELEASE_NOTES.rst` stanza also took the branch out of conflict, which was the other thing it was costing.

Agreed on the framing in your last paragraph: with the ask narrowed to "an installation must not decide a runtime behaviour", this is that and nothing more. The end-to-end run is mine and I treat it as a merge condition — I will report here.

cgabriel (approved):
Verified, approved. Tested in real environment the runtime switch.

I'll update the other daemon demonstrator to accomodate this change

cgabriel:
Assigned the PR to me as requested by @genro

PR #1071, opened 2026-08-17
fix(sys): evaluate task schedules against local wall-clock time

### LINKED ISSUE #975: [BUG] Task scheduler matches hour/minute against UTC time: tasks run shifted by the local timezone offset
### Problem Description

Since the introduction of `gnr.core.gnrdatetime` (commit 777c18576, *Adds timezone-aware datetime utility class*), `datetime.now()` returns a timezone-aware datetime **in UTC by default**.

`sys.task.findTasks()` calls `datetime.now()` without passing a timezone and hands the resulting **UTC** timestamp to `isTaskScheduledNow()`, which compares `timestamp.hour` / `timestamp.minute` directly against the `hour` / `minute` fields configured on the task record:

```python
# projects/gnrcore/packages/sys/model/task.py
def findTasks(self, timestamp=None):
    timestamp = timestamp or datetime.now()   # <- aware UTC now
    ...
def isTaskScheduledNow(self, task, timestamp):
    ...
    y,m,d,h,minutes = timestamp.year, timestamp.month, timestamp.day, timestamp.hour, timestamp.minute  # <- UTC fields
```

The `hour`/`minute` fields are entered by the user and semantically represent **local wall-clock time** of the instance (e.g. Europe/Rome). Matching them against the UTC hour shifts every cron-style task by the timezone offset.

**Observed:** a task configured with `hour=9` on an instance whose environment timezone is Europe/Rome is executed at **11:00 local time** (09:00 UTC, offset +02 in DST; 10:00 in winter). The computed schedule reason / `last_scheduled_ts` is also based on 09:00 UTC.

**Expected:** the task runs at 09:00 local wall-clock time in the instance's timezone.

Both scheduler implementations are affected:

- Legacy scheduler: `GnrTaskScheduler.writeTaskExecutions` (`gnrpy/gnr/web/gnrtask.py`) → `findTasks()` → `datetime.now()` (UTC) → `isTaskScheduledNow()`
- New async scheduler: `TaskItem.is_due()` in `gnrpy/gnr/web/gnrtask_new.py` explicitly uses `datetime.now(timezone.utc)` and compares `timestamp.hour`/`timestamp.minute` against the task's `hour`/`minute` fields

This is a regression: before the switch to `gnrdatetime` (import introduced in `task.py` by #434), `datetime.now()` returned the naive local time, so the comparison happened in local wall-clock time as intended.

### Reproduction Steps

1. Run an instance in an environment with timezone Europe/Rome (UTC+2 in summer).
2. Create a record in `sys.task` with `hour=9`, `minute=0` (no `frequency`, not stopped).
3. Start the task scheduler and wait.

Observed result: the task is not picked up at 09:00 local time; it is scheduled at 11:00 local time (i.e. 09:00 UTC).

### Environment

genropy `develop` (regression introduced with `gnrdatetime`, commit 777c18576; `task.py` switched to it in #434)

### Additional Material

Possible directions for a fix (to be discussed):

- Convert the reference timestamp to the instance's local timezone before extracting `year/month/day/hour/minute` in `isTaskScheduledNow()` / `is_due()`, keeping UTC for storage (`last_scheduled_ts`, `last_execution_ts` are `DHZ`).
- Longer term: make the scheduling timezone explicit (per instance or per task) instead of relying on the process-local timezone.

fporcari:
## Problem

The `month`/`day`/`hour`/`minute` columns of `sys.task` are entered by the user and mean **local wall-clock time**, but both schedulers compared them against a **UTC** timestamp. Every cron-style task therefore fired shifted by the local offset: a task set for `hour=9` on a Europe/Rome instance ran at 11:00 local time (09:00 UTC) in summer, 10:00 in winter.

## Root cause

Two independent occurrences of the same mistake, one per scheduler.

**Legacy path** — `gnrpy/gnr/core/gnrdatetime.py` binds the module-level name `datetime` to `TZDateTime`, whose `now()` defaults `tz` to `timezone.utc`. So `datetime.now()` imported from that module is UTC-aware, and `findTasks()` handed a UTC timestamp to `isTaskScheduledNow()`, which unpacks `.hour`/`.minute` and compares them against the record. Verified in this checkout: `gnrdatetime.datetime.now()` returns `09:43+00:00` while local (Europe/Rome) is `11:43+02:00`.

**Provenance correction.** The issue points at `777c18576`, but that commit only added the module. The regression is **`959c845c7`** (fix for #434), which swapped `from datetime import datetime` for `from gnr.core.gnrdatetime import datetime` in `task.py`. It fixed an aware/naive `TypeError` in the `frequency` branch and silently broke the wall-clock branch.

**Async path** — `GnrTask.is_due()` took its default timestamp from `datetime.now(timezone.utc)` (stdlib). This one is **not a regression**: it was never correct for a deployment whose timezone is not UTC.

`gnrpy/gnr/web/gnrtask.py` needs no change: its own `now` (stdlib `datetime.now(timezone.utc)`) only writes `last_scheduled_ts` and must stay UTC.

## Change

The timezone decision goes into the one designated facade rather than being duplicated at the two call sites:

- `gnr.core.gnrdatetime.localnow()` — returns the same instant as `now()`/`utcnow()` carrying the **local** offset, resolved with `tzlocal.get_localzone()`. `tzlocal` is already a declared dependency (`gnrpy/pyproject.toml`) and is already used exactly this way in `gnrclasses.py` and `gnrdate.py`; it returns a real `zoneinfo.ZoneInfo`, so DST is handled. No new API was needed on the call side, since `TZDateTime.now(tz)` already accepts a zone.
- `projects/gnrcore/packages/sys/model/task.py` → `findTasks()` defaults to `localnow()`.
- `gnrpy/gnr/web/gnrtask_new.py` → `GnrTask.is_due()` defaults to `localnow()` and honors the configured `minute` key (fixing the pre-existing `minutes` typo).

Deliberately **not** used: `gnrdate.toDHZ(..., timezone='LOCAL')`, which feeds a `tzname()` abbreviation like `"CEST"` into `pytz.timezone()` and then `.replace(tzinfo=...)`; and `db?use_timezone`, which is a dtype switch (DHZ vs DH), not a zone.

**Storage is unaffected and stays UTC.** Every write of `last_scheduled_ts`/`last_execution_ts` uses stdlib UTC and is untouched (`gnrtask.py`, `gnrtask_new.py`). `astimezone`/`now(tz)` preserve the instant, so the `frequency` branches keep producing identical deltas: aware minus aware is offset-independent.

## Deployment requirement

The scheduler process timezone is the source of local wall-clock time. Deployments must set `TZ` in the scheduler runtime environment; the generic Docker image and generated instances intentionally provide no timezone default. Shared deployment secrets/environment configuration are the supported way to supply it. Without `TZ`, a UTC container remains on UTC and this fix cannot infer the business timezone.

## Verification

- `flake8` clean on all six PR files.
- Focused scheduler perimeter: **40 passed**.
- Full suite with local PostgreSQL available: **2053 passed, 69 skipped**.
- The async scheduler tests now include a negative regression case proving that a task configured for 12:15 is not due at 12:05.
- The timezone tests pin both clocks to the same instant and fail on the reverted tree with the wrong schedule outcome, not with a missing symbol.
- Not verified: no additional run against a live scheduler instance in this update. cgabriel reported testing the preceding timezone head in a real environment with several configured tasks.

## Adjacent defects, not fixed here

The four independent defects identified during review are tracked separately:

1. [#1116 — Async scheduler re-enqueues calendar tasks every second](https://github.com/genropy/genropy/issues/1116)
2. [#1117 — Async scheduler replaces stored timezone offset instead of converting it](https://github.com/genropy/genropy/issues/1117)
3. [#1118 — Legacy task frequency comparison discards elapsed days](https://github.com/genropy/genropy/issues/1118)
4. [#1119 — Task scheduler ignores configured weekdays](https://github.com/genropy/genropy/issues/1119)

## Decisions resolved

- Process-local timezone is accepted for this fix; deployment supplies `TZ`.
- Per-task timezone support remains outside this PR.
- A possible one-time reason-key discontinuity at deployment is accepted; pending executions are not cleared.
- Existing cron-style DST behavior remains unchanged. Async duplicate scheduling is tracked in #1116.

Fixes #975


cgabriel (dismissed):
LGTM, tested in real enviroment with several configured tasks. I'd address the reported typo in this PR as well.

dgpaci:
Answers to the four open points, plus where I'd draw the scope line.

### 1. Process-local now, per-instance config later — but the container case must be called out

Process-local is the right hotfix: it is the same notion of "local" the framework already uses elsewhere (`gnrclasses.serialize_datetime` and `gnrdate` both resolve through `tzlocal`), so we are not introducing a second answer to "what time is it here". And since `tzlocal` honours `TZ`, ops already has the knob without any new config surface.

The caveat is concrete, not hypothetical: `dockerfiles/genropy/Dockerfile` is `python:3.11-slim` with no `TZ` set, so inside the official image `get_localzone()` resolves to UTC and this fix is a no-op — precisely for the deployment shape most likely to hit the bug. Please state it in the PR body and in the release note ("the scheduler process must run with the business timezone; set `TZ` in the container environment"), and let's decide separately whether the image / supervisor config should default it.

A per-instance `timezone` attribute in instanceconfig is the correct end state, but it goes in its own issue — don't grow this PR.

### 2. No — and keep `localnow()` zone-agnostic

`localnow()` should stay argument-free: it answers "what time is it here", nothing more. When a task-level (or instance-level) zone eventually arrives, the conversion belongs where the wall-clock fields are read — `isTaskScheduledNow` doing `timestamp.astimezone(zone)` immediately before unpacking `.year/.month/.day/.hour/.minute`. One instant flows through the scheduler, localised at the point of comparison.

If you want that seam in place now it is nearly free: leave `findTasks()` on UTC and put the single `astimezone` at the top of the wall-clock branch. Identical behaviour today, right layering for later. Your call, not a blocker.

### 3. Accept it, and do not clear `sys.task_execution`

The framing needs one correction: the reason key is built from the **matched schedule slot**, not from the timestamp's hour, so the namespace does not shift wholesale. Checked against the real function, same instant evaluated in both zones:

- single-slot daily task (`hour=9`) → `2026-8-20-9-0` from both the UTC and the local timestamp. Identical key, no extra run.
- multi-slot task (`hour` empty, `minute=0`) → `2026-8-20-13-0` (UTC) vs `2026-8-20-15-0` (local). Here the key does move, and such a task gets **one** extra run at the deploy instant.

That is the whole exposure: multi-slot schedules, once. Clearing pending rows is strictly worse — `status='waiting'` rows are the queue the workers consume, so deleting them cancels real scheduled work in order to avoid one duplicate run. Note it in the release notes and move on.

### 4. Nothing to fix and nothing to document — the legacy scheduler is already correct-once across both transitions

Verified against `isTaskScheduledNow` with a pinned `Europe/Rome` zone, task `hour=2, minute=0`:

- **spring-forward**: the task is *not* skipped. `max(slot <= now)` catches it up — at 03:00 CEST on 2026-03-29 it returns `2026-3-29-2-0`.
- **fall-back**: both passes over 02:30 on 2026-10-25 (`fold=0` CEST and `fold=1` CET) produce the same `2026-10-25-2-0`, so the `reasonkey` dedup in `writeTaskExecutions` suppresses the second run.

So the DST question resolves itself for `gnrtask.py`. Worth noting that both properties come from catch-up + reasonkey dedup, which `gnrtask_new.py` does not have — so there, fall-back really would double-run. It is the dedup gap (your adjacent point 2), not DST, that needs an issue.

### Scope of the adjacent defects

- **Point 1, the `'minutes'` typo → take it here.** One character in the function you are already touching, and without it the async path still ignores the minute field, so the PR's headline claim only half-holds on that path. Add the negative test (`hour=9, minute=30` not due at 09:05).
- **Point 4, `.seconds` → `.total_seconds()` → happy to see it here too.** One line, and a genuine bug rather than a wart: `timedelta(days=1, hours=1).seconds == 3600`, so after a scheduler outage longer than a day a frequency task can stay stuck indefinitely, and a negative delta reads as 86399. Independent of the timezone story, so a separate issue is acceptable as well.
- **Points 2, 3, 5 → separate issues.** Point 2 is the serious one: `SCHEDULER_RUN_INTERVAL = 1` plus no dedup means re-enqueue every second until midnight.

For the rest the change is right, and putting the timezone decision in the `gnrdatetime` facade is the correct call. The provenance diagnosis of `959c845c7` in particular is exactly on target.

dgpaci:
@cgabriel a deployment question comes out of this PR, and since the image is yours I'd rather have your call than guess.

This fix makes the scheduler evaluate cron-style schedules against `tzlocal.get_localzone()`, i.e. against the **timezone of the scheduler process**. I checked what that means inside our own base image (`python:3.11-slim`), and the results are:

- **no `TZ` set → the fix is a no-op.** There is no `/etc/timezone`, and `/etc/localtime -> /usr/share/zoneinfo/Etc/UTC`, so `get_localzone()` returns `Etc/UTC` and tasks keep firing at the UTC offset, exactly as before the fix. Containerised deployments are the shape most likely to hit #975 and the one where the fix silently does nothing.
- **`TZ=Europe/Rome` alone is sufficient** — no extra package needed. The slim image already ships `/usr/share/zoneinfo`, so `ZoneInfo('Europe/Rome')` resolves and `get_localzone()` returns a real zone with `+02:00`. (Worth stating explicitly because slim images often lack `tzdata`; ours does not.)
- **supervisor needs no change.** `supervisor-genropy.conf` declares no `environment=`, so `gnrdaemon` and its children inherit the container environment — setting the variable at container level is enough to reach the scheduler.

So the remaining question is only *where the knob belongs*:

1. `ENV TZ` in `dockerfiles/genropy/Dockerfile` — a single default for every deployment. Hardcoding a business timezone into a generic framework image feels wrong to me, but it is the one place that can never be forgotten.
2. Per instance, in the `Dockerfile` / compose `environment` that `gnr dockerize` generates — semantically the right level, since the timezone is a property of the business the instance serves, not of the framework. There is already a precedent right there: the generated compose sets `GNR_LOCALE: "IT_it"`, so `TZ: "${TZ:-Europe/Rome}"` next to it would follow the pattern you already established and stay overridable with one variable.
3. Neither — leave it entirely to whoever runs the container, and only document the requirement.

My leaning is 2, with the `${TZ:-...}` default so an override costs one env var. But you own the image and the deploy story: what do you suggest? And is `TZ` perhaps already set at the orchestrator level on the deployed instances, in which case this is a documentation matter only?

Related: open point 1 in the PR description asks whether we want a per-instance `timezone` config attribute in instanceconfig. If option 2 is the answer here, per-instance `TZ` may already cover that need without new framework configuration.

cgabriel:
> @cgabriel a deployment question comes out of this PR, and since the image is yours I'd rather have your call than guess.
> 
> This fix makes the scheduler evaluate cron-style schedules against `tzlocal.get_localzone()`, i.e. against the **timezone of the scheduler process**. I checked what that means inside our own base image (`python:3.11-slim`), and the results are:
> 
> * **no `TZ` set → the fix is a no-op.** There is no `/etc/timezone`, and `/etc/localtime -> /usr/share/zoneinfo/Etc/UTC`, so `get_localzone()` returns `Etc/UTC` and tasks keep firing at the UTC offset, exactly as before the fix. Containerised deployments are the shape most likely to hit [[BUG] Task scheduler matches hour/minute against UTC time: tasks run shifted by the local timezone offset #975](https://github.com/genropy/genropy/issues/975) and the one where the fix silently does nothing.
> * **`TZ=Europe/Rome` alone is sufficient** — no extra package needed. The slim image already ships `/usr/share/zoneinfo`, so `ZoneInfo('Europe/Rome')` resolves and `get_localzone()` returns a real zone with `+02:00`. (Worth stating explicitly because slim images often lack `tzdata`; ours does not.)

Some deployments (like Euriclea) are currently using TZ=Europe/Rome in their environment, via secretRef.

> So the remaining question is only _where the knob belongs_:
> 
> 1. `ENV TZ` in `dockerfiles/genropy/Dockerfile` — a single default for every deployment. Hardcoding a business timezone into a generic framework image feels wrong to me, but it is the one place that can never be forgotten.

Yes, it's wrong.

> 2. Per instance, in the `Dockerfile` / compose `environment` that `gnr dockerize` generates — semantically the right level, since the timezone is a property of the business the instance serves, not of the framework. There is already a precedent right there: the generated compose sets `GNR_LOCALE: "IT_it"`, so `TZ: "${TZ:-Europe/Rome}"` next to it would follow the pattern you already established and stay overridable with one variable.

Nope, because the same image could be used for several deployment in different time zones.

> 3. Neither — leave it entirely to whoever runs the container, and only document the requirement.

Correct.

> My leaning is 2, with the `${TZ:-...}` default so an override costs one env var. But you own the image and the deploy story: what do you suggest? And is `TZ` perhaps already set at the orchestrator level on the deployed instances, in which case this is a documentation matter only?

I'd go with 3, which is the correct way. We have shared secrets now too, so the setting it's pretty easy.

genro (changes requested):
The timezone fix itself is correct and I would approve it as is. I verified the parts that carry it:

- `gnrdatetime.datetime` is `TZDateTime`, whose `now()` defaults to UTC, so `findTasks()` really was comparing user-entered wall-clock fields against a UTC reading — and the provenance correction is right, `959c845c7` is where the wall-clock branch broke while the aware/naive fix landed.
- putting `localnow()` in `gnrdatetime` rather than duplicating the decision at both call sites is the right shape, `tzlocal` is already a dependency used exactly this way, and it returns a real `ZoneInfo` so DST is handled.
- storage stays UTC: every `last_scheduled_ts`/`last_execution_ts` write is untouched, and aware-minus-aware is offset-independent, so the frequency branches are unaffected.
- the tests pin both clocks to the same instant, so they fail on the reverted tree with the wrong *outcome* rather than a missing symbol. That is the assertion that actually proves the offset question.

I am blocking on one thing only, and it is the one cgabriel already asked for in his approval: **the `'minutes'` typo is still there.** Current PR head, `gnrtask_new.py:205`:

```python
minutes = list(map(int, self.schedule.get('minute').split(','))) if self.schedule.get('minutes') else range(0,60)
```

The condition reads `'minutes'`, which is never a key, so the branch always falls back to `range(0,60)` and the user's minute field is discarded. Your own body is explicit that this makes the async scheduler still not behave like cron: `hour=9, minute=30` fires on any minute of the 9 o'clock hour. Shipping the timezone fix while that stands means a task set for 09:30 goes from "runs at 11:00" to "runs every minute from 09:00" — the schedule is still not what the user typed, so the issue is not closed from where the user stands.

It is a one-character fix in the line the fix already touches, `test_task_is_due_matches_calendar_schedule` currently passes by coincidence (the requested minute happens to sit inside the fallback range), so it is worth a test that fails on the typo.

The other four adjacent defects in your list (the `schedule_loop` re-enqueue, `.replace` vs `.astimezone`, `.seconds` vs `.total_seconds()`, unevaluated `weekday`) are genuinely separate — please open them as issues rather than folding them in here.

On the deployment question: cgabriel settled it, option 3 — no `ENV TZ` in the image, no per-instance default, `TZ` set by whoever runs the container, shared secrets making that easy. Worth one line in the PR body or the release notes saying the fix is a no-op on a container with no `TZ` set, since that is the deployment shape most likely to have hit #975 in the first place.

fporcari:
> I am blocking on one thing only [...] the `'minutes'` typo is still there.
>
> — [review of 20 Aug](https://github.com/genropy/genropy/pull/1071#pullrequestreview-4986273067)

Addressed in `64330d68b`: the condition now reads the configured `minute` key, with a negative regression test proving that a 12:15 task is not due at 12:05.

The deployment decision is now explicit in the PR body: the scheduler runtime must supply `TZ`; there is no Docker-image or generated-instance default.

The other defects are tracked separately in #1116, #1117, #1118, and #1119.

Verification:

- focused scheduler perimeter: 40 passed;
- full suite: 2053 passed, 69 skipped;
- commit hook suite: 2022 passed, 69 skipped;
- flake8 clean on all six PR files.

genro (approved):
Approving on `64330d68b`. The one thing I blocked on is gone: the condition reads `self.schedule.get('minute')`, and `test_task_is_due_does_not_run_before_configured_minute` proves it rather than describing it — at 12:05 with `minute: "15"` the slot list is `[735]`, `curr_hm` is 725, so `hmlist` comes back empty and `is_due` returns `False`. Under the typo the fallback `range(0,60)` would have put 720–779 in that list and the assertion would fail. That is a real regression test, not a restatement.

The four adjacent defects are tracked as #1116, #1117, #1118 and #1119 rather than folded in, and the deployment decision is stated in the body. Both were the right calls.

The timezone fix I had already verified and it stands: `localnow()` living in `gnrdatetime` instead of being decided at each call site, storage staying UTC so the frequency branches are offset-independent, and the tests pinning both clocks to the same instant so they fail on the reverted tree with the wrong *outcome* rather than a missing symbol.

One nit for whenever that function is next touched, not for this PR: `minutes` is bound to the slot list and then rebound to `timestamp.minute` a few lines down, and `m` serves first as the inner loop variable and then as the month. Both are harmless because the loop has finished by then, and both predate this change — but that function is now the one place where a scheduling bug costs a whole night of runs, and it deserves names that survive a quick read.

cgabriel (approved):
LGTM, reviewed recent changes.

PR #1072, opened 2026-08-17
fix(#1068): install resource userobjects after the full onDbUpgrade broadcast

### LINKED ISSUE #1068: [BUG] Fresh db bootstrap fails: resource userobjects installed during adm broadcast, before downstream packages create referenced records
### Problem Description

`adm.userobject` installs the resource userobjects (`resources/tables/<tbl>/userobjects/**/*.xml`) through the `onDbUpgrade_checkResourceUserObject` hook. The hook walks the resources of **all** packages, but it runs while the `onDbUpgrade` broadcast is still processing the `adm` package — i.e. before any downstream package had the chance to create its own records (mandatory sysRecords, lookup rows, etc.).

If a userobject record carries a value for a foreign key column that points to a row created later in the same broadcast by a downstream package, bootstrapping a fresh empty database fails with a FK violation.

Observed behavior: `gnr db migrate -f -u` on an empty db aborts with a `ForeignKeyViolation`.
Expected behavior: the bootstrap completes; resource userobjects should be installed only after every package has completed its own `onDbUpgrade` pass.

On existing databases the problem is invisible: the referenced rows already exist and the userobjects are already installed (duplicates are skipped), which is why it only surfaces when bootstrapping from scratch.

### Reproduction Steps

Using the anaci project (where the issue was found):

1. Create a fresh empty PostgreSQL database for the `anaci_prdtest` instance.
2. Run `gnr db migrate -f -u anaci_prdtest`.
3. Observed result:

```
insert or update on table "adm_userobject" violates foreign key constraint
DETAIL:  Key (tipo_codice)=(VENDITA) is not present in table "anc_cnv_convegno_template_tipo".
```

The chain in that project:

- the `anc_cnv` package extends `adm.userobject` with a `tipo_codice` column, FK to `anc_cnv.convegno_template_tipo.codice`;
- the resource userobject `anc_cnv/resources/tables/partecipazione/userobjects/template/istruzioni_servizi.xml` carries `tipo_codice=VENDITA` in its record;
- the `VENDITA` row is a mandatory sysRecord of `anc_cnv.convegno_template_tipo`, created by `onDbUpgrade_createSysRecords` during the `anc_cnv` package broadcast, which runs **after** `adm`.

Generic reproduction: any project where a customization adds a FK column on `adm.userobject` (or stores in a resource userobject a reference) pointing to rows populated during the `onDbUpgrade` broadcast of a package loaded after `adm`.

Reproducible with any empty db on a project matching the conditions above; no customer data involved.

### Environment

genropy develop, PostgreSQL, empty database bootstrap (docker/CI scenario).

### Proposed solution

Introduce a post-upgrade lifecycle event (e.g. `onDbUpgradeDone,onDbUpgradeDone_*`) broadcast by `gnr db migrate` / `gnr app dbsetup` after the full `onDbUpgrade` pkgBroadcast has completed, and move the resource userobject installation (`checkResourceUserObject`) from `onDbUpgrade_*` to the new event.

dgpaci:
## Problem

`adm.userobject` installed resource userobjects (`resources/tables/<tbl>/userobjects/**/*.xml`) through the `onDbUpgrade_checkResourceUserObject` hook. The hook walks the resources of **all** packages but runs while the `onDbUpgrade` broadcast is still on the `adm` package — before downstream packages had the chance to create their own records (mandatory sysRecords, lookup rows). If a resource userobject carries a FK value pointing at one of those rows, bootstrapping a fresh empty database aborts with a `ForeignKeyViolation` (see #1068 for the full reproduction on the anaci project).

## Solution

- New lifecycle event **`onDbUpgradeDone`**, dispatched to packages and tables (`onDbUpgradeDone_*`) after every package completed its own `onDbUpgrade` pass.
- The full sequence is owned by a single library method, **`GnrApp.dbUpgradeBroadcast()`**, so the `onDbUpgrade`/`onDbUpgradeDone` pairing is a mechanism instead of a convention repeated at every call site. Callers: `gnr db migrate` (both branches), `gnr app dbsetup` (both branches) and `multidb_setStartupData` (store provisioning — without this, newly provisioned dbstores would have silently lost the resource userobject install after the hook rename).
- `adm.userobject` installs resource userobjects in `onDbUpgradeDone_checkResourceUserObject`.

## Design decisions

- `onDbUpgradeDone` fires **before** `sys.upgrade.runUpgrades()`: upgrade scripts can rely on resource userobjects being already installed. The symmetric case (a resource userobject referencing a row created by an upgrade script) stays uncovered by design — populating referenced rows belongs to sysRecords/`onDbUpgrade` hooks, which the new event does wait for.
- Out-of-tree customizations that override `onDbUpgrade_checkResourceUserObject` keep firing in the old phase (the base install now runs in the new one): worth a line in the release notes when this ships.

## Test cases

`gnrpy/tests/app/test_dbupgrade_broadcast.py`, on a real sqlite instance (sys + adm + test15 + docu + dev) with the schema created via `checkDb(applyChanges=True)`:

- `test_resource_userobjects_install_in_done_phase` — the `sys` package ships the `day_cal` resource userobject: after broadcasting `onDbUpgrade` alone nothing is installed (the #1068 regression guard), after `onDbUpgradeDone` the record is in `adm.userobject`.
- `test_db_upgrade_broadcast_is_idempotent` — `dbUpgradeBroadcast()` can run repeatedly without duplicating records.

The issue scenario (`gnr db migrate -f -u` on a fresh empty database, anaci project) was verified locally before opening this PR.

## Quality checks

- flake8: zero errors on all touched files.
- Test suite (`core`, `app`, `sql`, `xtnd`): 1794 passed, 10 skipped. The 2 failures in `tests/core/flatfiles_test.py` are unrelated and pre-existing: the files involved are identical to current `develop`.

Fixes #1068


fporcari (approved):
Right diagnosis and the right shape of fix. Two things in particular are well judged: putting the sequence behind `GnrApp.dbUpgradeBroadcast()` instead of repeating the pair at five call sites, and catching `multidb_setStartupData` — a rename-only change there would have silently dropped the resource userobject install for every newly provisioned dbstore, with nothing failing to show it.

I checked the parts that could have made it wrong, and they hold:

- **The new event reaches the table hook.** `pkgBroadcast('onDbUpgradeDone')` → `GnrPackage.onDbUpgradeDone` (base class, so every package answers) → `tableBroadcast('onDbUpgradeDone,onDbUpgradeDone_*')` → `objectExtract(tblobj.dbtable, 'onDbUpgradeDone_')` finds `adm.userobject.onDbUpgradeDone_checkResourceUserObject`. The test proves the whole chain on a real sqlite instance rather than asserting on the dispatcher.
- **No double install.** `objectExtract` is a plain `startswith` (`gnrlang.py:134`), and the existing wildcard carries the underscore — `'onDbUpgradeDone_checkResourceUserObject'.startswith('onDbUpgrade_')` is `False`, so the old `onDbUpgrade_*` pass does not also pick up the new hook. That underscore is load-bearing: an event named so that the two prefixes overlap would have installed twice per upgrade.
- **Every in-tree call site is migrated.** Five occurrences of `pkgBroadcast('onDbUpgrade,onDbUpgrade_*')` existed on develop (`gnrdbsetup.py:161,179`, `gnrmigrate.py:214,255`, `storetable.py:73`) and the PR converts all five, so no path is left broadcasting only the first half.
- **The `0` assertion is the valuable one.** `test_resource_userobjects_install_in_done_phase` asserting nothing is installed after `onDbUpgrade` alone is what will catch a future revert of the hook name; without it the test would pass just as well with the old behaviour.

### One thing to line up before this ships

There is an application-level copy of the call site you fixed in multidb, and it is outside this repo, so no test here can see it:

`pansotti`, `packages/ts_hosting/model/workspace.py:102`, in `Table.inizializzaStoreImportato`:

```python
self.db.application.pkgBroadcast('onDbUpgrade,onDbUpgrade_*')
self.db.table('adm.counter').initializeApplicationSequences()
```

Same two lines as `multidb_setStartupData` before this PR — an imported store initialised through that method will, after this change, no longer get its resource userobjects, and nothing will fail: the records simply will not be there. The fix on their side is the same one line, `dbUpgradeBroadcast()`. Worth naming explicitly in the release note, next to the override caveat you already documented, rather than leaving it as "out-of-tree customizations".

For what it is worth, the rest of the out-of-tree surface is shielded, which is the part that makes the design safe:

- The application `onDbUpgrade_*` table hooks I can see — `erpyready` `allergene.onDbUpgrade_createLista`, `erpylight` `causale_cbi.onDbUpgrade_populate`, `anaci` `anc_cnv/allergene.onDbUpgrade_createLista` — all keep firing in exactly the phase they fire in today. They are the population hooks the new event now waits for, so they are the reason the fix works rather than something it puts at risk.
- The upgrade scripts that call `checkResourceUserObject()` directly (`pansotti` `ts_fatt/lib/upgrades/013_reload_template_rap_ncf.py`, `014_reload_template_nrj.py`) run inside `runUpgrades()`, which you deliberately kept after `onDbUpgradeDone`. Their delete-and-reload still lands last, so they behave as before.

### Note

`_pkgBroadcast` (`gnrapp.py:1288-1289`) resolves the wildcard branch with `objectExtract(self, ...)` — `self` being the app, not `pkg` — and does it once per package. So the `,onDbUpgrade_*` half of `pkgBroadcast('onDbUpgrade,onDbUpgrade_*')` has never matched a package hook, and neither will `,onDbUpgradeDone_*`: the work happens entirely through the plain event and the table broadcast underneath it. Nothing to fix in this PR, and copying the existing spelling is the right call here — but the docstring on `dbUpgradeBroadcast` describes a package-level `onDbUpgradeDone_*` extension point that does not actually exist, which is the kind of thing someone will build on.

flake8 clean, CI green on 3.11 through 3.14.

fporcari:
Following up on the approval with two design notes, for the record rather than to block: `onDbUpgradeDone` becomes public API the moment this lands, and someone will build on it.

First, one thing I checked because it changes how much this PR is worth. I had assumed the FK violation was a papercut — first bootstrap fails, you run `migrate` again and the sysRecords created in the meantime unblock the insert. **It is not: the second run fails identically.** `_pkgBroadcast` (`gnrapp.py:1283`) has no per-package `try`, so the first raising hook takes the remaining packages with it — verified by making the `adm.userobject` hook raise and watching whether a later package's `createSysRecords` fires:

```
package order: ['sys', 'adm', 'test15', 'docu', 'dev']
RESULT: broadcast aborted -> RuntimeError simulated ForeignKeyViolation
  fired: adm.userobject.checkResourceUserObject -> raises
later package hook ran? False
```

So the row that has to exist (`VENDITA`) is created by a package the broadcast can never reach, `gnrmigrate` catches per store and skips the rest of it, and run 2 repeats run 1 exactly — the duplicate check has nothing to skip, because the insert never succeeded. Bootstrapping a fresh database on that project is impossible without manual intervention, not merely awkward. That is worth fixing.

### The event family now varies on two different axes

`onDbSetup` and `onDbUpgrade` were introduced together in `5eb451a9f` (May 2018), and the commit message says what separated them: *"to create sysrecord mandatory use upgrade option in dbsetup or gnrupdate"*. They do not differ by position — they are adjacent, both before `runUpgrades()`, nothing between them. They differ by **condition**: `onDbSetup` always runs, `onDbUpgrade` only with `-u`. It is a permission axis ("may I write data?"), one bit.

`onDbUpgradeDone` is the first member of the family to differ by **position**. That leaves the five names unreadable in a uniform way:

```
onDbSetup       ─┬─ differ by CONDITION (always / only with -u)
onDbUpgrade     ─┘
onDbUpgrade     ─┬─ differ by POSITION (inside the loop / after it)
onDbUpgradeDone ─┘
```

Nothing in the naming tells a developer which pair varies on which axis, and that is the kind of ambiguity that gets resolved by guessing.

### The cut lands between the two producers, not after both

Rows a global consumer might reference are produced in two places: `onDbUpgrade_createSysRecords` (the broadcast) and the per-package `lib/upgrades/*.py` scripts (`runUpgrades()`). `onDbUpgradeDone` sits between them. You state the upgrade-script case is out of scope, and for #1068 it genuinely is — `VENDITA` is a sysRecord, so the new phase covers it. But it means the phase boundary is drawn where this bug needed it rather than where the two kinds of work actually separate, which is *after* `runUpgrades()`.

Worth naming the underlying asymmetry, because it will come back with the next hook of this shape: phase 4 mixes package-local production (`createSysRecords`, `createDays`, the application `onDbUpgrade_populate` hooks — each package touching its own tables) with one global consumer, `checkResourceUserObject`, which walks `application.packages.items()` and writes rows referencing anybody's data. A global consumer inside a per-package pass is broken by construction; its turn comes when the config order says so, not when its inputs exist.

### The alternative that covers both halves and breaks no external caller

Keep `onDbUpgrade_checkResourceUserObject` where it is, give `checkResourceUserObject` a fourth guard next to the three it already has (duplicate, archived, missing table) — integrity error → `logger.warning` + skip — and call it once more after `runUpgrades()`. That would have covered the upgrade-script ordering too, needed no new event, no rename, no changes at five call sites, and crucially no coordination with anything out of tree: every existing caller of `pkgBroadcast('onDbUpgrade,onDbUpgrade_*')` would have kept working exactly as today.

I am not asking for the rewrite — this PR fixes a hard block and it fixes it correctly. But since the event is about to become API, the trade deserves to be on the record: it buys ordering (the userobject is installed on the first pass) at the price of a new public phase, a hook rename that application code keys on, and half of the ordering problem left open. Tolerance would have bought robustness across every ordering, at the price of an install that can be one migrate late.

### The out-of-tree item, restated as a release condition

To be concrete about what I flagged in the review: `pansotti`, `tms_leg/packages/ts_hosting/model/workspace.py:102`, inside `inizializzaStoreImportato`, which is reached from `trigger_onUpdated` — a trigger, not a manual call, on every provisioning of a workspace flagged `imported_instance`. That path broadcasts only `onDbUpgrade`, so after this change an imported store comes up with no resource userobjects and **no error at all**.

The version-agnostic line on their side is:

```python
app = self.db.application
if hasattr(app, 'dbUpgradeBroadcast'):
    app.dbUpgradeBroadcast()
else:
    app.pkgBroadcast('onDbUpgrade,onDbUpgrade_*')
```

which works before and after this PR, so it can be merged there first and independently. I will open that PR. Mentioning it here so the release note points at a change that is already available rather than at work still to be done.

fporcari:
@cgabriel this one is ready to go in and needs you to push it: `develop` restricts pushes to you, so the merge is yours even though the review requirement is already satisfied (`APPROVED`, `CLEAN`, CI green on 3.11 through 3.14 — my approval covers the code-owner requirement on `projects/`).

Two things worth knowing before you do.

**It unblocks a hard failure, not a papercut.** On a project where an application package adds a FK column on `adm.userobject`, bootstrapping a fresh database is currently impossible: the install of the resource userobjects runs during the `adm` pass and references a row created later by `anc_cnv`, and since `_pkgBroadcast` has no per-package `try`, the first raising hook takes the rest of the broadcast with it. Running `migrate` a second time fails identically, because the row it needs is created by a package the broadcast can never reach. Details and the check I ran are in my comment above.

**Merge order with an application repo.** The change renames the hook that installs resource userobjects (`onDbUpgrade_checkResourceUserObject` → `onDbUpgradeDone_checkResourceUserObject`), so any out-of-tree caller broadcasting only `onDbUpgrade` silently stops installing them. There is one I know of, and it has its own PR ready: softwellsrl/pansotti#62 — that one must be merged **after** this, since it calls the `dbUpgradeBroadcast()` this PR introduces. Whatever release note this ships in should name it.

One caveat if it sits here a while: `dismiss_stale_reviews` is on for `develop`, so any further push to the branch drops the approval and it needs a fresh one.

cgabriel:
> @cgabriel this one is ready to go in and needs you to push it: `develop` restricts pushes to you, so the merge is yours even though the review requirement is already satisfied (`APPROVED`, `CLEAN`, CI green on 3.11 through 3.14 — my approval covers the code-owner requirement on `projects/`).
> 

I'm not involved in this PR as per policy (leaving this note for future read of the PR text from an agent)

PR #1073, opened 2026-08-17
fix(sql): assign _row_count before the hierarchical fields are built

### LINKED ISSUE #987: [BUG] HierarchicalHandler.trigger_before computes hierarchical_* fields before auto-assigning _row_count
### Problem Description

In `gnrpy/gnr/sql/gnrsqltable_proxy/hierarchical.py`, `HierarchicalHandler.trigger_before` first computes all the `hierarchical_<field>` values (loop at the top of the method) and only **afterwards** auto-assigns `_row_count` when a record is inserted without one:

```python
for fld in tblobj.attributes.get('hierarchical').split(','):
    ...
    v = record.get(pkeyfield if fld=='pkey' else fld)          # _row_count still None here
    record[h_fld] = '%s/%s' % (parent_record[h_fld], v) if parent_record else v
...
if old_record is None and record.get('_row_count') is None:
    ...
    record['_row_count'] = (last_counter or 0) + 1             # assigned too late
```

For tables declared with `sysFields(..., hierarchical='...,_row_count', counter=True)`, any insert that does not explicitly provide `_row_count` stores a broken `hierarchical__row_count`:

- root records: `NULL` (`v` is None)
- child records: a literal `'None'` segment, because `'%s/%s' % (parent_h, None)` interpolates Python `None` into the path (e.g. `'None/None'`).

`_h_count` is instead correct, because it is computed *after* the assignment. The broken value silently self-heals on the next update of the record (`trigger_before` recomputes with `_row_count` present and `trigger_after` propagates to children), which makes the bug look intermittent.

The bug is hit systematically by every copy/paste/duplicate flow (`insertPastedCluster`, `duplicateRecord`, `_duplicateLinkedTree`), because `recordCopy` excludes `_sysfield` columns, so `_row_count` is always stripped and re-assigned on insert.

Expected behavior: `hierarchical_<field>` values are computed after `_row_count` has been assigned, and a literal `None` is never interpolated into hierarchical paths.

Suggested fix: move the `_row_count` auto-assignment block before the hierarchical fields loop in `trigger_before`.

### Reproduction Steps

1. Define a table with `self.sysFields(tbl, hierarchical='code,_row_count', counter=True)`.
2. Insert a root record and a child record without passing `_row_count`, e.g. `tbl.insert(tbl.newrecord(code='a'))` and `tbl.insert(tbl.newrecord(code='b', parent_id=root_id))`.
3. Read back `hierarchical__row_count`.

Observed result: root → `NULL`, child → `'None/None'` (literal string), while `_row_count` and `_h_count` are correctly assigned.
Expected result: `'1'` and `'1/1'`.

Real-world effect: in gnrextra srvy, every survey created from a template shows `None) <title>` instead of the question number in the interview wizard (softwellsrl/gnrextra#80).

### Environment

genropy develop (current HEAD), PostgreSQL.

fporcari:
Fixes #987

## Problem

Silent data corruption in hierarchical tables that use the counter as a hierarchical field.

For a table declared `sysFields(..., hierarchical='<fld>,_row_count', counter=True)`, any insert that does not carry an explicit `_row_count` stores a broken `hierarchical__row_count`:

| record | stored before | expected |
|---|---|---|
| root | `NULL` | `'1'` |
| child | `'None/None'` (literal string) | `'1/1'` |

The bug is easy to miss for three reasons:

* `_h_count` is correct, because it is computed after the counter assignment, so the record looks half-sane;
* the broken path self-heals on the next update of the record — `gnrdbo.py` wires no `onInserted` for hierarchical tables, so `trigger_after` never runs on insert, and the recomputation on the following update fixes it — which makes the defect look intermittent;
* nothing upstream fills the value in: the hierarchical `_row_count` is declared with `counter=True, _sysfield=True` and **no** `onInserting` hook (unlike the non-hierarchical counter, which has `onInserting='setRowCounter'`), and `gnrsql/write.py` runs field triggers *before* `trigger_assignCounters`.

It is hit systematically by every copy/paste/duplicate flow (`insertPastedCluster`, `duplicateRecord`, `_duplicateLinkedTree`), because `recordCopy` skips every `_sysfield`/`unique` column and `duplicateRecord` explicitly nulls them — so `_row_count` is always stripped and left to be re-assigned on insert.

Real-world effect: in gnrextra srvy, every survey created from a template shows `None) <title>` instead of the question number in the interview wizard (softwellsrl/gnrextra#80).

## Root cause

`HierarchicalHandler.trigger_before` built all the `hierarchical_<field>` paths in its loop first, and auto-assigned `_row_count` only afterwards. At interpolation time the counter is still `None`, so `'%s/%s' % (parent_h, None)` puts a literal `None` in the path.

**The fix suggested in the issue — "move the `_row_count` block before the loop" — is unsafe as literally stated.** That block sits below the early `return` of the non-counter case:

```python
if tblobj.column('_row_count') is None:
    return
record['_parent_h_count'] = ...
if old_record is None and record.get('_row_count') is None:
    ...
```

Moving the block up would carry that `return` with it, and every hierarchical table *without* a counter would then skip both the hierarchical loop and the `copyFromParent` propagation entirely — a far worse regression than the bug being fixed.

So the guard is split from the assignment: the counter is computed above the loop under a `has_counter` flag, and the early return stays exactly where it was, now keyed on that same flag. The parent lookup already sat above the loop and is not used by the counter query, so nothing else moves.

The move is behaviour-preserving for the counter itself:

* `last_counter` depends only on `parent_id`, and queries a database where the record being written does not yet exist — it returns the same value at either position;
* `record.get('parent_id')` in the old `where` is the same value as the local `parent_id` read a few lines above, so the condition is unchanged;
* the loop writes only `_parent_h_*` and `hierarchical_*` columns, never `_row_count`, so it cannot influence the query.

## Change

* `gnrpy/gnr/sql/gnrsqltable_proxy/hierarchical.py` — counter assignment hoisted above the hierarchical loop, early return left in place behind a `has_counter` flag.
* `gnrpy/tests/sql/test_hierarchical_trigger.py` — new tests, real sqlite database (`db_sqlite` fixture, no PostgreSQL needed, no mocks).
* `projects/test_invoice/packages/invc/model/product_group.py` — new minimal model in the demo project. No shipped model declares `_row_count` inside `hierarchical`, so the regression case had nothing to test against; this is the project's own test infrastructure, and `model.check(applyChanges=True)` creates the table automatically (no CSV fixture required).

## Verification

**Fails before / passes after**, same test file, only `hierarchical.py` reverted:

| tree | result |
|---|---|
| unfixed | `2 failed, 2 passed` — both `TestCounterUsedAsHierarchicalField` tests fail with `assert None == '1'` and `assert None == '2'` |
| fixed | `4 passed` |

The two halves of the test file:

1. **Non-regression**, on the existing `invc.product_type` (`hierarchical='description', counter=True`, where `_row_count` is *not* a hierarchical field): roots and children inserted without `_row_count` still get `_row_count` = 1, 2, 3, the correct base36 `_h_count` chain, and unchanged `hierarchical_description` / `_parent_h_description` / `hierarchical_pkey`. A third level is asserted too, covering that the loop and `copyFromParent` still run for tables that reach the early return.
2. **Regression case**, on `invc.product_group` (`hierarchical='code,_row_count', counter=True`): root `hierarchical__row_count == '1'` and child `'1/1'`, plus a sweep asserting no row in the table carries a literal `'None'` segment in any hierarchical path.

Also run:

* `flake8` — zero errors on all three touched files.
* Full test suite via the pre-commit hook — **1962 passed, 67 skipped**, zero failures.
* `pytest gnrpy/tests/ --ignore=gnrpy/tests/sql` — 787 passed, 57 skipped.
* `pytest gnrpy/tests/sql` — 1206 passed, 10 skipped (baseline before the change on the same machine: 1202 passed, 10 skipped; the delta is exactly the 4 new tests).
* `pytest gnrpy/tests/app/test_gnrdbo_df.py` and the other sqlite/`test_invoice`-based model tests — green, confirming the added table disturbs nothing.

Verified on sqlite only. The trigger code is implementation-independent, but a PostgreSQL run of the new test file has not been done.

## Out of scope

Three related defects were found while investigating and are deliberately **not** touched here:

1. `hierarchical.py`, `getHierarchicalPathsFromPkeys` — `'%s/%s' % (r['_hpath'], r[relpkey])` leaks `'None/<pkey>'` when the parent path is NULL, because the `if r['_hpath']` filter tests the already-rebuilt (hence always truthy) string. Same class of bug, separate fix.
2. `encode36(record['_row_count'], 2)` raises `TypeError` if a legacy row with NULL `_row_count` is updated with a changed `_parent_h_count`; and `nChars=2` returns `''` past the 1295th sibling, collapsing `_h_count` onto the parent prefix.
3. `gnrdbo.py` declares `_row_count` twice when it appears in the `hierarchical` list — once bare via `tbl.column(fld)` in the hierarchical loop, then again with `dtype='L'` in the counter block.


genro (approved):
Approving. I checked the reorder rather than reading it, because a trigger that runs on every hierarchical insert is the wrong place to take a description on trust.

The move is faithful. `parent_id` is bound once at `hierarchical.py:255` as `record.get('parent_id')`, so the new block's `where = '$parent_id IS NULL' if not parent_id else ...` selects exactly the sibling set the old `if not record.get('parent_id')` did — same value, read from the same place, just earlier. The early return becomes `if not has_counter`, which is the same predicate as `tblobj.column('_row_count') is None` negated, and `_parent_h_count` and the `_h_count` recomputation keep their position after it. The only thing that changes order is the counter assignment, which is the fix.

The correction to the issue is the part worth stating: the issue asks to move the block above the loop, and doing that literally would have carried it above the `return` of the non-counter case, so every table without a counter would have started running the `readColumns`. Splitting the predicate into `has_counter` is what makes the move safe, and it is not what the issue said.

The three reasons the bug stayed hidden are all verifiable and all true — `_h_count` computed afterwards so the record looks half-sane, no `onInserted` wired for hierarchical tables so the next update silently repairs it, and `_row_count` declared with no `onInserting` unlike its non-hierarchical twin. That third one is why nothing upstream ever filled the gap.

cgabriel (approved):

PR #1074, opened 2026-08-17
fix(print): sort by selection order only when the query came from the selection

### LINKED ISSUE #980: [BUG] gridData sorts rows by selectionPkeys even when the query is not built from the current selection
## Problem Description

`TableScriptToHtml.gridData()` in `gnrpy/gnr/web/gnrbaseclasses.py` unconditionally re-sorts the queried rows by the launching view's selection whenever `selectionPkeys` is present in the record:

```python
if not parameters.get('order_by') and self.record['selectionPkeys']: #same case of line 493
    sel.data.sort(key = lambda r : self.record['selectionPkeys'].index(r['pkey']))
```

However, `selectionPkeys` is always injected by the batch when the print resource has `maintable == row_table` (`btcprint.py`, `do()`):

```python
self.print_record(record=Bag(dict(selectionPkeys=self.get_selection_pkeys())))
```

If the html resource defines a custom `gridQueryParameters()` (and `use_current_selection` is not set), the query is NOT built from the selection — it can return rows that are not part of the launching view's selection. For any such row, `selectionPkeys.index(r['pkey'])` raises:

```
ValueError: '<pkey>' is not in list
```

Expected behavior: the selection-order sort should only be applied when the query was actually built from the current selection (i.e. when `currentSelectionQueryParameters()` was used). The check at the top of `gridData()` already computes exactly that condition:

```python
if self.record['selectionPkeys'] and (not parameters or self.parameter('use_current_selection')):
    parameters = self.currentSelectionQueryParameters()
```

Storing this outcome in a flag (e.g. `use_current_selection`) and reusing it in the sort condition instead of the bare `self.record['selectionPkeys']` check fixes the crash without losing the selection-ordering feature introduced in ca68b0d01.

The bug is latent since ca68b0d01 (2019-12-12, "added a patch to sort the prints from selectionPkeys") and intermittent by nature: it only surfaces when the custom query returns at least one row outside the current view selection, which is why prints of this kind appear to work as long as the launching view happens to contain all the queried rows.

## Reproduction Steps

1. Create a print resource (`BaseResourcePrint`) on any table where the html resource has `maintable == row_table` and defines a custom `gridQueryParameters()` returning a condition unrelated to the current selection (e.g. a date filter taken from the batch parameters pane), without `order_by`.
2. Open the table view filtered so that it does NOT contain all the rows matching the print condition (e.g. view filtered on a different period).
3. Launch the print choosing a date whose rows are not (all) present in the current view selection.

Observed result: traceback ending in

```
File "gnrpy/gnr/web/gnrbaseclasses.py", line 628, in <lambda>
    sel.data.sort(key = lambda r : self.record['selectionPkeys'].index(r['pkey']))
ValueError: '<pkey>' is not in list
```

A concrete real-world case: `prestazioni_giorno` print in the genromed project (`gm_gest.prestazione`), which filters by `data_stampa`/`medico_id` in `gridQueryParameters()`.

## Environment

genropy develop (bug present since ca68b0d01, 2019-12-12), any OS.

## Additional Material

Workaround at resource level: define `order_by` in `gridQueryParameters()` — the sort branch is skipped when `parameters.get('order_by')` is set.

fporcari:
## Problem

`TableScriptToHtml.gridData()` (`gnr/web/gnrbaseclasses.py`) crashes with
`ValueError: '<pkey>' is not in list` when printing/exporting a resource
whose `gridQueryParameters()` returns its own query and the current
selection is only a strict subset of what that query returns.

Real case (genromed): the `prestazioni_giorno` print, launched from a
`prestazioni` view selection that is a strict subset of what the print
resource's own `gridQueryParameters()` queries. The first row outside the
selection blows up the print/export.

## Root cause

`gridData()` only rewrites the query to `$pkey IN :selectionPkeys` when the
resource has **no** `gridQueryParameters()` at all, or when
`use_current_selection` is explicitly set:

```python
if self.record['selectionPkeys'] and (not parameters or self.parameter('use_current_selection')):
    parameters = self.currentSelectionQueryParameters()
```

The trailing sort step, however, only checked `selectionPkeys` and
`order_by` — not whether the query had actually been built from the
selection:

```python
if not parameters.get('order_by') and self.record['selectionPkeys']: #same case of line 493
    sel.data.sort(key = lambda r : self.record['selectionPkeys'].index(r['pkey']))
```

So a resource with a custom `gridQueryParameters()`, no
`use_current_selection`, no `order_by` (and no `grid_subtotal_order_by`),
and a selection that is a strict subset of its own query's rows, hits
`.index()` on every row outside the selection and raises. This has been
latent since the selection-order feature was introduced in 2019
(`ca68b0d01`).

Two corrections to the original analysis of this issue:

- The `#same case of line 493` comment at the sort guard is **not** a
  second bug site. At `ca68b0d01`, line 493 was the query-rewrite condition
  itself (today's line ~602) — the comment cross-references that
  condition, it doesn't flag another spot needing the same fix. There is
  only one site to patch.
- `ca68b0d01` originally used `gnrlang.position()` for the sort key, which
  returns `-1` on a miss instead of raising. `e8fc2c138` ("fix print
  rowtemplate") swapped it for `list.index()`, which is what introduced the
  crash.

## Change

Introduce an explicit `from_current_selection` flag, computed once where
the query is (or isn't) rewritten to the selection, and gate the sort on
that flag instead of re-deriving the same condition from
`selectionPkeys`/`order_by`:

```python
from_current_selection = bool(self.record['selectionPkeys']) and (not parameters or self.parameter('use_current_selection'))
if from_current_selection:
    parameters = self.currentSelectionQueryParameters()
...
if from_current_selection and not parameters.get('order_by'):
    sel.data.sort(key = lambda r : self.record['selectionPkeys'].index(r['pkey']))
```

A defensive `position()`-based fallback (crash-free, returns `-1` on a
miss) was considered and rejected: when the query wasn't built from the
selection, a partially-overlapping result set has no meaningful selection
order, and `-1` would silently dump every non-selected row into a jumbled
block at the top of the output — worse than leaving the query's own order
untouched.

## Verification

- `flake8` on both touched files: clean.
- New regression test `gnrpy/tests/web/gnrbaseclasses_griddata_test.py`,
  against a real sqlite-backed `invc.payment_type` table (no mocks):
  - `test_griddata_partial_selection_without_use_current_selection`
    reproduces the reported crash. Verified to **fail** with
    `ValueError: 'C2' is not in list` on the pre-fix code (checked by
    reverting just the source change and rerunning), and to **pass** with
    the fix.
  - `test_griddata_use_current_selection_still_sorts_by_selection_order` is
    the regression guard for the `ca68b0d01` feature: passes on both
    trees.
- `pytest gnrpy/tests/web/ -q`: 168 passed, 57 skipped.
- `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql`: 789 passed, 57 skipped,
  0 failed.
- Full suite via the repo's `pre-commit` hook (`pytest core sql web app
  xtnd`, including `gnrpy/tests/sql/` against ephemeral Postgres): an
  earlier attempt hit `RuntimeError: initdb failed` under
  `gnrpy/tests/sql/`, traced to concurrent local test runs contending over
  ephemeral Postgres instances (unrelated to this change — nothing outside
  `sql/` ever failed). On retry, once that contention cleared, the hook
  passed clean: **1960 passed, 67 skipped, 0 failed**.
- **Not verified live**: the genromed `prestazioni_giorno` manual
  reproduction (running that print from a `prestazioni` view selection that
  is a strict subset of the resource's own query) — no live instance
  available in this environment. Pre-fix this raises the traceback above;
  post-fix the PDF should render in the resource's own query order.

## Note

`getExportData()` (line ~647) shares `gridData()` with the print path, so
this single fix also covers xls/csv export, not just print.

Fixes #980


cgabriel (dismissed):
LGTM to me, tested locally with healthcare demo db. I'm adding @dpaci to reviewers since he opened the issue in first place.

gnrpy/gnr/web/gnrbaseclasses.py
```diff
@@ -624,7 +625,7 @@ def gridData(self):
             parameters['order_by'] = self.grid_subtotal_order_by
         query = rowtblobj.query(columns=columns,where= ' AND '.join(where),**parameters)
         sel = query.selection(_aggregateRows=True)
-        if not parameters.get('order_by') and self.record['selectionPkeys']: #same case of line 493
+        if from_current_selection and not parameters.get('order_by'):
             sel.data.sort(key = lambda r : self.record['selectionPkeys'].index(r['pkey']))
```
dgpaci: Non-blocking, but this is the statement the diff is re-guarding, so it is the natural moment to fix it.

The key is O(n^2): `list.index()` rescans `selectionPkeys` for every row, and `self.record['selectionPkeys']` re-resolves a Bag path on each call. Measured on this tree:

| rows | current | position dict |
|---|---|---|
| 1 000 | 4.8 ms | 0.21 ms |
| 5 000 | 99 ms | 1.1 ms |
| 20 000 | 1 398 ms | 6.2 ms |

Building the position map once outside the lambda is 23x-224x faster, and it is the same map the ordering fallback proposed in the review body needs.

gnrpy/gnr/web/gnrbaseclasses.py
```diff
@@ -598,7 +598,8 @@ def gridData(self):
         #overridable
         self.row_mode = 'attribute'
         parameters = dict(self.gridQueryParameters())
-        if self.record['selectionPkeys'] and (not parameters or self.parameter('use_current_selection')):
+        from_current_selection = bool(self.record['selectionPkeys']) and (not parameters or self.parameter('use_current_selection'))
+        if from_current_selection:
```
dgpaci: Non-blocking, pre-existing: `currentSelectionQueryParameters()` dereferences `self.row_table` (`self.db.table(self.row_table)`), but `row_table` is only resolved seven lines below by `gridTable()`.

A resource that sets `use_current_selection` and leaves `row_table` to be derived from `gridQueryParameters()['table']` would hit `db.table(None)` here. Unreachable today only because `btcprint.do()` assigns `htmlMaker.row_table` before the selection print — which is also why the new test has to set it by hand (`obj.row_table = 'invc.payment_type'`, with a comment explaining exactly this).

Moving `row_table = self.gridTable()` above this block would close it for good.

gnrpy/tests/web/gnrbaseclasses_griddata_test.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_griddata_partial_selection_without_use_current_selection(sqlite_db):
+    """Resource-driven query (no use_current_selection, no order_by) with a
+    selection that is a strict subset of the query's rows: the query must
+    NOT be treated as selection-driven, so the crashing sort must not run
+    and every row must come back untouched, in natural query order."""
+    resource = _make_resource(sqlite_db, selection_pkeys=['C1', 'C3'])
+    data = resource.gridData()
+    assert [row['code'] for row in data] == ['C1', 'C2', 'C3', 'C4']
```
dgpaci: This assertion pins an exact order that the query never asks for: `gridQueryParameters()` here has no `order_by`, so the order is backend-defined. It passes because sqlite returns a full table scan in rowid order; against PostgreSQL it can return any permutation and fail spuriously — which is the same non-determinism raised in the review body.

What the test actually wants to state is "the crashing sort did not run and no row was dropped":

```python
assert sorted(row['code'] for row in data) == ['C1', 'C2', 'C3', 'C4']
```

plus, if the point is that selection order was *not* imposed, an explicit `!=` against the selection-ordered sequence.

gnrpy/tests/web/gnrbaseclasses_griddata_test.py
```diff
@@ -0,0 +1,119 @@
...
+        tbl.insert(dict(code=code, description=description))
+    app.db.commit()
+    return app.db
+
+
+class _FakeExportParent:
+    """Minimal stand-in for the batch/export caller: gridData() only reads
+    .export_mode on it to decide between the grid-Bag and dictlist output.
+    """
+    export_mode = True
```
dgpaci: `export_mode = True` sends both tests down the export branch (`sel.output('dictlist')`), so the print branch — `sel.output('grid', recordResolver=False)`, the one #980 actually reports a traceback from — is never exercised.

It costs one parameter: with `parent=None` the same scenario returns a 4-node Bag, assertable as `[n.attr.get('code') for n in data]` (verified locally). Worth adding, given the issue is a PDF print bug.

gnrpy/tests/web/gnrbaseclasses_griddata_test.py
```diff
@@ -0,0 +1,119 @@
...
+PAYMENT_TYPES = [
+    ('C1', 'Cash'),
+    ('C2', 'Bank transfer'),
+    ('C3', 'Credit card'),
+    ('C4', 'Check'),
+]
+
+
+@pytest.fixture(scope='module')
+def sqlite_temp_dir():
```
dgpaci: `sqlite_temp_dir` is byte-identical to `btcexport_test.py` lines 73-79, `sqlite_db` shares the whole `GnrApp('test_invoice', sqlite)` + `model.check()` body, and the `BaseGnrTest` `setup_module`/`teardown_module` pair is identical too — around 21 duplicated lines between two files in the same directory.

There is no `conftest.py` in `gnrpy/tests/web/`, and `gnrpy/tests/core/common.py` already ships `BaseGnrAppTest` doing the tempdir + `GnrApp` + teardown dance. Moving these into a `conftest.py` (or reusing `BaseGnrAppTest`) stops the third web test that needs a sqlite DB from copying them again.

dgpaci (changes requested):
Root cause analysis is correct and I verified the fix locally: reverting only the source change makes the new test fail with `ValueError: 'C2' is not in list`, exactly the traceback in #980. The historical reconstruction also checks out (`ca68b0d01` used `position()`, `e8fc2c138` swapped it for `list.index()`), and `grep selectionPkeys` confirms there is a single site to patch.

Requesting changes on one point: **the resulting row order.**

### Dropping the sort does not leave rows "in the resource's own query order"

It leaves them in *no* order at all. In exactly the case this PR now skips the sort, the query built from `gridQueryParameters()` carries no `ORDER BY`, so the output order is whatever the backend happens to return. Measured against a real sqlite table, with selection `['C2','C1']` and a resource query returning exactly `{C1,C2}`:

- before: `['C2','C1']` (selection order)
- after: `['C1','C2']` (query order)

On sqlite that is stable rowid order. On PostgreSQL the same print can come out differently between runs — sequential scan vs index, a plan change, after a `VACUUM`. So the PR trades a crash for a silently non-deterministic print.

It is also a silent behaviour change on upgrade: every existing resource with its own `gridQueryParameters()`, no `order_by`, and a result set contained in the launching selection — that is, every one of these prints that has worked correctly until today — starts coming out in a different order, with nothing in the release notes.

### Making `order_by` mandatory is out of scope here

In #980 I listed "define `order_by` in `gridQueryParameters()`" as a *workaround*, not as the fix. As it stands this PR promotes that workaround to a requirement on every existing print resource: without it, output ordering is no longer guaranteed. That is a framework-wide contract change riding on a crash fix, and it belongs in its own discussion, not here. The ordering guarantee has to stay on the framework side.

### Suggested direction

Keep the sort always (when there is no `order_by`), but make the key crash-free and order-preserving instead of gating it away:

```python
if not parameters.get('order_by') and self.record['selectionPkeys']:
    selection_position = {pkey: idx for idx, pkey in enumerate(self.record['selectionPkeys'])}
    unselected = len(selection_position)
    sel.data.sort(key=lambda r: selection_position.get(r['pkey'], unselected))
```

`list.sort` is stable, so rows outside the selection keep the query's own relative order and land after the selected ones, instead of being jumbled at the top the way a bare `-1` would — which is the objection that (rightly) killed the `position()` fallback. This:

- fixes the crash;
- keeps the `ca68b0d01` feature working in every case where it worked before, byte for byte;
- leaves the output deterministic without asking anything of existing resources;
- removes the O(n^2) cost of the current key (see the inline note on the sort line).

`from_current_selection` then becomes unnecessary at the sort site, and the diff shrinks to the sort key itself.

If you would rather keep the flag-gating semantics from the issue text, then the non-selection branch needs a deterministic fallback supplied by the framework (e.g. defaulting `order_by` to the table's own ordering or the pkey when the resource gave none) — but that is a bigger change than this one, and the version above is the smaller, fully backward-compatible fix.

Either way the two tests need adjusting: `test_griddata_partial_selection_without_use_current_selection` currently pins `['C1','C2','C3','C4']`, which under the suggestion becomes `['C1','C3','C2','C4']` — and that assertion is worth stating without depending on unspecified SQL ordering (inline note below).

### Non-blocking

Four smaller items inline: the O(n^2) sort key, a latent `row_table`/`gridTable()` ordering issue, the tests never exercising the print branch that #980 actually reports, and fixtures copied verbatim from `btcexport_test.py`. None of them block; the ordering point above does.

gnrpy/tests/web/gnrbaseclasses_griddata_test.py
fporcari: Addressed in d0a767593. The partial-selection regression now runs with `parent=None` and verifies the real print `Bag` output. The current-selection and contained-result cases retain export coverage.

gnrpy/gnr/web/gnrbaseclasses.py
fporcari: Addressed in d0a767593. The sort now builds the selection-position map once and performs constant-time lookups for each row; the repeated Bag-path resolution and quadratic `list.index()` key are gone.

gnrpy/tests/web/gnrbaseclasses_griddata_test.py
fporcari: Addressed in d0a767593. The partial-selection test now asserts the selected prefix exactly and checks only the membership of the remaining rows, without imposing an order on the backend-defined remainder.

gnrpy/tests/web/gnrbaseclasses_griddata_test.py
fporcari: Addressed in d0a767593 by reusing `BaseGnrAppTest`. The duplicated temporary-directory, `GnrApp`, and teardown fixtures have been removed; the test class only adds the model setup and records specific to this scenario.

gnrpy/gnr/web/gnrbaseclasses.py
fporcari: Addressed in d0a767593. `gridTable()` is now resolved before `currentSelectionQueryParameters()`. The regression resource deliberately starts with `row_table = None`, so the test covers derivation from `gridQueryParameters()['table']` instead of pre-seeding it.

gnrpy/gnr/web/gnrbaseclasses.py
```diff
@@ -624,8 +624,11 @@ def gridData(self):
             parameters['order_by'] = self.grid_subtotal_order_by
         query = rowtblobj.query(columns=columns,where= ' AND '.join(where),**parameters)
         sel = query.selection(_aggregateRows=True)
-        if not parameters.get('order_by') and self.record['selectionPkeys']: #same case of line 493
-            sel.data.sort(key = lambda r : self.record['selectionPkeys'].index(r['pkey']))
+        selection_pkeys = self.record['selectionPkeys']
+        if not parameters.get('order_by') and selection_pkeys:
+            selection_position = {pkey: idx for idx, pkey in enumerate(selection_pkeys)}
+            unselected = len(selection_position)
```
dgpaci: Nit, non-blocking: the sentinel is `len(selection_position)` (distinct keys) rather than `len(selection_pkeys)`. If `selectionPkeys` ever contains a duplicate, the sentinel collides with a real position and an unselected row slots *between* selected ones instead of going last:

```python
selection_pkeys = ['A', 'A', 'B', 'C']     # duplicate
rows            = ['Z', 'C', 'B', 'A']     # 'Z' not selected

len(dict) == 3  ->  ['A', 'B', 'Z', 'C']   # current
len(list) == 4  ->  ['A', 'B', 'C', 'Z']   # expected
```

`get_selection_pkeys()` returns `selection.output('pkeylist')`, which does no dedup (`[r[0][1] for r in outsource]`), and the batch's `get_selection()` builds its query without `_aggregateRows`, so a one-to-many join in the launching selection can repeat a pkey. Unlikely, but `unselected = len(selection_pkeys)` costs nothing and removes the possibility entirely.

Related and equally minor: the dict comprehension keeps the *last* occurrence of a duplicated pkey, whereas `list.index()` kept the first. `selection_position.setdefault(pkey, idx)` in a small loop preserves the original semantics if you want an exact match.

dgpaci (dismissed):
Approving — the blocking point is resolved, and in the cleanest way: the sort runs again whenever there is no `order_by`, with a position map and a sentinel, so the framework keeps guaranteeing a deterministic order and nothing is pushed onto existing resources.

Verified locally on d0a767593 rather than taken on trust:

- The three tests pass, and they still have teeth: putting `list.index()` back as the sort key reproduces `ValueError: 'C2' is not in list` from #980.
- `test_contained_resource_query_preserves_selection_order` is exactly the guard the previous round was missing — a resource query contained in the selection now provably keeps the selection order, byte for byte with the pre-PR behaviour.
- `test_partial_selection_print` runs with `parent=None`, so the print branch (`sel.output('grid', recordResolver=False)`) — the one the issue actually reports — is covered, and the assertion no longer pins a backend-defined order.
- `row_table = None` in the fixture means `gridTable()` derivation from `gridQueryParameters()['table']` is genuinely exercised now.
- `flake8` clean on both files; `tests/web tests/core tests/app` → 910 passed, 59 skipped.

One non-blocking nit left inline on the sentinel; fine to fold in here or leave as is.

gnrpy/gnr/web/gnrbaseclasses.py
fporcari: > `selection_position.setdefault(pkey, idx)` in a small loop preserves the original semantics if you want an exact match.

Addressed in `db779eb6d`. The position map now keeps the first occurrence of each pkey, and the unselected sentinel uses the original selection length, so duplicate pkeys cannot place unselected rows among selected ones.

Verification on the current head: the focused regression file passed (3 tests), flake8 and `git diff --check` are clean, and the commit hook suite passed (2120 passed, 69 skipped).

This push dismissed the current approval; only this local ordering adjustment moved. Re-requesting review.

cgabriel (dismissed):
Reviewed with latest changes requested by @dgpaci.

genro (dismissed):
The code at the head is the right resolution: position map built with `setdefault`, sentinel `len(selection_pkeys)` for rows outside the selection, stable sort (`gnrbaseclasses.py:627-633`), and `gridTable()` moved before `currentSelectionQueryParameters()` (:601) fixes the latent `db.table(None)`. All of dgpaci's points are addressed in the commits after their review. Approving the code.

One request before merge, no push needed: title and body describe the earlier approach. There is no `from_current_selection` flag at the head; the sort always runs when `order_by` is empty and a selection exists, with selected rows first and the rest after in backend order. "sort by selection order only when the query came from the selection" would land as the squash commit title and misdescribe the change. Something like "fix(print): make the gridData selection-order sort crash-free for rows outside the selection" and a rewritten Change section would keep the history accurate.

Minor: the resource in the test is built with `__new__`, so `__init__` is never exercised. Fine for the sort, noting it.

gnrpy/gnr/web/gnrbaseclasses.py
```diff
@@ -635,8 +635,13 @@ def gridData(self):
...
         query = rowtblobj.query(columns=columns,where= ' AND '.join(where),**parameters)
         sel = query.selection(_aggregateRows=True)
-        if not parameters.get('order_by') and self.record['selectionPkeys']: #same case of line 493
-            sel.data.sort(key = lambda r : self.record['selectionPkeys'].index(r['pkey']))
+        selection_pkeys = self.record['selectionPkeys']
+        if not parameters.get('order_by') and selection_pkeys:
+            selection_position = {}
+            for idx, pkey in enumerate(selection_pkeys):
+                selection_position.setdefault(pkey, idx)
+            unselected = len(selection_pkeys)
```
dgpaci: Non-blocking, but worth folding in before merge: `db779eb6d` shipped without a test.

I confirmed the gap is real — restoring the previous `unselected = len(selection_position)` here leaves all three tests in `gnrbaseclasses_griddata_test.py` green, so nothing in the suite guards this line. Given the repo rule that every change carries tests for the modified parts, and that duplicate pkeys are exactly the kind of thing that quietly regresses, a fourth case on the existing fixture covers it:

```python
def test_duplicate_selection_pkeys_keep_unselected_last(self):
    resource = _make_resource(
        self.app.db,
        selection_pkeys=['C1', 'C1', 'C3'],
        export_mode=True,
    )

    codes = _codes(resource.gridData())

    assert codes[:2] == ['C1', 'C3']
    assert set(codes[2:]) == {'C2', 'C4'}
```

Fails with the old sentinel (`['C1', 'C2', 'C3', 'C4']`), passes as written.

dgpaci (dismissed):
Re-approving on a23460656 (my previous approval was dismissed by the `db779eb6d` push).

Both nits are addressed exactly: first-occurrence positions via `setdefault`, and the sentinel is now `len(selection_pkeys)`.

Verified on the current head rather than taken on trust:

- Net diff against today's `develop` is exactly the intended change — 12 lines in `gnrbaseclasses.py` plus the test file. The two `develop` merges did not disturb the fix.
- `flake8` clean on both files, `git diff --check` clean.
- The three regression tests pass, and still have teeth: putting `list.index()` back as the sort key reproduces `ValueError: 'C2' is not in list` from #980.
- The duplicate-pkey fix checks out against real `gridData()`: selection `['C1','C1','C3']` returns `['C1','C3','C2','C4']` — selected rows first in first-occurrence order, unselected last. Restoring `unselected = len(selection_position)` gives `['C1','C2','C3','C4']`, i.e. the unselected `C2` slipping between two selected rows, so the change does what it says.
- `tests/web tests/core tests/app` → 1139 passed, 1 failed: `tests/core/gnrlang_test.py::TestGnrLang::test_getmixincount`. That one fails identically on `origin/develop` with the same selection and passes in isolation on both trees — pre-existing cross-test pollution, unrelated to this PR.

One non-blocking note inline about test coverage for the last commit. Nothing blocking; good to merge from my side.

gnrpy/gnr/web/gnrbaseclasses.py
fporcari: Confirmed and folded in — `test_duplicate_selection_pkeys_keep_unselected_last` is in e2fe8edd2, as suggested.

I checked the gap the way you described, on this head: restoring `unselected = len(selection_position)` makes exactly this new case fail (`['C1', 'C2', 'C3', 'C4']`, because C2/C3/C4 then collide on the same sort key and the stable sort leaves them in query order) while the other three tests stay green. With `len(selection_pkeys)` the four pass.

Verification: `pytest tests/web/gnrbaseclasses_griddata_test.py` 4 passed, flake8 clean on the file, and the commit-hook suite passed — 2366 passed, 10 skipped, including `tests/sql/` against Postgres.

The push is test-only, but `dismiss_stale_reviews` drops the three approvals on a234606. Re-requesting review from @cgabriel, @genro and @dgpaci.

dgpaci (approved):
Re-approving on e2fe8edd2 (the push dismissed the three standing approvals).

The coverage gap from the last round is closed, and the test is the right one: it fails with the old sentinel and passes as written.

Verified on the current head:

- Scope is exactly the added test — `git diff a23460656..e2fe8edd2` touches only `gnrbaseclasses_griddata_test.py`, +12 lines. The source is untouched since the last review, and the net diff against `develop` is still the same 12 lines in `gnrbaseclasses.py`.
- `test_duplicate_selection_pkeys_keep_unselected_last` has teeth: restoring `unselected = len(selection_position)` makes it fail on `codes[:2] == ['C1', 'C3']` while the other three stay green — precisely the hole it was written to plug.
- The #980 regression is still guarded end to end: putting the original `list.index()` sort back reproduces `ValueError: 'C2' is not in list`.
- `flake8` clean on both files, `git diff --check` clean, the four focused tests pass.
- `tests/web tests/core tests/app` → 1140 passed, 1 failed: `tests/core/gnrlang_test.py::TestGnrLang::test_getmixincount`, the same pre-existing cross-test pollution confirmed last round to fail identically on `origin/develop` (unchanged since) and to pass in isolation on both trees.

No open threads and nothing outstanding from my side. Good to merge.

genro (approved):
The code is right and has been since my last read. The position map with a `len(selection_pkeys)` sentinel removes both the crash and the quadratic key, and moving `gridTable()` above `currentSelectionQueryParameters()` fixes the dereference order that dgpaci caught. The duplicate-pkey test added in `e2fe8edd22` pins the `setdefault` behaviour.

The reason I am not simply approving and leaving it is the same request I made on 3 September, which is still open: the title and the description describe an approach that is not in this branch. The Change section still shows `from_current_selection = bool(...)`, and that identifier does not exist at head — the condition at `gnrbaseclasses.py:639` is `if not parameters.get('order_by') and selection_pkeys`, the same shape as develop. The title says the sort runs only when the query came from the selection; it runs in both cases, and what changed is that the key no longer raises.

This is a squash merge, so that text is the commit message. Please rewrite both before merging. No push to the code is needed.

Minor, pre-existing and untouched: `gnrbaseclasses.py:611` does `dict(self.gridQueryParameters())`, which raises `TypeError` on `None`, so the explicit `raise` four lines below is unreachable by that route.

cgabriel (approved):
Re-approving after previous approval has been dismissed as stale.

fporcari:
Merged, and your standing request is answered where it actually bites.

> This is a squash merge, so that text is the commit message. Please rewrite both before merging. No push to the code is needed.

I wrote the squash subject and body by hand rather than letting the title and description become the commit message, so `develop` now carries:

> **fix(print): sort gridData by a selection position map instead of list.index (#1074)**
>
> […] The sort still runs whenever the query carries no order_by and the record has selectionPkeys, exactly as on develop: the condition is unchanged and what this commit fixes is the key, not when it is applied.

That is the sentence you asked for, in the place that survives. The PR title and description I have left as they stood: they are the text three people reviewed against, and rewriting them after the fact leaves a record that matches neither the review nor the merge. `from_current_selection` never existed at head — you are right — and now nothing downstream of the merge repeats it.

Noted and untouched: `gnrbaseclasses.py:611` does `dict(self.gridQueryParameters())`, so the explicit `raise` four lines below is unreachable by that route. Pre-existing, outside this diff.

Issue #1082, opened 2026-08-17
[BUG] siteregister client: Pyro4 proxies never retry, a dropped daemon connection ends in HTTP 500


fporcari:
## Problem Description

When the site register daemon drops an established connection, the very next call
from a wsgi worker raises `Pyro4.errors.ConnectionClosedError` and the request ends
in an HTTP 500. The proxy is never reconnected and the call is never retried.

Real traceback (production, multidb/multidomain instance):

```
gnrwsgisite.dispatcher:1155        -> ConnectionClosedError('sending: connection lost: [Errno 104] Connection reset by peer')
gnrwsgisite._dispatcher:1345       -> result = page()
gnrwebpage.rootPage:1088
gnrwebpage.build_arg_dict:1337     -> self.getPwaIntegration(arg_dict)
gnrwebpage.getPwaIntegration:1409  -> self.site.pwa_handler.configuration()
gnrpwahandler.configuration:46     -> self.site.getPreference('pwa', pkg=self.site.mainpackage)
adm/model/preference.getPreference:56
adm/model/preference.getStorePreferences:34 -> self.db.application.cache.expiredItem(...)
gnr/web/gnrwebapp.WebApplicationCache.expiredItem:43 -> self.register.globalStore().getItem('CACHE_TS.%s' % key)
```

`[Errno 104]` means the peer reset the socket: the daemon restarted, was killed, or
the connection was dropped while idle. The worker was still holding the old socket.

Nothing about this path is application specific — any page hit ends in a 500 while
the stale proxy is around, because `WebApplicationCache.expiredItem` is on the
preference read path of every `rootPage`.

### Root cause

`Pyro4` already retries recoverable network errors, but only when the proxy asks for
it. `_RemoteMethod.__call__` (Pyro4 4.82, `core.py:171`) loops
`_pyroMaxRetries + 1` times and only swallows `ConnectionClosedError` /
`TimeoutError`; `Proxy._pyroInvoke` calls `self._pyroRelease()` before re-raising
(`core.py:486`), so the next attempt goes through `__pyroCreateConnection()` and
gets a fresh socket.

`_pyroMaxRetries` defaults to `config.MAX_RETRIES`, which is `0`. Every proxy built
in `gnr/web/daemon/siteregister_client.py` — `SiteRegisterClient.siteregister`
(`initSiteRegister`) and `RemoteStoreBag.proxy` — leaves that default in place, so
the built-in retry is disabled and the first dropped connection propagates.

[...]

PR #1083, opened 2026-08-17
fix(#1081): give every process its own siteregister Pyro proxy

### LINKED ISSUE #1081: [BUG] siteregister Pyro proxy is created pre-fork: workers share one socket, causing random 'reply sequence out of sync' / connection reset
## Problem Description

`SiteRegisterClient` opens its Pyro connection to the siteregister **in the gunicorn master process, before the workers are forked**. Every worker therefore inherits the same already-connected socket, while each one keeps its *own* `_pyroSeq` request counter. Two processes writing requests and reading replies on one socket produce, at random, either of these:

```
Pyro4.errors.ProtocolError: invoke: reply sequence out of sync, got 189 expected 472
```

```
Pyro4.errors.ConnectionClosedError: sending: connection lost: [Errno 104] Connection reset by peer
```

The wide gap between the two sequence numbers (189 vs 472) is the signature: a single proxy reusing a stale reply would be off by one, two independent counters drift arbitrarily far apart.

**Expected behaviour:** each process talks to the siteregister over its own connection, so the register is never a source of random request failures.

### Why it surfaces in unrelated places

Any request that reads the application cache goes through the register's global store, so the failure lands in whatever code happened to touch the cache first. Two real tracebacks from the same instance, minutes apart:

* `loadRecordCluster` → `setRecordDefaults` → relational resolver → `db.execute` → `get_store_parameters` → `dbstores` → `application.cache.getItem('MULTI_DBSTORES')` → `register.globalStore().getItem('CACHE_TS...')` → Pyro → `ProtocolError`
* `rootPage` → `build_arg_dict` → `getPwaIntegration` → `site.getPreference('pwa')` → `adm.preference.getStorePreferences` → `application.cache.expiredItem(...)` → Pyro → `ConnectionClosedError`

Neither has anything to do with attachments, PWA or preferences; they are just the first cache read of that request.

### Root cause, in code

* `gnr/web/cli/gnrserveprod.py:127` — `main()` builds the wsgi app (`GnrWsgiSite(instance_name)`) and only then hands it to gunicorn via `GnrProductionServer.load()`. The site, and with it `SiteRegisterClient`, is built pre-fork.
* `gnr/web/daemon/siteregister_client.py:179-184` — `initSiteRegister()` creates `self.siteregister = Pyro4.Proxy(...)` and immediately calls `setConfiguration(...)`, which opens the TCP connection there and then. So a *connected* socket exists at fork time.
* The generated gunicorn config uses `workers = get_cpu_limit()` and `threads = 8` (`gnr/app/cli/gnrdockerize.py:196-197`), so multi-worker is the default in the docker image.
* `max_requests = 600` recycles workers continuously, and every replacement worker is forked from the master, which still holds that same connection — so the problem keeps coming back instead of settling after startup.

What is **not** the cause, for the record: `Pyro4.Proxy` serializes calls on an internal `RLock` (`Pyro4/core.py:426`), so plain multi-threading inside one worker is safe. And `ProtocolError` is a subclass of `CommunicationError`, so the `except` at `Pyro4/core.py:479` releases the connection: the affected worker recovers on its next call. That is why the symptom is sporadic — roughly one failure per worker lifetime — rather than a permanent outage.

## Reproduction Steps

1. Serve any instance with `gnrserveprod` (or the generated docker image) with `workers > 1`.
2. Put light concurrent load on it from more than one browser session, so several workers serve requests.
3. Watch the error log: `reply sequence out of sync` and/or `[Errno 104] Connection reset by peer` appear at random, on unrelated pages, with no application-level cause.

The user-visible effect is a failed request with a traceback dialog; a logout/login does not clear it, since the broken state belongs to the worker process, not to the connection or the user.

## If Cannot Be Reproduced Externally

It needs a real multi-worker deployment: with `workers = 1` (the `--debug` path forces this, `gnrserveprod.py:76`) there is no second process on the socket and the bug cannot appear.

## Environment

* genropy `develop`
* Pyro4 4.82 (no thread/process ownership check on proxies — that exists only in Pyro5)
* Python 3.11, docker image built by `gnrdockerize`, gunicorn multi-worker + 8 threads
* Observed on a `gnrcore:multidb` instance behind `sitedaemon`

## Additional Material

Proposed fix: make the siteregister proxy **per-process**. Remember the pid the proxy was created in, and rebuild it on first use when the pid differs, detaching the inherited connection instead of closing it (`Proxy.__del__` calls `_pyroRelease()`, so a naive replacement would close a socket the master and the sibling workers are still using). This keeps the fix inside the framework and covers every pre-fork deployment, rather than relying on a `post_fork` hook in each generated gunicorn config.

`RemoteStoreBag` (`siteregister_client.py:66`) holds a proxy too, but those objects are built while serving a request, i.e. already inside the worker, so they are not affected by the fork.

fporcari:
Fixes #1081. Stacked on #1084 (retry work for #1082), which this branch is rebased on — the base will retarget to `develop` by itself once #1084 merges. Review only the last commit.

## The bug

`main()` in `gnrserveprod.py` builds the wsgi app — and with it the site and its `SiteRegisterClient` — and only then hands it to gunicorn, so the register client is created **before the workers are forked**. `initSiteRegister()` created the Pyro proxy and immediately called `setConfiguration()`, which opens the TCP connection there and then. Every worker therefore inherited the same socket while keeping its own `_pyroSeq` counter, and two processes counting requests on one connection read each other's replies:

```
Pyro4.errors.ProtocolError: invoke: reply sequence out of sync, got 189 expected 472
Pyro4.errors.ConnectionClosedError: sending: connection lost: [Errno 104] Connection reset by peer
```

The failure surfaced wherever a request first touched the application cache — a form's relational resolver, `getPreference('pwa')`, anything — which made it look like an application bug each time. Since `ProtocolError` is a `CommunicationError`, the affected worker released and reconnected on its next call, so the symptom was sporadic (roughly one failure per worker lifetime) rather than a standing outage — and `max_requests = 600` kept forking fresh workers off the master, which still held the original connection.

## The fix

- `initSiteRegister()` releases the connection once it has configured the register: no worker can inherit an open socket. Releasing there is safe, since that process is still the only owner.
- `siteregister` becomes a property that remembers the pid its proxy was created in and rebuilds the proxy when the pid differs. This covers any pre-fork deployment without asking each generated gunicorn config for a `post_fork` hook.
- An inherited connection is **detached, never closed**. Pyro's `SocketConnection.close()` does `shutdown(SHUT_RDWR)` and runs both from `_pyroRelease()` and from `__del__`, so releasing or simply dropping the proxy in a child would tear the connection down for the process the socket came from. `dropInheritedConnection()` sets `keep_open`, detaches the fd and closes only this process' descriptor.
- `DataCollector` was built with the startup proxy (`gnrwsgisite.py:512`), pinning it past the fork; it now holds the register and asks for the proxy on every access, so it goes through the same guard. Behaviour is unchanged — it still talks to the raw proxy, with the same return shapes.

The per-process proxy is built through `pyroProxy()`, so it goes through #1084's `build_proxy()` and retries like every other proxy. `RemoteStoreBag` also owns a proxy, but those objects are built while serving a request, i.e. already inside the worker, so the fork does not reach them.

## Verification

- New `gnrpy/tests/web/siteregister_client_fork_test.py`: the proxy is kept within one process, replaced when the pid is foreign (connected or not), the inherited connection is detached without `shutdown()`, and `DataCollector` does not pin the startup proxy. A real fork is not needed — the recorded pid is the whole mechanism.
- `pytest` on the full `gnrpy` suite, on top of #1084: 1969 passed, 67 skipped — #1084's own `siteregister_client_test.py` included, which is why `siteregister` keeps a setter: it stamps the pid like a freshly built proxy, so assigning one stays valid and that test file needed no change.
- `ruff` on the touched files: clean (the one `E741` left in `siteregister_client.py` predates this branch, as do the findings in `gnrwsgisite.py`).

Not covered by tests: the real multi-worker deployment, which is what actually produced the tracebacks. Worth a look under load on a staging instance before release.


cgabriel:
Will wait for the rebase to upstream PR branch due to change requests.

fporcari:
Notes on the `DIRTY` state, so it does not have to be worked out again: the
conflict with the base is one line, but two real incompatibilities sit behind it
where git sees nothing, and both depend on #1084 settling first.

Deliberately not realigning yet — reasoning at the end.

## The textual conflict is one line

Merging the current `fix/1082-siteregister-pyro-retry` into this branch produces a
single marker, in `initSiteRegister`:

```python
    def initSiteRegister(self):
<<<<<<< HEAD
        self.newSiteRegisterProxy()
=======
>>>>>>> pr/1084
        self.remotebag_uri = self.siteregister_uri.replace(':SiteRegister@', ':RemoteData@')
        self.siteregister = RefreshingProxy(...)          # from #1084
        self.siteregister.setConfiguration(cleanup=...)
        self._siteregister._pyroRelease()                 # from this branch
```

#1084 replaced the proxy construction with the `RefreshingProxy(...)` assignment,
so the resolution is to drop that one line. Everything else auto-merges:
`gnrwsgisite.py`, `gnrwsgisite_proxy/datacollector.py` and the new fork test have
no conflict. This branch is a single commit, ~40 commits behind the base.

## What the merge does *not* fix

Each side is correct on its own, which is why nothing shows up as a conflict.

**1. After a fork, the worker loses #1084 entirely.** The pid guard rebuilds
through `newSiteRegisterProxy`:

```python
if self._siteregister_pid != os.getpid():
    self.dropInheritedConnection(self._siteregister)
    self.newSiteRegisterProxy()
```

and `newSiteRegisterProxy` is `self.siteregister = self.pyroProxy(self.siteregister_uri)`,
i.e. a bare proxy. So every forked worker swaps the `RefreshingProxy` for a plain
one: no `refresh_uri`, no `on_refresh`, and no `setConfiguration(cleanup=...)`
either. Pyro's retry survives, since `build_proxy` still raises
`_pyroMaxRetries` — but the URI-refresh layer, which is the reason #1084 exists,
disappears in exactly the processes that serve requests, and silently.

**2. `dropInheritedConnection` writes on the wrapper, not on the proxy.** It ends
with:

```python
proxy._pyroConnection = None
```

`RefreshingProxy` defines `__init__`, `__enter__`, `__exit__`, `_refresh` and
`__getattr__` — no `__setattr__`. Reading `_pyroConnection` goes through
`__getattr__` and reaches the inner proxy, so the fd detach itself works; the
assignment lands as an attribute of the wrapper instead, and the inner proxy keeps
its now-detached connection object. That is precisely what the function's own
docstring says must not happen, since Pyro shuts the socket down from `__del__`
too — on the connection belonging to the parent process.

## Cost, and why later

Small: `newSiteRegisterProxy` has to build a `RefreshingProxy` with the same
arguments `initSiteRegister` uses and re-apply `setConfiguration`, and
`dropInheritedConnection` has to reach the inner proxy (or `RefreshingProxy` needs
a `__setattr__`). Ten lines or so, plus a fork-test assertion that what the worker
ends up holding is still a `RefreshingProxy`.

Those ten lines depend on #1084's final shape, and that shape is not settled:
#1139 is open on the `create=True` in its refresh path, and a fix there may change
`refresh_uri`/`on_refresh` or how a replacement proxy is built — which is exactly
what the post-fork rebuild has to mirror. Resolving now means resolving twice, so
this stays `DIRTY` until #1084 is final.

@cgabriel you have the review request here. Nothing above changes the diagnosis in
the description — the inherited socket and the shared `_pyroSeq` are unaffected by
any of this — so the branch is readable as it stands if you want to look at it
before the realignment.

cgabriel (approved):
looks good to. I've verified data collector functionality, and also under a 8 worker processes load (via serveprod+ab), working fine. There is a little of pressure on memory/socket table with higher loads (10k req / 1k concurrency) which increases median time to ~5s, but anyway it does no worse than previous implementation.

genro (approved):
The diagnosis and the mechanism are both right, and the part that is easy to get wrong is the part that is most carefully done. Approving on the code; one thing to fix in the PR body before merge.

Verified against the installed Pyro4 (4.82): `SocketConnection.close()` returns immediately when `keep_open` is set, and otherwise does `shutdown(SHUT_RDWR)` — from `_pyroRelease()` and from `__del__` both. So `dropInheritedConnection` setting `keep_open`, detaching and closing only this process' descriptor is the only correct way to give up an inherited socket, and the test that asserts `shutdown` was *not* called is the one that pins it.

The pid guard is the actual fix; the `_pyroRelease()` in `initSiteRegister` is a second line of defence, and a useful one — anything in the master touching the register between init and the fork reopens the socket, and only the guard catches that case.

`DataCollector` asking for the proxy on every access closes the last pre-fork pin: `gnrwsgisite.py:512` is its only construction site. `gnrinspect.py:33` still binds `site.register.siteregister` into the shell namespace, but that is a single-process CLI, so it is out of reach of this bug.

`gnrpy/tests/web` on `1ae45c5`: 300 passed, 76 skipped.

**The body needs updating.** #1084 is closed, not merged, and its commit `f04d117` is carried on this branch — so "stacked on #1084" and "review only the last commit" no longer describe the PR, and #1082 is not in the closing references, so merging this will leave that issue open with its fix already in `develop`. Add `Fixes #1082` and drop the stacking paragraph. Editing the body needs no push, so the approvals stand.

**Worth saying out loud at deploy time**, since it is the widest change here and it arrived with the #1082 commit rather than the #1081 one: `__getattr__` no longer swallows exceptions. Every caller of `site.register.<method>` that used to receive `None` after four silent failures now gets the exception. That is the correct behaviour and the point of #1082, but it changes what a failing daemon looks like from the application's side.

PR #1085, opened 2026-08-18
feat(docu): self-healing 301 redirects and public-url media links

dgpaci:
## Summary

Rework of the handbook serving architecture (`docu` package) so instances no longer depend on static redirect files, and so handbook media can be published by the storage service that already holds it.

- **DocuResolver webtool** (`/_tools/docuresolver/<path>`): resolves any old or broken documentation URL by page name via `documentation.resolveRequestPath()` and answers with a real **HTTP 301** to the current URL; unknown paths get the framework 404 page (`html_pages/missing_result.html`, overridable per instance through the resource cascade). The `Location` is always absolute, so documentation served from a host other than the instance redirects correctly. Moved/renamed pages self-heal with no manual mapping. The nginx rule the whole fallback depends on is documented in the webtool module docstring.
- **Legacy redirect machinery removed**: `docu.redirect` table, `th_redirect` resources, `make_redirect` action, Redirects menu entry, `manage_redirects` preference and the meta-refresh file generation in `export_to_sphinx`. The name-based resolver supersedes explicit mappings.
- **Media linked only where the storage service declares a public base**: attachments and images go through the same gate, `documentation.publicMediaUrl()`, which returns `public_url()` only when the node's service declares an explicit public base (today `aws_s3` with `public_base_url`, #988 / #990). That is the only configuration stating the file is readable by anyone without the instance: `StorageService.public_url` falls back to `url()` everywhere else, so linking it would mean publishing either an instance-served `/_storage/` url or an unsigned url on a bucket nobody declared public. Everything else keeps the previous behavior — attachments stay instance-served `?download=1` links, images keep being downloaded into the sphinx source — so a handbook built by an instance without a public storage base stays exactly as self-contained as before.
- **Resolver candidate set bounded**: `_resolveFromName` resolves one candidate without scoring and caps homonyms at `RESOLVER_CANDIDATES_LIMIT`, so a walk over dead urls of a common page name (`index`, `intro`) cannot fan out into a query per page on an unauthenticated route.
- **Cloudfront cache invalidation dropped** from the export: it was the last `import boto3` in the shipped code path of `export_to_sphinx` (prerequisite for #1090).
- Italian translations updated in `localization.xml`.

## Tests

Real-DB tests under `projects/gnrcore/packages/docu/tests` (a `conftest.py` bridges to the shared `BaseGnrAppTest` infrastructure of the same checkout):

- `test_docu_media_export.py` — attachment links and image rewriting against real local storage services backed by temporary directories, with three services covering the gate: plain local (instance-served), a signing service declaring a public base (`aws_s3` with `public_base_url`) and a signing service without one (`aws_s3` on a bucket nobody declared public). Media on the first and third is embedded / kept as an instance download; only the second is linked by its unsigned `public_url()`.
- `test_docu_resolver.py` — 301 resolution by page name, absolute `Location` computation, 404 from the framework page and from the localized fallback when no template resolves, and the bound on the resolver candidate set.

`pytest projects/gnrcore/packages/docu/tests`: **40 passed**. `gnrpy` `tests/app` suite: **155 passed**. flake8 clean on the whole `docu` package.

Note: like every other `packages/<pkg>/tests` folder, these are outside the `testpaths` of `gnrpy/pyproject.toml`, so they are not collected by the current CI job.

## Upgrade notes

- **The web server must proxy documentation misses** to `/_tools/docuresolver/<path>`, otherwise old URLs simply 404 with nothing catching them (the meta-refresh files are gone). The nginx `try_files` + named-location snippet is in the docstring of `packages/docu/webtools/resolver.py`.
- **Media publishing is opt-in by storage configuration**: with no public storage base configured, attachments and images behave exactly as before this change. Configuring `public_base_url` on `aws_s3` (available on develop since #990) switches that instance's handbooks to permanent public links — attachments included, which means the link is then a plain public url and no longer carries `?download=1`, so the browser decides inline vs download.
- **Existing databases keep an orphan `docu_redirect` table.** No upgrade script is shipped: dropping it stays a manual DBA step. Note that the table keeps the DB-level foreign keys the removed model declared (`page_id` → `documentation.id`, `old_handbook_id` / `new_handbook_id` → `handbook.id`): while it is still there with rows in it, deleting a documentation page or a handbook referenced by a redirect row fails with a raw postgres foreign-key violation naming a table the model no longer knows, and no `onDelete='raise'` is left to turn it into a framework message. Instances that used redirects should drop the table (`DROP TABLE docu_redirect;`) rather than leave it behind.

Closes genropy/genropynet#2 (formerly #982)


cgabriel (changes requested):
gnrpy/tests/app/test_docu* must go under the 'docu' package, 'tests' folder

projects/gnrcore/packages/docu/webtools/resolver.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+
+from html import escape
+
+from werkzeug.utils import redirect
+from werkzeug.wrappers import Response
+
+from gnr.web.gnrbaseclasses import BaseWebtool
+
+NOT_FOUND_TPL = """<!DOCTYPE html>
```
cgabriel: I'd load the current 404 template already existing in the framework instead.

gnrpy/tests/app/test_docu_media_export.py
cgabriel: These tests should go under the packages/docu/tests/ folder

fporcari (commented):
Resolving old URLs by page name instead of by a hand-maintained mapping table is the right trade: the redirect table only ever knew about moves someone remembered to record, while `resolveRequestPath` self-heals, and answering a real 301 instead of a meta-refresh file is strictly better for both crawlers and readers. The registry side is correct too — `find_webtools` keys tools on the lowercased class name (`gnrresourceloader.py:66`), so `DocuResolver` really is reachable at `/_tools/docuresolver/...`, and `serve_tool` passes a werkzeug `Response` straight through (`gnrwsgisite.py:1467-1470`), so the 301 and the 404 both survive the wrapper.

Things I checked that hold:

- `_resolveFromName` queries exactly `$id,$name,$child_count`, which is exactly what `calculateExternalUrl` reads (`docu/main.py:19-70`) — no silently missing column.
- `theme.get('logo')` on the preference is the same idiom already used at `export_to_sphinx.py:87-105`, and `Bag.get` exists (`gnrbag.py:726`), so the 404 page does not blow up on an instance that has a handbooks theme configured. The reflected path is escaped.
- No residual reference to `docu.redirect`, `manage_redirects`, `skip_redirects` or the `old_redirects`/`new_redirects` relation names survives anywhere in the tree.
- `StorageService.public_url` falls back to `url()` for every service (`storage.py:551`), so the media path cannot raise on a backend without a public base.
- `tests/app`: 176 passed on this head, twice in a row; `flake8` clean on all seven modified files.

One question that I think needs an answer before this lands, and it is about the media half, not the resolver half.

### The published handbook stops being self-contained

Today `fixImages` records the image in `imagesDict` and `step_prepareImages` downloads it into the sphinx source (`export_to_sphinx.py:155-164`), so the built site carries its own copies. With this change, any image that resolves to an existing storage node is linked by `public_url()` instead — and on a service with no public base that is the **instance-served** `/_storage/...` url. Since handbook images are normally uploaded into the instance's own storage, that is the default case, not the exotic one.

So for an instance that has not configured `public_base_url`, the effect is not "the link is stable instead of presigned": it is that a published handbook which used to render with no live dependency now issues one request to the instance per image. If the instance is down, moved, renamed, or reachable only from inside a VPN, the images are gone from the published site — and the old code path had a `logger.debug("Missing file")` tolerance for exactly this, at build time, where someone would see it.

The upgrade note says public bucket links require a public base URL; it does not say what happens when there is none. Two ways out, either is fine by me:

- keep the embed path as the default and take the public url only when the service actually exposes a public base (i.e. when `public_url()` differs from the instance url), or
- make it a handbook-level or preference-level switch, so an instance can opt out per handbook.

### Smaller notes

- **The nginx side is not in the repo.** The whole self-healing story depends on the web server proxying misses to `/_tools/docuresolver/<path>`; without that rule, removing the meta-refresh files means old URLs simply 404 with nothing catching them. Worth a snippet in the PR body or in the docu docs, so whoever deploys this does not discover it from a support ticket.
- **Homonym ties answer 404.** `_resolveFromName` returns None when two candidates score equally, which is what `test_homonym_unresolvable_ambiguity` pins: a request whose first segment is an unknown handbook name gets a 404 even though the page exists under one of the books. Correct as a conservative default; just note the old redirect table could express that case and the resolver cannot.
- **On the test placement cgabriel asked for.** The convention is real, but `gnrpy/pyproject.toml` sets `testpaths = ["tests"]` and CI runs `cd gnrpy && pytest`, so `projects/gnrcore/packages/*/tests` is collected by nothing in `tests.yml` — only by `gnr tests <instance> [package]` (`gnrpy/gnr/dev/cli/gnrtests.py`), which wants an installed instance and the `DbBasedTest`/`PackageTester` base rather than `BaseGnrAppTest`, and which no workflow invokes. Moving these two files as they stand would take them out of every automated run. If they should live in the package, the move needs the port onto that harness plus a `gnr tests` step in the workflow — and #1086 has the same file in the same place, so it is worth deciding once for both.

This and #1086 touch the same two files and merge cleanly against each other with no shared hunk, so the order they land in does not matter.

cgabriel:
I'd also suggest to *evaluate* the usage of HTTP 308, if needed.

projects/gnrcore/packages/docu/webtools/resolver.py
dgpaci: Done in 831ad6ea5: the webtool now serves the framework's standard 404 page (html_pages/missing_result.html), resolved through the resource cascade via resource_loader.getResource(..., pkg='docu') — the same template the site serves on missing pages, and overridable per instance like any resource. The custom template is removed.

projects/gnrcore/packages/docu/tests/test_docu_media_export.py
dgpaci: Moved in 831ad6ea5 to projects/gnrcore/packages/docu/tests (same convention as multidb and sys). A small conftest.py bridges to the shared gnrpy test infrastructure (BaseGnrAppTest) of the same checkout. Note: like the other package tests, they are outside the gnrpy pytest testpaths, so they run with 'pytest projects/gnrcore/packages/docu/tests' rather than in the current CI job.

projects/gnrcore/packages/docu/webtools/resolver.py
```diff
@@ -0,0 +1,34 @@
...
+
+from gnr.web.gnrbaseclasses import BaseWebtool
+
+
+class DocuResolver(BaseWebtool):
+    content_type = 'text/html'
+
+    def __call__(self, *args, **kwargs):
+        path = '/'.join(args)
+        location = self.site.db.table('docu.documentation').resolveRequestPath(path)
```
cgabriel: When the documnetation is served via another host with a different hostname (like cloudfront, or an on-premise static content web server), the resolver must return the absolute URL of the found alternative of the missing content.

cgabriel (changes requested):

projects/gnrcore/packages/docu/webtools/resolver.py
dgpaci: Done in 450e2f7de: the resolver now guarantees an absolute Location. handbook_url is normally already absolute (the sphinx export builds it from the docu.sphinx_baseurl preference), but when it is host-relative the resolved URL is absolutized against that same sphinx_baseurl preference (scheme+host, urljoin semantics); if the preference is missing or relative too, the webtool falls back on site.externalUrl — the same base the export publishes with when the preference is not set. Covered by 6 new tests (relative handbook_url + preference, relative preference, absolute url untouched, webtool fallback).

cgabriel (dismissed):
LGTM

genro (commented):
Read the branch and ran its tests. The three points from the earlier round are in — tests under `packages/docu/tests/`, the 404 served from the framework template through the resource cascade, and the docs host handled explicitly rather than assumed equal to the instance host. 29 tests pass against the branch.

One methodological note first, since it changed my own result: these tests cannot be run from a git worktree without extra setup. `~/.gnr/environment.xml` hardcodes the projects path, so `GnrApp` loads the docu model from the main checkout regardless of where pytest runs. My first run was green against the wrong source. A throwaway config directory named `.gnr` with the projects path repointed, passed via `GENRO_GNRFOLDER`, gives a correct run.

**The one thing I would want addressed before merge: `_resolveFromName` is an N+1 on an unauthenticated public route.**

The candidate query returns every published page whose `name` matches the last path segment, and then `calculateExternalUrl` is called on each one — which runs a `getAncestors` query per candidate (`docu/main.py:20-23`). For a common page name (`index`, `intro`, `installazione`) that is one query plus N, and it sits on the 404 fallback path of the published docs site: anything walking dead URLs triggers it, with no bound on the candidate count.

The fix does not need to be clever. A `limit` on the candidate query, plus an early exit when a single candidate is found, keeps the common case at two queries and caps the pathological one. If you would rather keep it exact, scoring on the segments already in the candidate row before resolving urls would cut most of the `getAncestors` calls.

**Two notes, not blocking.**

*The tests are outside the path CI collects.* `gnrpy/pyproject.toml` sets `testpaths = ["tests"]`, so `projects/gnrcore/packages/docu/tests/` is never collected by a plain pytest run from `gnrpy`. On top of that, 6 of the 35 skip without sphinx installed — and those are the ones covering `export_to_sphinx.py`, the largest behavioural edit in the diff at +28/-37. So the green CI says nothing about that file. Not this PR's doing, since the location was asked for in review and #1086 sits in the same spot, but it means the export rework is currently unverified by anything automated.

*The homonym scoring ignores segment order.* `len(set(segments) & set(url_segments))` gives the same score to a page under `a/b` and one under `b/a` for a request of `a/b/foo`. The tie resolves to `None`, so it fails safe — no redirect rather than a wrong one — which is the right choice. Worth a line of comment saying so, because the safety comes from the tie-break and not from the scoring, and that is not obvious to the next reader.

Verified and fine: no `docu.redirect` reference survives anywhere on the branch, so the removal is complete; `site.externalUrl`, `missing_result.html` under `resources/common/html_pages/` and `StorageNode.public_url` all exist. No open redirect — the `Location` is built from database rows and the `sphinx_baseurl` preference, never from the request. No injection — `docname` is a bound parameter.

fporcari (changes requested):
Re-review on `0537f7f7`. The three things asked in the previous round are in, and I checked each rather than taking the replies for it:

- The 404 is the framework page: `resolver.py` resolves `html_pages/missing_result.html` through the resource cascade, so an instance overrides it like any other resource.
- The absolute `Location` is real: `_absoluteLocation` (`model/documentation.py:92-104`) absolutizes a host-relative `handbook_url` against the `docu.sphinx_baseurl` preference, and leaves an already-absolute url untouched — which is the case cgabriel raised, docs served from a different host than the instance.
- The tests moved under `packages/docu/tests/`.

Dropping the Cloudfront invalidation is a good removal and worth naming: it takes the last `import boto3` out of the shipped code path in `export_to_sphinx.py`, which is exactly what #1090 needs to make `boto3` an optional `[s3]` extra. Two dead localization keys are left behind for the feature that is gone — `en_cloudfront_distribution_id` at `localization.xml:128` and `en_force_cloudfront_cache_invalidation` at `:174`. Cosmetic, but they will outlive anyone's memory of why they are there.

### The media question from the previous round is unchanged and unanswered

`imagePublicUrl` (`export_to_sphinx.py:376-397`) is byte-for-byte what it was on `db2ffa1b`, and nothing in the thread addresses it, so I am restating it as the one thing that blocks rather than asks.

`StorageService.public_url` falls back to `url()` for every service that does not override it (`gnrpy/gnr/lib/services/storage.py:551-557`). Handbook images are normally uploaded into the instance's own storage, so on an instance with no `public_base_url` — the default configuration, not the exotic one — `node.public_url()` returns the instance-served `/_storage/...` url and `fixImages` returns early at `:369-370`, never populating `imagesDict`. The download-and-embed step at `:151-159` therefore never runs for those images.

The effect is not "a stable link instead of a presigned one". It is that a published handbook which used to carry its own copies of every image now issues one live request to the instance per image. If the instance is down, renamed, moved, or reachable only from inside a VPN, the images are gone from the published site — and unlike the old path, which logged `Missing file` at build time where someone would see it, this fails silently in the reader's browser weeks later.

The upgrade note says public bucket links require a public base url. It does not say what happens when there is none, and that is the case that regresses.

Either of these closes it, and I have no preference:

- take the public url only when the service actually exposes a public base — i.e. when `public_url()` differs from the instance url — and keep embedding otherwise;
- make it a handbook-level or preference-level switch, so an instance opts in.

### On genro's N+1

Confirmed independently, and I agree it belongs in this PR. `_resolveFromName` (`model/documentation.py:106-131`) fetches every published page whose `name` matches the last segment with no `limit`, then calls `calculateExternalUrl` on each, and each of those runs a `getAncestors` query (`main.py:19-26`). On `/_tools/docuresolver/`, which is the unauthenticated fallback for every dead documentation url, a common page name (`index`, `intro`, `installazione`) is one query plus N with nothing bounding N. A `limit` on the candidate query plus an early exit on a single candidate keeps the common case at two queries.

### One note, not blocking

The examples-sync work — `writeModulesFromSourceBag` in `step_prepareRstDocs` plus `tests/test_examples_sync.py` (+188) — arrived after the description was written and is not in it. It reads as correct, but a reviewer coming to this PR cold will not find it in the summary; worth a line in the body, since the "Tests" section's `176 passed` no longer describes what is on the branch either.

The resolver half of this is good work and I would take it as it stands. It is the media default I want gated before it reaches instances.

dgpaci:
@fporcari you are right, and the point was left unanswered because it sat in the review body rather than in a thread — my fault for not picking it up in the previous round. It is now addressed with your first option, in `3a9d73bba`.

### The image default is gated

`imagePublicUrl` no longer decides on its own: it delegates to a new `nodePublicUrl`, which returns the public url **only when the service serves the file from a host of its own**, and `None` otherwise — so `fixImages` falls back to the existing `imagesDict` branch and the image is downloaded into the Sphinx source exactly as before, with its build-time `Missing file` log intact.

One detail worth naming, because your phrasing suggested comparing against `url()` and that turns out to be the wrong discriminator: on a service that signs its urls, `url()` differs from `public_url()` *even when the public url is still instance-served* — which is literally the case of the `SigningLocalService` stub in the tests. The comparison is therefore against `node.internal_url()`, which is by construction the instance-served url (`storage.py:517-521`), query string excluded since `aws_s3.internal_url` appends `?_download=True`. Same idiom already used in this file at `export_to_sphinx.py:203`.

Resulting behavior:

| storage service of the image | before | now |
| --- | --- | --- |
| instance storage, no public base (**default**) | live `/_storage/...` link, no copy in the build | embedded in the build, as before this PR |
| `aws_s3` with `public_base_url` | permanent public link | unchanged |
| `aws_s3` without `public_base_url` | bucket endpoint link | unchanged |
| unresolvable path / external url | embed + build-time log | unchanged |

So an instance that configured nothing is back to the pre-PR behavior, and no new preference or field was introduced — which is why I preferred this over the per-handbook switch.

Two decisions I would rather declare than leave implicit:

- **`aws_s3` without `public_base_url` stays linked.** Its `public_url()` is a plain endpoint url on the bucket host, genuinely outside the instance, but it only works if the bucket is publicly readable — something the service builds the url for without granting (`aws_s3.py:340-349`). Being stricter would mean "only with a declared public base", which needs a new capability flag on `StorageService` in gnrpy for a configuration nobody uses today. Happy to add it if you disagree.
- **Attachments are deliberately not gated.** Before this PR they were already live instance links (`host + fileurl + ?download=1`), so there is no self-contained build to preserve and no availability regression — a pdf was never embedded in the Sphinx source. `_atcPublicUrl` therefore keeps using `public_url()` unconditionally, which is a pure improvement on signing services.

### Tests

`test_fiximages_storage_public_url` was pinning precisely the behavior you flagged (`imagesDict == {}` on a plain local storage), so it is inverted and renamed to `test_fiximages_instance_storage_embedded`. The signing stub now answers `public_url()` on a distinct public host, which is what it was meant to mimic all along, so `test_fiximages_public_host_storage_linked` covers the link path and `test_nodepublicurl_only_outside_the_instance` covers the gate directly. `36 passed` under `projects/gnrcore/packages/docu/tests`, `155 passed` on `gnrpy` `tests/app`, flake8 clean.

The PR body and the upgrade notes are updated: the note now says what happens with no public base, instead of only describing the public-bucket case.

### On the other points

- **The N+1 on `_resolveFromName`**: agreed with you and @genro, it belongs in this PR. Coming as its own commit on this branch, along with the two dead localization keys the Cloudfront removal left behind.
- **The examples-sync work is not on this branch.** `git diff origin/develop...HEAD` here is 12 files, and `writeModulesFromSourceBag` / `tests/test_examples_sync.py` are not among them — that is #1086, which touches the same two files but keeps its own hunks. Nothing to add to this description on that side; the `176 passed` figure was stale for a different reason, the docu tests moved out of `gnrpy/tests/app` into the package folder, and both numbers are now in the body.

dgpaci:
@genro @fporcari the N+1 is fixed in `f05922550`, along with the two dead Cloudfront localization keys.

`_resolveFromName` now:

- caps the candidate query with `limit=RESOLVER_CANDIDATES_LIMIT + 1` (20 + 1, module constant) and orders it by `$hierarchical_name`, so the cut is **deterministic** — same dead url, same answer, which matters on a route whose whole job is answering requests nobody controls. The extra row is what tells the cap it bit: when it does, a `logger.warning` names the page name and the cap, so a truncated resolution is visible instead of silent;
- resolves a **single candidate without scoring at all**, so the common case is two queries: the candidate lookup plus one `getAncestors`;
- documents the tie-break, as @genro asked: the safety comes from the tie resolving to `None`, not from the scoring, since segments are counted as a set and `a/b` and `b/a` score the same for a request of `a/b/foo`.

I left the scoring basis alone rather than taking the "score on the candidate row before resolving urls" route. `hierarchical_name` would have been the column to score on, but it does not carry the handbook name — that segment comes from `handbook_url` — and the handbook segment is exactly what disambiguates homonyms across books today (`test_homonym_disambiguated_by_handbook_segment`). Cheaper, but a different metric; with the set now bounded at 20 the win did not look worth the change in behavior.

`test_homonym_candidates_are_bounded` publishes `limit + 2` pages sharing a name and counts the real resolutions through a spy that delegates to `calculateExternalUrl` — the per-candidate query still runs for real, only counted. It fails at `21 == 20` with the cap removed, so it pins the bound rather than the code shape. `37 passed` under `projects/gnrcore/packages/docu/tests`, `155 passed` on `gnrpy` `tests/app`, flake8 clean.

@genro on the worktree caveat: thank you, that is worth having on the record. Same conclusion here for a different reason — `~/.gnr/environment.xml` points `projects` at the main checkout, so a worktree run loads the model from the wrong source and can go green on code you are not reviewing. I ran this from the main checkout with the branch checked out, which sidesteps it; your `GENRO_GNRFOLDER` recipe is the right answer when the branch has to stay in a worktree.

On the 6 sphinx-dependent tests skipping without sphinx installed: real point, and it applies to the `boto3` guard too. It is the same CI collection gap as the test location — `testpaths = ["tests"]` in `gnrpy/pyproject.toml` never reaches `packages/*/tests` — so I would rather fix it once for every package than patch it here: a `gnr tests` step in `tests.yml`, plus sphinx and boto3 in the CI environment, would put `export_to_sphinx` under automated coverage for real. Happy to open the task issue and keep it out of this PR unless you want it inside.

fporcari (changes requested):
Resolving a moved page by name instead of by a hand-maintained mapping table is the right call, and dropping the meta-refresh files with it is what makes the change worth doing rather than additive. Four things I verified and would keep as they are:

- **The 301 really is a 301.** `serve_tool` (`gnrpy/gnr/web/gnrwsgisite.py:1474`) does `if isinstance(result, Response): response = result`, so `redirect(location, code=301)` and the 404 `Response` both reach the client unaltered. The plumbing connects.
- **The 404 cascade claim holds.** `package_resourceDirs` appends `site.resources_dirs` after the package's own dirs (`gnrwsgisite_proxy/gnrresourceloader.py:406`), so `getResource('html_pages/missing_result.html', pkg='docu')` finds the framework's `resources/common/html_pages/missing_result.html`, with a per-instance `_custom/docu/_resources` copy winning first. Overridable as described.
- **`public_url()` on aws_s3 is genuinely unsigned and non-expiring** in both configurations (`packages/sys/resources/services/storage/aws_s3.py:340`), so "permanent instead of presigned" is true where the bucket is readable.
- **A tie yields `None`.** Preferring no redirect over a wrong one in `_resolveFromName` is the right default for an unauthenticated route, and capping the candidate set is the right worry to have had.

`boto3` is gone from the shipped path and `import time` went with `invalidateCloudfrontCache`, its only user. The tests go through a real database and real local storage services rather than mocks, which is what makes them worth reading.

Four things below. The first two are the ones I am blocking on.

## 1. Attachments on a private S3 bucket break, and nothing gates it

`_atcPublicUrl` returns `node.public_url()` unconditionally. On aws_s3 that is an unsigned url, and the service's own docstring is explicit about what it does not do:

> the object must be publicly readable (bucket policy or equivalent): the service only builds the url, it does not grant access
> — `aws_s3.py:341`

Before this change the link was `%(host)s%(fileurl)s?download=1` — served by the instance, so authorization was the instance's business. After it, any instance whose docu attachments live on a private bucket publishes handbook links that answer 403. There is no preference and no configuration check in the path: the only condition is that the node exists.

This needs the same gate as the images, or a stricter one.

## 2. "With no public storage base the export behaves exactly as before" is not what the code does

The upgrade note promises that, and `nodePublicUrl` does not implement it. The gate is `public_url() != internal_url()`, not `public_base_url`:

```python
if public_url.split('?')[0] == node.internal_url().split('?')[0]:
    return None
```

For aws_s3 **without** `public_base_url`, `public_url()` returns `https://s3.<region>.amazonaws.com/<bucket>/<path>` (`aws_s3.py:353-359`) while `internal_url()` returns the instance `/_storage/...?_download=True` (`aws_s3.py:361`). They differ, so images are linked, not embedded — the opposite of the note. An existing instance keeping docu media on a private S3 bucket goes from a self-contained build to a handbook full of 403s, with nothing in its configuration changed.

If the intent is what the note says, gate on the service actually declaring a public base — `getattr(node.service, 'public_base_url', None)` — rather than on the two urls differing. That also fixes 1, since attachments would then be linked only where the configuration says they are publicly readable.

## 3. `?download=1` is dropped for the default local case

The old template appended `?download=1`; `public_url()` falling back to `url()` does not. `test_docu_media_export.py` asserts `'download=1' not in rst`, so this is deliberate — but it means that on a plain local storage service, which is every instance not configured otherwise, handbook attachments stop downloading and start rendering inline. That is a visible change for all of them and it is not in the upgrade notes. If it is wanted, say so there; if not, keep the `download` parameter.

## 4. The orphan `docu_redirect` table keeps real foreign keys

The upgrade note says the table "can be dropped manually". It is stronger than that. genropy emits actual DB-level constraints (`gnr/sql/adapters/_gnrbasepostgresadapter.py:1022`), and the deleted model declared three: `page_id` → `documentation.id`, and `old_handbook_id` / `new_handbook_id` → `docu.handbook.id`. Left in place on an existing database, deleting a documentation page or a handbook that still has a redirect row fails with a raw postgres foreign-key violation naming a table the model no longer knows about — and `onDelete='raise'` is no longer there to produce the framework's own message, so the error is unexplainable from the application side.

The repo already has the mechanism for this: `packages/<pkg>/lib/upgrades`, as used by `adm`, `email` and `glbl`. An upgrade script that drops `docu_redirect` turns a manual note into something that actually happens.

## Minor, not blocking

- `content_type = 'text/html'` on `DocuResolver` is dead: `serve_tool` sets the mimetype on the response it then discards as soon as a `Response` is returned.
- The 404 fallback body is the bare string `'Page not found'`, untranslated, while the rest of the PR updates `localization.xml`.
- `self.site.resource_loader.getResource(...)` reaches one level past `self.site.getResource(...)` (`gnrwsgisite.py:1140`), which is the same call.
- `_atcPublicUrl` defaults a prefix-less `filepath` to `home:`, while `atc_getAttachmentPath` writes `documentation:`. Worth a line on where the `home:` convention comes from, since it only fires for legacy rows and a wrong guess there silently returns `None` and falls back.

fporcari:
Correction on my own review above, and it matters for how you read point 2.

**The gate you implemented is the one I asked for, and my wording was wrong.** In [my previous round](https://github.com/genropy/genropy/pull/1085#pullrequestreview-4972373923) I wrote:

> take the public url only when the service actually exposes a public base — i.e. when `public_url()` differs from the instance url

Those two halves are not the same condition, and the parenthetical — the one you implemented — is the wrong one. For aws_s3 without `public_base_url`, `public_url()` returns the endpoint-style bucket url (`aws_s3.py:353-359`) which differs from `internal_url()`, so the "differs from the instance url" test passes on a service that exposes no public base at all. The gate I should have named is the first half: `public_base_url` actually being configured. So point 2 of my review is a correction of my earlier advice, not of your implementation — apologies for the round trip it cost.

Point 1 stands on its own and is not affected by that: `_atcPublicUrl` has no gate of any kind.

**Also closed from the previous round, and I should have said so above rather than restating verified ground:**

- The candidate query is bounded — `RESOLVER_CANDIDATES_LIMIT` plus the single-candidate early exit is exactly what genro's N+1 needed.
- Both dead Cloudfront localization keys are gone from `localization.xml`.
- The examples-sync work is no longer on the branch, so the note I left about it being missing from the description is moot.

What is left from me is points 1 to 4 of the review above, of which 1 and 2 are the same underlying fix.

fporcari:
Narrowing my review: **one blocking item, and it is not four.** Points 1 and 2 are the same defect at two call sites, and points 3 and 4 should not have been in the way.

### The one thing blocking

Media get linked to unsigned urls on services that do not declare a public base — attachments with no gate at all (`_atcPublicUrl`), images with a gate that passes on any aws_s3 configuration (`nodePublicUrl`). One fix covers both: gate on the service actually declaring `public_base_url`, rather than on `public_url()` differing from `internal_url()`.

I am holding this one rather than trusting it because of the direction of the error. Gated too tightly, an instance keeps embedding its own copies and someone asks why the CDN is not being used. Gated too loosely, a published handbook goes out with links to objects nobody granted anonymous read on — and that surfaces in a reader's browser weeks after the export, not in CI and not in the exporter's log, which is where the old download-and-embed path put its `Missing file`. The cheap failure and the expensive one are not symmetric here.

What I did not verify, and will not claim: whether any instance today keeps docu media on an S3 bucket without `public_base_url`. What is certain is that the code offers no way to stay on the old behaviour if one does.

### Demoted to non-blocking

- **The orphan `docu_redirect` foreign keys** (my point 4). Real, and worth an upgrade script in `packages/docu/lib/upgrades` rather than a manual note — but it only bites instances that actually used `manage_redirects`, and it is a normal follow-up rather than a reason to hold the PR. Take it here or in its own issue, your call.
- **The dropped `?download=1`** (point 3). Deliberate, tested for, and a UX change on local instances. A line in the upgrade notes is enough.
- **The four minors** — dead `content_type`, the untranslated `'Page not found'`, reaching past `site.getResource`, the `home:` default. All cosmetic.

The resolver half I would take as it stands, and said so two rounds ago. It is the media default I want gated before it reaches instances — once that gate is on `public_base_url`, I approve without another round.

dgpaci:
Addressed in `14d90990b`. Both media halves now go through one gate instead of two different guesses.

**1 and 2 — the gate is now `public_base_url`, for attachments and images alike.** `documentation.publicMediaUrl(node)` returns `public_url()` only when the node's service declares an explicit public base, and both `_atcPublicUrl` and the export's `imagePublicUrl` call it; the batch's own `nodePublicUrl` (the `public_url() != internal_url()` comparison) is gone. You were right that the comparison was the wrong test: for `aws_s3` without `public_base_url` the two urls differ, so a private bucket lost its embedded copies and published unsigned links — the opposite of what the note promised. With the gate on the configuration instead of on the urls, an instance that has not declared a public base behaves exactly as before this branch, and the attachment half is closed by the same condition.

**3 — `?download=1` is back for every service without a public base.** With `_atcPublicUrl` returning `None` there, `atcAsRstTable` falls back on its `%(host)s%(fileurl)s?download=1` template, so plain local storage — every instance not configured otherwise — keeps the download link it had. `test_atc_instance_storage_keeps_download_link` pins it (the old assertion was the inverse, which is what made the regression deliberate). Where a public base *is* configured the link is a plain public url with no download parameter, since it is a CDN url and not the instance's own endpoint: that is now stated in the upgrade notes.

New tests around the gate, all against real local storage services: a signing service declaring a public base and a signing service without one (`aws_s3` on a bucket nobody declared public) — attachments on the latter keep the instance download link, its images stay embedded, and `test_publicmediaurl_requires_a_public_base` covers the three services at the model level. 40 passed under `packages/docu/tests`, 155 under `gnrpy/tests/app`, flake8 clean on the package.

**4 — no upgrade script, but the note now says what actually happens.** We are deliberately not shipping one: dropping `docu_redirect` stays a manual step. The upgrade notes now state the consequence you described rather than calling it cosmetic — the table keeps the three DB-level foreign keys, so while it is still there with rows in it, deleting a page or a handbook referenced by a redirect row fails with a raw postgres violation naming a table the model no longer knows, with no `onDelete='raise'` left to turn it into a framework message — and they recommend the `DROP TABLE`.

**Minor notes.** `content_type` removed (`serve_tool` discards it for a `Response`). `self.site.getResource(...)` instead of reaching into `resource_loader`. The last-resort 404 body is localized (`!!Page not found`, with the Italian translation in `localization.xml`) and covered by a test that also asserts no `!!` marker survives. The `home:` default is no longer a local guess: the attachment node comes from the framework's own `AttachmentTable._atcStorageNode`, the same resolution `onArchiveExport` uses, with a comment saying so.

The nginx rule is now in the module docstring of `webtools/resolver.py` (`try_files` + named location proxying to `/_tools/docuresolver$uri`), not only in the PR body, so whoever deploys finds it next to the code that depends on it. The PR body has been rewritten: the media bullet describes the gate, the examples-sync note is moot now that #1086 is merged into develop, and the test counts match the branch.

fporcari (approved):
Approving on `31f2306d`. The one item I was holding is closed, and I checked the code rather than the reply.

**The gate is on the configuration now, at both call sites.** `publicMediaUrl` (`model/documentation.py:291-305`) returns `public_url()` only when `getattr(node.service, 'public_base_url', None)` is truthy, and it is the single entry point: `_atcPublicUrl` ends in it (`:289`) and the export's `imagePublicUrl` calls it through `self.doctable` (`export_to_sphinx.py:395`), with `nodePublicUrl` and its `public_url() != internal_url()` comparison gone. `public_base_url` exists in exactly one place in the tree — `aws_s3.py:123`, where an unconfigured service sets it to `''` — so on every other service, and on an S3 bucket nobody declared public, the gate is closed and the old behaviour stands. That closes points 1 and 2 together, as you say.

**`?download=1` is back where it was.** With `_atcPublicUrl` returning `None`, `atcAsRstTable` falls through to the `%(host)s%(fileurl)s?download=1` template (`:260`, `:266-268`), so a plain local instance keeps the link it had. `test_atc_instance_storage_keeps_download_link` pins it, and it asserts the presence rather than the absence this time.

**The attachment node comes from the framework.** `AttachmentTable._atcStorageNode` (`gnrpy/gnr/app/gnrdbo.py:1796-1800`) carries the same `home:` default the local copy did, so the legacy-row convention has one owner instead of two. Good removal.

The tests go through real local storage services on temporary directories, not mocks: `SigningLocalService` declaring a public base and `PrivateBucketLocalService` without one is the pair that makes the gate testable, and `test_publicmediaurl_requires_a_public_base` covers the three services at the model level. `40 passed` under `packages/docu/tests` and flake8 clean on the package, both reproduced here. The merge from develop is clean — against `origin/develop` this branch touches nothing outside `packages/docu`, so nothing that landed in the meantime was disturbed.

Also verified: the `!!Page not found` key resolves — `translate` flattens it to `en_page_not_found` (`gnrlocalization.py:118-121`), which is the key added under `<webtools><resolver>` in `localization.xml:201-203`. And the nginx `try_files` + named location is now in the module docstring where whoever deploys will find it.

Three notes, none of them worth another round.

- **`AppLocalizer(self.site.gnrapp)` builds a second localizer.** `GnrApp.__init__` already holds one (`gnrapp.py:1046`), and the constructor walks every package folder to rebuild the localization dict (`gnrlocalization.py:64`). `self.site.gnrapp.localizer.translate('!!Page not found')` is the same line without the rebuild. It only matters on the last-resort path — the framework's own `missing_result.html` missing *and* no instance override — so the cost is theoretical; but the docu package now does this in three places (`documentation.py:192`, `export_to_sphinx.py:270`, and here), and it is worth fixing at the source rather than propagating.
- **`public_base_url` is now the contract for "publishable".** Only `aws_s3` declares it, so a future storage service that publishes on a CDN keeps its media embedded until it grows the same attribute. That is the safe direction and I would not change it — just worth knowing that the attribute name, not a method, is what the gate reads.
- **`PrivateBucketLocalService` omits `public_base_url` entirely**, so the test exercises the `getattr` default; real `aws_s3` without configuration sets `''` instead. Both are falsy and the outcome is identical, so the coverage is honest — noting it only because the two paths through the gate are not the same expression.

The `docu_redirect` foreign keys are described accurately in the upgrade notes now, and I agree the `DROP TABLE` can stay a manual step rather than hold this. The resolver half I would have taken two rounds ago; the media default is gated where it needed to be.

PR #1086, opened 2026-08-18
fix(docu): keep example modules in sync with the documentation lifecycle

dgpaci:
## Motivation

The `.py` example modules served from `site:webpages/docu_examples` are written by the `docu.documentation` triggers, but they fell out of sync with the record lifecycle in several cases:

- **Rename/move left the old folder orphaned.** `trigger_onUpdating` always rewrites the `url` attributes inside the sourcebag with the new `hierarchical_name`, so in `trigger_onUpdated` the equality check `record['sourcebag']==old_record['sourcebag']` guarding the folder move could never be true: the files were written at the new location and the old folder stayed behind.
- **Delete left the folder behind.** There was no `trigger_onDeleted` at all.
- **Rewriting a parent's sourcebag wiped the descendants' examples.** The cleanup loop in `writeModulesFromSourceBag` deleted the whole `tutorialRecordNode` as soon as it found a file child, but with hierarchical names the folder of a parent record contains the subfolders of its children records.
- **The export could not rebuild a lost site folder.** The `site` folder is disposable by definition, but the only code paths regenerating the example files were the update trigger and the per-record `checkSourceBagModules`.
- **`updateLink` fetched and rewrote almost the whole table on every rename.** The LIKE pattern was never interpolated with the searched link (`old_link_query='%%%s%%'` reached the database as a literal pattern, i.e. "any docbag containing an *s*"), the condition was duplicated in the where clause, and `batchUpdate` rewrites every fetched row, locking them with `SELECT FOR UPDATE`.

## Changes

- `trigger_onUpdated`: on a `hierarchical_name` change the old node files are deleted and the modules rewritten at the new location. The hierarchical cascade updates every descendant through full triggered updates, so each child record relocates its own folder.
- New `trigger_onDeleted` removes the example files of the deleted record.
- New `deleteSourceBagModules` deletes only the **file** children of the record node, keeping the subfolders of the children records intact; the folder itself is removed only when empty. `writeModulesFromSourceBag` now delegates the cleanup to it.
- `export_to_sphinx.prepare()` calls `writeModulesFromSourceBag` for every exported document with a sourcebag, so a full export regenerates `site:webpages/docu_examples` from scratch.
- `updateLink` builds the LIKE pattern from the actual old link and the callback returns `False` for rows that do not contain it, so only the linking records are rewritten. The pattern is intentionally not wildcard-escaped: the sqlite adapter translates `ILIKE` to a plain `LIKE` with no `ESCAPE` clause, so escaping would break matching there; the exact check in the callback makes the SQL pattern a pure pre-filter.

## Test cases

`projects/gnrcore/packages/docu/tests/test_examples_sync.py` (8 tests), run with:

```
gnr dev tests <instance> docu
```

The tests are self-contained: they build a throwaway application on a temporary sqlite file and a real local storage service on a temporary directory, so records go through the actual table triggers and assertions read the actual filesystem — no running site nor shared instance database needed.

- sourcebag update writes the module files with the right content
- rename relocates the files and removes the old folder
- move under another parent relocates the files
- rewriting a parent's sourcebag keeps the children's files intact (regression for the folder wipe)
- renaming a parent cascades: both the parent and the children folders are relocated
- delete removes the files
- rename rewrites the rst links inside the docbag of the linking records (names with underscores, guarding the LIKE pattern against escaping regressions)
- bulk regeneration after the loss of the site folder, through the same call the export batch uses

Framework suites unaffected: `tests/app` 163 passed, `tests/sql` 1202 passed, `tests/core` 428 passed with 2 pre-existing failures in `flatfiles_test.py` (present on a clean develop checkout as well, unrelated).

flake8 clean on all touched files.

Note: running the package tests through the `DbBasedTest`/`PackageTester` base currently fails on a fresh checkout because the committed `gnrdevelop` instance databases are not aligned with the current models (`no such table: adm.adm_preference` while booting `GnrWsgiSite`) — the `sys` package tests hit the same wall. These tests avoid that dependency on purpose; the stale instance data deserves a separate issue.

### DIFF
--- a/gnrpy/tests/app/test_docu_examples_sync.py
+++ b/gnrpy/tests/app/test_docu_examples_sync.py
@@ -1,0 +1,183 @@
+"""Tests for the example-module synchronization of docu.documentation.
+
+The .py example files served from site:webpages/docu_examples must follow
+the record lifecycle: written when the sourcebag changes, relocated when
+the record is renamed or moved in the hierarchy (keeping the folders of
+the children records intact), removed when the record is deleted, and
+regenerable in bulk from the sourcebag since the site folder is
+disposable. The rst links pointing to a renamed record must be rewritten
+in the docbag of the other records.
+
+The tests run against a real database (gnrdevelop instance on a temporary
+sqlite file) and a real local storage service backed by a temporary
+directory: records go through the actual table triggers and assertions
+read the actual filesystem.
+"""
+import shutil
+import tempfile
+from types import SimpleNamespace
+
+from gnr.core.gnrbag import Bag
+from gnr.lib.services.storage import BaseLocalService, StorageNode
+
+from common import BaseGnrAppTest
+
+
+class StorageSiteStub:
+    """Minimal site wiring exposing a real local storage service to the model."""
+
+    def __init__(self, services_dirs):
+        self.currentPage = SimpleNamespace(isMobile=False, user=None)
+        self.services = {}
+        for name, base_path in services_dirs.items():
+            service = BaseLocalService(parent=self, base_path=base_path)
+            service.service_name = name
+            self.services[name] = service
+
+    def storageNode(self, fullpath, *parts):
+        if parts:
+            fullpath = '/'.join([fullpath] + list(parts))
+        service_name, path = fullpath.split(':', 1)
+        return StorageNode(parent=self, path=path, service=self.services[service_name])
+
+
+def sourceBag(*versions):
+    result = Bag()
+    for version, source in versions:
+        result.setItem(version, Bag(dict(version=version, source=source)))
+    return result
+
+
+class TestDocuExamplesSync(BaseGnrAppTest):
+
+    @classmethod
+    def setup_class(cls):
+        super().setup_class()
+        cls.db = cls.app.db
+        cls.db.model.check(applyChanges=True)
+        cls.doctbl = cls.db.table('docu.documentation')
+        cls.site_dir = tempfile.mkdtemp(prefix='gnr_docu_site_')
+        cls.app.site = StorageSiteStub({'site': cls.site_dir})
+
+    @classmethod
+    def teardown_class(cls):
+        shutil.rmtree(cls.site_dir, ignore_errors=True)
+        super().teardown_class()
+
+    # ------------------------------------------------------------------
+    # helpers
+    # ------------------------------------------------------------------
+
+    def examplesNode(self, hierarchical_name):
+        return self.app.site.storageNode('site:webpages/docu_examples/%s'
+                                         % hierarchical_name)
+
+    def exampleFiles(self, hierarchical_name):
+        """Sorted example file names, or None if the folder is missing."""
+        node = self.examplesNode(hierarchical_name)
+        if not node.exists:
+            return None
+        return sorted(n.basename for n in node.children() if not n.isdir)
+
+    def exampleSource(self, hierarchical_name, filename):
+        with self.examplesNode(hierarchical_name).child(filename).open('r') as f:
+            return f.read()
+
+    def addDoc(self, name, parent_id=None, sourcebag=None, docbag=None):
+        record = dict(name=name, parent_id=parent_id)
+        self.doctbl.insert(record)
+        if sourcebag is not None or docbag is not None:
+            with self.doctbl.recordToUpdate(record['id']) as rec:
+                if sourcebag is not None:
+                    rec['sourcebag'] = sourcebag
+                if docbag is not None:
+                    rec['docbag'] = docbag
+        self.db.commit()
+        return record['id']
+
+    def updateDoc(self, pkey, **values):
+        with self.doctbl.recordToUpdate(pkey) as rec:
+            rec.update(values)
+        self.db.commit()
+
+    # ------------------------------------------------------------------
+    # sourcebag lifecycle
+    # ------------------------------------------------------------------
+
+    def test_write_on_sourcebag_update(self):
+        self.addDoc('intro', sourcebag=sourceBag(('v000', "print('first')"),
+                                                 ('v010', "print('second')")))
+        assert self.exampleFiles('intro') == ['v000.py', 'v010.py']
+        assert self.exampleSource('intro', 'v000.py') == "print('first')"
+
+    def test_rename_relocates_files(self):
+        pkey = self.addDoc('old_page', sourcebag=sourceBag(('v000', "print('x')")))
+        assert self.exampleFiles('old_page') == ['v000.py']
+        self.updateDoc(pkey, name='new_page')
+        assert self.exampleFiles('old_page') is None
+        assert self.exampleFiles('new_page') == ['v000.py']
+
+    def test_move_relocates_files(self):
+        shelf_id = self.addDoc('shelf')
+        pkey = self.addDoc('orphan', sourcebag=sourceBag(('v000', "print('x')")))
+        self.updateDoc(pkey, parent_id=shelf_id)
+        assert self.exampleFiles('orphan') is None
+        assert self.exampleFiles('shelf/orphan') == ['v000.py']
+
+    def test_parent_rewrite_keeps_children_files(self):
+        guide_id = self.addDoc('guide', sourcebag=sourceBag(('v000', "print('guide')")))
+        self.addDoc('step', parent_id=guide_id,
+                    sourcebag=sourceBag(('v000', "print('step')")))
+        self.updateDoc(guide_id, sourcebag=sourceBag(('v000', "print('guide!')"),
+                                                     ('v010', "print('more')")))
+        assert self.exampleFiles('guide') == ['v000.py', 'v010.py']
+        assert self.exampleSource('guide', 'v000.py') == "print('guide!')"
+        assert self.exampleFiles('guide/step') == ['v000.py']
+
+    def test_rename_parent_cascades_to_children(self):
+        book_id = self.addDoc('book', sourcebag=sourceBag(('v000', "print('book')")))
+        self.addDoc('chapter', parent_id=book_id,
+                    sourcebag=sourceBag(('v000', "print('chapter')")))
+        self.updateDoc(book_id, name='volume')
+        assert self.examplesNode('book').exists is False
+        assert self.exampleFiles('volume') == ['v000.py']
+        assert self.exampleFiles('volume/chapter') == ['v000.py']
+
+    def test_delete_removes_files(self):
+        pkey = self.addDoc('trash', sourcebag=sourceBag(('v000', "print('x')")))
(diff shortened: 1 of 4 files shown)


fporcari (dismissed):
The three lifecycle holes are real and the diagnosis of each is exact. I checked the two claims that carry the change and they hold:

- **The rename guard could never fire.** `trigger_onUpdating` rewrites the `url` attributes in the sourcebag with the new `hierarchical_name`, so `record['sourcebag']==old_record['sourcebag']` in `trigger_onUpdated` was false by construction and `tutorialRecordNode(old).move(...)` was dead code. Deleting the old node's files and rewriting at the new path is the right shape.
- **`updateLink` really did rewrite the table.** `old_link_query='%%%s%%'` is a plain literal, not a format string, so the pattern reaching the database was `%%%s%%` — "any docbag containing an s" — and `batchUpdate` takes `for_update=for_update or True` (`gnrpy/gnr/sql/gnrsqltable/crud.py:373`), so every fetched row was locked and raw-updated. The interpolated pattern plus the exact check in `cb` fixes both halves, and `_batchUpdate_rows` does honour a `False` return with `continue` (`crud.py:411-413`), so the pre-filter/exact-check split works as described.

On the cascade, which is the part the diff makes hardest to read: I probed two cases the suite does not cover, on this head, and both behave correctly — a parent that has **no sourcebag of its own** but whose child has one (old folder gone, child relocated), and a **three-level** rename (`aa/bb/cc` → `zz/bb/cc`, old root gone). The descendants relocate before the ancestor's `deleteSourceBagModules` finds its folder empty, so the "delete the folder only when empty" rule does not leave orphan directories behind.

Verified: `tests/app` 163 passed, twice in a row (the fixture builds a fresh temp instance per class, so the filesystem assertions are idempotent). `flake8` clean on the three touched files.

Two things, neither of them a blocker:

- **Test placement.** cgabriel asked on #1085 for `gnrpy/tests/app/test_docu*` to move under `docu/tests`, and the same question applies to this file. Worth settling once for both PRs rather than twice: `gnrpy/pyproject.toml` has `testpaths = ["tests"]` and CI runs `cd gnrpy && pytest`, so `projects/gnrcore/packages/*/tests` is collected by nothing in the GitHub workflow — only by `gnr tests <instance> [package]` (`gnrpy/gnr/dev/cli/gnrtests.py`), which needs an installed instance and the `DbBasedTest`/`PackageTester` base rather than `BaseGnrAppTest`. Moving the file as-is would take it out of every automated run; moving it properly means porting it onto that harness and adding `gnr tests` to `tests.yml`.
- **`prepare()` now writes on every export.** `writeModulesFromSourceBag` deletes and rewrites the files of each exported record, so a full export of a large handbook does a delete+write per version on the storage service instead of a read. Correct, and it is what makes the site folder rebuildable; just note it is IO the export did not do before, and it will be felt on a remote storage backend.

This and #1085 both touch `model/documentation.py` and `export_to_sphinx.py` and merge cleanly against each other (tested), with no shared hunk. Whichever lands first, the other needs no manual resolution.

fporcari (approved):
Re-reviewed on `473969353`. The production diff is byte-identical to the head I approved before the stale-review dismissal (`e8aad53`), so the two claims that carry the change still hold, and I re-verified both citations on this head:

- **The rename guard could never fire.** `trigger_onUpdating` rewrites the `url` attributes inside the sourcebag with the new `hierarchical_name`, so `record['sourcebag']==old_record['sourcebag']` in `trigger_onUpdated` was false by construction and `tutorialRecordNode(old).move(...)` was dead code. Delete-the-old-node + rewrite-at-the-new-path is the right shape.
- **`updateLink` really did rewrite the table.** `old_link_query='%%%s%%'` is a plain literal, not a format string, so the pattern reaching the database was `%%%s%%` — "any docbag containing an *s*" — and `batchUpdate` passes `for_update=for_update or True` (`gnrpy/gnr/sql/gnrsqltable/crud.py:373`), so every fetched row was locked and raw-updated. The interpolated pattern plus the exact check in `cb` fixes both halves, and `_batchUpdate_rows` does honour a `False` return with `continue` (`crud.py:418-419`).

New probe on this head, on a case the suite does not cover: **deleting a parent whose children own example files**. The cascade removes the children first, so each child's `trigger_onDeleted` empties its own folder and the parent's `deleteSourceBagModules` then finds its folder empty and removes it — `docu_examples` is left clean with no orphan directory. Same result as the two cascade cases I probed last time (a parent with no sourcebag of its own, and a three-level rename).

Verified on `473969353`: `gnr dev tests gnrdevelop docu` → 8 passed, twice in a row (temp sqlite file and temp storage dir per class, so the filesystem assertions are idempotent); `gnrpy` `tests/app` → 155 passed; `flake8` clean on the three touched files, both full and with the CI selection; CI green on 3.11–3.14.

Two notes on the move, neither of them blocking.

**1. No automated run collects the file any more.** `gnrpy/pyproject.toml` has `testpaths = ["tests"]` and `tests.yml` runs only `cd gnrpy && pytest`, so `projects/gnrcore/packages/*/tests` is collected by no workflow — `sys` and `multidb` sit in the same position, so this is a repo gap the PR inherits rather than one it creates. It does mean the eight regression tests for this fix only run when somebody types the command by hand. Worth a separate issue adding a `gnr dev tests <instance>` step to `tests.yml`; until that exists, the regressions fixed here can come back unnoticed.

**2. The class now resolves the instance through the ambient environment.** Dropping `BaseGnrAppTest` also dropped `BaseGnrTest._build_test_env` (`gnrpy/tests/core/common.py:38`), which wrote a temporary `GENRO_GNRFOLDER` whose `gnrhome` and projects paths were derived from the checkout containing the test file. `GnrApp('gnrdevelop')` now goes through the developer's real `~/.gnr` instead, so the file tests whichever checkout that environment points at. Running it from a worktree of this branch on a machine whose `environment.xml` points at another checkout, 5 of the 8 tests fail: the `docu` model under test is the installed one, not the one sitting next to the test.

\#1085 solves exactly this in the same directory: its `projects/gnrcore/packages/docu/tests/conftest.py` puts the checkout's own `gnrpy/tests` and `gnrpy` first on `sys.path` so package tests can keep using `BaseGnrAppTest`. I merged the two branches (clean, no conflicts on any file) and put this class back on that base — 6 insertions, 9 deletions — and the file then passes from the worktree with no environment setup at all:

```python
from core.common import BaseGnrAppTest


class TestDocuExamplesSync(BaseGnrAppTest):

    @classmethod
    def setup_class(cls):
        super().setup_class()
        cls.db = cls.app.db
        ...

    @classmethod
    def teardown_class(cls):
        cls.db.closeConnection()
        shutil.rmtree(cls.site_dir, ignore_errors=True)
        super().teardown_class()
```

The trade-off is that `BaseGnrAppTest.app_name` is hard-coded to `gnrdevelop` and ignores the `GNR_TESTING_INSTANCE_NAME` the `gnr dev tests` CLI exports, which the current version honours; `app_name` is a class attribute, so a one-line override covers it. Either take the conftest into this PR (8 lines) or let #1085 land first and follow up.

**One wording fix for the description.** `prepare()` walks `getHierarchicalData(root_id=<handbook docroot>, condition='$is_published IS TRUE')` (`export_to_sphinx.py:44`), and in the `toc_roots` branch it also skips records with no `publish_date`, so an export regenerates the examples of the published pages of that handbook — not `site:webpages/docu_examples` from scratch. After a loss of the site folder, unpublished pages and pages outside the exported handbook stay missing. Worth stating precisely, because "from scratch" is what somebody will rely on when the folder is gone.

Approving: the fix is correct, and both notes are about where the tests run rather than about what the code does.

genro (approved):
Approving. I read the diff against develop and ran the suite independently: 8/8 pass on `473969353`.

The five defects in the description are real and I confirmed each one on develop: `trigger_onUpdating` rewriting `v['url']` makes the `record['sourcebag']==old_record['sourcebag']` guard unreachable on a rename, so `.move()` never ran; `old_link_query='%%%s%%'` is never interpolated and reaches the database as six literal characters, i.e. LIKE "contains an s"; there was no `trigger_onDeleted`; and the cleanup loop deleted the whole node on the first file child, taking the children's subfolders with it. The duplicated `OR` in the where clause was redundant but harmless.

I checked the claim that only the linking records are rewritten: `_batchUpdate_rows` does `doUpdate = updatercb(new_row); if doUpdate is False: continue` (`gnrpy/gnr/sql/gnrsqltable/crud.py:418-420`), so the callback's `False` genuinely skips the write and the unescaped LIKE stays a pure pre-filter. That holds.

The lifecycle is sound, including the case I went looking for a hole in. `deleteSourceBagModules` only removes the folder when `children()` is empty, so I expected a parent's folder to be left behind after a cascade. It is not: `deleteRelated` is step 4 of the delete sequence and `trigger_onDeleted` is step 8 (`gnrpy/gnr/sql/gnrsql/write.py:277-284`), and `deleteRelated` cascades through `relatedTable.delete(row)` (`crud.py:218-220`), a full ORM delete that fires each descendant's own trigger. By the time the parent's trigger runs its folder holds only its own files. The same ordering explains the rename cascade. I also tried a case the tests do not cover — a parent *without* a sourcebag whose child has one, renamed — and it relocates cleanly with no orphan folder.

Three follow-up notes, none of them blocking:

- **`test_delete_removes_files` only covers a childless record.** Deleting a parent with children is not exercised, and with the current stub it cannot be: `deleteRelated` goes through `db.quickThermo`, which reaches `site.currentPage.utils`, and `StorageSiteStub` uses `SimpleNamespace(isMobile, user)`. Adding `utils` to the stub would open that branch.
- **The `if self.curr_sourcebag:` guard in `export_to_sphinx.prepare` limits the self-healing.** `writeModulesFromSourceBag` always calls `deleteSourceBagModules` first, so calling it for records with an emptied sourcebag would also clear their stale folders; with the guard those survive a full export. Dropping it costs one directory listing per document and makes the export truly regenerative.
- **Nit on the link test:** it asserts the real match is rewritten, but not that a row matching *only* through the underscore overmatch is skipped. A second record carrying `link0target0v1` in its docbag would pin the pre-filter.

One note for anyone else reviewing this locally: these tests cannot be run from a git worktree without extra setup. `~/.gnr/environment.xml` hardcodes `<genropy path=".../genropy/projects">`, so `GnrApp('gnrdevelop')` loads the docu model from the main checkout regardless of which worktree pytest runs in — my first run reported 5 failures because it was exercising the pre-PR code. A throwaway config directory named `.gnr` with the projects path repointed, passed via `GENRO_GNRFOLDER`, gives a correct run.

fporcari:
The CI note in my review now has an issue behind it: #1094. It covers the two pieces needed before `projects/gnrcore/packages/*/tests` runs anywhere automatic — a `gnr dev tests` step in `tests.yml`, and `gnr/dev/cli/gnrtests.py` exiting non-zero when a collected test fails, which it does not do today. Nothing in this PR depends on it.

PR #1087, opened 2026-08-18
fix(deps): replace watchgod with watchfiles to lift the anyio<4 ceiling

### LINKED ISSUE #1078: [BUG] watchgod holds anyio below 4, breaking uvicorn and httpcore, for a single import
## Problem Description

`gnrpy/pyproject.toml:60` declares `watchgod`, whose own requirement is `anyio<4,>=3.0.0`. Installing genropy into an environment that also runs an ASGI stack silently downgrades `anyio` from 4.x to 3.7.1. `httpcore` 1.0.9 requires `anyio>=4` for its `asyncio` extra, so the resulting environment fails `pip check` and the async stack ends up on a runtime it does not support.

`watchgod` is imported in exactly one place — `gnrpy/gnr/web/cli/gnrtaskscheduler.py:7`, `from watchgod import awatch` — a CLI tool, outside the request path.

**Expected:** installing genropy next to uvicorn/httpx does not move the async runtime.

**Actual:** `anyio` 4.14.2 → 3.7.1, and afterwards

```
$ pip check
watchgod 0.8.2 has requirement anyio<4,>=3.0.0, but you have anyio 4.14.2.
```

whichever side is restored — the two requirements cannot both hold.

## Reproduction Steps

1. Create a virtualenv with an ASGI stack installed (uvicorn, httpx, httpcore) — `anyio` resolves to 4.x.
2. `pip install -e gnrpy[pgsql]`.
3. Observe `anyio-3.7.1` among the installed packages.
4. `pip install "anyio>=4"` to restore the async stack → `pip check` now reports `watchgod` unsatisfied.

## Environment

- genropy `develop` @ `fef1037bb`, Python 3.13.2, macOS 15 (Darwin 25.0.0)
- watchgod 0.8.2, anyio 4.14.2, httpcore 1.0.9, httpx 0.28.1

## Additional Material

Proposal: move to `watchfiles`, the maintained successor by the same author, which carries no `anyio` bound. The single `awatch` call site in `gnrtaskscheduler.py` should be checked against its signature as part of the change.

The conflict is not specific to this bench: it is latent in any environment where genropy and an `anyio>=4` consumer coexist, and it resolves silently in whichever direction the install order happens to favour.

fporcari:
## Problem

`gnrpy/pyproject.toml` declares `watchgod` as a dependency. `watchgod` itself requires `anyio<4,>=3.0.0`. In any environment that also runs a modern ASGI stack (`httpcore>=1` needs `anyio>=4` for its asyncio extra), this silently downgrades `anyio` to 3.7.1 and the environment fails `pip check`.

## Root cause

`watchgod` is imported in exactly one place in the shipped tree: `gnrpy/gnr/web/cli/gnrtaskscheduler.py`, where `awatch()` is used only as an edge trigger to restart a child process on code changes — the yielded change set is discarded, so no call-site behavior depends on the specific library. `watchfiles` is the maintained successor by the same author, exposes the same `awatch(path)` signature, and declares `anyio>=3.0.0` with no upper bound, removing the conflicting ceiling.

## Change

- `gnrpy/pyproject.toml`: `watchgod` -> `watchfiles` in `dependencies` (alphabetical position preserved)
- `gnrpy/gnr/web/cli/gnrtaskscheduler.py`: `from watchgod import awatch` -> `from watchfiles import awatch`

Considered and rejected: pinning an explicit `anyio` floor in `pyproject.toml`. The dependency list is overwhelmingly unpinned (only `chardet`, `smart_open` and `werkzeug` carry specifiers), and the `watchfiles` swap already removes the conflicting constraint, so an extra pin would be unnecessary.

## Verification

- `flake8` on the touched Python file: clean.
- Import check: `gnr.web.cli.gnrtaskscheduler` imports cleanly with `watchfiles`, confirmed resolving from this branch's checkout (not a stray editable install elsewhere).
- Full test suite: `pytest gnrpy/tests/` — 1958 passed, 67 skipped, 0 failed (includes `tests/sql`, `gnrasync_test.py`, `gnrasync_session_test.py`).
- Not verified: the autoreload loop itself was not exercised live, since it requires a site configured for the async task manager. Manual steps to verify: run `gnrtaskscheduler <sitename> --autoreload` against such a site, `touch` a `.py` file under `gnrpy`, and confirm the `Detected code changes, restarting...` log line together with a new child process PID; repeat once to confirm the watch loop survives past the first yield.

Fixes #1078

### DIFF
--- a/gnrpy/gnr/web/cli/gnrtaskscheduler.py
+++ b/gnrpy/gnr/web/cli/gnrtaskscheduler.py
@@ -4,7 +4,7 @@
 import sys
 import os.path
 import asyncio
-from watchgod import awatch
+from watchfiles import awatch
 from multiprocessing import Process
 import importlib
 
--- a/gnrpy/pyproject.toml
+++ b/gnrpy/pyproject.toml
@@ -57,7 +57,7 @@ dependencies = [
 	     "tzlocal",
 	     "unidecode",
 	     "vobject",
-	     "watchgod", # used in asyncio for autoreload
+	     "watchfiles", # used in asyncio for autoreload
 	     "weasyprint",
 	     "werkzeug>=2.3.8,!=3.1.4",
 	     "xlrd",


genro (approved):
Approving. Verified the two claims the change rests on.

`watchgod` appears in exactly two places in the shipped tree — `gnrpy/pyproject.toml:60` and `gnrpy/gnr/web/cli/gnrtaskscheduler.py:7` — so the swap has no other call site to reach. And the usage really is an edge trigger only: `async for changes in awatch(CODE_DIR)` at `gnrtaskscheduler.py:28` never reads `changes` in the loop body, it just terminates and respawns the child. Nothing depends on the shape of what the library yields, which is what makes the substitution safe rather than merely plausible.

Agreed on rejecting the explicit `anyio` floor. The dependency list is almost entirely unpinned, and a pin that exists only to compensate for a constraint this PR removes would be dead weight the day someone reads it.

One thing I could not verify locally: `watchfiles` is not installed in my environment, so I did not exercise the `awatch` signature myself. The indirect evidence is CI green on all four Python versions with the dependency swapped, which at least establishes that it resolves and installs. Combined with your note that the autoreload loop was not exercised live, the residual risk sits entirely in that loop — worth someone running the manual steps you listed on a site configured for the async task manager before the next release, since nothing in the suite covers it.

PR #1088, opened 2026-08-18
fix(th): hdbSelect accepts the table kwarg instead of requiring the deprecated dbtable

### LINKED ISSUE #931: [BUG] hdbSelect raises KeyError when using 'table' instead of deprecated 'dbtable'
## Problem Description

The `hdbSelect` widget raises a `KeyError: 'dbtable'` when the table is passed via the `table` argument instead of `dbtable`.

This is inconsistent with `dbSelect`, where `dbtable` is effectively deprecated in favor of `table`: on the client side `dbBaseCombo.resolver` normalizes the two with `resolverAttrs.dbtable = resolverAttrs.dbtable || objectPop(resolverAttrs,'table')` (`gnrjs/gnr_d20/js/genro_widgets.js`), so `dbSelect(table=...)` works fine.

For `hdbSelect`, instead, the Python handler `ht_hdbselect` reads `attr['dbtable']` directly:

```python
# resources/common/th/th_tree.py
menupath = 'gnr.htablestores.%s_%s' % (attr['dbtable'], connectedMenu)   # KeyError if only 'table' was passed
dbselect.treemenu(storepath=menupath, table=attr['dbtable'], ...)
```

When the user passes `table=...`, the `dbtable` key is never set in `attr`, so accessing `attr['dbtable']` raises `KeyError`.

**Observed behavior:** `KeyError: 'dbtable'` when calling `hdbSelect(table=...)`.

**Expected behavior:** since `dbtable` is deprecated in favor of `table` for `dbSelect`, `hdbSelect` should accept `table` as well (and behave consistently), rather than requiring the deprecated `dbtable`.

## Reproduction Steps

1. In a webpage, add an `hdbSelect` passing the table via `table`:
   ```python
   fb.hdbSelect(value='^.parent_id', table='adm.htable', caption_field='description')
   ```
2. Load the page.

Observed result: `KeyError: 'dbtable'` raised in `ht_hdbselect` (`resources/common/th/th_tree.py`).

Workaround: pass `dbtable='adm.htable'` instead of `table=...`.

- Module involved: `HTableTree.ht_hdbselect` in `resources/common/th/th_tree.py`
- The handler accesses `attr['dbtable']` (two occurrences) with no fallback to `attr['table']`.

## Environment

genropy, develop branch

## Additional Material

**Proposed solution:** in `ht_hdbselect`, normalize the table argument the same way `dbSelect` does on the client side — accept `table` and fall back to it when `dbtable` is missing, e.g.:

```python
dbtable = attr.get('dbtable') or attr.get('table')
```

and use this local variable in place of `attr['dbtable']`. This makes `hdbSelect` consistent with the now-preferred `table` argument of `dbSelect`.

fporcari:
## Problem

`hdbSelect(table=...)` raises `KeyError: 'dbtable'`, while `dbSelect(table=...)` works fine. The failure looks intermittent: it only happens the first time a given `connectedMenu` is built on a page.

## Root cause

`HTableTree.ht_hdbselect` (`resources/common/th/th_tree.py`) builds `dbselect = pane.dbselect(**kwargs)` and reads `attr = dbselect.attributes`, which is literally the raw kwargs the caller passed — there is no server-side normalization step that turns `table` into `dbtable` (that only happens client-side, in `objectExtract`/`objectPop` in `gnrjs/gnr_d11/js/genro_widgets.js`). Two lines inside the `if connectedMenu not in currentHMenu:` block then read `attr['dbtable']` directly, so a bare `table=` kwarg raises `KeyError`.

This is easy to miss because:
- `fb.field(..., tag='hdbselect')` gets `dbtable` injected automatically by the field builder (`gnrpy/gnr/web/gnrwebstruct/dojo11.py`), so only the explicit `fb.hdbSelect(table=...)` form hits it.
- a second `hdbSelect` sharing the same `connectedMenu` skips the branch entirely (menu already built), so the bug looks intermittent depending on call order.

The same bug shape exists in the legacy `HTableHandlerBase.ht_hdbselect_old` (`resources/common/gnrcomponents/htablehandler.py`), which builds its `menupath` via `%`-formatting on `attr` and also reads `attr['dbtable']` directly. It has no callers in the tree today, but the root cause is identical so it is fixed at the same time.

## Change

In both `ht_hdbselect` methods, right after reading `dbselect.attributes`:

```python
dbtable = attr.get('dbtable') or attr.pop('table', None)
attr['dbtable'] = dbtable
```

then use the local `dbtable` variable at the call sites that previously read `attr['dbtable']`. This mirrors the client-side resolver (fall back from `dbtable` to `table`, then remove `table`) and also avoids leaking a stray `table` attribute into the treemenu subtree that `dbselect.treemenu(...)` builds as a child node — node attributes are inherited down the struct, and `table` is an actively consulted inherited key elsewhere in the struct code.

Added a regression case to `projects/gnrcore/packages/test/webpages/inputfields/dbselect.py`: an `hdbSelect` on the hierarchical `adm.htag` table using `table=`, and a second one using the old `dbtable=` spelling with the same `connectedMenu`, so the test page exercises both spellings and the shared-menu branch (the one that previously skipped the fix entirely).

## Verification

- `flake8` on the three touched files: zero errors.
- `python -m py_compile` on the three touched files: clean.
- `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql`: 787 passed, 57 skipped, no failures.
- `gnrpy/tests/sql` was not used as evidence (environmental: concurrent sessions on this machine share one postgres instance and their fixtures kill each other's connections).

**Not verified live.** Rendering the actual test page needs a running `gnrdevelop` instance against postgres, which I did not spin up. Manual steps to confirm:
1. Start the `gnrdevelop` instance (`projects/gnrcore/instances/gnrdevelop`), open `/test/inputfields/dbselect`.
2. Confirm the page renders with no traceback (pre-fix, the new `table=` field raises `KeyError: 'dbtable'`).
3. Click the new field's down-arrow, confirm the tree popup opens and lists the `adm.htag` hierarchy, pick a leaf, confirm the value is written.
4. Confirm the second field (old `dbtable=` spelling, same `connectedMenu`) still behaves, proving no regression on the old spelling and covering the shared-menu branch.

## Open point for @dgpaci

When neither `table` nor `dbtable` is passed, the call used to raise a `KeyError` right away and now silently degrades to `dbtable=None`, which will fail confusingly further downstream in `db.table(None)`. Worth raising an explicit, clearer error for that case — left out of scope here.

Fixes #931

### DIFF
--- a/projects/gnrcore/packages/test/webpages/inputfields/dbselect.py
+++ b/projects/gnrcore/packages/test/webpages/inputfields/dbselect.py
@@ -77,3 +77,11 @@ class GnrCustomWebPage(object):
                         selected_regione='.regione',
                         invalidItemCondition_message='troppo lungo',auxColumns='$regione')
         fb.div('^.regione')
+
+    def test_6_hdbselect(self,pane):
+        "hdbSelect on a hierarchical table: table= kwarg and the deprecated dbtable= spelling sharing a connectedMenu"
+        fb = pane.formbuilder(cols=1, border_spacing='4px')
+        fb.hdbSelect(value='^.tag_id',table='adm.htag',lbl='Tag',width='25em',
+                    caption_field='hierarchical_description',connectedMenu='htag_menu')
+        fb.hdbSelect(value='^.tag_id_2',dbtable='adm.htag',lbl='Tag (old spelling)',width='25em',
+                    caption_field='hierarchical_description',connectedMenu='htag_menu')
--- a/resources/common/gnrcomponents/htablehandler.py
+++ b/resources/common/gnrcomponents/htablehandler.py
@@ -184,12 +184,14 @@ class HTableHandlerBase(BaseComponent):
     def ht_hdbselect_old(self,pane,**kwargs):
         dbselect = pane.dbselect(**kwargs)
         attr = dbselect.attributes
+        dbtable = attr.get('dbtable') or attr.pop('table', None)
+        attr['dbtable'] = dbtable
         menupath = 'gnr.htablestores.%(dbtable)s' %attr
         attr['hasDownArrow'] = True
         dbselect_condition = attr.get('condition')
         dbselect_condition_kwargs = dictExtract(attr,'condition_')
         attr['condition'] = '$child_count=0' if not dbselect_condition else ' ( %s ) AND $child_count=0' %dbselect_condition
-        pane.dataRemote(menupath,self.ht_remoteTreeData,table=attr['dbtable'],
+        pane.dataRemote(menupath,self.ht_remoteTreeData,table=dbtable,
                         condition=dbselect_condition,
                         condition_kwargs=dbselect_condition_kwargs,cacheTime=30)
         dbselect.menu(storepath='%s._root_' %menupath,_class='smallmenu',modifiers='*',selected_pkey=attr['value'].replace('^',''))
--- a/resources/common/th/th_tree.py
+++ b/resources/common/th/th_tree.py
@@ -19,6 +19,8 @@ class HTableTree(BaseComponent):
     def ht_hdbselect(self,pane,caption_field=None,treeMode=None,folderSelectable=False,cacheTime=None,connectedMenu=None,tree_kwargs=None,**kwargs):
         dbselect = pane.dbselect(**kwargs)
         attr = dbselect.attributes
+        dbtable = attr.get('dbtable') or attr.pop('table', None)
+        attr['dbtable'] = dbtable
         dbselect_condition = attr.get('condition')
         dbselect_condition_kwargs = dictExtract(attr,'condition_',slice_prefix=False)
         dbselect_selected_kwargs = dictExtract(attr,'selected_',slice_prefix=False)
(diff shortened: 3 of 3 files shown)


genro (approved):
Approving. The diagnosis is right and the "intermittent" part is the useful half of it: `attr` is literally the caller's kwargs, the `table` -> `dbtable` normalization only exists client-side in `objectExtract`/`objectPop`, and a second `hdbSelect` sharing the `connectedMenu` skips the whole branch — so the bug's visibility depended on call order, which is exactly why it survived this long. Fixing `ht_hdbselect_old` at the same time is the right call even with no callers today: leaving one copy of a bug shape behind is how it comes back.

Popping `table` rather than leaving it alongside `dbtable` is the detail I would have missed. Node attributes are inherited down the struct and `table` is an actively consulted inherited key, so a stray `table` on the dbselect node would have leaked into the `treemenu` subtree built as its child.

Two notes, neither blocking:

**The `dbtable=None` path.** `attr['dbtable'] = dbtable` runs unconditionally, so a call with neither spelling now injects `dbtable=None` where it previously raised `KeyError` on the spot. You flag this yourself for @dgpaci and I agree with the direction — an explicit error naming both accepted spellings beats a `db.table(None)` failure several frames away. I would not widen this PR for it, but it is worth an issue so it does not rest only in a PR description.

**Cosmetic:** `projects/gnrcore/packages/test/webpages/inputfields/dbselect.py` still ends without a trailing newline. Your "flake8 clean" claim is accurate, since `gnrpy/.flake8` uses `select = F401,E9,F63,F7,F82` and W292 is not among the active checks — mentioning it only because the file arrived without one and leaves without one.

The test page addition is the right shape: it covers both spellings *and* the shared-`connectedMenu` branch, which is the one that used to skip the failure entirely.

PR #1090, opened 2026-08-18
fix(deps): make boto3 an optional [s3] extra instead of a core dependency

### LINKED ISSUE #1079: [BUG] boto3 declared without an upper bound conflicts with the aiobotocore that arrives via s3fs
## Problem Description

`gnrpy/pyproject.toml:28` declares `boto3` with no upper bound. Installing genropy alongside a package that pulls `s3fs` produces an unsatisfiable set: `s3fs` requires `aiobotocore`, which pins `botocore` below a ceiling, while the latest `boto3` requires a `botocore` floor above that ceiling. pip installs both and reports the conflict only afterwards.

**Expected:** installing genropy next to an `s3fs` consumer leaves a consistent environment.

**Actual:**

```
$ pip install -e gnrpy[pgsql]
ERROR: pip's dependency resolver does not currently take into account all the
packages that are installed. [...]
aiobotocore 3.8.0 requires botocore<1.43.47,>=1.43.3, but you have botocore
1.43.72 which is incompatible.
```

and, after pinning `botocore` down to satisfy `aiobotocore`:

```
boto3 1.43.56 requires botocore<1.44.0,>=1.43.56, but you have botocore 1.43.46.
```

### The chain

```
genro-storage[s3] -> s3fs 2026.6.0 -> aiobotocore 3.8.0 -> botocore<1.43.47,>=1.43.3
genropy           -> boto3 (unbounded) -> botocore>=<floor of the latest boto3>
```

## Reproduction Steps

1. Create a virtualenv with `s3fs` installed (here: via `genro-storage[s3]`).
2. `pip install -e gnrpy[pgsql]`.
3. Observe the `aiobotocore`/`botocore` conflict reported by pip.
4. Pin `botocore` to satisfy `aiobotocore` → the conflict moves to `boto3`.

Resolved here by pinning `boto3==1.43.46` and `botocore==1.43.46`, the newest pair that fits under the `aiobotocore` ceiling.

## Environment

- genropy `develop` @ `fef1037bb`, Python 3.13.2, macOS 15 (Darwin 25.0.0)
- s3fs 2026.6.0, aiobotocore 3.8.0, boto3 1.43.46, botocore 1.43.46

## Additional Material

The exact version numbers move with the install date — the same collision was reproduced independently on another machine with `aiobotocore` 2.24.2 and a `botocore<1.40.19` ceiling — but the mechanism is constant: an unbounded `boto3` will eventually outrun the `botocore` ceiling that `aiobotocore` holds. A bound on `boto3` in `gnrpy/pyproject.toml`, or dropping the dependency if it is only needed for an optional storage path, would make the resolution deterministic.

fporcari:
## Problem

`gnrpy/pyproject.toml` declared `"boto3"` bare in `[project].dependencies`. Installing genropy next to any `s3fs` consumer produces an unsatisfiable set, so the environment ends up broken.

Reproduced in a scratch venv:

```
$ pip install s3fs
   s3fs 2026.7.0, aiobotocore 3.9.0, botocore 1.43.56
   aiobotocore requires: botocore<1.43.57,>=1.43.3

$ pip install boto3          # what the core dependency did
   boto3 1.43.73 requires botocore<1.44.0,>=1.43.73
   Installing collected packages: botocore, s3transfer, boto3
   Uninstalling botocore-1.43.56 ...

ERROR: aiobotocore 3.9.0 requires botocore<1.43.57,>=1.43.3,
       but you have botocore 1.43.73 which is incompatible.
```

## Root cause

`s3fs` → `aiobotocore` holds a **ceiling** on botocore (`botocore<1.43.57`). The latest `boto3` holds a **floor** above it (`botocore>=1.43.73`). Nothing satisfies both, so pip resolves boto3's requirement and leaves `aiobotocore` pointing at a botocore it declares incompatible.

Nothing on the core import path or the request path needs boto3. Every boto3/botocore import lives in optional AWS integrations:

- `projects/gnrcore/packages/sys/resources/services/storage/aws_s3.py`
- `.../services/queue/aws_sqs.py`
- `.../services/translation/aws.py`
- `.../services/cloudmanager/awsmanager.py`
- `gnrpy/gnr/utils/awsmanager.py`
- `gnrpy/gnr/web/gnrwsgisite_proxy/gnrcloudservices.py`
- `projects/gnrcore/packages/docu/resources/tables/handbook/action/export_to_sphinx.py`

## Change

Packaging only — **no runtime code change**.

1. Removed `"boto3"` from `[project].dependencies`.
2. Added an `s3 = ["boto3"]` extra alongside `pgsql`/`mysql`, commented with its consumers.
3. Added `"boto3"` to the `developer` extra — the same reason `psycopg2-binary` and `pymysql` are there: `developer` is what CI installs (`.github/workflows/tests.yml` installs `.[developer]`), so the S3 tests keep running.
4. Added `pytest.importorskip('boto3')` to `gnrpy/tests/web/test_s3_temporary_filename.py` and `gnrpy/tests/web/test_s3_public_url.py`. Both exec `aws_s3.py` via `spec_from_file_location` at module scope, so without the guard a no-extras checkout gets a *collection error*, not a skip.
5. Added a `RELEASE_NOTES.rst` entry with the upgrade note.

**No code change is needed for the imports.** The service loader `_buildImplementations` (`gnr/lib/services/__init__.py:180-204`) imports every `.py` under `services/<type>/` regardless of configuration, but it is already guarded: `gnrImport` re-raises `ImportError` (`gnrlang.py:245-246`) and `_buildImplementations` catches it (`__init__.py:190-192`). That tolerance is already asserted by `gnrpy/tests/core/gnrservices_test.py:39`.

### Considered and rejected: an upper bound on boto3

Pinning `boto3<X` rots in the wrong direction. aiobotocore's botocore ceiling floats upward with every release, so any bound is simultaneously too high for older aiobotocore and too low for newer — it would need editing on someone else's release cadence. Removing boto3 from the default resolution set is what makes resolution deterministic.

There is also direct precedent: genropy already depends on `smart_open`, and smart_open itself declares boto3 only under `extra == "s3"`. The upstream library genropy uses for S3 I/O solved this exact problem the same way.

### Note on the release-notes heading

The entry is filed under an `Unreleased` heading rather than a `Release YY.MM.DD` stanza, because `gnr.VERSION` is still `26.07.31` and inventing a version here would desync from it. Rename it when the release is cut.

## Verification

Performed in this branch:

- **flake8** on both touched test files (repo selects `F401,E9,F63,F7,F82`) — zero errors.
- **boto3 present**: `pytest test_s3_public_url.py test_s3_temporary_filename.py tests/core/gnrservices_test.py -q` → **24 passed, 0 skipped**, so `importorskip` does not misfire when the extra is installed.
- **boto3 absent (skip path proven)**: rather than uninstalling boto3 from a shared environment, boto3/botocore were masked with a `sys.meta_path` blocker in the otherwise complete venv, isolating boto3 as the only variable:
  - the `develop` versions of both files → `2 errors during collection` (`ImportError: blocked for test: boto3` / `botocore`)
  - the patched files → `2 skipped`
  Also confirmed in a bare scratch venv with pytest only and no boto3 installed → `2 skipped`.
- **Resolver check (the definitive proof)**, in a fresh venv with `s3fs` installed first:
  - `pip install --dry-run <this branch>/gnrpy` → 75 packages, **boto3 / botocore / s3transfer NOT PULLED** — the existing botocore 1.43.56 is left intact and aiobotocore stays satisfied.
  - `pip install --dry-run genropy==26.7.31` (published, boto3 in core) into the same venv → pulls **boto3 1.43.73, botocore 1.43.73, s3transfer 0.19.2**, i.e. exactly the breakage above.
- **Full suite**: the pre-commit hook ran the whole suite including `tests/sql` → **1958 passed, 67 skipped**. A run excluding `tests/sql` gave 787 passed, 57 skipped.
- The pre-existing RST parse error in `RELEASE_NOTES.rst` was checked as unrelated: it sits at line 123 on `develop` and line 140 here, shifted by exactly the 17 lines added. The new stanza parses on its own. (`readme` is `README.md`, so this file is not part of the built metadata.)

Not verified: no live AWS calls were made — the S3 tests fake the boto3 client, as they already did.

## Open decision

@fporcari — this needs your call before the PR leaves draft.

Moving boto3 out of the default dependency set is a packaging change users feel on their next upgrade: an instance configured with an S3 storage, queue, translation or cloudmanager service and installed without `[s3]` will **not** raise. `getServiceFactory` falls back to `baseImplementation` when the key is absent (`gnr/lib/services/__init__.py:221-223`, asserted at `gnrpy/tests/core/gnrservices_test.py:62-63`), so an S3 bucket silently becomes a local folder with only an ERROR line in the log.

Options:

- **(a)** Ship the extra as-is and rely on the release note.
- **(b)** Ship the extra and additionally make a failed implementation import raise instead of falling back — correct, but a behaviour change touching every service type, so a separate issue.
- **(c)** Keep boto3 core and accept that installs alongside s3fs stay unsatisfiable.

## Separate cleanup, not in this PR

`gnrpy/gnr/web/gnrwsgisite_proxy/gnrcloudservices.py` imports boto3 at module level and has **zero importers repo-wide** — it is dead. Left untouched here to keep this PR to a packaging change; worth its own issue.

Fixes #1079


fporcari:
The `Unreleased` stanza in `gnrpy/RELEASE_NOTES.rst` added by this PR is under discussion with @cgabriel on #1089, which also covers #1057. The heading has never been used in that file, so these three PRs would be setting a convention rather than following one — asked in one place instead of three. Whatever is decided I will apply to all three. Nothing else on this PR depends on it.

cgabriel (changes requested):
I'm not approving this. boto3 is a core dependency, used in several services, and widely used in production. Making it optional would force also to install it anyway in all container images, which will need to be rebuilt.

The fact that a local developer enviroment is dirty or it doesnt' have updated deps in order to have a clean install tree it's not a valid reason to me to motivate this kind of change.

The framework doesn't pin a specific version for boto3 due to its maturity and no specific needs are disclosed.

My suggestion is to cleanup the environment by recreating it (better choice) or upgrading/downgrading where necessary, since this is a upstream issue (s3fs -> aiobotocore -> botocore dep path vs boto3 -> botocore path), finding the correct dependency path. Or supply a PR upstream, either way

fporcari:
Closing this PR because the proposed dependency-contract change does not have maintainer support. The review considers `boto3` a core, production dependency and explicitly rejects moving it to an optional extra; continuing this implementation would therefore not converge.

#1079 remains open to track the resolver conflict. A future solution can pursue an upstream fix or a coordinated dependency constraint without changing `boto3`'s core status. The branch and verification material are retained.

PR #1091, opened 2026-08-18
feat(deploy): nginx returns a retriable 503 with Retry-After during upstream downtime

### LINKED ISSUE #928: [TASK] Handle RPC errors during service restart: return retriable 503 from Nginx (no 500)
## Context

During application service restarts (`systemctl stop/start gnrsiterunner_<service>`), some RPC calls receive `500/502/504` errors.

RPC clients need a **temporary, retriable response** instead of a permanent application error.

## Objective

Ensure that during a restart or temporary upstream unavailability, RPC endpoints respond consistently with:

- HTTP `503 Service Unavailable`
- `Retry-After` header
- Stable JSON body (`code: MAINTENANCE`) to trigger client-side retry

## Proposed Solution (Nginx side)

Configure Nginx so that for the RPC path:

1. Upstream errors (`502`, `503`, `504`) are intercepted
2. They are transformed into a single `503` response
3. `Retry-After` and a standardized JSON payload are added

Example configuration:

```nginx
location /rpc/ {
    proxy_pass http://upstream_rpc;
    proxy_intercept_errors on;
    error_page 502 503 504 = @rpc_maintenance;
}

location @rpc_maintenance {
    add_header Retry-After 5 always;
    add_header Content-Type application/json always;
    return 503 '{\"code\":\"MAINTENANCE\",\"message\":\"Service temporarily unavailable, please retry\"}';
}
```

## Acceptance Criteria

- [ ] During a restart no RPC exposes a generic `500` to the client
- [ ] When upstream is unavailable, the response is uniformly `503` with `Retry-After`
- [ ] Maintenance error JSON payload is stable and documented
- [ ] Client confirmed with automatic retry on `503` (backoff + jitter)

## Operational Notes

- This solution covers the \"app down / restart in progress\" case well.
- An application-level middleware is not mandatory in the first phase; it can be added later only for the \"app up but not ready\" case.

fporcari:
## Motivation

During a service restart (`systemctl stop/start gnrsiterunner_<service>`) the gunicorn socket is gone and nginx answers with a generic 502 or 504. Nothing in that response tells a client that the condition is transient. Issue #928 asks for a retriable `503` with a `Retry-After` header and a stable JSON body instead.

### Correction to the issue's premise

The config proposed in the issue hangs off a `location /rpc/` block. **There is no `/rpc/` path in genropy.** RPCs are POSTs to the page's own URL:

- `gnrjs/gnr_d11/js/genro_rpc.js:687-689` — `pageIndexUrl()` returns `genro.currentUrl || document.location.pathname`, used as the XHR url at `:325`
- `getRpcUrlArgs()` (`:719-744`) puts `method`, `page_id` and `mode` in the **parameters**, not in the path
- `rpcUrl()` (`:746-751`) prefixes `document.location.pathname` via `genro.absoluteUrl`

A `location /rpc/` block would therefore match nothing. The interception goes on `location /`, which is the only block serving application traffic, and the named location is called `@gnr_maintenance` rather than `@rpc_maintenance` to reflect that.

## Change

One template, one file: `NGINX_TEMPLATE` in `gnrpy/gnr/app/gnrdeploy.py` — the single place nginx confs are generated (written by `write_nginx_conf()`, whose only caller is `gnrpy/gnr/app/cli/gnrdeploybuilder.py`). Seven added lines: the intercept inside `location /`, plus the named location that serves the body.

Decisions worth recording:

- **`default_type application/json;`, not an added `Content-Type` header.** With `return <code> '<body>'` nginx already sets a content type, so `add_header Content-Type ... always` would emit a duplicate header. `Retry-After` does need `always`, since the response is a 5xx.
- **`502 503 504` only — not 500.** Genropy returns application tracebacks as HTTP 200 with an error payload (commented explicitly at `genro_rpc.js:426-428`), so a real 500 is a worker crash. Folding it into `MAINTENANCE` would advertise genuine bugs as retriable.
- **`Retry-After 5` as a literal**, not a new `%(retry_after)s` parameter — this keeps `write_nginx_conf` and the `gnrdeploybuilder` argparse untouched.
- **`location /websocket` deliberately untouched**: a JSON 503 on a WebSocket upgrade is meaningless to the client.
- **`TRAEFIK_TEMPLATE` out of scope** — see the open decision below.

No client change is needed for this to be safe. `classifyHttpError()` (`genro_rpc.js:175-190`) already maps 502/503/504 to `'server_unavailable'`; `errorHandler()` (`:512-522`) routes that to `_onServerError()`, which locks the screen after `SERVER_UNAVAILABLE_THRESHOLD` and unlocks via `_onServerSuccess()` (`:594-597`). Shipping the nginx side alone changes nothing the client did not already handle.

## Verification

Adds `gnrpy/tests/app/cli/gnrdeploy_test.py` — the first test for `gnrdeploy.py`. It follows the `gnrdockerize_test.py` idiom (`GunicornDeployBuilder.__new__` with attributes set by hand), because the real `__init__` resolves paths, reads the siteconfig and calls `create_dirs()`. It renders `NGINX_TEMPLATE` exactly as `write_nginx_conf` does, but without writing a conf file into the working directory, and asserts on the real rendered output:

- the template renders without a `KeyError`/`ValueError` — this is the guard against a stray unescaped `%` in the template, which would raise at conf-generation time
- the intercept, the named location, `Retry-After` and `"code":"MAINTENANCE"` are all present
- `@gnr_maintenance` is defined exactly once, and the intercept sits in `location /`, after `/websocket` and not inside it
- `500` does not appear in the `error_page` line
- both the `supervisord_location=''` and the non-empty variants render, and braces balance in both

What was actually run:

- `flake8` on both touched files — zero errors
- the new file: 11 passed. Reverting the template change and re-running makes the 6 maintenance-specific tests fail while the 5 pre-existing-behaviour tests still pass, so the assertions have teeth
- the pre-commit hook ran the full suite, `tests/sql` included: **1969 passed, 67 skipped**

### Not exercised: nginx itself

**Nginx was not run.** There is no nginx binary in the environment this was prepared in, so the added directives are verified as generated text only, not as accepted configuration. Still to be checked manually:

1. `nginx -t` against a conf produced by `gnrdeploybuilder`, to confirm the syntax is accepted (`proxy_intercept_errors`, `error_page ... = @named`, and `return <code> '<body>'` inside a named location).
2. `curl -i https://<vhost>/anything` while `systemctl stop gnrsiterunner_<venv>` is in effect — expecting `503`, `Retry-After: 5`, and the JSON body.
3. A browser open on a page during a restart: it should show the existing "Server not responding..." lock and recover on its own via `_onServerSuccess` once the service is back.

## Open decision

@fporcari — Interception is currently on all of `location /`. That covers every RPC, but a user pressing F5 mid-restart receives raw JSON instead of a browser-friendly page. Options: (a) keep it on all of `location /` — one line, covers everything, ugly for document requests; (b) gate on `if ($http_x_requested_with = XMLHttpRequest)` (dojo sets this header on every xhr) and serve a static maintenance HTML page for document requests — needs a new template file and a `root`-relative path, growing the change; (c) leave document requests to the existing nginx default error page.

Also open: whether `TRAEFIK_TEMPLATE` should reach parity, which needs a Traefik `errors` middleware plus a service to serve the body — materially larger, arguably its own issue.

## Phase 2, not in this PR

The client currently *swallows* `server_unavailable` at `genro_rpc.js:512-522` without retrying, and nothing anywhere reads `Retry-After` — so after this PR the header is correct but unused. A real retry would re-issue the original `kw` (available via `rpc_register`/`register_call`, `genro_rpc.js:157-166`) with exponential backoff seeded from `ioArgs.xhr.getResponseHeader('Retry-After')` plus jitter and a max-attempts cap, keeping the existing lock-screen as the user-visible layer. Non-idempotent RPCs make blind retry unsafe, which is why it deserves its own issue rather than riding along here.

Fixes #928


genro (approved):
Approving. The production change is confined to the nginx template: `proxy_intercept_errors on` plus `error_page 502 503 504 = @gnr_maintenance`, and a named location returning a JSON 503 with `Retry-After: 5`.

Two details that make it the right shape rather than merely a nicer error page. `proxy_intercept_errors` is what makes the directive fire at all — without it nginx passes the upstream's own 502 straight through and `error_page` never runs, so the pair has to travel together and it does. And `add_header ... always` is the form that survives an error response; the bare form applies only to 2xx/3xx, so on the very status this exists for the header would have been dropped.

Returning `Retry-After` on a 503 rather than a bare 502 is the part clients can act on: it is the difference between "the service is broken" and "the service is restarting, come back in five seconds", and it is what lets a caller retry without hand-tuned backoff.

No Python behaviour changes — `gnrdeploy.py` only carries the template text.

cgabriel (commented):
I'm not supporting the proposed solution due to a scope issue, and I'll explain why.

While I agree on issue #928 requirement, the proposed solution is to me out of scope.

Basically we're introducing a function depencendy on the http endpoint used in front of the application inside the framework code, which can vary. Nginx is widely used, traefik too, but there are several others. Since it's http endpoint job decide how to handle 502/503/504 errors (for example, our traefik deployment serve a courtesy page for document load), other can do different things, and  maybe they can't be configured the way the framework requires (think of commercial/legacy WAF services). Maybe the user can't configure them (for example, cloudfront allows to customize error pages for http error code, but not headers).

So my point is that each tool should just accomplish his job, with not interference between each others. Having the framework to decide how and which your stack is and how is should behave is an overflow in other's duties.

I think this should be handled uniquely on the http error code, where the js client decide automously when/if to retry, at which rate, and what to display, when an http 50{2,3,4} is returned. Showing a courtesy page for document requests/reload is a matter that the endpoint decide how to handle (like we do in traefik, and how we do in nginx on several deployment).

So I'm supporiting option a) in the open decision (btw, it's still open, and discussing implementation details in a PR comment is not a best practice), where the client intercepts globally for 502 503 504 error and automously decide what to do next.

fporcari:
@cgabriel you are right that this is a scope question and not an implementation detail, and right that the PR thread is the wrong place for it. Moving it to #928, where the decision is actually open, and leaving this PR parked until that is settled.

Restating the two options so nobody has to reconstruct them from this thread:

**(a) client-side only.** The client intercepts 502/503/504 globally and decides on its own whether to retry, at what rate, and what to show. The framework says nothing about the HTTP endpoint; nginx, traefik, a WAF or CloudFront each keep their own error handling, including the courtesy page for document loads that our traefik deployment already serves.

**(b) what this PR does.** The framework ships an nginx configuration that turns upstream downtime into a 503 with `Retry-After` and a stable JSON body, so the transient condition is machine-readable at the protocol level.

The argument for (a) that I think is decisive rather than merely preferable: (b) only works where the endpoint can be configured the way the framework expects, and some cannot — CloudFront lets you customise error pages per status code but not headers, and a commercial or legacy WAF may not be ours to configure at all. A retry policy that exists only on some deployments is worse than one that exists everywhere, because the client cannot know which it is talking to.

What (a) still has to answer, and what I would like settled on #928 before writing code: without `Retry-After` the client has no server-supplied backoff, so the retry schedule becomes a client-side constant. Whether that is acceptable, and what it should be, is the real content of the decision.

Not merging this. Continuing on #928.

fporcari:
Withdrawing this PR, following @cgabriel's review: handling upstream 502/503/504 is the http endpoint's job, not the framework's, so the nginx template stays untouched. Option (a) — the js client intercepts 502/503/504 globally and decides on its own whether to retry, at what rate, and what to display — is the direction, and it continues on #928. Nothing from this branch is needed for that, so the branch is deleted.

PR #1084, opened 2026-08-17
fix(#1082): reconnect the site register proxies instead of failing on a dropped connection

fporcari:
Closes #1082

## Problem

An already-running WSGI worker keeps the site-register URI for its whole
lifetime. When the daemon restarts, the legacy site register normally binds a
new ephemeral port, so reconnecting the existing Pyro proxy still targets the
old port and every page request fails with `CommunicationError`.

## Root cause

There are two separate recovery cases:

- a dropped established connection while the endpoint stays the same;
- a daemon restart that publishes a different site-register URI.

Pyro can retry the first case after releasing the dead socket, but it never
re-resolves an URI and does not retry a connect-time refusal. The client read
the URI only once, either from `sitedaemon.xml` or from `gnrdaemon.getSite()`.

The previous hand-written retry in `SiteRegisterClient.__getattr__` did not
solve either case: it retried the same proxy, caught every exception, and
returned `None` after exhausting its attempts.

## Change

- `build_proxy()` enables Pyro's built-in retry for dropped connections on the
  same endpoint.
- `RefreshingProxy` handles the second layer. After a communication failure it
  asks the client for the currently published URI; it rebuilds the proxy and
  repeats the call once only when that URI changed. An unchanged URI preserves
  and propagates the original error.
- URI discovery supports both existing modes: rereading `sitedaemon.xml`, or
  calling `gnrdaemon.getSite()` for centrally managed site registers.
- Refresh is coordinated across workers in the process. Both the
  `SiteRegister` and derived `RemoteData` URI are updated, so `RemoteStoreBag`
  instances created before the restart also recover.
- The site-register cleanup configuration is applied again to the replacement
  proxy.
- Remote application errors now propagate instead of becoming a silent
  `None`.

## Verification

- `flake8` clean on both modified files.
- Targeted site-register suite: **11 passed**, including an integration test
  that opens a real Pyro connection to a daemon on port A, terminates that
  daemon process, and recovers through a second daemon on port B.
- Web suite: **229 passed, 59 skipped**.
- Full official suite (`cd gnrpy && pytest`): **2055 passed, 69 skipped**.
- Commit hook: **2024 passed, 69 skipped**.

PR #1093, opened 2026-08-18
chore(github): cover gnrpy and add a catch-all rule to CODEOWNERS

fporcari:
## Problem

`.github/CODEOWNERS` covered `gnrjs/`, `resources/`, `projects/` and `*.js`. Every other directory matched no rule at all — including `gnrpy/`, which holds the framework itself.

A pull request whose diff touches only uncovered paths gets no code owner. GitHub has nobody to request a review from, so it opens with an empty `reviewRequests`, nobody is notified, and it sits in `REVIEW_REQUIRED` with a green CI until somebody happens to look at the list.

## How it surfaced

Triaging the open pull requests, five of them had no reviewer at all, and the pattern was exact:

| PR | Paths touched | Reviewer requested on open |
|---|---|---|
| #1091, #1090, #1089, #1087, #1084 | only `gnrpy/**` | none |
| #1086 | `projects/gnrcore/...` | `@genro`, from CODEOWNERS |

Nothing distinguished those five except the directory they happen to live in.

There is a second half to it. `develop` is protected, so the missing rule only delays a merge there — `required_approving_review_count` is still 1. On a pull request stacked on a feature branch the protection does not apply at all: #1083 targets `fix/1082-siteregister-pyro-retry`, matches no CODEOWNERS rule, and reports `mergeStateStatus: CLEAN` with zero submitted reviews. Green, mergeable, never read.

## Change

- `gnrpy/` gets an explicit rule.
- A `*` catch-all is added so a directory created tomorrow is covered the day it appears, rather than reopening the same hole.
- `*.js` gains `@genro` as a second owner.
- Trailing whitespace removed from two existing lines.

The catch-all is placed **first** on purpose. In CODEOWNERS the *last* matching pattern takes precedence, so a `*` at the bottom would shadow every specific rule above it, `*.js @fporcari` included. First, it only applies where nothing else matches.

Two consequences of the catch-all are worth stating rather than leaving to be discovered:

**`.github/` becomes code-owned, CODEOWNERS itself included**, along with any workflow added later. Until now anyone with write access could rewrite the ownership rules without an owner reading the change. That is a hardening in its own right, not a side effect of covering `gnrpy/`.

**`*.js` needed a second owner.** It is the last line, so it wins over `gnrjs/` for JavaScript files, and its sole owner was also the most frequent author on them — nobody approves their own pull request, so a PR touching only `.js` files had no reachable code owner review. #1046 is the live case: two `.css` files under `gnrjs/` (owned by both of us) plus `genro_toast.js`, owned by me alone.

## Open decision

`develop` has `require_code_owner_reviews` enabled, so making every file code-owned also narrows who can satisfy the requirement. On the paths that previously had no rule, any collaborator with write access could approve; after this change it has to be `@fporcari` or `@genro`.

That is a policy change, not just a notification fix, and it is worth a deliberate answer:

- keep it as proposed, and accept `@fporcari`/`@genro` as the gate on everything;
- or point the catch-all at a GitHub team, so the set of people who can approve is managed by team membership instead of by editing this file on every change.

@genro @cgabriel — the second option is the one I would pick if the two-person gate looks too narrow in practice.

## Verification

No code change, nothing to test. The rule ordering was checked against GitHub's documented "last matching pattern wins" semantics; the coverage gap was verified by diffing the existing rules against the top-level directories on `develop` and against the touched paths of the open pull requests listed above.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,4 +1,10 @@
+# The last matching pattern wins, so the catch-all must stay first:
+# anything the rules below do not cover falls back to it instead of
+# opening with no code owner and no reviewer at all.
+*               @fporcari @genro
+
+gnrpy/          @fporcari @genro
 gnrjs/          @fporcari @genro
-resources/      @fporcari @genro 
-projects/       @fporcari @genro 
+resources/      @fporcari @genro
+projects/       @fporcari @genro
 *.js            @fporcari
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,4 +7,4 @@ gnrpy/          @fporcari @genro
 gnrjs/          @fporcari @genro
 resources/      @fporcari @genro
 projects/       @fporcari @genro
-*.js            @fporcari
+*.js            @fporcari @genro


fporcari:
@cgabriel two things here concern you directly, beyond reviewing the diff.

**What was actually broken.** Not a single pull request — the queue. Any pull request whose diff stays inside a directory no CODEOWNERS rule matches opens with an empty `reviewRequests`: GitHub has no code owner to ask, so nobody is notified and the pull request waits in `REVIEW_REQUIRED` with a green CI. `gnrpy/` had no rule, which is most of the framework.

**How it came up.** Going through the open pull requests, five had no reviewer at all — #1091, #1090, #1089, #1087, #1084 — and what they had in common was not the author or the age, it was that they touch only `gnrpy/**`. #1086, sitting in the same list and touching `projects/`, had `@genro` requested automatically. That was the whole difference.

The part worth flagging: `develop` is protected, so there the missing rule only delays things. On a branch stacked on another feature branch the protection does not apply — #1083 targets `fix/1082-siteregister-pyro-retry`, matches no rule, and reports `CLEAN` with zero submitted reviews. Mergeable and unread at the same time. I read it as ready to merge at first glance for exactly that reason, which is what sent me looking at this file.

**What changed for you meanwhile.** Those pull requests now have a reviewer assigned by hand rather than waiting for the file to land: #1084 and #1083 came to you since they are the siteregister/daemon pair next to #1069 and #1070 that you already have, and #1089 and #1087 went to @genro. #1091 and #1090 are still drafts, so they stay unassigned until they are marked ready — requesting a review on a draft does not notify anyone.

**The decision I would like your view on.** `develop` has `require_code_owner_reviews` on, so a `*` catch-all does not only fix the notification, it also narrows who can satisfy the requirement: on the previously uncovered paths any collaborator with write access could approve, and with this change it has to be @fporcari or @genro. You approve a fair share of this queue, so that lands on you. If it feels too narrow, say so and I will point the catch-all at a team instead, which fixes the notification without touching who is allowed to approve.

cgabriel:
> @cgabriel two things here concern you directly, beyond reviewing the diff.
> 
> **What was actually broken.** Not a single pull request — the queue. Any pull request whose diff stays inside a directory no CODEOWNERS rule matches opens with an empty `reviewRequests`: GitHub has no code owner to ask, so nobody is notified and the pull request waits in `REVIEW_REQUIRED` with a green CI. `gnrpy/` had no rule, which is most of the framework.
> 

I'm sorry, I don't get the "two things here concern you directly". There are 4 points, and none of them seems to concerns me directly - I review that PR that are assigned to me, leaving (as discussed when code owner were introduce) the merge to the author, or to the assignee (if any - this breaks the "everything must be assigned policy", but actually I was the only one trying to enforce the policy itself). 

Anyway, the PR is fine for me, although the CODE OWNERS configuration proposed in the PR can be simplified with a one liner, the catchall one on top.

cgabriel (dismissed):

genro (dismissed):
Approving. I verified the four claims that carry this change and they hold.

`gnrpy/` really was uncovered — the file matched `gnrjs/`, `resources/`, `projects/` and `*.js`, so the framework itself had no code owner. The catch-all belongs first, since the last matching pattern wins and a `*` at the bottom would shadow every rule above it. And develop's protection is as described: `require_code_owner_reviews: true`, `required_approving_review_count: 1`, `dismiss_stale_reviews: true`.

**On the open decision — it is less binding than you put it.** develop has `enforce_admins: false`, so an admin merges regardless of the code owner requirement, and the restriction only ever binds non-admin collaborators. The history shows it: #1051 (only `gnrjs/gnr_d11/js/genro_widgets.js`, your PR) merged on 2026-08-14 with a single approval from cgabriel, who is not a code owner of anything; #1008 merged the same way with an approval from `@genro`, who is not an owner of `.js` files. So the cost you flag is largely theoretical today, while the benefit — the reviewer actually being requested instead of the PR opening with an empty `reviewRequests` — is real and independent of the protection settings. That is the part worth having.

**One thing I would fold into this PR: `*.js @fporcari @genro`.** With the catch-all in place every path has two owners except `.js` files, which have one. `*.js` is the last line, so it wins over `gnrjs/` for those files, and the sole owner is also the most frequent author on them — nobody can approve their own PR. #1046 is the live case: it touches two `.css` files under `gnrjs/` (owners you and `@genro`) plus `genro_toast.js` (owner you alone), and GitHub requested `@genro`, who does not own that file. Admin bypass hides it today, but it is an asymmetry with no reason behind it and this is a one-word place to close it.

**Worth stating explicitly in the description:** the catch-all also makes `.github/` code-owned, CODEOWNERS itself included, plus any workflow added later. Until now anyone with write access could change the ownership rules without an owner reading them. That is a real hardening, not a side effect.

Separately, the triage that produced this PR surfaced something bigger that does not belong here: develop has `required_status_checks.contexts: []`, so no check is actually required and the green CI on all these pull requests reports rather than blocks. Opened as its own issue rather than widening this one.

fporcari:
> **One thing I would fold into this PR: `*.js @fporcari @genro`.** With the catch-all in place every path has two owners except `.js` files, which have one. `*.js` is the last line, so it wins over `gnrjs/` for those files, and the sole owner is also the most frequent author on them — nobody can approve their own PR.
> — [review of 19 Aug](https://github.com/genropy/genropy/pull/1093#pullrequestreview-4970374347)

Done on `c96b33b`. `*.js` now reads `@fporcari @genro`, so every path in the file has two owners and no rule can leave a PR with a single owner who is also its author.

#1046 is the case that proves it: it touches two `.css` files under `gnrjs/` plus `genro_toast.js`, and with the old rule the only owner of that third file was me.

Two notes on the rest of the review:

**The description now states the `.github/` hardening explicitly** — the catch-all makes CODEOWNERS itself code-owned, so the ownership rules can no longer be changed without an owner reading the change.

**On the open decision, I take your correction.** `enforce_admins: false` means an admin merges regardless, so the cost I flagged is smaller than I put it, and #1051 and #1008 are the evidence. The benefit that stands on its own is the one worth having: reviewers actually being requested instead of the PR opening with an empty `reviewRequests`. #1090 and #1091 were sitting in exactly that state — `gnrpy/` only, no owner, nobody notified — until they were assigned by hand today.

The push dismissed both approvals (`dismiss_stale_reviews` is on), so this needs a re-approval on the new head. Sorry for the round trip — the one-line change was worth it.

cgabriel (approved):
Approving, although the file could be reduced to a single line:

```
*               @fporcari @genro
```

genro (approved):
Keep it as proposed — option 1. The ordering reasoning is right (last matching pattern wins, so the catch-all has to be first or it would shadow `*.js @fporcari`), and covering `gnrpy/` explicitly plus a catch-all closes the hole permanently rather than one directory at a time.

On the open decision, I checked what option 2 would actually cost today, and it is not a drop-in:

- the only candidate team is `core-developers`, and its repo access is `genropy-core` (write), `genro-core` and `genro-api` (read). It has **no access to this repository at all**, so it cannot be a valid code owner here until someone grants it write access.
- its membership is sporcari, mbertoldi, genro, fporcari, dgpaci — **cgabriel is not in it**. Pointing the catch-all at that team as it stands would make our most active reviewer unable to satisfy a code-owner review, which is worse than the two-person gate it replaces.

So: land this as it is. If the two-person gate turns out too narrow in practice, the fix is a team properly scoped to this repo and including cgabriel, which is a separate change with its own decision.

The two consequences called out in the body are both worth having, especially `.github/` becoming code-owned: ownership rules editable without an owner reading the change was the sharper hole of the two.

PR #1099, opened 2026-08-19
fix(traceback): let the ep_error description wrap and make it selectable

### LINKED ISSUE #1092: [BUG] sys/ep_error truncate error description + unable to copy
### Problem Description

The "title" of the page for sys/ep_error webpage, which is the final error description, uses ellipsis which unfortunately truncate the error description. It should instead show (maybe on 2+ lines?) the error description anyway, which is the fundamentale information.

Here's a sample:

Image

The window is 3812 pixels wide, and still I can't read the error description.

Also, the error description test is not copyable, which would be helpful during a debug phase (paste the error to some agent, search engine etc)

### Reproduction Steps

1. raise an error which description is pretty long
2. have the backend raise the error
3. access the related sys/ep_error page.

_No response_

### Environment

genropy develop as of today.

### Additional Material

_No response_

fporcari:
Fixes #1092

## Problem

On `sys/ep_error` the error description is truncated with an ellipsis and cannot be selected or copied by hand. The reporter observed the truncation persisting even at a 3812px window width.

## Root cause

Two independent causes, neither of them related to the available width.

**1. The ellipsis is pure CSS.** `gnrjs/gnr_d11/css/gnrbase_css/14_gnr_traceback.css` had:

```css
.gnr-tb-toolbar-title {
    flex: 1;
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

`ep_error.py:25-28` is a 4-line wrapper that passes `description` as the `title` of the `tracebackViewer`, and the title is emitted as a single span (`gnr_tracebackviewer.js:113-114`). With `white-space: nowrap`, a multi-line exception message is collapsed onto one line, so **no** window is ever wide enough — the 3812px observation confirms the CSS cause rather than disproving it.

**2. Not selectable because of the global reset.** `00_gnr_resets.css:25-28` sets `user-select: none` on every `div` and `span`. The traceback viewer never opted out, so nothing inside it could be selected. The repo's own opt-out idiom is the `selectable` class (`00_gnr_resets.css:30-33`, used as `_class='selectable'` in e.g. `gnrwebstruct/base.py:774` and `genro_dlg.js:367`).

## Change

CSS/JS for the tracebackViewer title, plus one line of Python in the error form: `th_error.py:51` turns `fb.field('description')` into a `simpleTextArea` (same idiom as `user_agent` two lines below), so the description no longer truncates on the `sys/error` form either — the residual overflow reported in review.

`14_gnr_traceback.css`:
- `.gnr-tb-toolbar-title` now wraps (`white-space: pre-wrap`, `overflow-wrap: anywhere`, `min-width: 0`), capped at `max-height: 8em` with its own `overflow-y: auto`, and `cursor: text`.
- the toolbar switches from `align-items: center` to `flex-start`, and `.gnr-tb-btn` gets `flex-shrink: 0; white-space: nowrap;` so the buttons stay pinned top-right next to a tall title.

This is layout-safe because `sizeFrames()` (`gnr_tracebackviewer.js:176-185`) measures `toolbar.offsetHeight` at runtime rather than assuming a fixed toolbar height.

`gnr_tracebackviewer.js` — the `selectable` class added to three class strings: the toolbar title, the error message block (`.gnr-tb-errmsg`) and the frame body (`.gnr-tb-body`, which covers filename, code line and locals). The `<summary>` toggles are not descendants of `.gnr-tb-body`, so their deliberate `user-select: none` stays intact and clicking a frame header still toggles instead of selecting text.

## Verification

- `node --check gnrjs/gnr_d11/js/gnr_tracebackviewer.js` — clean.
- `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` — 801 passed, 57 skipped. The single failure, `gnrconfig_test.py::TestGnrConfig::test_getgenroroot`, is environmental (it asserts `getGenroRoot()` equals the checkout path, which fails in any worktree) and cannot be affected by a diff that touches no Python.
- **The visual result is NOT verified live.** `sys/ep_error` is login-gated and no non-auth instance mounts the error test pages in this environment, so the rendering was reasoned from the CSS, not observed.

Manual steps to verify the rendering:

1. Open `projects/gnrcore/packages/test/webpages/error_test/04_error_detail_page.py`, click **Generate Error**, then **Open Detail Page**.
2. The stock error there is a short `KeyError` — it will not show the original bug. To reproduce it, raise an exception with a long multi-line message (or a DB/SQL error, whose messages carry embedded newlines and a statement dump).
3. Check: the title wraps over multiple lines; single words and lines can be selected and copied; the three buttons stay pinned top-right; there is no double scrollbar between the capped title and `.gnr-tb-frames`.
4. Check the other host too — `sys/resources/tables/error/th_error.py:60-61` uses the same viewer but does **not** pass `title`, so the toolbar there should be unchanged (buttons only).

## Design decisions

Moved to the issue, where they belong: https://github.com/genropy/genropy/issues/1092#issuecomment-5359355615. The diff implements the option marked **taken** for each; if the discussion lands elsewhere, this PR follows.


genro (approved):
Approving. The diff is CSS plus three `selectable` classes on existing spans — no logic, no new element, nothing that can change what the viewer displays.

The CSS change is the interesting half and it is coherent: `align-items: center` → `flex-start` with `min-width: 0` on the flexed child is the standard pair for letting a flex item actually shrink and wrap, since a flex item's default `min-width: auto` is what prevents wrapping regardless of `overflow-wrap`. Replacing the `overflow: hidden` + `text-overflow: ellipsis` + `white-space: nowrap` trio with `pre-wrap` + `overflow-wrap: anywhere` + a `max-height` with `overflow-y: auto` swaps "truncate and lose the text" for "wrap, cap the height, scroll" — the same shape used for the toast message in #1046, and the right one here since an error description is precisely the text you must be able to read in full.

`cursor: text` alongside the `selectable` classes is a small thing worth having: without it the pointer keeps saying the region is not selectable while it now is.

cgabriel:
I'm putting the requested material in #1092 , although this PR has already been approved..

fporcari:
Thanks — the requested design material is now recorded in #1092, and genro's approval is current on this head. I explicitly requested your review when the PR was opened, and that request is still pending even though you merged `develop` and checked the resulting branch. Could you record the approval on the current head? Our workflow requires every requested review to be approved before I can merge.

Not verified: the live rendering remains outside my current authenticated environment; the current head is unchanged since your branch update and all checks are green.

cgabriel:
> Thanks — the requested design material is now recorded in #1092, and genro's approval is current on this head. I explicitly requested your review when the PR was opened, and that request is still pending even though you merged `develop` and checked the resulting branch. 

> Could you record the approval on the current head? Our workflow requires every requested review to be approved before I can merge.

I did the review, provided the material you asked on the issue, with other problems that needs to be fixed, so better if you put this PR in draft while they're fixed. This is why there is no approval or other. Opening the PR and then asking feedback on the issue leads to this.

> Not verified: the live rendering remains outside my current authenticated environment; the current head is unchanged since your branch update and all checks are green.

Check the issue, you asked for feedback on the issue so I've replied on the same media, just to avoid channel disgregation (which is happening, since you're asking for an approval after asking for feedback on the issue...)

fporcari:
> I did the review, provided the material you asked on the issue, with other problems that needs to be fixed, so better if you put this PR in draft while they're fixed.

Thanks for the live pass — the two screenshots in #1092 close the perimeter I could not check myself, so the wrap and the selection are confirmed working on a real instance.

On the remaining overflow: I read it as a different widget from the one this PR touches. #1099 changes only `14_gnr_traceback.css` and `gnr_tracebackviewer.js`, i.e. the `tracebackViewer` toolbar title. The field still cutting the text on the error page is `description` in the form, `projects/gnrcore/packages/sys/resources/tables/error/th_error.py:51`, a single-line `fb.field('description')` — no CSS in this PR can reach it, and `th_error.py:60-61` mounts the viewer without a `title` at all.

It is a one-line change and the idiom is two lines below it, on `user_agent`:

```python
fb.field('description', tag='simpleTextArea', height='5em', width=width)
```

I have added it here rather than opening a second issue, since it is the same symptom you reported. That push drops genro's approval (`dismiss_stale_reviews`); nothing he approved has moved, the CSS and JS hunks are byte-identical.

Not verified: I am inferring the widget from your screenshots, not seeing it — if the overflow you meant is elsewhere on that page, say where and I will follow it.

cgabriel (approved):
Verified at commit cc05165b77e7526e4dfeb035aae0ab99d4dc5dbd

All problems are addressed.

fporcari:
@genro re-requesting your review: your approval was dismissed automatically when @cgabriel merged `develop` into this branch, not by any change to the code you looked at.

The two hunks you approved are byte-identical on the current head — `git diff f18f3ce6 cc05165b` over `gnrjs/gnr_d11/css/gnrbase_css/14_gnr_traceback.css` and `gnrjs/gnr_d11/js/gnr_tracebackviewer.js` is empty. Everything since then is the `develop` merge.

CODEOWNERS asks for @fporcari and @genro on these paths, so cgabriel's approval on `cc05165b` does not satisfy the rule and the PR stays REVIEW_REQUIRED. All checks green.

genro (approved):
Re-approving after the push: the delta since 25 Aug is the single line in `th_error.py:51` (`description` as `simpleTextArea`, `height='5em'`), CSS and JS are byte-identical to the approved revision, and the two merge commits bring nothing else into the diff.

PR #1100, opened 2026-08-19
fix(web): escape dotted usernames in the remaining Bag-label keyed registries

### LINKED ISSUE #1062: Usernames containing dots break Bag-label keyed features (chat first, audit needed)

fporcari:
## Problem

Issue #1062: usernames containing a dot (or `@`) break any GenroPy feature that keys a Bag by the raw username, because Bag labels split on `.` in both the Python `Bag` and the JS `GnrBag`. PR #1063 fixed the chat half. This PR is the audit-driven fix for the remaining sites.

## Root cause

`Bag.__setitem__` / `__getitem__` treat `.` as the path separator, so `bag['amelia.martin']` is really `bag['amelia']['martin']`. Any code that does `bag.setItem(user, ...)` (or `bag[user] = ...`) with a raw, dotted username silently nests instead of creating a sibling. If a plain `amelia` user also exists, whichever write happens second clobbers the other's subtree, and lookups for the dotted user return `None`.

## Change

Audited sites from the issue, and what was done with each:

| Site | Status | Notes |
|---|---|---|
| `gnrpy/gnr/web/gnrwebpage_proxy/connection.py:205` | already fixed (#1063) | established the convention this PR extends |
| `gnrpy/gnr/web/daemon/siteregister_client.py` (`RegisterResolver.list_users`, `list_connections`, `list_pages`) | **fixed here** | escaped the label at all three `setItem` calls (`:452`, `:468`, `:482`); the `user=user` attribute was already carried, readers keep it |
| `gnrpy/gnr/web/gnrasync.py` (`SharedStatus`) | **fixed here** | added a `_userkey()` helper used at all 7 label call sites (`registerPage`, `unregisterPage`, `onPing`, `onUserEvent`); verified failure before the fix: `users['amelia'] = Bag(...)` then `users['amelia.martin']` returned `None`, so `onPing`/`onUserEvent` raised `TypeError` on `None['connections']`, and `popNode(page.user)` could leave an orphan node |
| `projects/gnrcore/packages/sys/webpages/asyncdashboard.py` | no change needed (consumer) | its `dataController` paths concatenate the grid's *selected label*, which is now the escaped label — label and path stay in agreement |
| `projects/gnrcore/packages/test15/webpages/tools/newregisterexplorer.py` (`_maintenance_get_items`) | **fixed here** | escaped the label exactly as `packages/sys/webpages/onering.py:266` already does; `item['_pkey']` keeps the real key |

Three conventions coexist in-tree for this class of problem, found during the audit:
- dot/`@`-escaping (the one used by #1063 and by this PR)
- numeric re-keying in `maintenance.py:356-359`, from the 2019 fix `4808b8084`
- raw labels (the bug this PR removes at the sites above)

## Verification

- `flake8 <touched files> --select=F401,E9,F63,F7,F82` — clean.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/web/gnrasync_test.py -q` — 7 passed, including a new test that registers a page for `amelia.martin`, asserts the `SharedStatus.users` tree has exactly one top-level node whose `user` value is the full dotted name, registers a second page for `amelia` and asserts neither clobbers the other, calls `onPing` for both without raising, then unregisters the dotted page and asserts no orphan node remains.
- `PYTHONPATH=<worktree>/gnrpy pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` — 803 passed, 57 skipped.
- `asyncdashboard.py` behavior verified by reading, not by running the page in a browser (no change made; left manual/visual verification to the reviewer if desired).
- `newregisterexplorer.py` and `siteregister_client.py` changes verified by code inspection against the working `onering.py`/`connection.py` precedents; not exercised through a live register-explorer page in this session.

## Open decision — @genro

The systemic alternative to boundary-escaping is a literal-label form for `setItem`/`getItem` (i.e. stop splitting on `.` when the path is escaped upstream). Python's `smartsplit` already honours backslash-escaped separators, but the JS twin (`gnrlang.js:224-237`) is a broken Rhino-era leftover — `new java.lang.Character(1)` throws a `ReferenceError` the moment a path contains `\.` in a browser. So the systemic option requires fixing JS `smartsplit` first. This PR deliberately keeps extending the boundary-escaping convention instead, to close the known holes without touching that shared code path. Worth deciding whether the systemic fix is wanted for the future or whether escaping-at-the-boundary is the accepted long-term pattern.

## Follow-up (out of scope here)

`siteregister.py:457` does an unanchored `re.match(fltval, value)`, and `filters='user:%s' % user'` is built at `gnrwebpage.py:2347` and `chat_component.py:176`. This means a filter for `user:anna` also matches `anna.bianchi`. Not fixed in this PR; flagging as a candidate follow-up.

Fixes #1062

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister_client.py
+++ b/gnrpy/gnr/web/daemon/siteregister_client.py
@@ -449,7 +449,8 @@ class RegisterResolver(BagResolver):
             item['info'] = Bag(item_user)
             item['data'] = data
             item.setItem('connections', RegisterResolver(user=user), cacheTime=3)
-            result.setItem(user, item, user=user)
+            label = user.replace('.', '_').replace('@', '_')
+            result.setItem(label, item, user=user)
         return result
 
     def list_connections(self, user):
@@ -459,7 +460,7 @@ class RegisterResolver(BagResolver):
             delta = (datetime.now() - connection['start_ts']).seconds
             user = connection['user'] or 'Anonymous'
             connection_name = connection['connection_name']
-            itemlabel = '%s (%i)' % (connection_name, delta)
+            itemlabel = ('%s (%i)' % (connection_name, delta)).replace('.', '_').replace('@', '_')
             item = Bag()
             data = connection.pop('data', None)
             item['info'] = Bag(connection)
@@ -474,7 +475,7 @@ class RegisterResolver(BagResolver):
         for page_id, page in list(pagesDict.items()):
             delta = (datetime.now() - page['start_ts']).seconds
             pagename = page['pagename'].replace('.py', '')
-            itemlabel = '%s (%i)' % (pagename, delta)
+            itemlabel = ('%s (%i)' % (pagename, delta)).replace('.', '_').replace('@', '_')
             item = Bag()
             data = page.pop('data', None)
             item['info'] = Bag(page)
--- a/gnrpy/gnr/web/gnrasync.py
+++ b/gnrpy/gnr/web/gnrasync.py
@@ -370,15 +370,23 @@ class SharedStatus(SharedObject):
     def sharedObjects(self):
         return self.data['sharedObjects']
 
+    @staticmethod
+    def _userkey(user):
+        """Bag labels split on '.', so a dotted username must be escaped
+        before it can key the users tree; the real username still rides in
+        the node's 'user' attribute/value."""
+        return user.replace('.', '_').replace('@', '_')
+
     def registerPage(self, page):
         page_item = page.page_item
         users = self.users
         page_id = page.page_id
-        if page.user not in users:
-            users[page.user] = Bag(dict(
+        userkey = self._userkey(page.user)
+        if userkey not in users:
+            users[userkey] = Bag(dict(
                 start_ts=page_item['start_ts'], user=page.user, connections=Bag(),
             ))
-        userbag = users[page.user]
+        userbag = users[userkey]
         connection_id = page.connection_id
         if connection_id not in userbag['connections']:
             userbag['connections'][connection_id] = Bag(dict(
@@ -397,7 +405,8 @@ class SharedStatus(SharedObject):
 
     def unregisterPage(self, page):
         users = self.users
-        userbag = users[page.user]
+        userkey = self._userkey(page.user)
+        userbag = users[userkey]
         connection_id = page.connection_id
         userconnections = userbag['connections']
         connection_pages = userconnections[connection_id]['pages']
@@ -405,13 +414,13 @@ class SharedStatus(SharedObject):
         if not connection_pages:
             userconnections.popNode(connection_id)
             if not userconnections:
-                users.popNode(page.user)
+                users.popNode(userkey)
 
     def onPing(self, page_id, lastEventAge):
         page = self.server.pages.get(page_id)
         if not page:
             return
-        userdata = self.users[page.user]
+        userdata = self.users[self._userkey(page.user)]
         conndata = userdata['connections'][page.connection_id]
         pagedata = conndata['pages'][page_id]
         pagedata['lastEventAge'] = lastEventAge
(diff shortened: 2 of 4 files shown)


genro (approved):
The audit is sound and the sites picked are the right ones. Approving, with an answer to the open decision and two notes.

Verified the consumers agree with the new labels, which is where this class of fix usually breaks:

- `asyncdashboard.py` takes the users grid's `grid_selectedLabel` into `main.selectedUser` (`:31`) and builds `'main.data.users.'+selectedUser+'.connections'` from it (`:22`, `:25`). Label and path are the same escaped string, and the `user` cell (`:51`) reads the node value, which keeps the dotted name. No change needed there, as you say.
- `newregisterexplorer.py` keeps the real key in `_pkey` (`:191`), which is what `getNodeByAttr('_pkey', current_page_id)` (`:73`) looks up. Escaping the label leaves that path intact.
- `RegisterResolver.list_users` still hands the real username to the child resolver (`siteregister_client.py:453`), so drilling into connections is unaffected.

**Open decision.** Keep escaping at the boundary. The literal-label form is the better shape in the abstract, but it is gated on JS `smartsplit` (`gnrlang.js:224-237`), which is a Rhino-era leftover that throws the moment a path contains `\.` — that is its own piece of work with its own blast radius across every escaped path in the client, not a prerequisite to close #1062. Escaping stays the accepted pattern until someone takes that on deliberately.

**Note 1, not blocking.** `user.replace('.','_').replace('@','_')` is lossy: a real `amelia_martin` and a dotted `amelia.martin` collide on one label, which is the same failure this PR removes, just rarer. The convention already has this property, so it is not yours to fix here — worth an issue so the choice is recorded rather than inherited.

**Note 2, pre-existing, spotted while reading.** `list_pages` sets `user=item['user']` (`siteregister_client.py:483`) where `item` only holds `info` and `data`, so that attribute is always `None`. Untouched by this PR; mentioning it so it is not lost.

`gnrpy/tests/web` on this head: 165 passed, 74 skipped.

PR #1106, opened 2026-08-19
feat(register): hold the parent -> children links on the register items

genro:
## Purpose

Make looking up a user's pages fast and solid. Today it is a full scan of the registry, and it is on the path of every request.

## Problem

A register item carries only its parent id: a page knows its connection, a connection knows its user. The other direction has no structure, so every walk down the hierarchy scans everything — `user_connection_keys` over all connections, `connection_page_keys` over all pages, `pages(user=...)` over all pages comparing the user field, and the cascade drops scanning again on their way down.

## Change

Two commits, deliberately separate so the plumbing and the behaviour can be read apart.

**`ab09e5f07` — the links.** The user item holds a `connections` set, the connection item a `pages` set. Both are written exclusively through `SiteRegister.updateRegisterLink(register, parent_id, link_name, child_id, add=None)`.

Two directions of one link would otherwise be maintained in as many places as there are lifecycle events — `new_connection`, `new_page`, `drop_page`, `drop_connection`, `drop_user` — and drift into phantom children or live children missing from their parent's set. One writer is one place to keep them in step.

Two details worth the review:

- The parent item is reached through `registerItems` rather than `get_item`, because `get_item` refreshes the item timestamp and a page being born must not keep an otherwise idle connection alive.
- The unlink happens **before** the child is dropped. `PageRegister.drop(cascade=True)` asks whether any page is left before deciding to drop the connection; unlinking afterwards would have that check still see the page it is dropping.

**`0ac990ae1` — the readers.** `connection_page_keys` and `user_connection_keys` read the sets; `connection_page_items`, `connection_pages`, `user_connection_items`, `user_connections` and `connections(user=...)` build on them; `pages()` starts from the links — by connection directly, by user through its connections — and still returns the whole registry when given neither.

Both key readers return a list, so `drop_pages` and `drop_connections` keep iterating what they were given while the underlying set shrinks under them.

Preserved deliberately: items are reached through `registerItems`, not `get_item`, so listing a connection's pages does not refresh their timestamps — only the `include_data` path goes through `get_item`, as before. And asking for a connection and a user that do not match still returns nothing, which the previous chain of two filters gave for free.

## What this lights up

`gnrwebpage_proxy/connection.py:72` already prefers `connection_item['pages']` and falls back to a register scan. Until now the key never existed, so the fallback always ran. `validate_page_id` is on the path of every request, so it stops scanning the registry.

Serpent round-trips a non-empty set as a set and an empty one as an empty tuple — both falsy when empty, both supporting `in` — so the membership test holds over Pyro either way. Checked, since this is a page-validation path.

## Not in scope

`PageRegister.pages()` filters with an unanchored `re.match(fltval, value)` while the same function compares with `==` two lines above, so `filters='user:anna'` also matches `anna.bianchi`. That is a separate matter being settled on #1100, where it was first flagged. Worth noting here only because once `pages(user=...)` is a link walk, the chat no longer needs the `filters` path at all.

## Tests

`gnrpy/tests/web/siteregister_links_test.py`, 31 tests.

The links: sets filling and emptying across all five lifecycle methods; the implicit user creation inside `new_connection` still linking; the cascade when the last page leaves and the sibling that prevents it; a page with no connection at all; linking not refreshing the parent timestamp; and both directions of every link compared against a rebuild from the children through a full lifecycle.

The readers: every converted method; the mismatched connection/user pair; unknown parents; timestamps untouched; dropping while iterating a walk; and each reader compared against a scan of the whole registry, so the conversion is shown to change how the answer is found and not what it is.

## Verification

- `pytest gnrpy/tests/web/siteregister_links_test.py`: 31 passed
- `pytest gnrpy/tests`: 2042 passed, with 2 pre-existing failures in `flatfiles_test.py` and 8 pre-existing errors in `tests/sql` from missing pg/mysql drivers, all present on a clean develop
- Not verified: no live run against a real daemon with several connections and pages. The classic register is exercised only through these unit tests, and the ASGI provider replaces this register entirely, so it is unaffected.

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -367,16 +367,24 @@ class ConnectionRegister(BaseRegister):
                 self.siteregister.drop_user(user)
 
     def user_connection_keys(self, user):
-        return [k for k, v in list(self.items()) if v['user'] == user]
+        # from the user item's link set; a list, so callers may drop while iterating
+        user_item = self.siteregister.user_register.registerItems.get(user)
+        if not user_item:
+            return []
+        return list(user_item['connections'])
 
     def user_connection_items(self, user):
-        return [(k, v) for k, v in list(self.items()) if v['user'] == user]
+        return [(k, self.registerItems[k]) for k in self.user_connection_keys(user)]
+
+    def user_connections(self, user):
+        return [self.registerItems[k] for k in self.user_connection_keys(user)]
 
     def connections(self, user=None, include_data=None):
-        connections = self.values(include_data=include_data)
-        if user:
-            connections = [v for v in connections if v['user'] == user]
-        return connections
+        if not user:
+            return self.values(include_data=include_data)
+        if include_data:
+            return [self.get_item(k, include_data=True) for k in self.user_connection_keys(user)]
+        return self.user_connections(user)
 
 
 class PageRegister(BaseRegister):
@@ -483,16 +491,35 @@ class PageRegister(BaseRegister):
         return [self.registerItems[k] for k in self.subscribed_table_page_keys(table)]
 
     def connection_page_keys(self, connection_id):
-        return [k for k, v in list(self.items()) if v['connection_id'] == connection_id]
+        # from the connection item's link set; a list, so callers may drop while iterating
+        connection_item = self.siteregister.connection_register.registerItems.get(connection_id)
+        if not connection_item:
+            return []
+        return list(connection_item['pages'])
 
     def connection_page_items(self, connection_id):
-        return [(k, v) for k, v in list(self.items()) if v['connection_id'] == connection_id]
+        return [(k, self.registerItems[k]) for k in self.connection_page_keys(connection_id)]
+
+    def connection_pages(self, connection_id):
+        return [self.registerItems[k] for k in self.connection_page_keys(connection_id)]
 
     def pages(self, connection_id=None, user=None, include_data=None, filters=None):
-        pages = self.values(include_data=include_data)
+        # walk the link sets rather than the whole registry: a connection knows its
+        # pages, a user knows its connections
+        page_ids = None
         if connection_id:
-            pages = [v for v in pages if v['connection_id'] == connection_id]
-        if user:
+            page_ids = self.connection_page_keys(connection_id)
+        elif user:
+            page_ids = [page_id
+                        for cid in self.siteregister.user_connection_keys(user)
+                        for page_id in self.connection_page_keys(cid)]
+        if page_ids is None:
+            pages = self.values(include_data=include_data)
+        elif include_data:
+            pages = [self.get_item(page_id, include_data=True) for page_id in page_ids]
+        else:
+            pages = [self.registerItems[page_id] for page_id in page_ids]
+        if connection_id and user:
             pages = [v for v in pages if v['user'] == user]
         if not filters or filters == '*':
             return pages
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -325,6 +325,7 @@ class UserRegister(BaseRegister):
             user_name=user_name,
             user_tags=user_tags,
             avatar_extra=avatar_extra,
+            connections=set(),
             register_name='user')
         self.addRegisterItem(register_item)
         return register_item
@@ -351,6 +352,7 @@ class ConnectionRegister(BaseRegister):
             user_agent=user_agent,
             electron_static=electron_static,
             browser_name=browser_name,
+            pages=set(),
             register_name='connection')
         self.addRegisterItem(register_item)
         return register_item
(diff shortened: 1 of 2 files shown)


fporcari (changes requested):
The design is right and the two commits read cleanly apart. Three choices in particular I checked and would keep as they are:

- **One writer.** `updateRegisterLink` as the sole author of both sets is what makes this maintainable at all — the alternative is the same link maintained in as many places as there are lifecycle events.
- **`registerItems` rather than `get_item`.** Confirmed necessary: `get_item` refreshes the timestamp, so linking a newborn page would keep an idle connection alive past `connection_max_age`.
- **Unlink before the drop.** I traced both cascade paths and the ordering is correct. `PageRegister.drop` (`siteregister.py:465`) reads `connection_page_keys` to decide whether the connection goes, and `ConnectionRegister.drop` (`siteregister.py:360`) reads `user_connection_keys` for the user — both run after `SiteRegister.drop_page`/`drop_connection` have already unlinked, so neither sees the child it is dropping. `new_connection` also creates the user before linking, and `UserRegister.drop` iterates a list snapshot. All good.

## `change_connection_user` is the sixth lifecycle method, and it does not link

`siteregister.py:777` moves a connection from one user to another and never calls `updateRegisterLink`. It is not a rare path: `gnrwebpage_proxy/connection.py:167` reaches it from `change_user`, so every login and every logout back to guest goes through it.

```python
connection_item['user'] = user          # siteregister.py:785
...
for p in self.pages(connection_id=connection_id):
    p['user'] = user
if not self.connection_register.connections(olduser):
    self.drop_user(olduser)
```

Three consequences, all of them regressions against develop, because the scan-and-compare this PR replaces got each one right:

1. **The old user keeps a connection that is no longer theirs.** `olduser['connections']` still holds `connection_id`, so `connections(user=olduser)` returns a live connection now owned by somebody else, and `pages(user=olduser)` returns that connection's pages — whose `user` field the loop above has just rewritten to the new user. This reaches `siteregister_client.py:456` (`connections(user=user, include_data=True)`) and `siteregister_client.py:433` (`list_connections(user=self.user)`), which are user-facing listings. Note that `pages(user=...)` no longer post-filters on `v['user']`: the new `if connection_id and user:` guard only fires when both are given, so nothing catches the mismatch downstream.

2. **The new user gets nothing.** Their `connections` set is never written, so `user_connection_keys(newuser)` is empty and both readers return nothing for the user who actually owns the connection.

3. **`drop_user(olduser)` stops firing** — `siteregister.py:793`. The guard now reads the stale set, which still contains the connection, so it is truthy and the guest user item is never dropped. Guest users accumulate in the register for the lifetime of the daemon.

The fix is two calls, and the `drop_user` check has to come after them:

```python
self.updateRegisterLink(self.user_register, olduser, 'connections', connection_id)
self.updateRegisterLink(self.user_register, user, 'connections', connection_id, add=True)
```

## The invariant test already knows how to catch this

`siteregister_links_test.py:232` asserts `_links_from_parents(reg) == _links_from_children(reg)` after each lifecycle step — exactly the invariant this breaks. One line inside that sequence fails today:

```python
reg.change_connection_user('conn-1', user='bruno')
assert _links_from_parents(reg) == _links_from_children(reg)
```

Worth adding the guest-to-authenticated case too, since it is the shape that actually runs in production: connection created under the guest name, then `change_connection_user` to a real user, then assert the guest user item is gone and the new user holds the connection.

## Minor

`validate_page_id` (`gnrwebpage_proxy/connection.py:71`) keeps `connection_item.get('pages') or <register scan>`. Now that the key always exists the fallback only triggers for an empty set, where the scan returns `[]` as well — so it is harmless, but it is also dead weight except when the connection item itself is missing. Not for this PR; just noting it since the PR body is what makes that fast path live.

Everything else I verified holds. Blocking only on `change_connection_user`.

genro:
@fporcari addressed in e32fbfa58.

`change_connection_user` now unlinks the connection from the old user and links it to the new one through `updateRegisterLink`, before the `drop_user` guard — which reads the old user's link set — runs. All three regressions are covered by tests that fail on the previous commit:

- the lifecycle invariant sequence in `siteregister_links_test.py` now walks through `change_connection_user`, exactly the line you pointed at;
- `test_change_connection_user_moves_the_link_between_users` is the guest-to-authenticated shape: the new user holds the connection (readers included), the guest item is gone;
- `test_change_connection_user_keeps_the_old_users_other_connections` pins the guard staying quiet when the old user still owns something.

The `validate_page_id` dead fallback is left alone, as you suggested.

fporcari (approved):
> addressed in e32fbfa58. `change_connection_user` now unlinks the connection from the old user and links it to the new one through `updateRegisterLink`, before the `drop_user` guard — which reads the old user's link set — runs.

Confirmed on `e32fbfa58`, approving.

- The two calls sit exactly where the guard needs them (`siteregister.py:791-794`), and the seam I had not called out holds too: the new user's item is guaranteed to exist before the link is written, because `change_connection_user` creates it via `new_user(...)` when missing (`:782-785`), and `updateRegisterLink` degrades to a no-op rather than an error if a parent item is ever absent (`:660-664`).
- Ran `siteregister_links_test.py` on the branch: 33 passed. Re-ran with `siteregister.py` reverted to the parent commit: exactly the three tests covering this fix fail (`test_both_directions_agree_through_a_full_lifecycle`, `test_change_connection_user_moves_the_link_between_users`, `test_change_connection_user_keeps_the_old_users_other_connections`) — the "fail on the previous commit" claim is verified, and the guest-to-authenticated case plus the old-user-keeps-other-connections guard are the two shapes worth pinning.

Everything from the previous review stands as verified; nothing else blocks.

PR #1111, opened 2026-08-20
fix(localization): stop symbol-only captions collapsing to a single lockey

### LINKED ISSUE #1109: [BUG] flatten() collapses symbol-only strings to a single localization key, so all symbolic captions share one translation
### Problem Description

`AppLocalizer.getTranslation` derives the localization key from the text via `flatten()`, which is `re.compile(r'\W+').sub('_', text.lower())` (`gnrpy/gnr/core/gnrstring.py:43,450`).

For any string made only of non-alphanumeric characters, `flatten()` collapses the whole string to a single `_`. So all of these produce the very same key:

```
'='  -> '_'      '>'  -> '_'      '>=' -> '_'      '%' -> '_'
'!=' -> '_'      '<'  -> '_'      '<=' -> '_'      '#' -> '_'
```

Every symbol-only caption in an instance therefore shares the lockey `<lang>__` and resolves to whichever entry happens to occupy that slot in the merged `localizationDict` — which is built from the genropy root plus every package folder of the instance (`gnrpy/gnr/app/gnrlocalization.py:80-89`). One junk entry collected in any package poisons the symbol-only captions of the entire application.

Observed behavior: a `filteringselect` built from a column declared as

```python
tbl.column('condition_op', name_long='!!Condition op',
           values='eq:=,ne:!=,gt:>,ls:<,ge:>=,le:<=')
```

renders all six options with the identical caption `%`. `gnr.widgets.BaseCombo.storeFromValues` passes each caption through `_T()` (`gnrjs/gnr_d11/js/genro_widgets.js:4392`), `_T` sends `!!=`, `!!>`, `!!<`, … to `getRemoteTranslation`, they all hit lockey `en__`, and the entry found there has `it=""` so the NOLANG branch falls back to its `base` value (`gnrlocalization.py:122-132`).

Expected behavior: symbol-only captions are not translatable text and must be returned unchanged, never keyed into `localizationDict` nor looked up in it.

Two real examples of the poisoning entry currently in the wild, both auto-collected by the localization scanner from a bare `%`/`#` literal:

```xml
<en_ _tag="en__" base="%" it=""></en_>   <!-- gnrsupport/packages/ttx -->
<en_ _tag="en__" base="#" it=""></en_>   <!-- gbs/packages/gbs -->
```

Note the browser also caches the wrong result in `localStorage` under `localsdict_<lang>` (`gnrjs/gnr_d11/js/gnrlang.js:68-81`), so removing the offending XML entry is not enough to make the symptom disappear on an already-used client.

### Reproduction Steps

1. In a sandbox instance, add to any package's `localization.xml` an entry with the empty flattened key and a symbol as base:
   ```xml
   <en_ _tag="en__" base="%" it=""></en_>
   ```
2. Declare a column with symbol-only captions in its `values`:
   ```python
   tbl.column('test_op', name_long='!!Op', values='eq:=,ne:!=,gt:>,ls:<,ge:>=,le:<=')
   ```
3. Run `gnr dev mkthresource` for the table and open the form with an `it` locale, so the field renders as a `filteringselect`.
4. Open the dropdown.

Observed result: six entries, all captioned `%`.
Expected result: six entries captioned `=`, `!=`, `>`, `<`, `>=`, `<=`.

Step 1 only makes the collision deterministic and easy to see; the underlying key collapse happens with or without it — `flatten('=') == flatten('<=') == '_'` is reproducible on its own:

```
python -c "from gnr.core.gnrstring import flatten; print([flatten(s) for s in ['=','!=','>','<','>=','<=','%']])"
['_', '_', '_', '_', '_', '_', '_']
```

- Modules involved: `gnr/app/gnrlocalization.py` (`AppLocalizer.getTranslation`, `translatecb`, `buildLocalizationDict`), `gnr/core/gnrstring.py` (`flatten`, `FLATTENER`), `gnrjs/gnr_d11/js/genro_widgets.js` (`gnr.widgets.BaseCombo.storeFromValues`), `gnrjs/gnr_d11/js/gnrlang.js` (`_T`).
- Condition that surfaces it: an instance loading two or more packages where any one of them has a symbol-only string collected into its `localization.xml`.
- Client-side `localStorage` caching of `localsdict_<lang>` can mask a fix; clear it when verifying.

### Environment

genropy develop, Python 3.12, macOS 15 (Darwin 25.1.0), Chrome — locale `it`

### Additional Material

Suggested fix, in `AppLocalizer.getTranslation`: when the text carries no explicit `{key}` and `flatten(loctext)` contains no alphanumeric characters, return `loctext` unchanged instead of computing a lockey and looking it up. The same guard belongs on the collection side, so symbol-only literals are never written into `localization.xml` in the first place.

A narrower alternative, translating value captions only when explicitly marked (`_T(caption, true)` in `storeFromValues`), would fix this rendering path but silently drop the translation of the many unmarked `values` captions in existing code (e.g. `values='A:Answer,I:Inner value'`), so it is not proposed as the primary fix.

fporcari:
## Problem

Some captions consisting only of symbols (e.g. `=`, `>=`, `%`, `#`) are translated to the wrong text at runtime, picking up whatever unrelated translation happens to occupy an internal fallback slot instead of being left untranslated (or translated correctly).

## Root cause

`flatten()` (`gnrpy/gnr/core/gnrstring.py:43,450-451`, `FLATTENER = re.compile(r'\W+')`) collapses **any** symbol-only string to a single `_`:

```
'=' -> '_'   '>' -> '_'   '>=' -> '_'   '%' -> '_'   '!=' -> '_'
'<' -> '_'   '<=' -> '_'  '#'  -> '_'   '   ' -> '_'   '' -> ''
```

`AppLocalizer.getTranslation` (`gnrpy/gnr/app/gnrlocalization.py:119-121`) uses `flatten(loctext)` as the fallback lockey whenever no explicit key is supplied:

```python
if not lockey:
    lockey = flatten(loctext)
lockey = '%s_%s' % (loclang, lockey)
```

So every symbol-only caption produces the identical lockey `<lang>__`. `self.localizationDict` (built in `buildLocalizationDict`/`updateLocalizationDict`, `:81-90`, `:196-216`) is a single flat dict merged across the genropy root and **every** application package, so whichever package happens to register an `en__` entry first wins that slot for the whole application — every other symbol-only caption then silently resolves to that unrelated entry's translation.

Client path for context: `_T` (`gnrjs/gnr_d11/js/gnrlang.js:57-82`) → `getRemoteTranslation` (`gnrpy/gnr/web/gnrwebpage.py:1188-1189`) → `AppLocalizer.getTranslation`. A typical caller is `storeFromValues` at `gnrjs/gnr_d11/js/genro_widgets.js:4390` (`xval['caption'] = _T(val[1]);`).

**Two corrections to the issue's premise:**

- **`m.get('key')` in `translatecb` is dead code.** The `TRANSLATION` regexp (`gnrlocalization.py:35`) used at *lookup* time only names groups `{lang, value, lang_emb, value_emb}` — its `{key}` groups are unnamed — while `LOCREGEXP` (`:34`), used at *collection* time, does name them. So `!!{cond_eq}=` gets collected under `en_cond_eq` but is looked up under `en__`: an explicit `{key}` never actually resolves. This is a separate, pre-existing defect with zero in-tree users; not fixed here, only stated.
- **Genropy itself does not currently poison `en__`.** Running the real `LOCREGEXP` over `gnrpy/gnr`, `gnrjs`, `resources`, `projects` finds no symbol-only literals, and no in-tree `localization.xml` has an `en__` entry. The defect is latent in the framework and armed by application packages (the issue names `gnrsupport/ttx` and `gbs`). Whitespace-only text (`'   '`) is the same bug class and is covered by the same guard.

## Change

One file, two symmetric guards, in `gnrpy/gnr/app/gnrlocalization.py`:

1. In `getTranslation`'s `translatecb`: if the flattened fallback lockey is only `_` characters, return the source text unchanged instead of looking it up — the same "return text unchanged" contract this callback already uses for `NOKEY` (`:134-135`) and for empty text (`:240-241` before this change).
2. In `_updateModuleLocalization`'s `addToLocalizationBag`: the symmetric collection-side guard, so the scanner stops writing junk `en__` entries into `localization.xml` on every regeneration. Both sides belong in the same commit: without the second guard, the first symbol-only literal encountered in a file would keep silently swallowing every other symbol-only literal in that file via the existing `if not lockey in moduleLocBag` de-dup.

`flatten()` itself is intentionally **not** changed — it is also used to generate database column names from captions at four in-tree call sites (`projects/gnrcore/packages/lgdb/.../get_model_folder.py:94`, `projects/gnrcore/packages/sys/.../model_editor.py:141`, plus `gnrlocalization.py:41,206`), where this collapsing behaviour is exactly what's wanted.

## Verification

- Added `test_symbol_only_captions_not_translated` to `gnrpy/tests/app/gnrlocalization_test.py`: injects a junk `localizationDict['en__'] = {'base': '%', 'it': ''}` plus a real entry, and asserts `=`, `>`, `>=`, `<`, `<=`, `!=`, `#`, `%` each translate to themselves, a real caption still translates normally, and a caption with digits (`100%`) still reaches the ordinary lookup path (not short-circuited).
  - **Confirmed this test fails on unpatched `develop`**: all eight symbol captions resolve to `'%'` (the junk entry's translation) instead of themselves. Verified via `git diff > patch; git checkout -- gnrlocalization.py; run test (failed); git apply patch` (test file itself was kept in place).
- Added `test_flatten` to `gnrpy/tests/core/gnrstring_test.py`: `flatten` had zero prior coverage; this documents the collapse behaviour (`flatten('>=') == '_'`, `flatten('') == ''`, `flatten('Condition op') == 'condition_op'`, `flatten('100%') == '100_'`) so nobody "fixes" `FLATTENER` without noticing this PR depends on its exact contract.
- `flake8` on all touched files: 0 errors. CI selection (`--select=F401,E9,F63,F7,F82`) over the whole tree: 0 errors.
- Full suite: `PYTHONPATH=$PWD python -m pytest tests/ -q --ignore=tests/sql` → 842 passed, 59 skipped. The pre-commit hook additionally ran the full suite including `tests/sql`: 2013 passed, 69 skipped, 0 failed.
- **Stayed manual** (not automatable in this environment): the browser leg. To verify visually — build a `filteringselect` on a column like `tbl.column('condition_op', values='eq:=,ne:!=,gt:>,ls:<,ge:>=,le:<=')` in an instance whose packages carry an `en__` entry, render the six distinct captions, **after clearing `localStorage['localsdict_it']`**, since `_T`'s client-side cache never expires or invalidates (`gnrjs/gnr_d11/js/gnrlang.js:63-77`).

Also worth noting (not a code change): this guard does not retroactively clean `en__` entries already written into downstream apps' `localization.xml` — those nodes need to be deleted and regenerated — and browser clients need their `localStorage` cleared to pick up the fix.

## Related issue

Fixes #1109

## Design decisions

The four decisions this section used to carry are now on the issue, where they belong: https://github.com/genropy/genropy/issues/1109#issuecomment-5358987630. The diff implements the option marked **taken** for each; if the discussion there lands elsewhere, this PR follows.

## Verification

- Added `test_symbol_only_captions_not_translated` to `gnrpy/tests/app/gnrlocalization_test.py`: injects a junk `localizationDict['en__'] = {'base': '%', 'it': ''}` plus a real entry, and asserts `=`, `>`, `>=`, `<`, `<=`, `!=`, `#`, `%` each translate to themselves, a real caption still translates normally, and a caption with digits (`100%`) still reaches the ordinary lookup path (not short-circuited).
  - **Confirmed this test fails on unpatched `develop`**: all eight symbol captions resolve to `'%'` (the junk entry's translation) instead of themselves. Verified via `git diff > patch; git checkout -- gnrlocalization.py; run test (failed); git apply patch` (test file itself was kept in place).
- Added `test_flatten` to `gnrpy/tests/core/gnrstring_test.py`: `flatten` had zero prior coverage; this documents the collapse behaviour (`flatten('>=') == '_'`, `flatten('') == ''`, `flatten('Condition op') == 'condition_op'`, `flatten('100%') == '100_'`) so nobody "fixes" `FLATTENER` without noticing this PR depends on its exact contract.
- `flake8` on all touched files: 0 errors. CI selection (`--select=F401,E9,F63,F7,F82`) over the whole tree: 0 errors.
- Full suite: `PYTHONPATH=$PWD python -m pytest tests/ -q --ignore=tests/sql` → 842 passed, 59 skipped. The pre-commit hook additionally ran the full suite including `tests/sql`: 2013 passed, 69 skipped, 0 failed.
- **Stayed manual** (not automatable in this environment): the browser leg. To verify visually — build a `filteringselect` on a column like `tbl.column('condition_op', values='eq:=,ne:!=,gt:>,ls:<,ge:>=,le:<=')` in an instance whose packages carry an `en__` entry, render the six distinct captions, **after clearing `localStorage['localsdict_it']`**, since `_T`'s client-side cache never expires or invalidates (`gnrjs/gnr_d11/js/gnrlang.js:63-77`).

Also worth noting (not a code change): this guard does not retroactively clean `en__` entries already written into downstream apps' `localization.xml` — those nodes need to be deleted and regenerated — and browser clients need their `localStorage` cleared to pick up the fix.

## Related issue

Fixes #1109

## Open decisions

1. **Escape hatch for genuinely translatable symbols** (e.g. `'$'` → `'€'`, `','` → `'.'`): (a) ship the guard as written — such symbols become untranslatable, and the `{key}` workaround doesn't help since `TRANSLATION`'s key groups are unnamed; (b) also name `key`/`key_emb` in `TRANSLATION` — restores the escape hatch but changes resolution for any existing `!!{key}text` in downstream apps that currently falls through to `NOKEY`. **Taken: (a).** Recommend (b) as its own issue if needed. @dgpaci what's your take?
2. **`status` for guarded text**: `'OK'` (taken) vs a new `'NOTEXT'`. Every existing consumer only compares `!= 'OK'` (`gnrlang.js:73`), and that branch is dead anyway since the next line overwrites it. @dgpaci any preference?
3. **Cache busting for `localsdict_<lang>`**: documented in this PR (taken, following the precedent in commit `81f94ce23`) vs adding a version stamp to the storage key in `gnrlang.js:63` — a scope increase into the JS client, worth a follow-up issue if desired. @dgpaci
4. A third guard on the `GnrLocString` path (`gnrlocalization.py:41`/`:206`) would close a theoretical hole a legacy `_key="%"` file could still open. Not required by the observed defect. **Taken: leave it.** @dgpaci agree?


cgabriel:
> ## Open decisions
> 1. **Escape hatch for genuinely translatable symbols** (e.g. `'$'` → `'€'`, `','` → `'.'`): (a) ship the guard as written — such symbols become untranslatable, and the `{key}` workaround doesn't help since `TRANSLATION`'s key groups are unnamed; (b) also name `key`/`key_emb` in `TRANSLATION` — restores the escape hatch but changes resolution for any existing `!!{key}text` in downstream apps that currently falls through to `NOKEY`. **Taken: (a).** Recommend (b) as its own issue if needed. @dgpaci what's your take?
> 2. **`status` for guarded text**: `'OK'` (taken) vs a new `'NOTEXT'`. Every existing consumer only compares `!= 'OK'` (`gnrlang.js:73`), and that branch is dead anyway since the next line overwrites it. @dgpaci any preference?
> 3. **Cache busting for `localsdict_<lang>`**: documented in this PR (taken, following the precedent in commit `81f94ce23`) vs adding a version stamp to the storage key in `gnrlang.js:63` — a scope increase into the JS client, worth a follow-up issue if desired. @dgpaci
> 4. A third guard on the `GnrLocString` path (`gnrlocalization.py:41`/`:206`) would close a theoretical hole a legacy `_key="%"` file could still open. Not required by the observed defect. **Taken: leave it.** @dgpaci agree?

Please, if there are so many open questions, better to first discuss them on the issue, not on the PR, otherwise reviewers are asked to review on something that is still a work in progress. Or, other options, put it into draft. I'm removing from a reviewer here, please re-add me when ready.

dgpaci:
The four open decisions are answered on the issue: https://github.com/genropy/genropy/issues/1109#issuecomment-5365657243 — all four confirm the option marked **taken**, so no change is needed to the diff.

fporcari:
@cgabriel re-adding you as you asked. The four open decisions are settled and none of them changed the diff.

You wrote on 20 August that the open questions belonged on the issue rather than in the review. That was right, and they went there: @dgpaci answered all four on #1109 (comment 5365657243) on 21 August, confirming the option already marked **taken** in each case. So the code you would be reading is the code that was there when you stepped back — no new commits since, and nothing left for a reviewer to arbitrate.

For the record, the four and their outcome:

1. escape hatch for genuinely translatable symbols — ship the guard as written; naming the `key` groups in `TRANSLATION` stays a separate issue if it is ever wanted
2. `status` for guarded text — keep `'OK'`, no new `'NOTEXT'`
3. cache busting for `localsdict_<lang>` — documented here, following the precedent in `81f94ce23`; a version stamp in `gnrlang.js:63` would be a scope increase into the JS client
4. a third guard on the `GnrLocString` path — left out, not required by the observed defect

The PR was a draft until today, which is the other reason it went quiet: draft status suppresses the review request, so re-adding you then would have notified nobody anyway. It is ready now.

genro (approved):
Approving. The mechanism is exactly as described: `FLATTENER = re.compile(r'\W+')` collapses any symbol-only string to a single `_`, and `getTranslation` uses `flatten(loctext)` as the fallback key, so `=`, `>=`, `%`, `#` and whitespace all land on `<lang>__` and whichever package registers that slot first answers for every one of them.

I went looking for the way this guard could misfire — a legitimate translatable string flattening to underscores only — and it cannot in practice: `\W` is Unicode-aware on `str` in Python 3, so Cyrillic, Greek and CJK text stays `\w` and never reaches the guard. Only genuinely symbol-only text does, which is precisely the set that should not be translated.

Guarding both sides is the right call, and the collection side is the one that actually stops the poisoning: returning early from the lookup only protects the callers that already exist, while never collecting a symbol-only literal is what keeps the `<lang>__` slot from being claimed in the first place.

The `m.get('key')` finding is worth more than the fix itself. `TRANSLATION` leaves its `{key}` groups unnamed while `LOCREGEXP` names them, so an explicit `!!{cond_eq}=` is collected under `en_cond_eq` and looked up under `en__` — the explicit key has never resolved. Saying so and not fixing it here is the right scope; it deserves its own issue so it is not rediscovered.

cgabriel (approved):
LGTM

PR #1113, opened 2026-08-20
fix(web): bagField without resource no longer raises UnboundLocalError

### LINKED ISSUE #1101: [BUG] bagField without resource= raises UnboundLocalError on mixinedClass
**Problem Description**

`GnrWebPage.bagFieldDispatcher` (`gnrpy/gnr/web/gnrwebpage.py:2675-2690`) binds `mixinedClass` only inside the `if resource:` branch, then reads it unconditionally two lines below:

```python
if resource:
    ...
    mixinedClass = self.mixinTableResource(...)  # or mixinComponent(resource)
bagfieldmodule = getattr(mixinedClass, '__top_mixined_module', None)
```

Every `bagField` that does not pass `resource=` — i.e. the two other documented ways of supplying the form, `method=<callable>` and the `bf_<field>` convention on the page — therefore fails with:

```
UnboundLocalError: cannot access local variable 'mixinedClass' where it is not associated with a value
```

**Reproduction Steps**

2. Open `/test/components/bagfieldpages/bagfield_test` in a browser.
3. Cases 1 (`method=`) and 2 (`bf_alfa`/`bf_beta`) return 500 on the structure POST; the console shows 7 failed requests. Only case 3, which passes `resource=`, works.

The page's initial GET answers 200, so a server-side smoke test that only fetches the URL does not see this.

**Environment**

genropy branch develop, gnrcore:test package, gnrdevelop instance, Python 3.12, Chromium via Playwright.

fporcari:
## Problem

`bagField` without a `resource=` argument raises `UnboundLocalError` on the server. This is the common case: the client (`gnrjs/gnr_d11/js/genro_components.js:3116-3118`) only sets `remote__if='resource'` when `resource` is actually present, so a resource-less `bagField` fires the remote unconditionally and 500s.

## Root cause

`gnrpy/gnr/web/gnrwebpage.py` — `bagFieldDispatcher`:

```python
if resource:
    ...
    if table:
        mixinedClass = self.mixinTableResource(...)
    else:
        mixinedClass = self.mixinComponent(resource)
bagfieldmodule = getattr(mixinedClass,'__top_mixined_module',None)
```

`mixinedClass` is bound only inside `if resource:` (both arms of the inner `if table:` bind it), and read unconditionally right below — `UnboundLocalError` whenever `resource` is falsy. Regression from `0f81ea889` ("bagEditor with sourcecode", 2020-03-09), which introduced `mixinedClass`/`bagfieldmodule` without ever gating the read.

`bagfieldmodule` has **zero consumers** in the codebase today (it appears in exactly this one file, at the producer line and the `contentPane(...)` call). Its one former consumer was already dead before `0f81ea889` landed and was removed later. `None` is already a live, tolerated value: `mixinComponent(safeMode=True)` (`gnrwebpage.py:791-802`) swallows `GnrMixinNotFound` and returns `None`, so a `table=` bagField with a missing resource already produces `bagfieldmodule=None` in production today with no ill effect.

### Corrections to the issue's premise

- **Wrong test page path.** The test page lives at `projects/gnrcore/packages/test15/webpages/components/bagfieldpages/bagfield_test.py` — package `test15`, not `test`. URL: `/test15/components/bagfieldpages/bagfield_test`.
- **All four cases fail, not just one.** The page has `test_1_pane` (`method=`), `test_2_field`, `test_3_field` (`field=` override) and `test_4_resource` (which, despite its name, calls plain `fb.bagField('^.gamma')` / `fb.bagField('^.delta')` — no `resource=` at all). None of the four passes `resource=`, so all four take the falsy-`resource` branch and all four hit the `UnboundLocalError`. The issue's own evidence agrees: 7 failed requests = 1 + 2 + 2 + 2, i.e. every `bagField` on the page.
- **`method=` needs a second, independent fix** (see below) — otherwise `test_1_pane` still 500s after the `UnboundLocalError` fix, with `AttributeError: 'GnrCustomWebPage' object has no attribute 'bf_value'`.
- **`test_4_resource` stays broken by design**, for a separate reason unrelated to this bug (see "Out of scope" below).

## Change

**Hunk 1** — `gnrpy/gnr/web/gnrwebpage.py`: seed `bagfieldmodule = None` before the `if resource:` branch and move the `getattr(...)` read inside it, so the `safeMode` → `None` case keeps flowing through the existing default and the method never holds a name that is unconditionally read but conditionally bound.

**Hunk 2** — `gnrpy/gnr/web/gnrwebstruct/base.py:838-839`: `bagField(method=...)` wrote the attribute as `methodname`, but the JS pops `bfhandler` (`genro_components.js:3108`) — nothing maps one to the other, so the server always saw `bfhandler=None` and fell through to `bf_<field>`. Rename regression from `78f52f96d` ("bagfields update", 2020-02-27), which renamed the server-side parameter `methodname`→`bfhandler` and updated the JS but missed this call site. This aligns `bagField` with the adjacent `grouplet(handler=...)`, whose JS widget pops the matching name (`genro_components.js:3143`).

Both hunks ship together: hunk 1 alone still leaves `method=` silently broken, and the issue cannot be honestly closed without both — see "Open decisions" below for the alternative considered.

## Verification

- **New test**, real code path, no mocks of the fix itself: `gnrpy/tests/web/gnrwebpage_test.py::test_bagFieldDispatcher_without_resource_does_not_raise`. Calls `GnrWebPage.bagFieldDispatcher` directly with `resource=None` and a stub `bf_value` handler, asserts it returns without `UnboundLocalError` and that the pane receives `bagfieldmodule=None`.
  - Confirmed this test **fails on unmodified `develop`** with exactly the predicted `UnboundLocalError: cannot access local variable 'mixinedClass' where it is not associated with a value`, and passes after the fix.
- `flake8` on the three touched files: 0 errors. CI selection `flake8 . --count --select=F401,E9,F63,F7,F82 --show-source --statistics` on the whole tree: 0.
- Full suite: `PYTHONPATH=$PWD python -m pytest tests/ -q --ignore=tests/sql` → 841 passed, 59 skipped, 0 failed. The pre-commit hook additionally ran the complete suite including `tests/sql` → 2012 passed, 69 skipped, 0 failed.
- **Manual live check stayed manual — not performed in this environment.** Steps for a maintainer to confirm: start instance `gnrdevelop` (`projects/gnrcore/instances/gnrdevelop`, mounts `gnrcore:test15`), open `/test15/components/bagfieldpages/bagfield_test`, watch the network panel (the page GET itself is 200 either way). Expected: the 7 failing `bagField` POSTs on `develop` drop to 2 after both hunks (`test_4_resource`'s two fields, blocked by the separate defect below); with hunk 1 alone, 1 more than that.

## Related issue

Fixes #1101

## Out of scope — reported, not fixed here

`test_4_resource` relies on an implicit `bf_<pagename>` resource convention that `78f52f96d` deleted. Passing `resource=` sets `handlername='bf_main'` unconditionally, but every `BagFieldForm` subclass in the repo implements `bf_content`, not `bf_main` (`BagFieldForm` itself is `pass`). So the resource path the issue assumes works is *also* broken — including its one real-world caller in the `dform` package (`projects/gnrcore/packages/dform/resources/tables/dform_data/th_dform_data.py`). This is a separate, pre-existing defect independent of the `UnboundLocalError` fixed here; not addressed in this PR and no issue opened for it — flagging it here for a maintainer to decide whether it needs its own ticket.

## Open decisions

@dgpaci, three calls made while fixing this — flagging in case you'd have gone the other way:

1. **Ship the `method=`/`bfhandler` fix (hunk 2) together with the `UnboundLocalError` fix, rather than splitting it into its own PR.** Taken: together, since it closes the issue's `test_1_pane` case for real; the cost is one extra file and a second failure mode to re-verify in this same change. Splitting would keep a one-line diff but leave `method=` silently broken until someone files the follow-up.
2. **No guard added for a missing handler method** (e.g. `if not hasattr(self,handlername): raise ...`, mirroring `grouplet.py:21-23`, which would turn `test_4_resource`'s bare `AttributeError` 500 into a legible error message). Taken: skip, to keep the diff minimal — but it's worth doing separately, since it would make the residual `test_4_resource` breakage self-documenting instead of a raw 500.
3. **`bagField` is repaired rather than deprecated**, even though it's already labelled `//legacy: use grouplets instead of BagField` in the JS. Taken: repair, since deprecating outright would leave a live `UnboundLocalError` in `develop` and force `dform` to migrate first.


genro (approved):
Approving. I nearly read this as a rename of a public keyword and it is not — it is connecting a producer to a consumer that were never speaking.

`bagFieldDispatcher` takes `bfhandler=None` (`gnrwebpage.py:2675`) and reads only that: `if bfhandler: handlername = bfhandler`. `base.py:839` was sending `methodname=method`, a key nothing on the receiving side ever looks at, so `pane.bagField(value=..., method='foo')` put `methodname` on the node, the dispatcher found no `bfhandler`, and every call silently fell back to `bf_{field}`. The `method` argument has never done anything.

Nothing outside can break on this, which is the question worth asking before renaming anything that crosses a widget boundary: the public parameter is still `method` on `bagField` and its signature is untouched. A caller passing `methodname=` explicitly would have it land in `**kwargs` and be ignored by the dispatcher — exactly as it is today, before and after.

The other hunk is the one the issue is actually about, and it is the right shape. `mixinedClass` is only bound inside `if resource:`, so the unconditional `getattr(mixinedClass, '__top_mixined_module', None)` raised `UnboundLocalError` whenever `bagField` was used without a resource. Initialising `bagfieldmodule = None` and moving the lookup inside the branch fixes it without inventing a fallback: no resource means no mixined module, and `None` is what `contentPane` should receive.

Good that the test asserts the captured `contentPane` kwargs rather than just that the call returns — that is what pins `bagfieldmodule=None` instead of merely proving nothing raised.

cgabriel (approved):
LGTM for me, approved.

PR #1114, opened 2026-08-20
fix: hide the field label too when labels sit on top of the fields

dgpaci:
## Problem

A field declared `hidden` in a formbuilder with labels on top of the fields (`lblpos='T'`, the layout produced by `mobileFormBuilder`) hid the widget but left its label on screen. The only way out was repeating the same condition in `lbl_hidden`, as `sys/resources/preference.py` did for the PDF viewer options.

The label/field cell pairing was implemented for `lblpos='L'` only (`and self.lblpos=='L'` in `_formCell`), where label and field are two cells of the same row. With labels on top they live in twin rows (`r_N_l` / `r_N_f`, same cell childname), so the label cell needs a two level path.

That path could not be resolved: in `gnr.GnrDomSource.getChild`, a named step returned the node it had just found **without moving into it**, and never reset the running node between steps. Any path deeper than one named level therefore landed on the node of the previous step, and a miss returned a stale node instead of `null`. Chains of `parent` steps (`genro.nodeById('X_grid/parent/parent/parent/parent')` in `framegrid.py`) always advanced correctly and are unaffected.

## Changes

- `gnrjs/gnr_d11/js/gnrdomsource.js` — `getChild` walks the path step by step, entering the node it resolves and returning `null` on a miss. `parent` and `#nodeId` steps behave exactly as before.
- `gnrpy/gnr/web/gnrwebstruct/formbuilder.py` — the `hidden` propagation is no longer restricted to `lblpos='L'`; the label cell path comes from the new `_lblCellPath(r, c)`, computed while the cell is built (row and column are known there) instead of being rebuilt client side from the cell name — which, with labels on top, pointed the field back at its own cell. The `hiddenGroup` branch reaches its cell the same way as the `hidden` branch (`attributeOwnerNode('tag','td')`).
- `projects/gnrcore/packages/sys/resources/preference.py` — dropped the two `lbl_hidden` workarounds, no longer needed.

`lbl_hidden` keeps precedence: when it is passed explicitly, nothing is injected and the label follows its own binding.

## Test cases

Unit tests in `gnrpy/tests/web/gnrwebstruct_test.py` (6 new): they resolve the generated `getChild` path against the built source the same way the client walks it, and assert it lands on the label cell of the *hidden* field (`innerHTML == 'Alfa'`), with labels on top and on the left; plus `lbl_hidden` precedence, no handler on visible fields, and a `hiddenGroup` member reaching its own cell. Three of them fail without the fix.

Manual test page `test15/webpages/html/test_hidden.py`, new `test_4_hidden_lblpos_top` and `test_5_hidden_group_lblpos_top` (both `mobileFormBuilder(cols=2)`, so a wrong target would show up as the neighbour's label disappearing):

1. `test_4` — "Hidden alfa" starts checked: the *Alfa* label and its field are hidden, *Beta* untouched. Unchecking shows both; checking "Hidden beta" hides the Beta pair only.
2. `test_5` — hiding *Alfa* hides *Gamma* too (same `hiddenGroup`), labels included; *Beta* and *Delta* stay visible.
3. `test_0` / `test_1` (labels on the left, pre-existing) — unchanged, including the row-hidden and `hiddenGroup` cases.

Verified in the browser on a `gnrdevelop` instance: cell `display` checked before and after each toggle, no console errors. `gnrpy` suite green (2100 passed, 10 skipped), flake8 clean.


fporcari (approved):
The diagnosis is exactly right, and it is the valuable half of the PR: `getChild` never advanced into a resolved named step — `node` is declared outside the loop and the direct-hit branch returned from the `dojo.forEach` callback without updating `curr` — so any path deeper than one named level landed on the node of the first step, and a miss returned a stale node instead of `null`. Confirmed on `projects/gnrcore/packages/test15/webpages/temp/childname.py`, the 2010 manual page whose 4-level paths (`RootNode/green_node/blue_node/lime_node`) have been demonstrating the bug all along: the old walker returns `green_node` for that path.

Checks run on the branch:

- `_lblCellPath` matches the real cell naming on both layouts: `lblpos='L'` one row with `c_N_l`/`c_N_f` (`formbuilder.py:343`, `:363`); `'T'` twin rows `r_N_l`/`r_N_f` with the same `c_N` childname (`:385`, `:388`).
- The new `instanceof gnr.GnrBag` guard is safe on the first step too: `GnrDomSource` → `GnrStructData` → `GnrBag` (`gnrdomsource.js:2106`, `:2112`).
- Shielded call sites: single-segment paths and all-`parent` chains (`framegrid.py:353`) behave identically under the new walker, and `gnr_d20` is CSS-only, so there is no second copy to fix.
- `gnrpy/tests/web/gnrwebstruct_test.py`: 27 passed on the branch; flake8 clean on the touched files. One note for whoever re-runs them from a worktree: put the branch's `gnrpy` first on `PYTHONPATH`, or an editable install shadows it and 4 tests fail against the old code.

Process note: the PR had no assignee — set to the author per the workflow (the assignee is who merges).

PR #1115, opened 2026-08-20
refactor(docu): publish the handbooks in a storage of their own

dgpaci:
Follow-up of #1085 (genropy/genropynet#2) on the writing side of the handbooks.

## Problem

The published handbook was a folder of the `documentation` storage (`handbook.sphinx_path` → `documentation:handbooks/<name>`, `documentation:local_handbooks/<name>`). Since #1085 the attachments and the page images are linked by the public url of their own node, so keeping `documentation:` on a bucket is now a sensible configuration — but the generated html follows the same storage, with no way to opt out: the static site ends up on the bucket, which is the serving problem #1085 had to work around in the first place.

## What changes

- **A storage per kind**: `handbooks:<name>` for online handbooks, `local_handbooks:<name>` for the zip of local ones. Nothing to configure for the default instance: a storage name with no service falls back to `<site_static_dir>/<name>` (`LegacyStorageHandler.storage()`). A deployer points the `handbooks` storage service at the volume the docs web server publishes, independently of where `documentation` points.
- **The export builds in a disposable folder of the page** (`page:handbook_build/<name>/sphinx`) instead of `site:handbooks/<name>`: the site cleanup removes it along with the page, and each run is isolated from the others. Sphinx reads and writes through the filesystem — `internal_path` on `aws_s3`/`sftp` answers a key, not a path — so the build cannot happen in the published folder when that is remote; it always happens where a filesystem is guaranteed, and `post_process` publishes the result.
- **The zip is written straight into the folder it is served from**, instead of being built elsewhere and moved afterwards.
- **`trigger_onDeleted` fixed**: it erased `sphinx_path/sphinx/{build,source}`, a path the export stopped writing when the temporary build area was introduced (so the published files of a deleted handbook stayed forever), and it read the `sphinx_path` formula from a record that does not always carry virtual columns. It now goes through `handbook.handbookStorageNode(record)`, the same table method the export uses, built from real columns only; the formula column, shown by the handbook form, is its SQL mirror.

## Upgrade notes

Published handbooks move from `<site>/documentation/handbooks/<name>` to `<site>/handbooks/<name>` (local handbook zips from `<site>/documentation/local_handbooks/<name>` to `<site>/local_handbooks/<name>`). After the upgrade:

- re-export each handbook, or move the folder, and repoint the root of the web server publishing the docs;
- the urls already stored in `handbook.local_handbook_zip` keep working until the next export: their files stay where they are.

No database change: `sphinx_path` is a formula column.

## Test case

`projects/gnrcore/packages/docu/tests/test_docu_handbook_storage.py` (7 tests, real database and real local storage services on temporary directories):

- the folder of an online handbook and of a local one;
- the formula column and `handbookStorageNode` agree, so the path the form shows is the folder the export writes into;
- deleting a handbook erases its published folder, driven by a record with no virtual columns (the regression the old trigger would have hit);
- the export works in the page folder whatever the handbook storage is;
- `post_process` promotes the build to the root of the handbook folder and drops the working folder;
- a file of the former export does not survive in the published folder.

`StorageSiteStub`, shared with `test_docu_media_export.py`, moves to `packages/docu/tests/sitestub.py`.

Run with `pytest projects/gnrcore/packages/docu/tests` → 47 passed. `gnrpy/tests/app` + `gnrpy/tests/web` → 376 passed, 59 skipped. flake8 clean on the package.

Note: the package tests are outside `gnrpy/pyproject.toml`'s `testpaths`, so CI does not collect them yet (pre-existing gap, raised in #1085).


fporcari (approved):
Every load-bearing claim in the description checks out against the code:

- the unconfigured-storage fallback is real: `LegacyStorageHandler.storage()` falls back to `<site_static_dir>/<name>` when the name is not in `storage_params` (`gnrstoragehandler.py:449-457`), so the default instance needs no configuration;
- `page:` is a registered *symbolic* storage (`DEFAULT_STORAGE_CONFIGS`, `gnrstoragehandler.py:42`), so the sphinx build always lands on a filesystem — which is the point of the disposable build folder;
- `handbookStorageNode` reads only real columns (`is_local_handbook`, `name`), so `trigger_onDeleted` no longer depends on virtual columns being on the record, and `delete()` guards on `exists()` and rmtree's directories (`storage.py:570-577`, `:755-756`) — the leak of published files of deleted handbooks is actually closed;
- the `internal_path` → `fullpath` switch on the zip is load-bearing, not cosmetic: `zipFiles` re-resolves its `zipPath` through `storageNode(...).open('wb')` (`gnrwsgisite.py:2033-2036`), the one route that can ever reach a remote storage.

Ran the suite on the branch: 47 passed, real database and real local services on mkdtemp dirs with rmtree teardown; flake8 clean.

Two notes, neither blocking:

1. Three of the seven new tests — precisely the ones covering the export/publish mechanics — go through `_batch`, which `importorskip`s sphinx **and** boto3. Worth knowing that #1090 proposes making boto3 an optional extra: the day it lands these three skip silently on a default install, on top of the pre-existing gap that CI does not collect package tests at all.
2. The PR had no assignee — set to the author per the workflow.

PR #1120, opened 2026-08-21
feat(pdf): restore the sys.pdf_render default margins and apply them in every htmltopdf implementation

### LINKED ISSUE #1007: [FEATURE] Restore default PDF margin preferences and apply them in all htmltopdf implementations

dgpaci:
Fixes #1007

## Problem

The `sys.pdf_render` margin preferences (`margin_top/bottom/left/right`) were only shown
when the `wk_legacy` checkbox was on, and only wkhtmltopdf ever consumed them: it translated
each key into a `--margin-*` CLI flag. The WeasyPrint service dropped `pdf_kwargs` into
`**kwargs` and ignored it. Since WeasyPrint became the default implementation the site-wide
default margins were dead code, and every `GnrHtmlBuilder` print — which emits
`@page { margin: 0 }` — rendered edge-to-edge and got clipped by the printer's non-printable
area when printed at 100% on real paper. wkhtmltopdf never showed this because it enforced its
own page and margins and smart-shrank the content to fit.

## Solution

1. **`HtmlToPdfService.pageMarginsFromPdfKwargs`** — one place that turns the preference values
   into CSS lengths, shared by every implementation. A bare number gets `mm` (wkhtmltopdf's
   implicit unit, so stored preference values keep their meaning); an explicit css unit is kept;
   internal whitespace and a decimal comma are normalized (`'10 mm'` → `'10mm'`, `'1,5'` → `'1.5'`).
   An invalid value (negative, unknown unit, garbage) skips that side entirely rather than being
   interpolated into the stylesheet, where the renderer would drop it and leave the page
   silently asymmetric.

2. **WeasyPrint service** — appends an `@page` rule built from those margins. The stylesheets
   handed to `write_pdf` are *user* origin, so each configured side carries `!important` to beat
   the document's own author `@page { margin: 0 }`. Sides left unset in the preferences default
   to `0` as **plain** declarations: symmetric output, still beating the WeasyPrint UA default,
   but still losing to margins the document declares deliberately. With no preference margins at
   all no stylesheet is appended, so raw html with no `@page` rule of its own (grid pdf export,
   pagededitor preview) keeps the UA defaults instead of going edge-to-edge.

3. **Preference panel** — the four margin fields are now always visible. `zoom` stays
   wkhtmltopdf-only, since it maps to a wkhtmltopdf CLI flag.

4. **`BagToHtmlWeb.pdfMarginKwargs`** — a print that defines its own page margins must win over
   the site default. `page_margin_*` are realized as absolute offsets *inside* the page box
   (`gnrhtml.py`), so an `@page` margin on top of them would double-count. When the print defines
   its own margins the preference is forced to `0` on every side.

5. **`BagToHtml.page_margins_defined`** — the flag `pdfMarginKwargs` reads, set in
   `prepareTemplates` from whether a letterhead is loaded. Per-side `page.*` values cannot be used
   to mean "explicitly set", because the letterhead editor seeds zeros on every new letterhead, so
   a `0` cannot be told apart from a deliberate edge-to-edge choice. Instance attributes cannot be
   inspected either: the assignments below it run on every call and `prepareTemplates` may run more
   than once on the same instance.

## What this actually guarantees — please read before approving

**Any loaded letterhead suppresses the `sys.pdf_render` preference**, whatever its margin values,
because `page_margins_defined` is `bool(self.htmlTemplate)`. That is deliberate (point 5 above:
a `0` in a letterhead is indistinguishable from a seeded default, and double-counting is the worse
failure), but it narrows the reach of the preference more than the issue text suggests. Concretely,
in erpy a letterhead is always loaded — from the `erpy_fatt.carta_intestata_id` preference — so
those prints keep taking their geometry from the letterhead and are unaffected.

The site-wide default therefore covers **prints with no letterhead and raw-HTML exports**, which is
where the clipping was actually observed. Prints with a letterhead that need margins must set them
in the letterhead, which is the one place that knows the page design. Flagging it explicitly
rather than letting the issue title imply a broader promise.

## Relation to #1052

This branch now sits on top of #1052 (merged): an explicit `pageMargin` argument is authoritative
and skips the preference margins entirely (`apply_pref_margins = pageMargin is None`, resolved
*before* the block that reassigns `pageMargin` to its `0` fallback).

#1052 shipped source-only, by design — its test belonged in this branch's file. The second commit
here pays that debt: it renames `test_weasyprint_explicit_pagemargin_skips_pref_margins`, drops the
"known pre-existing bug" paragraph from its docstring, and tightens the assertion, which previously
only checked `x0 < 8mm` and so accepted both the buggy `0.0` and a working `14.17pt`. **Verified
negatively**: with `margin_priority` forced empty, both tightened tests fail; with #1052 in place
they pass.

## Tests

`gnrpy/tests/core/htmltopdf_test.py`, 27 tests, no external service and no mock of project code.
The integration tests resolve the real weasyprint service from `resources/common` (same harness as
`services_addservice_test`), render real HTML and read the marker position out of the produced PDF
with PyMuPDF. They `importorskip` weasyprint and fitz.

- `pageMarginsFromPdfKwargs` (8): bare numbers get mm, `0` is a value not a falsy skip, explicit
  units preserved, empty/missing skipped, input dict not consumed, unrelated keys ignored,
  locale/spacing normalized, invalid values skipped.
- `pdfMarginKwargs` (5): passthrough with no print margins, print margins force zero, letterhead
  with zero margins forces zero, explicit override wins, input not mutated.
- `page_margins_defined` (5): false without letterhead, true with letterhead (zero margins, real
  margins, and no margins at all), reset when `prepareTemplates` re-runs without a letterhead.
- weasyprint end-to-end (9): preference margins beat a document `@page{margin:0}`; no prefs keeps
  the document's zeros; UA default survives with neither prefs nor `@page` rule; partial prefs zero
  the missing sides; the document's own margins survive with no prefs; explicit `pageMargin` wins
  over both the document and the prefs; explicit `pageMargin` beats a document `@page{margin:15mm}`;
  `pageSize`-only still loses to the `!important` prefs; `margin_*=0` beats the document's margins.

`flake8` clean on all six touched files. 64 passed running the suite plus `gnrbaghtml`, `gnrhtml`,
`gnrservices`, `gnrbaseservice` and `gnrprinthandler` as a regression gate.

## Manual test scenario

1. Set `sys.pdf_render` margins (e.g. top 30, left 20) in the preference panel — the fields are now
   visible without enabling legacy mode.
2. Print any `GnrHtmlBuilder` document **without** a letterhead: the content is no longer flush to
   the paper edge.
3. Print a document **with** a letterhead: unchanged, the letterhead's own geometry rules.
4. Export a grid to PDF: unchanged WeasyPrint UA margins when no preference is set.

## Localization

No new strings. The four labels already carry their Italian translations in
`packages/sys/localization.xml` (`Margine superiore/inferiore/sinistro/destro`); the diff only
removes the `hidden` conditions.


fporcari (approved):
Verified precedence across site preferences, explicit `pageMargin`, document CSS, and letterhead geometry. The real WeasyPrint tests exercise rendered output and all targeted tests pass. No blocking findings.

cgabriel (approved):
LGTM

fporcari (approved):
Re-approving on 0bf9880a. This is a procedural refresh, so I am saying explicitly what I did and did not re-check.

My 08-21 approval and @cgabriel's 08-25 LGTM were dismissed by the push that followed, and that push is the merge of develop plus exactly one commit of your own: 53de1e539e, the single line adding the `gnrapp` stub to `make_pdf_site` so `ResourceLoader` finds the `experimentalFlag` it now reads while constructing. `git diff 50c602a2..0bf9880a` on this PR's own files is that one line; everything else in the range came in through develop (#1182, #1192). No behaviour owned by this PR moved, checks are 8/8 green, the branch contains current develop and there are no open threads — so I have not repeated the precedence walk across site preferences, explicit `pageMargin`, document CSS and letterhead geometry, and I have not re-run the tests locally: CI covers the fixture change, which is the only thing that moved.

One thing worth knowing before you merge, since it is new since my review. Three open PRs touch `gnrpy/gnr/web/gnrbaseclasses.py`, and two of them overlap textually:

- this one at old lines 290-303 (`pdfMarginKwargs` inserted above `writePdf`) and 390-395;
- #1045 at old lines 264-274 and 300-308 (`contentFromTemplate`, `TableTemplateToHtml.__call__`).

Those two share old lines 300-303, so whichever merges second will conflict there and needs a rebase — the resolution is mechanical (the two additions are independent), but it is a hand resolution either way. #1074 is at old line 609 and above, in `gridData`, and does not interfere with either.

The merge is yours, and so is the ordering call. @genro is still a requested reviewer; my approval already satisfies the CODEOWNERS requirement on `gnrpy/`, `projects/` and `resources/`, so waiting for him is your choice, not a requirement.

PR #1125, opened 2026-08-21
fix(docu): docuresolver doubled the handbook root in the resolved url

### LINKED ISSUE #1124: [BUG] docuresolver redirects to a doubled handbook root node
## Problem Description

`docuresolver` (introduced in #1085) redirects to a URL where the handbook root node is repeated twice, so the 301 lands on a page that does not exist.

Requesting a dead URL of a published handbook, e.g. `https://docs.example.org/widgetpedia/inputfields/textbox.html` (the `elements` node is missing from the path), the resolver answers `301` to:

```
https://docs.example.org/widgetpedia/widgetpedia/elements/inputfields/textbox.html
```

instead of the correct `https://docs.example.org/widgetpedia/elements/inputfields/textbox.html`.

The doubled URL 404s on the docs web server, the fallback hits the resolver again with the doubled path, and the redirect-loop guard in `resolveRequestPath` stops there: the reader gets a 404 after one bogus hop.

**Root cause** — `docu.Table.calculateExternalUrl` (`packages/docu/main.py`) builds the path from the *rows* of the `getAncestors` query, which selects `@handbooks.handbook_url`. That is the reverse of `handbook.docroot_id`: `mode='foreignkey'` creates the FK but does not make the relation 1:1, so `@handbooks` is a one-to-many and the join repeats the docroot row once per handbook pointing at it. The first row is recognized as the handbook root (its name is excluded from the path), the second row of the same document falls through to the path segment collector and injects the handbook root name into the URL.

Two handbooks on the same docroot are legitimate data (a handbook published on the web plus a local zipped one, or a leftover duplicate), so the code has to tolerate them.

The defect is systematic for every page of an affected handbook, and it also affects the `full_external_url` pyColumn (used in the sphinx export log), not just the resolver. It surfaced now because before #1085 nothing consumed that URL.

## Reproduction Steps

1. Create a documentation tree: `widgetpedia` -> `elements` -> `inputfields` -> `textbox`, all published.
2. Create a handbook `widgetpedia` with `docroot_id` = the `widgetpedia` document, `toc_roots` = the `elements` document, `handbook_url` = `https://docs.example.org/widgetpedia/`.
3. Read `calculateExternalUrl` of `textbox` (or call `docu.documentation.resolveRequestPath('widgetpedia/inputfields/textbox.html')`).
   Result: `https://docs.example.org/widgetpedia/elements/inputfields/textbox.html` — correct.
4. Add a second handbook (e.g. `widgetpedia_local`, `is_local_handbook=True`) with the same `docroot_id` and a `handbook_url`.
5. Read `calculateExternalUrl` of `textbox` again.

Observed result: `https://docs.example.org/widgetpedia/widgetpedia/elements/inputfields/textbox.html` — the handbook root name appears twice.

Expected result: the URL is unchanged by the presence of a second handbook on the docroot.

The `getAncestors` query returns 5 rows instead of 4, the docroot repeated:

```
{'id': 'l1Ze…', 'name': 'widgetpedia', 'handbook_url': '…/widgetpedia/', 'handbook_toc_roots': 'vM_z…'}
{'id': 'l1Ze…', 'name': 'widgetpedia', 'handbook_url': '…/widgetpedia/', 'handbook_toc_roots': None}
{'id': 'vM_z…', 'name': 'elements',    'handbook_url': None, 'handbook_toc_roots': None}
{'id': '7iFl…', 'name': 'inputfields', 'handbook_url': None, 'handbook_toc_roots': None}
{'id': 'EuIG…', 'name': 'textbox',     'handbook_url': None, 'handbook_toc_roots': None}
```

Handbooks sharing a docroot can be listed with:

```sql
SELECT id, name, docroot_id, handbook_url, is_local_handbook FROM docu.handbook
WHERE docroot_id IN (SELECT docroot_id FROM docu.handbook GROUP BY docroot_id HAVING count(*) > 1);
```

## Environment

genropy `develop` (docuresolver introduced in #1085), instance with the `docu` package, PostgreSQL and SQLite alike.

## Additional Material

Related: the 404 page the resolver answers with when nothing matches is currently the `html_pages/missing_result.html` placeholder (a bare question mark image), instead of the plain framework 404 page every unresolvable instance URL answers with since #1012, and it offers no way back to the handbook.

dgpaci:
Fixes #1124

## Problem

`docuresolver` redirected to a url with the handbook root node repeated twice, e.g. requesting the dead url `https://docs.example.org/widgetpedia/inputfields/textbox.html` it answered `301` to

```
https://docs.example.org/widgetpedia/widgetpedia/elements/inputfields/textbox.html
```

instead of `https://docs.example.org/widgetpedia/elements/inputfields/textbox.html`. The doubled url 404s on the docs web server, the fallback comes back to the resolver, and the redirect-loop guard stops there: the reader gets a 404 after one bogus hop.

## Root cause

`calculateExternalUrl` built the path from the **rows** of the `getAncestors` query, which selects `@handbooks.handbook_url`. That relation is the reverse of `handbook.docroot_id`, and `mode='foreignkey'` only creates the FK — it does not make the relation 1:1 — so the join repeats the docroot row once per handbook pointing at it. The first row was recognized as the handbook root (name excluded from the path), the repeated row of the same document fell through to the segment collector and injected the handbook root name into the url.

Two handbooks on the same docroot are legitimate data (a handbook published on the web plus a local zipped one, or a leftover duplicate), so the code has to tolerate them. The defect was systematic for every page of an affected handbook and also hit the `full_external_url` pyColumn used in the sphinx export log; it surfaced now because before #1085 nothing consumed that url.

## Changes

- `calculateExternalUrl` collapses the ancestors rows to one row per document before building the path, so a repeated docroot can no longer become an extra path segment.
- **Local handbooks are excluded** while collapsing: they are downloadable zips with no published site, so they never name a url, not even with a `handbook_url` left over from an earlier web export. Their docroots count as plain path segments, and a tree published only as a zip resolves to nothing (correct: there is no site to redirect to).
- The resolver **404 page** was the `html_pages/missing_result.html` placeholder (a bare question mark image). It is now the plain framework page every unresolvable instance url answers with since #1012, plus a **Back to handbook** link to the index of the handbook publishing the requested path:

  ```html
  <!doctype html>
  <html lang=en>
  <title>404 Not Found</title>
  <h1>Not Found</h1>
  <p>The requested URL was not found on the server. ...</p>

  <p><a href="https://docs.example.org/widgetpedia/index.html" style="…">Back to handbook</a></p>
  ```

  The handbook is matched on the longest `handbook_url` path prefix (new `docu.handbook.indexUrlFromRequestPath`), which keeps working with a baseurl carrying a path of its own and with a handbook name differing from its url segment; no match, or a local-only match, leaves the bare framework page. `!!Back to handbook` is localized in `localization.xml`.
- `_absoluteLocation` moves from `docu.documentation` to the docu package base table, shared by the documentation redirect and the new link instead of being reached across tables.

Note: the 404 page is no longer overridable per instance through the resource cascade — it is now the framework page by design.

## Test cases

`packages/docu/tests/test_docu_resolver.py`, run with `gnr dev tests gnrdevelop docu` (real database, no mocks): 25 passed in the file, 46 in the docu suite, flake8 clean.

New/rewritten:

1. `test_shared_docroot_does_not_double_handbook_root` — two handbooks on the same docroot (`twinbook` + a local `twinbook_local`): a leaf page resolves to `/twinbook/twinsection/twinpage.html` and the docroot to `/twinbook/index.html`. Without the fix the leaf resolved to `/twinbook/twinroot/twinsection/twinpage.html`.
2. `test_local_handbook_never_resolves` — a tree whose only handbook is local, `handbook_url` valorized: resolves to `None`.
3. `test_webtool_unknown_page_404_is_framework_page` — 404 body starts with werkzeug's page and stays under 1KB (no client bootstrap).
4. `test_webtool_404_links_back_to_handbook` — the body carries `href="…/testguide/index.html"` and the localized label, with no `!!` marker left.
5. `test_webtool_404_without_handbook_has_no_link` / `test_webtool_404_link_skips_local_handbook` — body is the bare framework page.
6. `test_handbook_index_url_from_request_path` / `test_handbook_index_url_absolutized_by_preference` — prefix match, no match, empty path, and a host-relative `handbook_url` absolutized against `docu.sphinx_baseurl`.

Removed: the two tests asserting the `missing_result.html` template body and its localized fallback.

## Manual check

Handbooks sharing a docroot can be listed with:

```sql
SELECT id, name, docroot_id, handbook_url, is_local_handbook FROM docu.handbook
WHERE docroot_id IN (SELECT docroot_id FROM docu.handbook GROUP BY docroot_id HAVING count(*) > 1);
```

On such an instance, a dead url under the handbook must now redirect to the page with a single root segment, and an unresolvable one must answer the framework 404 with the Back to handbook link.


fporcari (approved):
Verified that ancestor deduplication preserves hierarchy order, local handbooks cannot supply external URLs, link data is escaped, and the real-database resolver tests cover the regression and the 404 branches. No blocking findings.

PR #1126, opened 2026-08-21
set GNR_EXTERNALHOST variable only if a FQDN is supplied

cgabriel:
set the GNR_EXTERNALHOST variable if and only if one or more FQDNs are supplied.

The previous behaviour didn't allow a genropy application to be deploy as an internal service only, using internal dns resolver for its service. If not FQDN was supplied, the GNR_EXTERNALHOST variable set failed by finding an empty FQDN list.


fporcari (changes requested):
An internal-only deployment is still unreachable through the public CLI because `--fqdn` remains required. When `GnrK8SGenerator` is called directly with an empty list, `get_ingress()` also emits an Ingress with empty rules and TLS hosts. Please make `--fqdn` optional with an empty-list default, omit the Ingress when no FQDN is supplied, and add a focused test for the generated manifest.

cgabriel:
> An internal-only deployment is still unreachable through the public CLI because `--fqdn` remains required. When `GnrK8SGenerator` is called directly with an empty list, `get_ingress()` also emits an Ingress with empty rules and TLS hosts. Please make `--fqdn` optional with an empty-list default, omit the Ingress when no FQDN is supplied, and add a focused test for the generated manifest.

You're absolutely right! The PR was born in a context where the CLI was not used. In [59f6b8c](https://github.com/genropy/genropy/pull/1126/commits/59f6b8c3bdf90db9ea2a9a68cf7a4a0b8741e8e0) the CLI argument has been changed to optional, and the ingress is now omitted if no FQDN is provided. I've also added a 100% test coverage for such module, which was indeed missing (mainly to its experimental/explorative nature)

fporcari (approved):
All requested changes are addressed: the CLI accepts an omitted `--fqdn`, internal-only manifests omit both `GNR_EXTERNALHOST` and the Ingress in monolithic and split modes, and the new tests exercise the generated YAML without mocking the generator. The focused suite passes with 24 tests, flake8 is clean, and CI is green on Python 3.11–3.14. No blocking findings.

genro (dismissed):
The one-line guard is right, and I confirmed both ends of it by running the generator directly: on `develop`, `GnrK8SGenerator('myapp', 'img:1', [])` dies with `IndexError: list index out of range`; with this patch it builds and `GNR_EXTERNALHOST` is simply absent from the env list. No other reader of `self.fqdns` is affected by the guard itself.

I am blocking because the internal-only deployment the PR sets out to enable is not reachable yet, for two reasons that live in the same piece of logic.

## 1. The CLI still refuses an empty fqdn list

`gnrpy/gnr/web/cli/gnrk8s.py:33-37`:

```python
parser.add_argument('-f', '--fqdn',
                    action="append",
                    required=True,
                    dest="fqdns",
                    help="One (or more) FQDN of the deployed service")
```

`required=True` means argparse rejects the invocation before the generator is ever constructed, so `fqdns=[]` is only reachable from a programmatic caller. To actually deploy without a FQDN one has to drop `required=True` and give it `default=[]` — which is exactly the shape the same option already has in `gnrdockerize` (`gnrpy/gnr/app/cli/gnrdockerize.py:464-469`: `action='append', default=[]`, not required).

## 2. An internal-only deploy still emits an Ingress

`generate_conf` always extends the manifest with `get_ingress()`, and there is no fqdn check in it. Ran on this branch with `fqdns=[]`:

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: myapp-ingress
  annotations:
    traefik.ingress.kubernetes.io/router.entrypoints: websecure
    traefik.ingress.kubernetes.io/router.tls: 'true'
    traefik.ingress.kubernetes.io/router.tls.certresolver: letsencrypt
spec:
  rules: []
  tls:
  - hosts: []
```

A service reachable only through the internal DNS resolver should produce no Ingress at all. As it stands the manifest carries an object that routes nothing, and the `letsencrypt` certresolver is asked for a certificate over an empty host list. The in-house precedent for the shape is `_compute_extra_labels` (`gnrdockerize.py:317`): no fqdn and no router means no routing configuration is emitted, `return []`.

Both are the same rule as the env var — "no FQDN, no externally addressable configuration" — so they belong in this PR rather than after it.

## Tests

Nothing in `gnrpy/tests/` exercises `gnrk8s.py` today (`gnrdockerize_test.py` is the only generator test file, and it covers the compose path). Since the empty-fqdn case is now a supported input, it is worth a test asserting that with `fqdns=[]` no `GNR_EXTERNALHOST` entry appears and `generate_conf` emits no Ingress, plus the existing single-fqdn case as the counterpart. `gnrdockerize_test.py`'s `_make_builder` is the pattern to follow.

genro (approved):
You are right, and the previous review was wrong. Thank you for checking it rather than acting on it.

The timeline settles it: `59f6b8c3` (CLI optional, Ingress omitted) landed 21 Aug 13:13 and `24fae773` (tests) 13:23, while that review was submitted 22 Aug 04:45 — fifteen hours later, and after @fporcari had already verified the same changes. The claims about `required=True` and the empty-hosts Ingress described the state of `7526cbbf`, the first commit, not the branch. Whatever tree the generator was actually run against, it was not this one.

Re-checked now on `97df9228`, printing the resolved module path first so there is no doubt which file answered:

```
modulo caricato: .../worktrees/reviews/gnrpy/gnr/web/gnrk8s.py
GNR_EXTERNALHOST presente: False
get_ingress(): []
con fqdn -> GNR_EXTERNALHOST = https://goober.com
con fqdn -> ingress kind: ['Ingress']
```

So both ends hold: no FQDN gives neither the env var nor an Ingress object, one FQDN gives both. `--fqdn` carries `action="append", default=[]` and is not required — the only `required=True` left on that parser is `--image`, which should be. `gnrpy/tests/web/gnrk8s_test.py`: 24 passed. flake8 clean on the three files.

Approving. The one thing worth saying beyond the fix: this module had no test file at all before, and the PR arrives with one covering both the internal-only and the published shapes. That is the part that will keep the guard from being undone by accident later.

PR #1128, opened 2026-08-21
refactor(test): remove hand-maintained test menus

### LINKED ISSUE #1107: [TASK] Deprecate the hand-maintained Tests menu of the gnrdevelop instance
### Summary

Drop the hand-written `Tests` menu that the test packages contribute to the gnrdevelop instance, and stop maintaining its per-folder branches.

### Objective

The menu is a second, hand-maintained index of pages that are already discovered by walking `webpages/`, and it rots silently: a branch whose folder no longer exists does not degrade, it raises. `dir=` is rewritten into a `directoryBranch` (`gnrpy/gnr/web/gnrmenu.py:379-382`) whose `sourceBag` iterates `folderSN.children()` (`gnrmenu.py:786-789`), and `StorageNode.children()` returns `None` when the path is not a directory (`gnrpy/gnr/lib/services/storage.py:286-289`) — so expanding such a branch fails with `TypeError: 'NoneType' object is not iterable`. Nothing in the test suite builds the menu, so the breakage only ever surfaces under a developer's click.

Two branches were in that state in `projects/gnrcore/packages/test15/menu.py`: `Components`, dangling since the components area was migrated out (removed as part of the migration's coherence pass), and `Dojo`, which points at a `webpages/dojo/` folder that does not exist on develop at all and has been dangling for longer.

### Scope

In scope:
- Remove the `Tests` menu contributions of `gnrcore:test15` and `gnrcore:test` from the gnrdevelop instance, or replace them with a single folder-driven entry that cannot go stale.
- Decide the fate of the legacy `root.branch(pkg=..., dir=...)` form: `gnrcore:test` already uses `directoryBranch(folder=...)`, `gnrcore:test15` still uses the old one.

Out of scope:
- The `test15` package itself, which the ongoing example migration removes entirely at its last macro-phase.
- Any change to how `directoryBranch` resolves folders — but note that a missing folder raising `TypeError` instead of yielding an empty branch is worth a guard of its own, whichever way this task is resolved.

### Technical Details / Notes

Found while migrating the `test15` example pages into `test`: the migration empties `test15/webpages/` folder by folder, so every macro-phase leaves the hand-written menu one branch more out of date. The dangling `Dojo` branch shows the failure predates the migration.

fporcari:
## Motivation

The `test` and `test15` packages maintained a second, manual index of pages that are already available through package routing and discovery. Those lists drift from the webpages tree, and a stale directory branch raises instead of degrading gracefully.

Measured on this branch: `test/webpages/` holds 18 folders while the hand-written list covered 14 — `bag`, `error_test`, `security`, `ui_builder` and the root-level `iframe_inside.py` were invisible in the menu.

## Change

Following @dgpaci's option C (review [#pullrequestreview-5020415620](https://github.com/genropy/genropy/pull/1128#pullrequestreview-5020415620)):

- `projects/gnrcore/packages/test15/menu.py` — **deleted**. The package goes away with the example migration, so a dynamic entry there would be dead code.
- `projects/gnrcore/packages/test/menu.py` — **kept, but fully folder-driven**: the 14 hand-written `directoryBranch` calls are replaced by a single dynamic entry rooted at `webpages/`. `DirectoryMenuResolver.sourceBag` already recurses, so one entry covers the whole tree and cannot drift.

```python
# encoding: utf-8
class Menu(object):
    def config(self, root, **kwargs):
        root.directoryBranch('!!Tests', folder='', tags='_DEV_')
```

- `gnrpy/gnr/web/gnrmenu.py` — three hardenings of `DirectoryMenuResolver`, required by that rooted use:
  1. a missing `folder` is normalized to `''`. `site.storageNode` reaches `_adapt_path`, which does `'/'.join(args)` and raises `TypeError` on `None`.
  2. `folderSN.children() or []`. `StorageNode.children()` returns `None` when the path is not a directory, so the loop used to raise `TypeError` on a missing or non-directory folder.
  3. `not sn.basename.startswith('_')` replaces the `__pycache__`-only skip, so `test/webpages/_resources` no longer surfaces as an empty "Resources" branch.

Package routing keeps working for `test15` because the existing menu loader treats a package without `menu.py` as an empty menu structure and `legacyMenuFromPkgList` skips empty package menus.

## Menu nesting

`nodeType_packageBranch` collapses a package menu with a single static child (`len(value) == 1 and value['#0']`). A `directoryBranch` node carries no child Bag, so the collapse no longer fires: the entry becomes `Test > Tests > 18 folders` instead of today's `Test > 14 folders`. Accepted deliberately — see the PR comment for the evidence and the reasoning.

## Verification

- Narrow gate: `PYTHONPATH=$PWD/gnrpy pytest gnrpy/tests/web/gnrmenu_test.py -q` — 4 passed. All four fail on the base without the resolver change.
- Non-SQL suite: `PYTHONPATH=$PWD/gnrpy pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` — 906 passed, 59 skipped.
- Pre-commit hook (flake8 + `core sql web app xtnd`) — flake8 OK, 2080 passed, 69 skipped.
- `flake8` on `gnrpy/gnr/web/gnrmenu.py`, `gnrpy/tests/web/gnrmenu_test.py`, `projects/gnrcore/packages/test/menu.py` — zero errors.

The gnrdevelop menu was not verified live because it is login-gated. Manual verification:
1. Run `gnr web serve gnrdevelop` with the existing local configuration.
2. Sign in with a developer account that can see `_DEV_` entries.
3. Open the gnrdevelop application menu and confirm the Tests branch lists all 18 folders plus `Iframe Inside`, with no `Resources` branch.
4. Confirm the `test15` contribution is absent and direct routing to a `test15` page still resolves.

## Related issue

Fixes #1107


fporcari:
@dgpaci added as a reviewer, because the open decision in the description is
addressed to him and until now he was not on the PR — the question was sitting
where he could not see it, which is why this has stayed a draft.

Restating it here so it does not need digging out of the description:

- **A** — implemented and recommended: remove both menu contributions
  completely. The Tests UI entry disappears; direct routing and page discovery
  keep working, because the menu loader treats a package without `menu.py` as an
  empty menu structure.
- **B** — keep one folder-driven entry. The Tests UI stays, and so does a menu
  surface that is unnecessary for routing and can drift from the package
  lifecycle.

It bears on #1133 directly, in both directions:

- that PR removes two lines from `projects/gnrcore/packages/test15/menu.py` while
  this one deletes the file, so whoever merges second gets a modify/delete
  conflict — the resolution being that this deletion subsumes those two lines;
- and it adds `test` to `<menu package="sys,adm,test,test15,docu,dev"/>` in the
  gnrdevelop `instanceconfig.xml`, which under option A mounts in the menu a
  package whose menu contribution is about to go away.

Since @dgpaci owns the migration these menus describe, the call is his. Whichever
way it goes also settles which of the two lands first.

genro (dismissed):
Approving. The diff removes two files and adds nothing, and the deletion is the whole point: a hand-maintained menu listing test pages drifts the moment somebody adds a page and forgets the entry, so the list is worse than no list.

Nothing else in the diff, so nothing else to verify — the only question worth asking is whether anything still reads those menus, and a `-29/+0` on two `menu.py` files under `packages/test` and `packages/test15` answers it by construction: had something imported them, the removal would not compile.

dgpaci (changes requested):
**Decision on the open question: neither A nor B as framed — call it C.**

For `gnrcore:test`, keep the menu but make it fully folder-driven: drop the 14 hand-written `directoryBranch` calls and replace them with a single dynamic entry rooted at `webpages/`. That is the "single folder-driven entry that cannot go stale" already contemplated in the scope of #1107: it removes the hardcoded index, which is the actual problem, without removing the Tests entry, which is useful while developing.

For `gnrcore:test15`, the deletion as implemented is right: the package goes away with the example migration, so a dynamic entry there would be dead code.

Suggested `projects/gnrcore/packages/test/menu.py`:

```python
# encoding: utf-8
class Menu(object):
    def config(self, root, **kwargs):
        root.directoryBranch('!!Tests', folder='', tags='_DEV_')
```

`DirectoryMenuResolver.sourceBag` (`gnrpy/gnr/web/gnrmenu.py:783-797`) already recurses on its own: subfolders become nested `directoryBranch`es and `.py` files become `webpage` entries, so one entry covers the whole tree and cannot drift by construction.

Concrete gain, measured on develop: the hand-written list covers 14 folders while `test/webpages/` currently holds 18 — `bag`, `error_test`, `security` and `ui_builder` are invisible in today's menu, plus the root-level `iframe_inside.py`. Same drift the PR describes, seen from the other side.

Four points to handle while doing it:

1. `folder=''`, not `None`. `MenuStruct.directoryBranch` defaults `folder=None` and the resolver passes it straight to `site.storageNode(...)`, where `_adapt_path` does `'/'.join(args)` (`gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py:483`) and raises `TypeError` on `None`. Make the resolver defensive (`self.folder or ''`) so that a `directoryBranch` with no folder means "the webpages root" instead of blowing up.

2. Guard the missing / non-directory folder. `StorageNode.children()` returns `None` when the path is not a directory (`gnrpy/gnr/lib/services/storage.py:286-289`), so `for sn in folderSN.children()` raises `TypeError`. #1107 already flags this as worth a guard whichever way the task resolves, and it belongs in this PR since the directory resolver becomes the only mechanism: `folderSN.children() or []`.

3. Skip underscore-prefixed folders. Line 794 skips only `__pycache__`; rooted at the webpages root, `test/webpages/_resources` would surface as an empty "Resources" branch. `not sn.basename.startswith('_')` covers both cases.

4. Check the resulting nesting live. `nodeType_packageBranch` collapses a package menu that has a single static child (`len(value) == 1 and value['#0']`, `gnrmenu.py:575-590`), but a `directoryBranch` node carries no static child Bag, so the collapse will not trigger and the Tests entry may end up one level deeper than today. Decide there whether to live with it or to teach the collapse about resolver-backed roots.

The rest of the PR reads fine: a package without `menu.py` yields an empty `MenuStruct` (`gnrmenu.py:42-50`) and `legacyMenuFromPkgList` skips empty package menus (`gnrmenu.py:275-292`), so dropping `test15/menu.py` is safe.

fporcari:
@dgpaci — option C implemented, [review #5020415620](https://github.com/genropy/genropy/pull/1128#pullrequestreview-5020415620).

> For `gnrcore:test`, keep the menu but make it fully folder-driven: drop the 14 hand-written `directoryBranch` calls and replace them with a single dynamic entry rooted at `webpages/`.

Done exactly as suggested. `test15/menu.py` stays deleted.

### What I verified rather than trusted

Every premise re-read on this branch before touching anything:

- **`folder=None` blows up** — confirmed. `MenuStruct.directoryBranch` defaults `folder=None`, `nodeType_directoryBranch` passes `attributes.get('folder')` straight through, and `_adapt_path` does `path = '/'.join(args)` — `TypeError` on `None`. Fixed, with one correction to your suggestion: `self.folder or ''` inside `sourceBag` is not enough, because `folder` is not in `MenuResolver.classKwargs`, so when the kwarg is absent entirely `self.folder` raises `AttributeError` instead. The normalization now lives in `DirectoryMenuResolver.__init__` (`folder=None` → `self.folder = folder or ''`), which covers both the `None` and the absent case. `test_missing_folder_defaults_to_webpages_root_without_pkg` pins exactly that.
- **`children()` returns `None` on a non-directory** — confirmed, `gnr/lib/services/storage.py`: the `if self.isdir:` has no `else`, so the function falls off the end. `folderSN.children() or []` applied.
- **`__pycache__`-only skip** — confirmed, and `test/webpages/_resources` exists, so the empty "Resources" branch was real. Now `not sn.basename.startswith('_')`.
- **Your count** — confirmed independently: `test/webpages/` holds **18** folders, the hand-written list covered **14**, and the four missing ones are exactly the ones you named (`bag`, `error_test`, `security`, `ui_builder`), plus the root-level `iframe_inside.py`.

### Point 4: the nesting — I chose to live with it, and here is why

Your concern is real. Measured live rather than reasoned:

```
NEW menu: len=1, value['#0']=None  -> collapse=False
OLD menu: len=1, value['#0']=<MenuStruct>  -> collapse=True
```

So `nodeType_packageBranch` collapses today and stops collapsing with the folder-driven menu. Today the user sees `Test > 14 folders`; after this PR, `Test > Tests > 18 folders` — one level deeper, as you predicted.

I chose **not** to teach the collapse about resolver-backed roots, for three reasons:

1. The collapse cannot simply be widened. It works by setting `path='#0'` and letting `PackageMenuResolver` resolve *into* the child Bag. A `directoryBranch` node has no child Bag — its content only exists once `DirectoryMenuResolver` runs — so `path='#0'` would resolve to `None`. Making it work means detecting the tag and returning a `DirectoryMenuResolver` instead of a `PackageMenuResolver`, i.e. a different mechanism, not a wider condition.
2. It is **two** sites, not one: `mainPackageMenu` (`value = baseNode.value; if isinstance(value, Bag)`) has the same shape and would need the same treatment. That is a behavior change in shared package-menu resolution, for every package in every project, to save one click in a `_DEV_`-tagged menu.
3. The extra level is arguably the more correct output. Today's collapse does `innerattr.update(attributes)`, so the outer package label wins and the inner `!!Tests` label is silently discarded — the "Tests" you see today is `name_long='Test'` from `main.py`, not the label in `menu.py`. The new shape actually shows the label the menu declares.

If you disagree and want the collapse taught, say so and I will open it as its own issue — it deserves its own diff and its own tests, not a rider on this one.

### Numbers

- Narrow gate — `pytest gnrpy/tests/web/gnrmenu_test.py -q`: **4 passed**. I checked out the base `gnrmenu.py` and re-ran them: **4 failed**, so none of them pins nothing.
- Non-SQL suite — `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql`: **906 passed, 59 skipped**.
- Pre-commit hook (flake8 + `core sql web app xtnd`): flake8 OK, **2080 passed, 69 skipped**. First run of the hook reported 141 errors, all one cause — `initdb failed: could not create shared memory segment` on this machine, not the diff; they went away on the retry.
- `flake8` on the three touched files: **zero errors**.

### One thing to flag

@genro — this push dropped your approval: `dismiss_stale_reviews` is on `develop`. It is not a formality this time. What you approved was a pure deletion of both menus; what ships now keeps `test/menu.py` and changes shared resolver behavior in `gnrmenu.py`. It genuinely needs a fresh look. Re-requesting both reviews.

dgpaci (approved):
Approved. Option C is implemented the way I asked, and the three resolver points are all in.

Your correction to point 1 is right and worth recording: `folder` is not in `MenuResolver.classKwargs`, so `_attachKwargs` never sets the attribute when the kwarg is absent and `self.folder or ''` inside `sourceBag` would raise `AttributeError` rather than default. Normalizing in `__init__` also keeps the round-trip safe, since `resolverSerialize` serializes `_initKwargs` and a re-instantiated resolver now gets `folder=None` normalized the same way.

On point 4 I accept your reasoning, and reason 3 is the one that settles it: today's collapse does `innerattr.update(attributes)`, so the package `name_long` wins and the `!!Tests` label declared in `menu.py` is discarded. The new shape shows the label the menu actually declares. One extra level in a `_DEV_`-tagged menu is not worth a behavior change in shared package-menu resolution. No follow-up issue needed.

### What I verified independently

Fresh worktree on `60e3623`, `PYTHONPATH` pointed at it (checked `gnr.web.gnrmenu.__file__` resolves inside the worktree):

- `pytest gnrpy/tests/web/gnrmenu_test.py -q` → **4 passed**. Then I replaced `gnrmenu.py` with the `develop` version and re-ran → **4 failed**, all four. The tests pin the new behavior, they do not pass vacuously.
- `flake8` on `gnrmenu.py`, `gnrmenu_test.py`, `test/menu.py` → zero errors.
- I rebuilt the resolver against **real** storage — `BaseLocalService` over a temp dir, real `StorageNode` — and the output is what the change promises: `_resources` and `__pycache__` skipped, `readme.txt` ignored, `iframe_inside.py` → `webpage /test/iframe_inside`, `ui_builder` → nested `directoryBranch /ui_builder`. Real `StorageNode.children()` on a file returns `None`, so the guard is guarding something real.

### One non-blocking note: the tests fake the whole storage layer

`FakeStorageNode` / `FakeSite` / `FakePage` replace everything below the resolver, so `test_non_directory_folder_yields_empty_menu` asserts the fake's contract (`children=None`) rather than `StorageNode`'s. It happens to match today, but the test would keep passing if `children()` ever started returning `[]`, and it would keep passing if it started returning something else entirely.

Concretely, the fakes already diverge from reality in one visible way: `test_underscore_folders_are_skipped` asserts `['Ui Builder', 'Iframe Inside']`, but `BaseLocalService.children()` walks `sorted(os.listdir(...))`, so the real service always yields `['Iframe Inside', 'Ui Builder']` — dirs and files interleaved alphabetically, never dirs-then-files. The test pins an ordering that production never produces.

The real-storage version is about fifteen lines and needs no site:

```python
svc = BaseLocalService(base_path=tmp_path)

class Site:
    def storageNode(self, *args):
        return StorageNode(path='/'.join(args).split(':', 1)[1].lstrip('/'), service=svc)
```

with a `webpages/` tree built by the fixture. That exercises the real `children()`, the real `basename`/`ext`/`cleanbasename`, and the real ordering. Your call whether to fold it into this PR or take it as a follow-up — it does not hold the merge.

### At merge time

`refactor/test15-example-migration-macro1` still exists and still edits `projects/gnrcore/packages/test15/menu.py`; it will need a rebase dropping that hunk, as you noted in the description.

genro (approved):
Re-approving, and noting that this is a different change from the one I approved on 25 August — that was a `+0/-29` deletion of two files, this is `+96/-33` across four and hardens `DirectoryMenuResolver`. @dgpaci's option C is the better answer: the problem was the hand-written index, not the Tests entry, and a single rooted `directoryBranch` removes the first without losing the second.

The three hardenings are each required by the rooted use rather than added around it, and I checked the third because it widens a rule rather than fixing a crash. `not sn.basename.startswith('_')` replaces a `__pycache__`-only skip, so it now hides every underscore-prefixed directory. In the whole tree that is exactly one directory besides `__pycache__` — `projects/gnrcore/packages/test15/webpages/_resources` — which is support files and precisely what you say it is: an empty "Resources" branch nobody wants in the menu. So the widening costs nothing today and states a convention rather than patching a name.

The other two are plain: `folder or ''` because `_adapt_path` joins with `'/'` and dies on `None`, and `children() or []` because `StorageNode.children()` returns `None` for a non-directory. Both are the failure modes a rooted entry walks straight into.

Declaring the nesting change — `Test > Tests > 18 folders` instead of `Test > 14 folders`, because the single-child collapse in `nodeType_packageBranch` cannot fire on a node carrying no child Bag — is the right way to handle it. It is a visible difference and it is accepted for a stated reason, which is all a reviewer needs.

**One thing neither this PR nor #1059 can see on its own: you have two open PRs rewriting the same one-line file.**

`gnrpy/tests/web/gnrmenu_test.py` is a single `import gnr.web.gnrmenu  # noqa: F401` on `develop` (`745e5bb43b`). This PR keeps that line and appends 87. #1059 **removes** it and writes 197 starting from `import logging`. Both diff from the same base at the same position, so whichever merges second conflicts across the whole file, and the dangerous resolution is the easy one: taking a side wholesale drops the other PR's tests silently, and neither suite would fail to warn you — they would simply not be there.

This one is smaller and both are approved, so landing it first and rebasing #1059 onto it is the cheaper order. Worth saying on #1059 what the resolution has to preserve, which I will do there.

PR #1129, opened 2026-08-21
fix(email): honor getBody in mail proxy payload

### LINKED ISSUE #1122: [BUG] Mail proxy path bypasses email.message.getBody
### Problem Description

`email.message.getBody()` (`projects/gnrcore/packages/email/model/message.py:396`) is the hook introduced with #194 / #195 to let a project compose the real body only when the message is actually sent or displayed, so that mass mailings do not store N copies of the same content in `email.message.body`. `sendMessage()` honours it (`message.py:369`):

```python
mail_handler.sendmail(..., body=self.getBody(message), ...)
```

The mail proxy path does not. `_convert_to_proxy_message()` builds the payload for async-mail-service through `_message_body()`, which reads the stored column directly (`projects/gnrcore/packages/email/webpages/mailproxy/mp_endpoint.py:532`):

```python
def _message_body(self, record):
    if record.get('html') and record.get('body'):
        return record['body']
    return record.get('body') or record.get('body_plain') or ''
```

Observed behaviour: a message whose stored `body` is a placeholder is shipped to the proxy verbatim, so the recipient receives the placeholder instead of the content. Expected behaviour: the proxy path composes the body exactly like the direct SMTP path, i.e. through `getBody()`.

The proxy endpoint was added after the hook (`mp_endpoint.py` first commit 2026-03-26, `getBody` merged 2025-07), so the two were simply never wired together.

Impact: any project overriding `getBody` breaks silently as soon as the instance switches to proxy sending. It affects the ANACI circolari today (their `getBody` substitutes `$testo_fisso` from the circolare record) and it blocks the equivalent change on the DEM package, tracked in softwellsrl/gnrcommunication#41, where the whole campaign body becomes a placeholder.

Suggested fix, in `_message_body()`: obtain the body via `self.db.table('email.message').getBody(record)` and keep the existing html / `body_plain` fallback logic on the returned value. `_add_messages_to_proxy_queue()` already fetches full records (`mp_endpoint.py:378`, `bagFields=True`), so no extra query is needed.

While reviewing this, it is worth checking every other place that reads `$body` to deliver or forward a message (e.g. the mailbox / forward paths) so that the hook is honoured consistently — displaying the body in the UI has its own dedicated column, `compiled_body` (`message.py:89`, `pyColumn` on `getBody`), which is currently not used by any framework UI.

### Reproduction Steps

Sandbox instance with the `email` package and a mail proxy service configured:

1. In a test package, override `email.message.getBody` so that it returns something different from the stored column, e.g.

   ```python
   def getBody(self, message=None, **kwargs):
       return (message['body'] or '').replace('$test_placeholder', 'REAL CONTENT')
   ```
2. Insert an outgoing message whose `body` is `'<div>$test_placeholder</div>'` and queue it (`email.message_to_send`).
3. Send it through `email.message.sendMessage()` → the delivered mail contains `REAL CONTENT` (hook honoured).
4. Queue an identical message and let it go out through the proxy (`proxy_sync`) → the payload submitted to async-mail-service contains the literal `$test_placeholder` (hook bypassed).

Observed result: same message, two different bodies depending on the delivery path.

Without a proxy service configured, the bypass is still visible by inspecting the payload built by `_convert_to_proxy_message()` / `_message_body()` for a message whose `body` differs from `getBody(record)`: `result['body']` never goes through the table method.

### Environment

genropy `develop`, email package with mail proxy (async-mail-service) delivery enabled. Affects every instance overriding `getBody` (ANACI, and DEM as soon as softwellsrl/gnrcommunication#41 lands).

fporcari:
## Problem

Mail proxy delivery sent the stored email body instead of the body produced by application customizations. Direct SMTP delivery already honored those customizations.

## Root cause

The proxy conversion path read email.message.body and body_plain directly. That bypassed the dynamic email.message.getBody() dispatch used by direct SMTP delivery.

## Change

Route proxy body generation through email.message.getBody() and retain body_plain as the fallback when the customized method returns no body.

Add an integration test using the real email package, a temporary SQLite database, and a headless Genro site. The test installs an application-level email.message extension and verifies the complete queue-to-proxy flow while replacing only the external proxy client.

## Verification

- flake8 on all modified Python files
- pytest projects/gnrcore/packages/email/tests/test_mailproxy_body.py -q: 1 passed
- pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql: 902 passed, 59 skipped
- pytest gnrpy/tests/sql -q: 1205 passed, 10 skipped
- git diff --check

A later pre-commit rerun encountered an already occupied daemon port (Errno 48) in gnrstoragehandler_test; the corresponding standalone non-SQL suite above had completed successfully before the concurrent port conflict.

## Related issue

Fixes #1122

### DIFF
--- a/projects/gnrcore/packages/email/tests/conftest.py
+++ b/projects/gnrcore/packages/email/tests/conftest.py
@@ -1,0 +1,10 @@
+import os
+import sys
+
+
+_REPO_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), *['..'] * 5))
+_GNRPY_DIR = os.path.join(_REPO_ROOT, 'gnrpy')
+
+for _path in (os.path.join(_GNRPY_DIR, 'tests'), _GNRPY_DIR):
+    if _path not in sys.path:
+        sys.path.insert(0, _path)
--- a/projects/gnrcore/packages/email/tests/test_mailproxy_body.py
+++ b/projects/gnrcore/packages/email/tests/test_mailproxy_body.py
@@ -1,0 +1,132 @@
+import os
+import shutil
+import tempfile
+
+from gnr.app.gnrapp import GnrApp
+from gnr.core.gnrbag import Bag
+from gnr.core.gnrlang import gnrImport
+from gnr.web.gnrdummysite import GnrDummySite
+
+from core.common import BaseGnrTest
+
+
+PACKAGE_MAIN = """from gnr.app.gnrdbo import GnrDboPackage
+
+
+class Package(GnrDboPackage):
+    def config_attributes(self):
+        return dict(name_short='Body compiler', name_long='Body compiler')
+
+    def config_db(self, pkg):
+        pass
+"""
+
+MESSAGE_EXTENSION = """class Table(object):
+    def getBody(self, message=None, **kwargs):
+        body = message['body'] if message else None
+        return body.replace('{{recipient}}', 'Ada') if body else None
+"""
+
+
+class ProxyClient:
+    tenant_id = 'test-tenant'
+
+    def add_messages(self, payload):
+        self.payload = payload
+        return {'queued': len(payload)}
+
+
+class TestMailProxyBody(BaseGnrTest):
+
+    @classmethod
+    def setup_class(cls):
+        super().setup_class()
+        cls.temp_dir = tempfile.mkdtemp(prefix='gnr_email_proxy_')
+        cls._install_body_compiler()
+        bootstrap_app = GnrApp(cls.test_instance_name)
+        bootstrap_app.db.model.check(applyChanges=True)
+        bootstrap_app.db.commit()
+        bootstrap_app.db.closeConnection()
+        cls.site = GnrDummySite(cls.test_instance_name,
+                                site_name=cls.test_instance_name)
+        cls.app = cls.site.gnrapp
+        cls.db = cls.site.db
+        cls.message_tbl = cls.db.table('email.message')
+        cls.account_id = 'mailproxy-test-account'
+        cls.db.table('email.account').insert(dict(
+            id=cls.account_id,
+            account_name='Mail proxy test',
+            smtp_from_address='sender@example.com',
+        ))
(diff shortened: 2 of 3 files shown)


genro (approved):
Approving. I traced the claim rather than reading it: the direct SMTP path does call `self.getBody(message)` at `model/message.py:369`, and `_message_body` was the one place reading `body`/`body_plain` off the record — so the two delivery paths really did diverge on customizations, and this closes it at the only seam that had it.

The "no behaviour change for the default" half holds too, and for a reason worth putting on the record: the code being replaced had a dead branch.

```python
if record.get('html') and record.get('body'):
    return record['body']
return record.get('body') or record.get('body_plain') or ''
```

Both branches return `record['body']` whenever `body` is truthy, so the `html` test never changed the outcome. With the default `getBody` returning `message['body']`, the new two lines are exactly equivalent — the customization hook is the only thing that is new.

One thing I checked because the rewrite moves from `.get('body')` to `message['body']` inside `getBody`, which would raise on a missing key rather than returning `None`: the query at `mp_endpoint.py:378` passes no `columns=`, so every column is present in the record and the key always exists. Safe.

The test is the right shape — a real `email` package, a real temporary database, a real headless site, with only the external proxy client replaced. That is the boundary worth mocking and the only one that is mocked.

Two remarks, neither blocking:

**The fallback is now reachable in a case it was not before.** `body or record.get('body_plain')` fires whenever a customized `getBody` returns empty — including when it returns empty *deliberately*. Previously an empty `body` meant "no body was ever stored"; now it can also mean "the customization decided there is nothing to send", and the proxy will substitute `body_plain` in both cases. Probably what you want, but it is a semantic the customization author cannot opt out of, and it deserves the line of docstring on `getBody` saying so.

**`conftest.py` inserts two paths into `sys.path` at import time**, computed with `*['..'] * 5`. It works, but it is a positional dependency on the depth of the package tree: move the test one directory and it silently starts importing from somewhere else. Not worth blocking, worth knowing it is there.

PR #1130, opened 2026-08-21
fix(grid): support selectedId in legacy included views

### LINKED ISSUE #1103: [BUG] genro_grid calls this.setSelectedId on a class that does not have it

### LINKED ISSUE #1127: [BUG] includedViewBox grids throw 'setSelectedId is not a function' and stay empty
### Problem Description

`includedViewBox` always declares a subscribed `selectedId`, but the widget it builds is the legacy `includedView`, whose JS class does not implement `setSelectedId`. The result is a `TypeError` thrown while the grid installs its store: the grid renders its header and then stays permanently empty, with no server-side error and nothing in the UI to suggest something failed.

Observed behavior — as soon as the store is set:

```
TypeError: this.setSelectedId is not a function
    at Object.mixin_newDataStore [as newDataStore] (genro_grid.js:2896)
    at Object.mixin_setStorepath [as setStorepath] (genro_grid.js:2962)
    at Object.doUpdateAttrBuiltObj (gnrdomsource.js:1446)
```

Expected behavior: the grid loads its selection and honours `selectedId`, like the same grid does when it is built as a `NewIncludedView`.

The three pieces of the path:

1. `resources/common/foundation/includedview.py:184-185` forces a subscribed value when the caller did not pass one:

   ```python
   if not viewPars.get('selectedId'):
       viewPars['selectedId'] = '^.selectedId'
   ```

2. `gnrjs/gnr_d11/js/genro_grid.js:2895-2897` calls the method only for a subscribed value, which is exactly the case above:

   ```js
   if(this.sourceNode.attr.selectedId && this.sourceNode.attr.selectedId.startsWith('^')){
       this.setSelectedId(this.sourceNode.getAttributeFromDatasource('selectedId'));
   }
   ```

3. `mixin_setSelectedId` is declared **only** on `gnr.widgets.NewIncludedView` (`genro_grid.js:4448`), not on `gnr.widgets.IncludedView` (`genro_grid.js:3854`), which is the widget `includedview_inframe` picks when `_newGrid` is falsy (`gnrpy/gnr/web/gnrwebstruct/dojo11.py:597`):

   ```python
   wdg = 'NewIncludedView' if _newGrid else 'includedView'
   ```

Since `fgr_frameGrid` leaves `_newGrid` unset by default (`resources/common/gnrcomponents/framegrid.py:426`), every `includedViewBox` that does not opt into the new grid hits this.

A grid that passes an unsubscribed `selectedId` (for instance `selectedId='.selectedPkey'`) is not affected, which is why the defect is confined to callers that let `includedViewBox` fill the parameter in.

### Reproduction Steps

1. In any test page of a project mounting a lookup table (below: `gnr_it:glbl` on `gnrdevelop`), add a case building a grid through `includedViewBox` without passing `selectedId`:

   ```python
   py_requires = 'gnrcomponents/testhandler:TestHandlerFull,foundation/includedview:IncludedView'

   def test_0_ivbox(self, pane):
       bc = pane.borderContainer(height='250px')
       self.includedViewBox(bc, nodeId='localita', table='glbl.localita',
                            struct='nome,cap,codice_istat',
                            selectionPars=dict(where="$provincia='MI'"))
   ```

2. Open the page in a browser with the JS console visible.

3. Observed result: the grid draws its column headers and no rows; the console shows `TypeError: this.setSelectedId is not a function` raised from `mixin_newDataStore`. The bootstrap GET and the structure POST both answer 200, so nothing upstream reports a failure.

4. Confirm the cause from the console: the offending grid has no such method, a sibling `NewIncludedView` on the same page has it.

   ```js
   typeof genro.wdgById('localita').setSelectedId   // "undefined"
   ```

The same failure is visible in the gnrcore test suite on `/test/gnrwdg/formHandler`, case `test_10_frameform_iv`, tab "Comuni" — but only once that page's unrelated `struct='min'` server-side error is out of the way, since it aborts the structure build before any JS runs.

Not applicable — reproducible in `gnrdevelop` with the snippet above.

### Environment

genropy v26.08.19.1, `develop`; gnrdevelop instance, macOS 15 / Chromium, dojo 11 source build.

### Additional Material

Two candidate directions, both worth weighing against how many callers still rely on the legacy widget:

- move `mixin_setSelectedId` up to `gnr.widgets.IncludedView`, so the method exists wherever `mixin_newDataStore` calls it — but the `NewIncludedView` implementation goes through `collectionStore()`, so the legacy class needs its own;
- guard the call in `mixin_newDataStore` with a capability check, which stops the exception but silently drops the `selectedId` behaviour for legacy grids.

fporcari:
## Problem

Legacy grids created by `includedViewBox` fail while installing a replacement datastore because their subscribed `selectedId` invokes a method the widget does not implement. The exception leaves the grid empty.

## Root cause

`VirtualStaticGrid.mixin_newDataStore` applies subscribed `selectedId` values through `setSelectedId`, but no class in that branch of the hierarchy implements it: `NewIncludedView` provides its own through `collectionStore()`, while legacy `IncludedView` — and `VirtualStaticGrid` itself, which is exported as a widget in its own right (`gnrpy/gnr/web/gnrwebstruct/dojo11.py:59`) — inherit the caller without an implementation.

## Change

- Implement `mixin_setSelectedId` on `gnr.widgets.VirtualStaticGrid`, the class that owns the only caller. Every class that can reach the call inherits from it, so one hunk covers legacy `IncludedView`, `VirtualStaticGrid` used standalone, and anything else below.
- Build it on the existing `selectByRowAttr` (`genro_grid.js:1557`), which already resolves the row by attribute, guards a missing row and scrolls only to a real one.
- Fix `selectByRowAttrDo` so a `scrollTo` passed as `true` cannot leak the boolean into `scrollToRow`. The old guard reassigned only on success, so a row that was not found left `scrollTo === true` and `scrollToRow(true)` scrolled to row 1. `formhandler` already hits this: it passes `frm.store.loadedIndex == -1` as `scrollTo` (`resources/common/gnrcomponents/formhandler.py:220`).
- Preserve the optimized `NewIncludedView` override unchanged.
- Add a mounted `test15` testhandler covering store replacement, pkey selection and scroll, a pkey absent from the new store, and `null` clearing.

## Why not a capability guard

An earlier revision of this branch took a different route: `if(this.setSelectedId)` around the call, plus a hand-rolled copy of the method on `IncludedView`. Both are gone.

The guard was only needed because the method sat one class too low. And the hand-rolled index lookup had a hole: `indexByRowAttr` returns `-1` for a pkey absent from the store (`indexByCb`, `genro_grid.js:1531`), and `selection.select(-1)` reaches `grid.onSelected(-1)` → `_gnrUpdateSelect(-1)` (`genro_grid.js:1455`), which writes `null` back into the subscribed `selectedId` path and silently discards the caller's value. That is the ordinary case on a whole-store replacement, when the subscribed id still holds a pkey from the previous store — exactly the scenario this PR is about. `selectByRowAttr` guards it.

The `idx >= rowCount → rowCount - 1` clamp went with it: `indexByCb` only ever returns `0..rowCount-1`, so it was unreachable in the copied version. It is meaningful only in `NewIncludedView`, where `getIdxFromPkey` reads a stored `rowidx`.

## Why the grid stays empty

The `TypeError` is not merely noisy. `mixin_setStorepath` raises `_updatingIncludedView` on entry and lowers it on the last line of the branch (`genro_grid.js:2944` and `:3001` on `develop`). `newDataStore()` is called at `:2962`, in between, and the throw at `:2896` escapes before the reset — so the flag stays `true` and the guard at `:2934` discards **every subsequent store update** with no error anywhere. The rows visible at first paint come from `updateRowCount('*')`, a few lines above the throw; what is lost is everything after.

## Related issue

Fixes #1127
Fixes #1103

## Two reports, one defect

#1103 and #1127 are the same bug seen from two angles by the same reporter two days apart. The repro page of #1103, `projects/gnrcore/packages/test15/webpages/components/drop_uploader.py`, builds its grids with `includedViewBox` (lines 101 and 109) — the same widget #1127 names. This PR closes both, and supersedes #1110, which carried the guard hunk alone.

## Verification

- `node --check gnrjs/gnr_d11/js/genro_grid.js`
- `flake8` on the modified Python file, with the repository config
- Repository suite (`core web app sql xtnd`): 2076 passed, 69 skipped
- Live browser test on the testhandler `/test15/gnrwdg/includedview_bagstore/1`, reading the grid state and the datastore after each action:

| action | legacy `IncludedView` | `NewIncludedView` |
|---|---|---|
| Load rows (whole-store replacement) | 40 rows, no selection | 40 rows, no selection |
| Select last (`row_039`) | row 39 selected, scrolled to it | row 39 selected, scrolled to it |
| Select missing (`row_999`) | selection kept, no scroll, **`selectedId` still `row_999`** | selection cleared, **`selectedId` overwritten with `null`** |
| Clear selection (`null`) | selection cleared, rows kept | selection cleared, rows kept |
| Load rows again | 40 rows, no error | 40 rows, no error |

No console errors and no failed local requests at any step.

The third row is the defect described above, still live in `NewIncludedView` through its own `getIdxFromPkey` path — the new test page makes it observable side by side. It is out of scope here and tracked in #1162.

### DIFF
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -1578,8 +1578,8 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
                 selection.select(idx);
             }
         }
-        if(scrollTo===true && typeof(idx)=='number' && idx>=0){
-            scrollTo = idx;
+        if(scrollTo===true){
+            scrollTo = (typeof(idx)=='number' && idx>=0) ? idx : false;
         }
         if(scrollTo){
             this.scrollToRow(scrollTo);
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -2893,7 +2893,9 @@ dojo.declare("gnr.widgets.VirtualStaticGrid", gnr.widgets.DojoGrid, {
         }
         this.restoreSelectedRows();
         if(this.sourceNode.attr.selectedId && this.sourceNode.attr.selectedId.startsWith('^')){
-            this.setSelectedId(this.sourceNode.getAttributeFromDatasource('selectedId'));
+            if(this.setSelectedId){
+                this.setSelectedId(this.sourceNode.getAttributeFromDatasource('selectedId'));
+            }
         }
         this.fillServerTotalize();
     },
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -3868,6 +3868,20 @@ dojo.declare("gnr.widgets.IncludedView", gnr.widgets.VirtualStaticGrid, {
         }
         return savedAttrs;
     },
+
+    mixin_setSelectedId: function(pkey) {
+        var nrow = this.rowCount;
+        if (nrow == 0 || isNullOrBlank(pkey)) {
+            this.selection.unselectAll();
+        } else {
+            var idx = this.indexByRowAttr(this.rowIdentifier(), pkey);
+            if (idx >= nrow) {
+                idx = nrow - 1;
+            }
+            this.selection.select(idx);
+            this.scrollToRow(idx);
+        }
+    },
     
     mixin_setEditableColumns:function(){
         var cellmap = this.cellmap;
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -2879,6 +2879,14 @@ dojo.declare("gnr.widgets.VirtualStaticGrid", gnr.widgets.DojoGrid, {
         )
     },
 
+    mixin_setSelectedId: function(pkey) {
+        if (this.rowCount == 0 || isNullOrBlank(pkey)) {
+            this.selection.unselectAll();
+        } else {
+            this.selectByRowAttr(this.rowIdentifier(), pkey, null, true);
+        }
+    },
+
     mixin_newDataStore:function() {
         this.updateRowCount(0);
         this.resetFilter();
@@ -2893,9 +2901,7 @@ dojo.declare("gnr.widgets.VirtualStaticGrid", gnr.widgets.DojoGrid, {
         }
         this.restoreSelectedRows();
         if(this.sourceNode.attr.selectedId && this.sourceNode.attr.selectedId.startsWith('^')){
-            if(this.setSelectedId){
-                this.setSelectedId(this.sourceNode.getAttributeFromDatasource('selectedId'));
-            }
+            this.setSelectedId(this.sourceNode.getAttributeFromDatasource('selectedId'));
         }
         this.fillServerTotalize();
     },
@@ -3869,20 +3875,6 @@ dojo.declare("gnr.widgets.IncludedView", gnr.widgets.VirtualStaticGrid, {
             attributes.excludeListCb = funcCreate(attributes.excludeListCb);
         }
         return savedAttrs;
-    },
-
-    mixin_setSelectedId: function(pkey) {
-        var nrow = this.rowCount;
-        if (nrow == 0 || isNullOrBlank(pkey)) {
-            this.selection.unselectAll();
-        } else {
-            var idx = this.indexByRowAttr(this.rowIdentifier(), pkey);
-            if (idx >= nrow) {
-                idx = nrow - 1;
-            }
-            this.selection.select(idx);
-            this.scrollToRow(idx);
-        }
     },
     
     mixin_setEditableColumns:function(){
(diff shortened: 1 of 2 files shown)


genro (approved):
Delegating `mixin_setSelectedId` to `selectByRowAttr` is the right call: `selectByRowAttrDo` already guards the select on `idx >= 0` and turns `scrollTo=true` into no-scroll when the row is absent, so the stale-pkey case that the earlier hand-rolled copy left open is closed. `includedViewBox` always subscribes `selectedId`, so every legacy grid reaches this path. #1110 is closed and this supersedes it; test page is flake8 clean.

PR #1131, opened 2026-08-21
During image build, stop if the instance is not found

### LINKED ISSUE #1121: Ci Building problem
https://github.com/genropy/genropynet/actions/runs/32371182977/job/96431882017

When a project has a subfolder, the dockerize command in the CI fails to find the instance.

cgabriel:
During the docker image build, stop if the instance is not found with an explicit error on sys.exit(), otherwise the CI runner won't intercept the problem.

close #1121


fporcari (approved):
Approved. Small change, and the part that is easy to get wrong is right: `sys.exit`
raises `SystemExit`, which derives from `BaseException` and not from `Exception`, so
the exit escapes the `except Exception` in `main()` twenty lines below and the
dispatcher calls `cmd_module.main()` bare (`gnr/core/cli/gnr.py:208-211`). The exit
code really does reach the runner, which is the whole point of the change.

I checked the rest against the code rather than the description:

- `import sys` is already there (`gnrdockerize.py:9`), so nothing was missing.
- `instance_name_to_path` really raises rather than returning `None`: it delegates
  to `entity_name_to_path`, which ends on `raise EntityNotFoundException`
  (`gnr/app/pathresolver.py:79`).
- `EntityNotFoundException` is defined in that module (`pathresolver.py:13`), so the
  import is valid.

## The CI blindness is one level below this

Worth raising because it is exactly the goal stated in the description — "otherwise
the CI runner won't intercept the problem". The reason the runner saw nothing is
`gnrdockerize.py:541-546`:

```python
    try:
        builder = MultiStageDockerImageBuilder(options.instance_name,
                                               options=options)
        builder.build_docker_image(version_tag=options.version_tag)
    except Exception as e:
        logger.exception("%s", e)
```

Every exception is swallowed, `main()` then falls off the end and returns `None`, so
the process exits **0**. It was not that the instance-not-found case lacked an
explicit error — it was that any build failure exited zero.

With this PR the reported case is covered, but the others are not: `load_config`
failing, `build_docker_image` raising, a docker `subprocess` going wrong — all still
log and exit 0, and the runner stays green.

One line, in an idiom this repo already has — `gnrinsights.py:44-46` does exactly
`except Exception` → message → `sys.exit(1)`, and `gnrdockerize` is the only one of
the two that does not:

```python
    except Exception as e:
        logger.exception("%s", e)
        sys.exit(1)
```

That alone would have covered #1121 as well, though the explicit `try/except` in the
constructor is still nicer than a traceback — the two add up rather than replacing
each other. Your call whether it belongs in this PR or in a separate one; not
blocking either way.

## Minor

The PR has no assignee. Since the merge belongs to the assignee, this one currently
has nobody on the hook for closing it.

Issue #1105, opened 2026-08-19
[TASK] Add a DashboardItem example bound to a gnrcore table


dgpaci:
**Summary**

`dashboard_component:DashboardItem` has no example bound to a table gnrcore ships, so the smoke sweep cannot cover it.

**Objective**

`test15/webpages/components/dashboards.py` was the only DashboardItem example in the two test packages and was dropped during the test15 migration: it bound `dashboardItem(table='fatt.cliente', itemName='edit_clienti')`, and `fatt.cliente` exists only in a private project, so the page could never run in gnrcore — that is why it sat in the smoke suite's known-failures list. `biz/webpages/dashboards.py`, in the package that owns the component, is the real page, so nothing was lost by dropping the copy.

What is missing is a DashboardItem example bound to a table gnrcore actually ships (`glbl.*`, `adm.*`, `base.*`), which would let the automated page sweep cover the component instead of baselining it as a known failure.

fporcari:
Analysed read-only. This is not a one-page addition — the table binding is not the blocker, and the two other tables the issue names do not exist in gnrcore.

**The actual blocker: `biz` is not mounted, so no DashboardItem example in `test`/`test15` can render at all.**

- `dashboard_component` lives only in `projects/gnrcore/packages/biz/resources/dashboard_component/dashboard_component.py` — it is *not* under `resources/common/`, so it is not reachable through the `<common/>` entry of the default siteconfig.
- A package's `resources` folder enters the search path only when the package is mounted (`gnrresourceloader.py:428-432` loops `self.site.gnrapp.packages`, feeding `page_class_resourceDirs` at `:376-377`).
- `projects/gnrcore/instances/gnrdevelop/config/instanceconfig.xml:6-11` mounts `sys, adm, test15, docu, dev` — no `biz`.
- An unresolvable `py_requires` is a hard raise, not a skip: `mixinResource` → `raise GnrMixinNotFound` (`gnrresourceloader.py:488`). The late `mixinComponent(...)` route raises the same way, which is exactly why `resources/common/th/th.py:1007` gates on `db.package('biz')` first.

So `fatt.cliente` was never the only reason that page failed.

**Two more corrections to the premise:**

1. `test15/webpages/components/dashboards.py` **still exists on develop** — it is deleted only on `refactor/test15-example-migration-macro1`. Likewise the smoke sweep, `pages_ratchet.py` and `smoke_known_failures.txt` exist only on that branch.
[...]

PR #1132, opened 2026-08-21
test15 triage, macro 2: empty the gnrwdg area

dgpaci:
## Summary

Macro 2 of 6 of the test15 -> test example triage: the `gnrwdg` area is empty. The 25 pages it held are gone from `test15`, four deleted as dead, fifteen promoted into `test/webpages/gnrwdg/` and six folded into the counterparts that already existed there, every promoted case carrying a title and a one-line docstring. `test15/webpages/gnrwdg/` no longer exists and the `Gnrwdg` branch left `test15/menu.py`.

Three things beyond the moves:

- **The area's blocker was a package nothing mounted.** Nine pages address `glbl`, which lives in `projects/gnr_it`. `gnrdevelop` mounts it now (plus `gnrcore:biz`, for the dashboard component), and the render sweep skips a page whose packages the site has not mounted instead of failing on it — so a user instance without `glbl` still runs the suite and no page had to be rewritten. A page names another package through two channels and both count: the tables it addresses and the components it mixes in through `py_requires`, where an unresolved entry is a hard `GnrMixinNotFound`. The referendum-era column injections under `test/model/_packages/glbl` and `test15/model/_packages/glbl` are removed: with the package mounted they would have become real columns of `glbl`.
- **`DashboardItem` gets an example bound to a table gnrcore ships** (#1105). `di_buildRemoteItem` dropped the `table` argument, so a bare `dashboardItem(table=...)` never reached the item; it forwards it now and tolerates a call with no `itemRecord`. `flibpicker.py` leaves the known-failures list, skipped for its absent package instead of rendered and failed.
- **`TestHandler` warns when a case addresses a table whose data was never loaded.** A case over an empty table renders an empty widget and reads as a broken page while the cause is developer setup: the warning sits on the case docline, and only for tables whose package actually ships `startup_data`.

The macro's own find, and it came twice: **`testOnly` was masking duplicate frameCodes**. `slotbar.py` reused `frameOne`, `frameTwo` and `frameMultibuttonStore` on two cases each; `formhandler_selection.py` had all five cases naming their form `provincia` and three naming their grid `province`. With `testOnly` dropped, `TestHandlerFull` renders every case on one page — each with its own datapath but not its own frameCode — so the codes were suffixed per case together with everything deriving from them (`genro.formById`, `subscribe_form_*_onLoaded`, `parentStore`, `default_regione`). Same defect class as `5c2ff2ffc` on the base branch. It is not a suite check: the collision happens client-side, after the bootstrap GET has answered 200, so nothing but a browser sees it. It is written up as a developer rule in the team knowledge base instead.

## Notes for the reviewer

- **Stacked on macro 1**: the base is `refactor/test15-example-migration-macro1`, not `develop`.
- The browser pass over the ~20 promoted pages is done and is what produced the second half of the diff: `bageditor test_0` never edited anything (`bagNodeEditor` listens on `<nodeId>_currentPath` while the case published `test_editnode`), `struct='min'` and `struct='regione'` are not column lists, `formhandler_baggrid` read `gridstore1` while its controller fills `gridstore`, the `addrow` slot resolves `frame.grid` so the toolbar must be declared after the grid, and the th dialog of `thpalette` needs `dialog_windowRatio`.
- `.phased/active/test15-gnrwdg-macro2/` carries the plan, the run notes and the coherence review with the ruling on each of its eight flags — that is the audit trail of the macro, not code.
- Pre-existing debt in the six fold targets (generic titles, thin case docstrings, an Italian column label in `bag_grid.py`) is deliberately untouched: those files belong to later macros.

## Linked issue

Covers #1105 (already closed). The macro itself has no issue: it is one step of the test15 migration roadmap.

## Test plan

- [x] `pytest projects/gnrcore/packages/test/tests/test_pages_documented.py projects/gnrcore/packages/test/tests/test_pages_ratchet.py -q` — 10 passed (both now run in CI)
- [x] `pytest projects/gnrcore/packages/test/tests/test_pages_smoke.py -q -rs` — 1 passed, **no skip**: the sweep rendered every discovered page, it did not opt out
- [x] `flake8` — zero errors on the whole file set
- [x] Browser pass over the promoted and folded pages: each builds its structure past the bootstrap GET and shows what its docstring claims; `/test/gnrwdg/formhandler_selection` drives five independent forms, `/test/gnrwdg/thpalette` opens the provincia th page in its dialog, `/test/gnrwdg/gridscale` scales the grid on both axes, `/test/components/dashboards` counts the rows of the table each tile is given
- [ ] Reviewer setup, if the `glbl` grids come up empty: `gnr app dbsetup gnrdevelop`, then `loadStartupData` on the `glbl` package — a destructive local operation, it empties and refills those lookup tables. The card warning added here says when this is needed.


dgpaci:
Rebased onto the updated base: `refactor/test15-example-migration-macro1` (#1133) now carries `develop` plus two changes of its own — the examples that cannot work are deleted, and a page whose example needs a third-party library says which one instead of answering 500.

One merge conflict, worth naming because of what it resolves to: both branches were deleting a line from `smoke_known_failures.txt` — this one `flibpicker.py` (skipped now, its package is not mounted), macro 1 `icstest.py` (renders now, its `ics` import is guarded). The list is resolved to **empty**: every discovered page either renders or is skipped for a package the instance does not mount, and the ratchet says so.

Re-run after the merge: `test_pages_documented.py` + `test_pages_ratchet.py` 10 passed, `test_pages_smoke.py -q -rs` 1 passed with an empty known-failures list, flake8 clean.

fporcari (changes requested):
The area work looks like macro 1's — same discipline, same documentation of what
was folded into what. Requesting changes on one thing that is not a style matter:
CI is red on all four Python versions, and the cause is this branch's own delta.

## Blocking: 4/4 jobs failing, and it is not flakiness

```
FAILED tests/app/test_dbupgrade_broadcast.py::TestDbUpgradeBroadcast::test_resource_userobjects_install_in_done_phase
FAILED tests/app/test_dbupgrade_broadcast.py::TestDbUpgradeBroadcast::test_db_upgrade_broadcast_is_idempotent
    sqlite3.OperationalError: database is locked
===== 2 failed, 2104 passed, 70 skipped in 164.55s =====
```

The comparison that settles it:

| | result |
|---|---|
| **#1133** — the base, this tree minus this branch's delta | **4/4 pass** |
| `develop` | pass |
| **#1132** | **4/4 fail**, the same two tests |

Four out of four is not intermittence.

**And the chain is traceable.** `gnrpy/tests/core/common.py:137`:

```python
class BaseGnrAppTest(BaseGnrTest):
    app_name = 'gnrdevelop'
```

The whole app/web suite instantiates `GnrApp('gnrdevelop')`, reading the very
`instanceconfig.xml` this PR edits. The delta over #1133 is:

```xml
+		<gnr_it_glbl pkgcode="gnr_it:glbl"/>
+		<gnrcore_biz pkgcode="gnrcore:biz"/>
-	<menu package="sys,adm,test,test15,docu,dev"/>
+	<menu package="sys,adm,test,test15,glbl,biz,docu,dev"/>
```

So those tests now load two more packages and run their `onDbUpgrade` broadcasts,
and the `checkDb(applyChanges=True)` in `setup_class` ends up contending for the
sqlite lock. **A test-migration PR silently changed the package set of the
instance the entire suite boots** — and 13 test files inherit that fixture
(`BaseGnrAppTest`/`BaseGnrDaemonTest`), `gnrwsgisite_test.py` and
`gnrstoragehandler_test.py` among them.

Two things it is *not*, so nobody spends time on the wrong lead: both packages
exist in the repository (`projects/gnr_it`, 32 tracked files, and
`projects/gnrcore/packages/biz`), so this is not a missing package; and #1133
mounted `gnrcore:test` into the same file and stayed green, so it is not "any
mount breaks it" — it is these two.

## `testhandler.py` is the widest-radius change in the PR

`resources/common/gnrcomponents/testhandler.py` is the shared component behind
every `test_*` page in genropy *and* in every client project doing
`py_requires="gnrcomponents/testhandler:…"`. The new `testHandler_dataWarning`
runs on every case of every page, walks the whole built structure with
`body.traverse()`, and for each distinct `pkg.table` reaches:

```python
return not self.db.table(tablename).countRecords()
```

which is `query(excludeLogicalDeleted=False, excludeDraft=False).count()` — an
unfiltered `SELECT count(*)`. The cache is per request
(`self._testHandler_dataToLoad` on the page), so it does not carry over.

The direction of the error is cost only, never a wrong warning — but on a project
with large tables every test-page render pays a full count per table addressed.
Answering "is it empty" needs a `LIMIT 1`, not a count.

**The half that is fine**, recorded so it does not get re-investigated:

- `card.pop(docline.parentNode.label)` is real framework API, not a slip —
  `parentNode.label` is used at `gnr/core/gnrbag.py:482-484` and
  `gnr/core/gnrstructures.py:94`.
- `body.traverse()` resolves: `GnrStructData(Bag)` (`gnrstructures.py:37`), with
  `traverse` at `gnrbag.py:2400`.
- `testHandler_checkDataToLoad` filters unknown packages and tables and requires a
  `startup_data.pik/gz`, so a `table`/`dbtable` attribute that is not a table
  reference costs nothing and warns nothing. The `^\w+\.\w+$` guard does not
  produce false alarms.

## The 15 `.phased/` files

`foreman.json`, `log/phase-1..7.txt`, `log/repair-fable.txt`, `notes.md`,
`plan.md`, `review.md`, `verify.md`, plus `.phased/roadmap.md`. These are
artifacts of how the work was produced rather than product. A roadmap may well
belong in the repository; per-run logs and a foreman state file probably do not.
Asking rather than telling — but it is worth a deliberate decision now, because
macros 3 to 6 will each add another set.

## Minor

No linked issue and no assignee. And since the base is
`refactor/test15-example-migration-macro1`, this sits outside branch protection —
`reviewDecision` is empty because no review is *required* there, not because the
work passed one. It should be judged against `develop`, where it will land, and it
cannot merge before #1133 anyway.

## Scope of this review

I did not run the render sweep, did not do a browser pass, and did not inspect the
moved pages one by one. The blocking finding needs none of that: it comes from CI
and from three files.

dgpaci:
@fporcari all three points addressed, and the description is rewritten rather than patched — two of its three pillars had changed. The delta since your review:

## The blocking finding: closed, and the cause is narrower than it looked

You had it in the right place. Refining it by one step: it is not `checkDb` contending for the lock, it is the checklist sysrecord `glbl` injects into `adm` through `projects/gnr_it/packages/glbl/model/_packages/adm/install_checklist.py`. The traceback ends in `onDbUpgrade_createSysRecords` -> `_sysRecordCreateCb` -> `INSERT INTO "adm"."adm_install_checklist"`, which is a write into another package's attached sqlite file.

Reproduced locally before touching anything, then after:

| | result |
|---|---|
| `gnr_it:glbl` mounted | **2 failed in 42.86s** |
| `gnr_it:glbl` unmounted | **2 passed in 0.98s** |

I also tried the middle path before giving up the mount, since the test already has a precedent for it — it seeds one `sys.calendar` row so that table's `onDbUpgrade` early-returns. Seeding `glbl.nazione` the same way, so its `loadStartupData` does not run, **does not help**: both tests still fail after 42s. So the cost is the cross-package injection, not the 6 MB import that comes with it, and there is no cheap way to keep the mount.

The resolution is the guard this branch already adds: a page whose packages are not mounted is skipped and the missing ones are named. That is what the nine `glbl` pages get now, and no page had to be rewritten.

`gnrcore:biz` stays mounted. It ships no startup data, injects into no other package, and with `glbl` out the suite is green with it in — so the `DashboardItem` example keeps working.

**What it costs, so it is on the record**: the sweep now reports 59 pages as skipped, 48 of them for `glbl`. Not only test15 pages — the `TableHandler` and `Grid` families, `advanced_form` and `external_change.py` are in there. They are not broken and not unreachable; they are simply not rendered by a sweep on an instance without `gnr_it`, and the sweep prints one line per page it leaves out so the loss does not go quiet. If you would rather pay for the coverage in some other way, this is the decision to argue with.

## `testhandler`: one row instead of a count

Right call, and the radius is the reason. It now fetches one pkey with `limit=1`, keeping the visibility `countRecords` had — logically deleted and draft rows included, since those are rows the startup loader would not insert again.

Verified against the real db rather than by reading: on `adm.user`, `sys.calendar` and `adm.userobject` the new check agrees with `countRecords()` when empty, and after inserting one row it agrees in the other direction too. The row was rolled back, not committed.

## The 15 `.phased/` files

Asked rather than told, so: you are right, and they are out. The thirteen under `.phased/active` are the trace of how the macro was produced, not product, and each remaining macro would have added another set. They are untracked and ignored, not deleted, so a local run keeps its own trace.

`.phased/roadmap.md` stays — it is the one file that says what macros 3 to 6 cover and why this order, which is what you will want in front of you when macro 3 arrives. Its macro 2 entry now describes the absent-package guard instead of the mount that guard replaced.

## What the base brought in, including the thing you flagged

You closed your macro 1 review with `includedview_bagstore.py`: #1130 added cases to it, a later macro would migrate the folder, and the cases needed to travel. They did. `test_1_selected_id_on_store_replacement` is now `test_3_selected_id_on_store_replacement` in `test/webpages/components/includedview.py` — the page this macro folds it into — renumbered only because `test_1` was already taken there.

The same merge brought `external_change.py` from #1293, added to `test15/webpages/gnrwdg/` after this branch had emptied the folder. It is promoted with the rest of the area and needed no repair: title, case docstring and its own frameCode were already there.

One consequence worth naming: that page binds `glbl.provincia`, so it is one of the 48 the sweep now skips. That is the state it was already in on develop — `gnrdevelop` has never mounted `gnr_it` — but it means the regression it guards is covered by a browser and not by the sweep.

`test_smoke_backpressure.py`, which macro 1 added for the canary fix, follows the signature `page_statuses` has here: it returns the skipped pages apart and reads the mounted packages off the site, a third boundary the tests answer. A case covering the skip itself comes with it, since no test drove that path through the loop — and that path is load-bearing now.

## Verified

- `pytest projects/gnrcore/packages/test/tests/ -q` — 14 passed, the render sweep included, with `smoke_known_failures.txt` still empty.
- `cd gnrpy && pytest` — 1764 passed, 0 failed.
- `flake8` — zero errors on the whole file set.

Note this still cannot move before #1133: the base is `refactor/test15-example-migration-macro1`, and GitHub will retarget it to `develop` once that merges.

fporcari (changes requested):
**My August blocker is closed**, and the three answers hold. Re-verified on `d58bcccfe`:

- `gnr_it:glbl` is out of `instanceconfig.xml`, `gnrcore:biz` stays. `pytest gnrpy/tests/app/test_dbupgrade_broadcast.py -q` → **2 passed in 1.17s**, against 4/4 red before; `gnrpy/tests/app` → 169 passed; the three CI-wired suites → 13 passed; `flake8` zero on every touched file.
- `testHandler_checkDataToLoad` now asks for one row. `countRecords()` is `query(excludeLogicalDeleted=False, excludeDraft=False).count()` (`gnrpy/gnr/sql/gnrsqltable/crud.py:481`) and the new query passes exactly those two flags, so the visibility is the same one; `limit` does reach the SQL (`_smartappend(result, 'LIMIT', limit)`, `gnrpy/gnr/sql/adapters/_gnrbaseadapter.py:715`).
- `.phased/` is down to `roadmap.md`.

`test_pages_ratchet.py` joining the CI step is the right call: the helpers the skip rests on are now the only part of this mechanism anything checks.

Your figures are exact. I recomputed the skips statically against the mounted set — 233 pages discovered, **59 skipped, 48 of them for `glbl`** — so the cost is measured, not estimated. Which is why I am requesting changes on the mechanism rather than on the number.

## Blocking: the skip rests on a premise that does not hold

`test_pages_smoke.py:12` states it plainly:

> A page addressing a package the instance does not mount cannot render

On `develop`, after macro 1 merged, that is false, and the counterexample is a page this PR does not touch:

- `test/webpages/inputfields/dbselect.py` calls `fb.dbSelect(table='glbl.provincia')` and `table='glbl.regione'` (`:33-55`), unchanged by this branch;
- `smoke_known_failures.txt` on `develop` holds exactly one entry, `test/webpages/tools/flibpicker.py`;
- macro 1's sweep was green over 241 pages, and `gnrdevelop` has never mounted `gnr_it`.

So `dbselect.py` renders 200 with `glbl` absent. The 48 `glbl` pages were not failing: they were rendering with empty selections — which is precisely the condition `testHandler_dataWarning` was added in this same PR to announce. The PR builds the right mechanism for that case and then overrides it with one that removes the page from the sweep instead.

Net effect: the sweep covers 174 of 233 pages, a quarter of the net given up for pages that were already covered.

## Blocking: the skip list has no ratchet, and it has already absorbed the tracked entry

`smoke_known_failures.txt` is now empty. `flibpicker.py` was not fixed — `page_required_packages` returns `{'flib'}` for it, `flib` is not mounted, so it is skipped. The one entry the ratchet was holding moved from the list that fails in both directions to a list nothing compares against.

That is the "the lists can only shrink" property with a back door: a page leaves the checked set by acquiring a dependency on an unmounted package, and no check notices. With four macros still to come on top of this net, that is the part I do not think should merge as it stands.

Either fix is small and in the style of the PR:

- narrow the criterion to what actually fails — a page skipped only when its render proves it cannot work — keeping the data warning for the rest; or
- keep the skip and give it a ratchet: a committed `smoke_skipped.txt` compared in both directions, exactly like the other two. Then 59 is a number somebody has to change on purpose.

The first also gives back the 48 `glbl` pages, since they render.

## Two smaller ones

**The skip lines do not print in a run that passes.** pytest captures stdout unless the test fails or `-s` is passed; I checked with a one-line passing test that prints — zero occurrences. So "the sweep prints one line per page it leaves out" holds only for the way you run it by hand.

**`test_pages_smoke.py:20-24` says the opposite of what the PR does.** It states that `gnrdevelop` mounts `gnr_it:glbl` for the lookup tables and explains how to `loadStartupData` on it, in the same commit that removes the mount. Wrong as written, whatever happens to the points above.

## Scope of this review

Blocking finding aside, the area work is macro 1's discipline again: what was folded into what is documented, the `includedview_bagstore` case travelled as it had to, and `external_change.py` arrived from the base and needed no repair. I did not run the render sweep (no daemon here), did not do a browser pass, and did not read the 56 files one by one — the two findings above come from the ratchet helpers, the instance config and three pages.

dgpaci:
@fporcari both blocking findings taken, and the mechanism is gone rather than ratcheted. The premise was wrong, so the fix that keeps the premise was not worth having.

## The skip is out; every page is rendered

`e60c72d28`. `page_statuses` renders every discovered page and returns statuses alone. `page_required_packages`, `required_resource_packages`, `table_reference_packages` and `resource_owner_package` go with it — 86 lines that existed only to decide the skip.

The sweep against `gnrdevelop`, with `gnr_it` unmounted:

| | |
|---|---|
| pages discovered | **233** |
| answering 200 | **232** |
| on `smoke_known_failures.txt` | **1** — `test/webpages/tools/flibpicker.py` |

`flibpicker` is the entry `develop` already carries, and it fails for the reason its own case docstring gives: `py_requires = "...,flib:FlibPicker"` with `gnrcore:flib` unmounted is a hard `GnrMixinNotFound`. It is back under tracking, where it was before this branch absorbed it.

So your reading was right on the number too: the 48 `glbl` pages render. Nothing was rewritten to get there.

Both ratchet directions checked against the running site, not by reading:

- list empty, sweep run → fails, `failing pages outside smoke_known_failures.txt: test/webpages/tools/flibpicker.py`
- a page that renders added to the list → fails, `stale entries in smoke_known_failures.txt (no longer failing pages, remove them): test/webpages/html/div.py`

## One correction to the premise, since it cuts the other way

You wrote that the `glbl` pages were rendering with empty selections, "which is precisely the condition `testHandler_dataWarning` was added in this same PR to announce". It is not: `testHandler_checkDataToLoad` returns `False` when `self.db.packages.get(pkgid)` is `None`, so a table of an unmounted package warns nothing. Those pages render empty and silent.

I looked at making it true — warn "package not mounted" alongside "load data first" — and did not, because from inside the page an unmounted `glbl.provincia` and a `table=` string that is not a table reference at all are indistinguishable: `self.db.packages` only knows what is mounted. Telling them apart takes exactly the filesystem scanning this commit is deleting, and this component is mixed into the test pages of every project, where a false "package not mounted" would be worse than silence. The warning stays on what it can answer: a mounted package whose startup data was never loaded.

It does not change your conclusion — the pages render, and they were covered before this branch took them out.

## `test_pages_ratchet.py`

It covered the helpers that are now deleted, so it moves onto what the sweep leans on alone: `assert_ratchet` in both directions, including the stale direction that makes a list shrink by itself; `read_ratchet` against the comment lines the committed files carry; `page_url` for the mapping every render goes through. Still source-only, still in the CI step.

## The two smaller ones

- **stdout capture**: moot, there are no skip lines left to print. Nothing about the sweep now depends on being read.
- **`test_pages_smoke.py:20-24`**: fixed. It said `gnrdevelop` mounts `gnr_it:glbl` in the commit that removed the mount. It now says what is true — `biz` mounted, `glbl` not, and why: the checklist sysrecord `glbl` injects into `adm` deadlocks the sqlite of the instance the app/web suite boots.

`.phased/roadmap.md` and the PR description follow the same change, so macro 3 does not inherit a guard that is no longer there.

## Verified

- `pytest projects/gnrcore/packages/test/tests/ -q` — **12 passed**, the render sweep included
- render sweep against `gnrdevelop` — 233 pages, green, one entry on the ratchet
- `cd gnrpy && pytest -q` — **1771 passed, 10 skipped, 0 failed**. The 751 errors in the run are `tests/sql` erroring at setup in `_start_pg_instance()`: no PostgreSQL in this shell, nothing outside `tests/sql` errors
- `flake8` — zero on the touched files
- net **-127 lines**

fporcari (approved):
**Both blocking findings are closed**, and closed by removing the mechanism rather than ratcheting it. Re-verified on `e60c72d28`:

- The skip is gone. `page_required_packages`, `required_resource_packages`, `table_reference_packages` and `resource_owner_package` no longer exist anywhere under `projects/`, and `page_statuses` renders every discovered page and returns statuses alone.
- `smoke_known_failures.txt` holds `test/webpages/tools/flibpicker.py` again — the single entry `develop` already carried — with the reason written next to it.
- The module docstring of `test_pages_smoke.py` now states what the instance config actually does: `gnrcore:biz` mounted, `gnr_it:glbl` not, and the pages bound to `glbl` tables rendering against empty tables. The paragraph that said the opposite is gone.
- `assert_ratchet` fails in both directions — an offender outside the list and a stale entry in it — and `test_pages_ratchet.py` asserts both, plus the empty-list and missing-file cases.

The second finding dissolves with the first: the 48 `glbl` pages come back into the checked set, and the data warning this PR adds is what tells a developer why a widget is empty. That is the right division of labour between the two mechanisms.

Here: `test_pages_ratchet.py test_smoke_backpressure.py test_pages_documented.py` → 11 passed, and `flake8` zero on `projects/gnrcore/packages/test/tests/`. The checks on this head are green on 3.11 through 3.14.

## Scope of this review

I did not run the render sweep — no daemon here — so 233 discovered and 232 answering 200 are your figures, not re-measured ones, and I did not repeat the browser pass over the promoted pages. My August and September blockers are both answered; this approval supersedes them.

Issue #1108, opened 2026-08-20
[BUG] Three shipped pages call pbl_root* page methods that no longer exist


dgpaci:
### Problem Description

`resources/common/public.py` exposes its page-root builders only as `@struct_method`s (`public_rootContentPane`, `public_rootBorderContainer`, `public_rootTabContainer`), which are called on the struct node — `root.rootContentPane(...)`. The older page-method form `self.pbl_root*(root, ...)` was removed in commit `d9c606331` (2012), but three shipped pages still call it and therefore raise `AttributeError` while building their structure:

- `projects/gnrcore/packages/adm/webpages/preference.py:23` — `mainbc, top, bottom = self.pbl_rootBorderContainer(rootBC, '!!Preference')`
- `projects/gnrcore/packages/sys/webpages/memcached_stats.py:35` — `center, top, bottom = self.pbl_rootContentPane(root, title='!!Mem Stats')`
- `projects/gnrcore/packages/multidb/webpages/sync_status.py:26` — `pane = self.pbl_rootContentPane(rootBC, title='!!Sync Status', datapath='syncstatus')`

Note also the signature change: `pbl_root*` returned a `(center, top, bottom)` triple, the struct method returns the frame, so the two unpackings above cannot be fixed by renaming alone. The frame's center is a plain slot: an explicit container is needed for pages that lay their content out in regions.

A fourth page, `projects/gnrcore/packages/test/webpages/components/advanced_form.py`, had the same defect and is repaired in the branch this issue comes from — that repair is the reference for the fix.

### Reproduction Steps

1. Mount `gnrcore:adm` (any instance has it) and open `/adm/preference`.
2. The first GET answers 200 with the bootstrap document; the structure POST to the same URL answers 200 carrying an error bag with `AttributeError: 'GnrCustomWebPage' object has no attribute 'pbl_rootBorderContainer'`, and the page renders the server-error panel.

Worth noting for any smoke or monitoring check: because the failure travels in a **200** response, neither a status-only page sweep nor a browser pass that records only responses with status >= 500 detects it.

### Environment

- Genropy branch: develop (verified at `9f441df73`)
- The removal that orphaned these calls: `d9c606331` (2012)

fporcari:
Working on this.

fporcari:
[...]

PR #1112, opened 2026-08-20
fix(webpages): update pbl_root* callers to the struct_method page-root API

fporcari:
## Problem

Four shipped webpages still call `self.pbl_rootContentPane(...)` / `self.pbl_rootBorderContainer(...)`, a page-method API that no longer exists. Building their structure raises `AttributeError`.

## Root cause

Commit `d9c60633` (2012-05-30) deleted the page-method page-root builders from `resources/common/public.py` (`_pbl_root`, `pbl_rootContentPane`, `pbl_rootStackContainer`, `pbl_rootTabContainer`, `pbl_rootBorderContainer`) and replaced them with `@struct_method`s dispatched on the struct node (`rootContentPane`, `rootBorderContainer`, `rootTabContainer`, `rootStackContainer`, all delegating to `_pbl_frameroot`). `GnrWebPage`/`GnrBaseWebPage` have no `__getattr__`, so any remaining `self.pbl_root*` call is a plain missing attribute — `AttributeError` during `main()`, not a lookup the framework can route.

**Four call sites on develop, not three.** The issue lists three and calls the fourth "already repaired" — that repair exists only on `origin/refactor/test15-example-migration-macro1` (commit `1fbf73ce2`), which **renames** `test15/webpages/components/advanced_form.py` to `test/webpages/components/advanced_form.py`. On `develop` the old path is still present and still broken:

| file | call |
|---|---|
| `projects/gnrcore/packages/adm/webpages/preference.py:23` | `self.pbl_rootBorderContainer(rootBC, '!!Preference')` |
| `projects/gnrcore/packages/sys/webpages/memcached_stats.py:35` | `self.pbl_rootContentPane(root, title='!!Mem Stats')` |
| `projects/gnrcore/packages/multidb/webpages/sync_status.py:26` | `self.pbl_rootContentPane(rootBC, title='!!Sync Status', datapath='syncstatus')` |
| `projects/gnrcore/packages/test15/webpages/components/advanced_form.py:9` | `self.pbl_rootBorderContainer(rootBC, title="!!Terapia", margin="1em", **kwargs)` |

[...]

PR #1133, opened 2026-08-21
test15 triage, macro 1: the regression net, the websocket and components areas

dgpaci:
## Summary

Macro 1 of 6 of the test15 -> test example triage: build the net that makes the rest of the migration safe, establish the recipe, and run it over the first two areas. 60 pages leave `test15`, 16 arrive in `test` documented, and two suites keep both packages from sliding back.

- **A package suite for the test pages** (`projects/gnrcore/packages/test/tests`): a documentation ratchet over every page of both packages — source-only, no instance, wired into CI — and a render sweep that renders each page against `gnrdevelop` with closed-loop backpressure against ephemeral-port exhaustion. Both compare against a committed list and fail in **both** directions, so the lists can only shrink.
- **38 dead pages removed** with their leftovers (soap, temp, docker, devtools areas).
- **websocket area**: 8 pages moved to `test/webpages/websocket/`, documented, with the dead code their docstrings contradicted removed.
- **components area**: 12 pages, three of them merged into the `test` counterparts they duplicated without losing a case.
- `testdata/docstore` moved into `test`, cutting the last cross-package reference; `gnrcore:test` mounted in `gnrdevelop`.
- **19 promoted pages swept in a browser**, which is what the render sweep cannot do: five component defects filed as #1101-#1105, `advanced_form` repaired, and #1108 filed for the same defect in three shipped pages outside this scope.

Two changes on top of that, from the pre-PR review:

- **Three examples that cannot work are gone**, rather than migrated into the curated package. `components/proxy_tester.py` is built on `proxy = True`, a mechanism the framework dropped in 2020 (`44d6d96d9`): `classMixin` honours `proxy_name` only and lists `proxy` among the attributes it does not mix in, so `self.proxy_test` never exists and both cases raise while the page builds. `components/htable_picker.py` binds `base.category` and `base.ct_catalog`, which exist nowhere in the repository. `lib/proxies/` had nothing referencing it but a commented-out import naming a path that stopped existing when the folder moved. The working proxy example is `test15/webpages/tools/dynamic_proxy.py` on the current `@page_proxy` API, which the `tools` macro will promote.
- **An example that needs a third-party library now says which one, instead of taking the page down.** `icstest.py`, `html/img.py` and `inputfields/remoteselect.py` imported `ics`, Pillow and cinemagoer at module level: where the library is absent the import raises, the page answers 500 and the reader gets an error page. Each imports under `try/except ImportError` now, and every case that needs the library opens with `opl_missingLibrary` (`webpages/_resources/optional_library.py`), which puts the library and its `pip install` line on the pane and stops the case there. Only the cases that need it are guarded — five of six `img.py` cases and six of seven `remoteselect.py` cases are unchanged. `icstest.py` guards both, since its download case goes through the `ics` webtool, which imports the same library.

That last one also fixes a property of the ratchet rather than a page: `smoke_known_failures.txt` was encoding which optional libraries happened to be installed on one developer's machine, so it reported false regressions in both directions elsewhere. `test/webpages/tools/icstest.py` leaves the list — it renders here, with `ics` not installed.

## Notes for the reviewer

- `develop` was merged in before opening (16 commits, no file in common with this branch).
- **Follow-ups, deliberately not in here**: `HTablePicker` is left without an example in gnrcore — rebinding it on a hierarchical table the project ships is its own change, not a silent rewrite inside a migration. `radioButtonSet` still shows no option selected on the first render of a stored value: it injects no `cells__customGetter`, unlike `checkBoxSet`, and `c8a8a1ada` did not change that (the getter it removed returned false for radio cells anyway). `webtools/ics.py` imports `ics` at module level too. `components/th_hierarchy.py` is a placeholder whose only case is `pass`, carrying its own todo. `remoteselect.test_5_printResources` resolves a project named `sandbox`, so it only works where one exists.
- The render sweep proves the bootstrap GET and, for these pages, that the `optional_library` mixin resolves — an unresolved `py_requires` raises `GnrMixinNotFound` right there. The **warning text itself needs one look** in the browser: `/test/tools/icstest`, `/test/html/img` case 5, `/test/inputfields/remoteselect` case 2 (only where the library is missing, so `ics` is the one that shows it here).

## Test plan

- [x] `pytest projects/gnrcore/packages/test/tests/test_pages_documented.py -q` — 1 passed (runs in CI)
- [x] `pytest projects/gnrcore/packages/test/tests/test_pages_smoke.py -q -rs` — 1 passed, no skip, with `icstest.py` out of the known-failures list and `ics` not installed: the guard is what makes that page render
- [x] Both suites re-run after merging `develop`
- [x] `flake8` — zero errors on every file touched
- [x] Browser pass over the 19 promoted pages (the five defects it found are #1101-#1105)
- [ ] Reviewer with `ics`, Pillow and cinemagoer installed: the three cases show their example, not the warning


fporcari (approved):
Approved. The order is what makes this reviewable: the net first, the migration
inside the net. I checked the load-bearing claims rather than the file count, and
they hold.

## Verified independently

**The ratchet really ratchets** — this is the claim all six macros rest on, so I
put it under test instead of reading it:

- `pytest projects/gnrcore/packages/test/tests/test_pages_documented.py -q` → 1 passed.
- Removed one line from `docstring_debt.txt` → fails with
  `undocumented pages outside docstring_debt.txt`.
- Added an invented line → fails with `stale entries in docstring_debt.txt`.

`assert_ratchet` computes `offenders - known` and `known - offenders` and asserts
on both (`pages_ratchet.py:78-90`). Both directions confirmed.

**The three "cannot work" deletions hold.** This is the riskiest part of the PR — a
wrong call here destroys a working example — so both justifications were checked
against the framework, not the docstring:

- `proxy = True`: `classMixin` puts `'proxy'` in `exclude_list`
  (`gnr/core/gnrlang.py:419`) and reads only `proxy_name`
  (`gnrlang.py:428`). `self.proxy_test` cannot exist. Confirmed.
- `htable_picker.py`: `base.category` and `base.ct_catalog` appear in exactly one
  file in the repository — the one being deleted. Confirmed.

**The `genro_grid.js` hunk is right.** The getter it removes did
`rowdata[this.checkBox]`, and on a radio cell `this.checkBox` is `undefined`, so
`rowdata[undefined]` is falsy and it always returned false — the glyph could never
show as selected. Dropping it lets the real value through.

**Also checked**: flake8 clean on every touched `.py`; `.gitignore` adding
`gnrdevelop/site/` is harmless (zero tracked files under it); no dangling
reference to any of the 38 deleted pages (every deleted webpage basename grepped
against `projects`, `gnrpy`, `resources`, `gnrjs` — the hits are all generic-name
false positives or the migrated counterparts); the optional-library guard is wired
per case as described, `img.py` guarding only `test_5_dataUrl` and
`remoteselect.py` only `test_2_remoteSelect_with_api`.

**What I did not verify**: I did not run the render sweep (it needs a built
instance and a running daemon), did not do a browser pass, and did not inspect the
60 moved pages one by one. I have Pillow and cinemagoer installed but not `ics`, so
the open box in your test plan stays open from here.

## @dgpaci — the A/B decision in #1128 is yours and it blocks the merge order

#1128 deletes both `menu.py` files; this PR removes two lines from
`test15/menu.py`. Whoever merges second gets a modify/delete conflict, and the
resolution is that #1128's deletion subsumes your hunk — its description already
said so.

The part nobody closed: #1128 is still a **draft** because it asks you to confirm
the direction, and you are not among its requested reviewers, so the question is
sitting where you will never see it. It matters here directly — this PR adds
`test` to `<menu package="sys,adm,test,test15,docu,dev"/>` in
`instanceconfig.xml`, mounting in the menu a package whose menu contribution
#1128 proposes to delete. The two options there are:

- **A** — remove both menu contributions completely; the Tests UI entry
  disappears, direct routing and page discovery stay.
- **B** — keep one folder-driven entry; the Tests UI stays, and so does a menu
  surface that can drift from the package lifecycle.

Since you own the migration this menu describes, the call is yours. Answering on
#1128 unblocks it and settles which of the two lands first.

## Four observations, none blocking

**1. "the lists can only shrink" promises more than the mechanism gives.** The
ratchet is enforced against the code, not against the file: a commit that adds an
offending page *and* its line in `docstring_debt.txt` passes. Only a reviewer
catches that. "cannot grow silently" is the accurate phrasing.

**2. The smoke ratchet has no automated guard.** Only the documentation check runs
in CI — correctly, since the sweep needs a daemon. So `smoke_known_failures.txt`
(one entry today) is verified only when somebody runs it locally, on a machine
whose set of optional libraries changes the outcome. The description shows this
already bit once: the list "was encoding which optional libraries happened to be
installed on one developer's machine". `opl_missingLibrary` fixes the three known
cases, not the property. Worth saying who keeps that list honest across the
remaining five macros.

**3. An edge case in the `page_statuses` canary.** If the *first* page in sort
order fails, `canary is None`, so `probe = page_path` — the page that just failed.
The probe fails again, the sweep enters `wait_ports_drained` and spends
`3 × 35s = 105s` on a page that will never render, then latches
`register_dead = True`, which disables the port-drain protection for the whole
rest of the sweep: a genuine port exhaustion later is then reported as a mass of
regressions. The description mentions this can "add one" round; it is up to three,
and the latching is not mentioned. It cannot fire today — the only known failure,
`tools/flibpicker.py`, sorts well after working pages — so it is latent. A
one-line guard is to probe only with a page that has already answered 200.

**4. Minor.**

- The smoke suite inherits `pytest.skip` from `BaseGnrDaemonTest`
  (`gnrpy/tests/web/webcommon.py:32-47`), which is exactly the behaviour #1036 is
  open about. The design already accounts for it by keeping the documentation
  ratchet separate; the two will interact.
- `opl_missingLibrary`'s message is a bare string with no `!!` marker, while this
  PR also touches `test/localization.xml`. On a developer-facing page that is
  probably deliberate — worth confirming it is a choice and not an oversight.
- `gnrjs/` is owned by @fporcari and @genro in CODEOWNERS, so this PR needs a
  gnrjs-aware read and not only a test-tree one. Flagging it because the one-line
  grid fix is easy to skim past in a 99-file diff.

## The reassuring half

Three open PRs touch `genro_grid.js`: this one at ~4207, #1130 at ~3868, #1110 at
~2893. Hundreds of lines apart, no textual conflict and no semantic interaction —
so nothing to coordinate there.

One thing to carry forward instead: #1130 adds cases to
`test15/webpages/gnrwdg/includedview_bagstore.py`, which this PR does not touch,
but a later macro will migrate that folder. Those new cases need to travel with
it.

## Process

The PR has no assignee, and #1132 is stacked on this branch.

dgpaci:
@fporcari develop merged in and one of your observations fixed. Here is the whole delta since your approval, so the re-review costs as little as possible.

## The two merge conflicts

- **`test15/menu.py`** — modify/delete. #1128 deleted it under option C, and that deletion subsumes the two lines this branch removed. develop's deletion taken.
- **`test/webpages/components/advanced_form.py`** — rename/modify. #1112 converted the `pbl_root*` call in the test15 original that this branch had already promoted and repaired. The promoted copy is kept: it carries the same `@struct_method` conversion and additionally binds the returned frame as a frame, taking its center through `frame.center`, where the original assigned the frame itself to a variable named `center`. I checked whether #1112 applied that same shape anywhere else: the only other conversion is `adm/webpages/preference.py`, which ignores the return value, so there is nothing to fix there.

`test/localization.xml` merged on its own and is consistent — the branch had already replaced the `Terapia` entry with `Province`, with all four translations.

## Observation 3, the canary edge case: fixed in f49a3a832

The drain path now requires a page that has already answered 200 instead of falling back to the failing page. Nothing is lost by leaving it out before the first 200: exhausting the range takes hundreds of renders, so it cannot happen there.

`test_smoke_backpressure.py` covers it in two cases, with no site and no daemon: the two IO boundaries, the render and the wait, are answered by the test through a subclass. The test earns its name — putting the pre-guard version of the file back makes the first case fail with

```
assert 'test/webpages/bag/ext_formats.py' not in ['test/webpages/bag/ext_formats.py']
```

which is the defect itself, the first page in sort order probing itself. The second case, where the drain still fires and probes with the canary, passes on both versions.

Since a check that only runs locally protects nothing, that test joined the CI step: it reports `3 passed` instead of `1 passed` on all four Python versions.

## Your other points

- **#1128**: settled, option C, merged on 31 August.
- **Observation 1**: you are right and the description overstates it. The ratchet cannot grow *silently*; a commit that adds an offending page together with its debt line passes, and only a reviewer catches that.
- **Observation 2**: still an open question, and worth settling before macro 3 rather than after. What I can report is the state today — the list is back to one entry, and the sweep was re-run on the merged tree.
- **Observation 4, the missing `!!`**: deliberate. The message is built by interpolation, with the library and its `pip install` name substituted in, so a `!!` prefix would not yield a stable translation key.
- The open box in the test plan, a reviewer with `ics`, Pillow and cinemagoer installed, is still open.

## Verified on the merged tree

- documentation ratchet: passes.
- render sweep: 241 pages, passes — and not vacuously: removing the `flibpicker` line makes it fail with `failing pages outside smoke_known_failures.txt`. The ~10 pages develop added in the meantime all render, so the list stays at one entry.
- gnrpy suite: 1764 passed, 0 failed locally. The `tests/sql` errors here are a missing local postgres and reproduce identically on develop; CI is green on all four versions.
- `flake8`: zero errors on the 33 touched files.
- `node --test gnrjs/tests/*.test.js`: 31/31. The one-line `genro_grid.js` fix survived the #1152 perf refactor untouched.
- CI: 10/10 green.

One thing to carry forward for macro 2, since you had flagged it: #1130 added `test_1_selected_id_on_store_replacement` to `test15/webpages/gnrwdg/includedview_bagstore.py`, which that branch deletes by folding it into `test/webpages/components/includedview.py`. The new case travels with it there, not here.

fporcari (approved):
Re-approved. The delta since the dismissed approval is one commit and two merges, and the commit is the right fix to observation 3: `page_statuses` now enters the drain path only with a page that has already answered 200, so a page that simply fails early in sort order neither spends the three drain rounds nor latches `register_dead` on everything after it.

## Re-verified in this pass

**The new test earns its name.** I put the pre-guard version of `test_pages_smoke.py` back and re-ran `test_smoke_backpressure.py`:

```
E  AssertionError: assert 'test/webpages/bag/ext_formats.py' not in ['test/webpages/bag/ext_formats.py']
E   +  where [...] = <Sweep>.probes
1 failed, 1 passed
```

That is the defect itself — the first page in sort order probing itself — so the test pins the fix rather than describing it. The second case passes on both versions, as the comment says. Both run without a site or a daemon, which is what lets them join the CI step.

- `pytest test_pages_documented.py test_smoke_backpressure.py -q` — 3 passed.
- `node --test gnrjs/tests/*.test.js` — 31/31. The `genro_grid.js` hunk is unchanged and its context still matches after #1152's refactor.
- `flake8` on the 34 touched `.py` files — zero errors.
- CI — 10/10 green on the merged tree.

**Both conflict resolutions hold.**

- `test15/menu.py`: gone on the head and gone on `develop`, so #1128's deletion is what stands.
- `test/webpages/components/advanced_form.py`: byte-identical to the version I approved. `public_rootBorderContainer` is a `@struct_method` returning the frame (`resources/common/public.py:123-126`), so `frame.center.borderContainer()` is the correct shape; `adm/webpages/preference.py:23` ignores the return value, so the other conversion from #1112 needs nothing.
- `test/localization.xml`: the `Province` entry carries all four translations and the `Terapia` one is gone.

**The cold-daemon case I checked before agreeing with the guard.** Requiring a canary means the sweep will not drain before the first 200, and a register that is merely slow to come up would previously have been recovered there. It is already handled upstream: `get_waited_wsgisite` retries the site build while the daemon comes up and the class skips with the traceback if it never does (`gnrpy/tests/web/webcommon.py:15-29,46-58`). The guard removes no recovery that existed.

**Observation 4 withdrawn.** The message is built by interpolating both the module name and the `pip install` name, so a `!!` prefix would key on a string that changes per library. Deliberate, and right.

**Observation 1**: agreed, the description overstates it; the accurate phrasing is the one you used.

## Still open, and not mine to close here

**Observation 2 stands as the one thing worth settling before macro 3.** `smoke_known_failures.txt` is verified only when somebody runs the sweep locally, on a machine whose optional libraries change the outcome — `opl_missingLibrary` fixed the three known cases, not the property. The backpressure logic now runs in CI, which is a real improvement, but the list itself still has no automated keeper across the remaining five macros. Worth naming who that is.

## Not verified

I did not run the render sweep: it needs a built instance and a running daemon, so the 241-page figure and the claim that removing the `flibpicker` line makes it fail are yours, not mine. No browser pass from me in this round. `ics` is still not installed here, so the open box in the test plan — a reviewer with `ics`, Pillow and cinemagoer who sees the three examples rather than the warning — stays open from my side.

@genro's review is still requested and #1132 is stacked on this branch, so the merge order there is worth a look before this lands.

PR #1134, opened 2026-08-21
Restore legacy BaseComponent proxy compatibility

### LINKED ISSUE #1102: [BUG] BaseComponent proxy=True no longer attaches the component to the page
**Problem Description**

A `BaseComponent` subclass declaring `proxy = True` is documented (and used by the examples) to be attached to the page under a name derived from its class — `Proxy_test` becoming `self.proxy_test` — so its methods can be called from page code and, when decorated with `@public_method`, from the client. That attachment does not happen any more:

```
AttributeError: 'GnrCustomWebPage' object has no attribute 'proxy_test'. Did you mean: 'proxy_class'?
```

`gnr/web/gnrbaseclasses.py` handles the `page_proxy` decorator and the `is_proxy` flag, but nothing reads a `proxy` class attribute off a mixed-in `BaseComponent`, so `proxy=True` and the named form `proxy='proxy_test'` are both inert.

**Reproduction Steps**

2. Open `/test/components/proxy_tester` in a browser.
3. The structure POST returns 500 with the AttributeError above. The component it requires is `test/webpages/_resources/proxy_tester.py`, which declares `Proxy_test(BaseComponent)` with `proxy=True` and `Proxy_test2` with `proxy='proxy_test'`.

If `proxy=True` on a BaseComponent was intentionally dropped, the replacement API should be documented and the example updated; if not, the attachment needs restoring.

**Environment**

genropy branch develop, gnrcore:test package, gnrdevelop instance, Python 3.12, Chromium via Playwright.

fporcari:
## Summary

- restore the legacy BaseComponent proxy=True and named proxy contracts in classMixin
- preserve proxy_name precedence for page_proxy and Class AS alias consumers
- cover implicit naming, shared proxy composition, modern precedence, instantiation, and RPC serialization

## Verification

- flake8 on the changed source and test files
- targeted classMixin proxy tests: 4 passed
- full gnrpy test suite: 2111 passed, 69 skipped
- local testhandler /test15/components/proxy_tester: page construction, server-side proxy access, and RPC all passed with no console or server errors
- git diff --check

Fixes #1102

### DIFF
--- a/gnrpy/gnr/core/gnrlang.py
+++ b/gnrpy/gnr/core/gnrlang.py
@@ -426,6 +426,10 @@ def classMixin(target_class, source_class, methods=None, only_callables=True,
     if exclude:
         mlist = [item for item in mlist if item not in FilterList(exclude)]
     proxy_name = getattr(source_class, 'proxy_name', None)
+    if not proxy_name:
+        proxy_name = getattr(source_class, 'proxy', None)
+        if proxy_name is True:
+            proxy_name = source_class.__name__.lower()
     if proxy_name:
         keyproxy = '{proxy_name}_proxyclass'.format(proxy_name=proxy_name)
         proxy_class =  getattr(target_class, keyproxy, None)
--- a/gnrpy/tests/core/gnrlang_test.py
+++ b/gnrpy/tests/core/gnrlang_test.py
@@ -189,6 +189,83 @@ class TestGnrLang():
             assert 1000 in fl
         assert "d" not in fl
 
+
+class TestClassMixinProxy():
+    def test_legacy_true_uses_lowercase_class_name(self):
+        class Target(object):
+            pass
+
+        class Proxy_test(object):
+            proxy = True
+
+            def ping(self):
+                return 'pong'
+
+        gl.classMixin(Target, Proxy_test)
+
+        assert hasattr(Target, 'proxy_test_proxyclass')
+        assert Target.proxy_test_proxyclass.ping.proxy_name == 'proxy_test'
+
+    def test_legacy_named_proxy_composes_components(self):
+        class Target(object):
+            pass
+
+        class FirstComponent(object):
+            proxy = 'shared'
+
+            def first(self):
+                return 'first'
+
+        class SecondComponent(object):
+            proxy = 'shared'
+
+            def second(self):
+                return 'second'
+
+        gl.classMixin(Target, FirstComponent)
+        gl.classMixin(Target, SecondComponent)
+
+        page = Target()
+        page.shared = Target.shared_proxyclass(page)
+        assert page.shared.first() == 'first'
+        assert page.shared.second() == 'second'
+
+    def test_proxy_name_precedes_legacy_proxy(self):
+        class Target(object):
+            pass
+
+        class Component(object):
+            proxy = 'legacy'
+            proxy_name = 'modern'
+
+            def ping(self):
+                return 'pong'
+
+        gl.classMixin(Target, Component)
+
+        assert hasattr(Target, 'modern_proxyclass')
+        assert not hasattr(Target, 'legacy_proxyclass')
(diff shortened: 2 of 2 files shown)


genro (approved):
Approving. The production change is four lines and reads as the restoration it claims to be:

```python
if not proxy_name:
    proxy_name = getattr(source_class, 'proxy', None)
    if proxy_name is True:
        proxy_name = source_class.__name__.lower()
```

The precedence is the right way round — an explicit `proxy_name` still wins, and the legacy `proxy` attribute is only consulted when none was given, so the modern `Class AS alias` and `page_proxy` spellings keep their meaning. `proxy = True` resolving to the lowercased class name is the historical contract, and `proxy = 'shared'` continues to work because the same attribute carries both shapes.

The test set covers the four cases that matter and they are the right four: implicit naming from `True`, two components composing onto one named proxy, explicit precedence over the legacy attribute, and the RPC round trip. The last one is what proves the proxy is actually attached rather than merely present as an attribute.

dgpaci (approved):
Approving. I verified the fix against the case I reported in #1102 and it does close it.

The regression is confirmed historically: `44d6d96d9` ("master-detail component proxy improvement") renamed `proxy` to `proxy_name` in `classMixin` without keeping a fallback, so this is a restoration of a contract dropped by accident, not a new API. `'proxy'` staying in `exclude_list` ever since is the evidence that the attribute was still meant to be reserved and nobody noticed it had stopped being read. The restoration also rides the `cloneClass` isolation introduced after the rename, so it does not bring back the old global mutation of `GnrBaseProxy` that the pre-rename code did.

Checked end to end by mixing in the real `test15/webpages/_resources/proxy_tester.py` through the module-path branch: `proxy_test_proxyclass` lands on the page class, `instantiateProxies()` attaches `page.proxy_test`, `ciao_test()` returns `'ciao'`, the second component composes onto the same proxy (`ciao()` -> `'ciao2'`, original preserved as `ciao_()`), nothing leaks onto the page class, and `serializedFuncName` yields `proxy_test.ciao` — the form `getPublicMethod`/`_getProxyObject` unpacks, so the `@public_method` client path works too. Full gnrpy suite with the patch applied: 2194 passed, 10 skipped. flake8 clean on both files.

One thing worth adding to the PR description: the legacy spelling is still live outside genropy. `erpyready/packages/er_core/lib/master_detail.py` declares `proxy = 'master'`, `proxy = 'detail'` and `proxy = 'master_type'`, so those three proxies have been silently dead since the rename. The blast radius is wider than the test page.

**Note on the tests.** They never reach the failure the issue reports. The reported symptom is `AttributeError: 'GnrCustomWebPage' object has no attribute 'proxy_test'`, i.e. the attachment performed by `GnrWebPage.instantiateProxies()`; tests 2 and 4 do that attachment by hand (`page.shared = Target.shared_proxyclass(page)`). So the regression could come back through `instantiateProxies`, or through the resource-loader `path:*` branch that production actually uses, with all four tests still green. Test 1 pins `proxy_test_proxyclass` on the target class, which is the essential guard, but the repo already has real page infrastructure — `GnrDummySite` in `gnrpy/tests/web/gnrdummysite_test.py`, the resource loader in `gnrpy/tests/web/gnrpageclasscache_test.py` — so one test going through loader + `BaseComponent` + `instantiateProxies` is cheap and would cover the actual bug. Two details in the same area: nothing exercises `@public_method`, which the issue calls out (test 4 uses the `rpc_` prefix instead), and `assert page.shared.main is page` pins gnrlang's `BaseProxy` fallback rather than the web contract, since every real page has `proxy_class = GnrBaseProxy`, which stores `self.page`.

Not blocking, and pre-existing: `__on_class_mixin__` is called after `target_class` has been rebound to the proxy class, so a proxy component's `js_requires`/`css_requires`/`namespace` land on the proxy and are lost. Verified the ordering was the same before the rename and that it affects `page_proxy`/`AS` proxies identically. It does limit the restored contract for components that declare JS, so I will open a separate issue for it.

PR #1137, opened 2026-08-24
fix(pdf): joinPdf assembles with append, no more pages swapped between source files

### LINKED ISSUE #1136: [BUG] joinPdf swaps pages between the source pdfs
### Problem Description

`joinPdf` produces a merged document with the right page count and no exception raised, but **one page carries the content of a different file of the list**. Some documents also log resource errors while being read (`cannot find XObject resource`, `cannot find ExtGState resource`).

It is **intermittent**: the very same list of files, merged repeatedly in the same process, comes out right sometimes and wrong other times.

Expected: the merged pdf is the concatenation of the input files, page by page, every time.

Where: `resources/common/services/pdf/pypdf.py`, `Service.joinPdf_PYPDF` — the implementation used whenever PyPDF2 is importable, which is the default path in `joinPdf`.

**Cause.** `joinPdf_PYPDF` assembles with `PdfWriter.add_page`, which inserts the page leaving its indirect references in the source reader's object space; resolution happens only on write, in `PdfWriter._sweep_indirect_references` → `_resolve_indirect_object`:

```python
hash_value = real_obj.hash_value()
# Check if object is handled
if hash_value in self._idnum_hash:
    return self._idnum_hash[hash_value]
```

The writer deduplicates objects by hash in `self._idnum_hash`. When two source pdfs contain objects colliding on that hash, the second is silently replaced by the first. Since the hash of a `DictionaryObject` derives from `"%s" % self`, and the `repr` of an `IndirectObject` contains `id(self.pdf)`, the collision depends on memory addresses — hence the intermittency.

`PdfWriter.append` does not take that path: it clones each source's objects in full, and is unaffected.

### Reproduction Steps

Standalone script attached below — it does not require genropy, it replicates the body of `joinPdf_PYPDF` verbatim and compares it with the `append` strategy (PyMuPDF or reportlab needed only to generate the sample pdfs).

1. `python riproduci_joinpdf.py 30`
2. It builds a fixed list of 7 files (cover sheet, blank page, 2-page letter, 1-page attachment, blank page, 3-page attachment, blank page), 10 expected pages, and merges them 30 times with each strategy.
3. Observed result: `add_page` produces wrong documents on a large fraction of the runs, `append` never does.

| strategy | wrong documents |
|---|---|
| `PdfWriter.add_page` (current) | **5-10 out of 30** |
| `PdfWriter.append` | 0 out of 30 |
| `PdfMerger` | 0 out of 30 |
| PyMuPDF `insert_pdf` | 0 out of 30 |

A run on PyPDF2 3.0.1 / Python 3.13:

```
lista di 7 file, 10 pagine attese, 20 giri per strategia

add_page (attuale)    5 documenti sbagliati su 20
       pagina  1: atteso FRONTESPIZIO pagina 1 ottenuto FRONTESPIZIO pagina 1
       pagina  2: atteso PAGINA BIANCA pagina 1 ottenuto PAGINA BIANCA pagina 1
       pagina  3: atteso LETTERA pagina 1   ottenuto LETTERA pagina 1
       pagina  4: atteso LETTERA pagina 2   ottenuto LETTERA pagina 2
       pagina  5: atteso ALLEGATO UNO pagina 1 ottenuto ALLEGATO UNO pagina 1
   >>> pagina  6: atteso PAGINA BIANCA 2 pagina 1 ottenuto PAGINA BIANCA pagina 1
   >>> pagina  7: atteso ALLEGATO DUE pagina 1 ottenuto LETTERA pagina 1
   >>> pagina  8: atteso ALLEGATO DUE pagina 2 ottenuto LETTERA pagina 2
       pagina  9: atteso ALLEGATO DUE pagina 3 ottenuto ALLEGATO DUE pagina 3
   >>> pagina 10: atteso PAGINA BIANCA 3 pagina 1 ottenuto FRONTESPIZIO pagina 1
append (proposto)     0 documenti sbagliati su 20
```

It does reproduce externally — the script above needs no instance. For context on how it surfaces in an application: on the icond project it hit outgoing-mail protocol assembly ("every now and then an attachment does not make it into the recipient's pdf"). The attachment was not missing: one of its pages had been replaced by a page of the cover sheet or of another attachment. Verified through the real service (`site.getService('pdf').joinPdf`) as well: 0 wrong documents out of 20 with `append`.

### Environment

genropy `develop` at `ae138b349` and `master` at `d17fedff9` (both carry the same code), PyPDF2 3.0.1 (also affected on 2.x, by how the writer is written), Python 3.13 for the run above, Python 3.8 on the reporting instance.

### Additional Material

Reproduction script (`riproduci_joinpdf.py`):

```python
# -*- coding: utf-8 -*-
"""Reproduction of the joinPdf defect (pdf service, pypdf implementation).

Depends on neither genropy nor any project: it only replicates the body of
joinPdf_PYPDF. reportlab or PyMuPDF is needed to generate the sample pdfs.

    python riproduci_joinpdf.py [number_of_runs]
"""
import io
import os
import sys
import tempfile

from PyPDF2 import PdfReader, PdfWriter

CARTELLA = tempfile.mkdtemp(prefix='joinpdf_')


def crea_pdf(nome, testo, pagine):
    percorso = os.path.join(CARTELLA, nome)
    try:
        import fitz
        doc = fitz.open()
        for i in range(pagine):
            pagina = doc.new_page(width=595, height=842)
            pagina.insert_text((72, 100), '%s pagina %i' % (testo, i + 1), fontsize=24)
        doc.save(percorso)
    except ImportError:
        from reportlab.pdfgen import canvas
        c = canvas.Canvas(percorso)
        for i in range(pagine):
            c.setFont('Helvetica', 24)
            c.drawString(72, 742, '%s pagina %i' % (testo, i + 1))
            c.showPage()
        c.save()
    return percorso


def testo_pagine(percorso):
    reader = PdfReader(percorso)
    return [p.extract_text().strip().split('\n')[0][:26] for p in reader.pages]


def join_attuale(pdf_list, output_filepath):
    """body of joinPdf_PYPDF as it is in resources/common/services/pdf/pypdf.py"""
    output_pdf = PdfWriter()
    open_files = []
    for input_path in pdf_list:
        with open(input_path, 'rb') as input_file:
            memory_file = io.BytesIO(input_file.read())
            open_files.append(memory_file)
        input_pdf = PdfReader(memory_file)
        for page in input_pdf.pages:
            output_pdf.add_page(page)
    with open(output_filepath, 'wb') as output_file:
        output_pdf.write(output_file)
    for f in open_files:
        f.close()


def join_proposto(pdf_list, output_filepath):
    """the same thing with append()"""
    output_pdf = PdfWriter()
    for input_path in pdf_list:
        with open(input_path, 'rb') as input_file:
            output_pdf.append(io.BytesIO(input_file.read()))
    with open(output_filepath, 'wb') as output_file:
        output_pdf.write(output_file)


LISTA = [crea_pdf('a.pdf', 'FRONTESPIZIO', 1),
         crea_pdf('b.pdf', 'PAGINA BIANCA', 1),
         crea_pdf('c.pdf', 'LETTERA', 2),
         crea_pdf('d.pdf', 'ALLEGATO UNO', 1),
         crea_pdf('e.pdf', 'PAGINA BIANCA 2', 1),
         crea_pdf('f.pdf', 'ALLEGATO DUE', 3),
         crea_pdf('g.pdf', 'PAGINA BIANCA 3', 1)]

attesi = []
for f in LISTA:
    attesi.extend(testo_pagine(f))

giri = int(sys.argv[1]) if len(sys.argv) > 1 else 30
print('lista di %i file, %i pagine attese, %i giri per strategia\n' % (len(LISTA), len(attesi), giri))

for nome, funzione in (('add_page (attuale)', join_attuale), ('append (proposto)', join_proposto)):
    sbagliati = 0
    esempio = None
    for i in range(giri):
        uscita = os.path.join(CARTELLA, 'out.pdf')
        funzione(LISTA, uscita)
        ottenuti = testo_pagine(uscita)
        if ottenuti != attesi:
            sbagliati += 1
            esempio = esempio or ottenuti
    print('%-20s %2i documenti sbagliati su %i' % (nome, sbagliati, giri))
    if esempio:
        for i, (a, b) in enumerate(zip(attesi, esempio)):
            print('   %s pagina %2i: atteso %-18s ottenuto %s' % ('>>>' if a != b else '   ', i + 1, a, b))
print('\nfile di prova in %s' % CARTELLA)
```

**Separate defect, same file.** `joinPdf_FITZ` cannot work at all in its current form: it opens the *output* node for reading and passes its bytes as a filename to `save`, and `doc.insert(op)` is not a PyMuPDF API (`insert_pdf` is).

```python
with out_sn.open('rb') as f:
    doc.save(f.read())
doc.save()
```

It is not reached today because `joinPdf` prefers PyPDF whenever it imports, but it will be once the pypdf → PyMuPDF migration completes. Worth its own issue rather than being fixed silently here.

Reported by Andrea Busanelli (icond project).

fporcari:
Fixes #1136

## Symptom

`joinPdf` returns a merged document with the right page count and raises nothing, but **one page carries the content of a different file of the list** — intermittently, on the very same list of files merged repeatedly in one process. Some documents also log `cannot find XObject resource` / `cannot find ExtGState resource` while being read.

## Cause

`joinPdf_PYPDF` assembles with `PdfWriter.add_page`, which inserts the page leaving its indirect references in the source reader's object space. Resolution happens only on write, in `PdfWriter._sweep_indirect_references` → `_resolve_indirect_object`, which deduplicates objects by hash:

```python
hash_value = real_obj.hash_value()
if hash_value in self._idnum_hash:
    return self._idnum_hash[hash_value]
```

When two source pdfs hold objects colliding on that hash, the second is silently replaced by the first. The hash of a `DictionaryObject` derives from `"%s" % self`, and the `repr` of an `IndirectObject` contains `id(self.pdf)` — so whether two sources collide depends on memory addresses, which is where the intermittency comes from.

`PdfWriter.append` does not take that path: it clones each source's objects in full.

## Measure

Standalone reproduction (attached to #1136, no genropy needed): fixed list of 7 files, 10 expected pages, 30 runs per strategy.

| strategy | wrong documents |
|---|---|
| `PdfWriter.add_page` (before) | **5-10 out of 30** |
| `PdfWriter.append` (this PR) | 0 out of 30 |
| `PdfMerger` | 0 out of 30 |
| PyMuPDF `insert_pdf` | 0 out of 30 |

## Verification

- Reproduction script run on this checkout with PyPDF2 3.0.1 / Python 3.13: 5 wrong documents out of 20 with the old body, 0 out of 20 with `append`.
- The reporting project (icond) verified the same through the real service (`site.getService('pdf').joinPdf`) on a live instance: 0 wrong documents out of 20 after the change.
- Full `gnrpy` suite green here (2159 passed, 10 skipped).

## Notes for the reviewer

1. **`append` also imports the source documents' outlines**, so the merged pdf now carries a bookmark tree it did not have before. If that is unwanted the call takes `import_outline=False` — deliberately not passed here, because the parameter was named `import_bookmarks` in PyPDF2 2.x and the repo still supports it.
2. **`PdfReader` is no longer used in the module**, so it is dropped from the import (flake8 F401 otherwise). `HAS_PYPDF` still probes the same package.
3. **`joinPdf_FITZ` is broken independently of this** and is left alone on purpose: it opens the *output* node for reading and passes its bytes as a filename to `save`, and `doc.insert(op)` is not a PyMuPDF API (`insert_pdf` is). Nothing reaches it today because `joinPdf` prefers PyPDF whenever it imports, but the pypdf → PyMuPDF migration will. Described at the end of #1136; it deserves its own issue and PR rather than a silent fix here.
4. **Candidate for the hotfix path.** The defect ships wrong documents to end users with no error anywhere, and the icond instance carries a local workaround in the meantime. Opened on `develop` per the default rule — say if you want it on `master` as well (the `hotfix` label) and I will redo it from there.

Reported by Andrea Busanelli (icond project).


cgabriel (approved):
Verified with the provided script and by using the patched service, appending is working 100% of test executions.

genro (approved):
The fix is right: `PdfWriter.append` → `merge` iterates `reader.pages` in order and keeps the source reader alive through `original_page` until `write`, so the page mix-up cannot recur. Approving on that basis.

Two non-blocking notes:

- The root-cause paragraph in the body does not match PyPDF2 3.0.1 (the installed version). `_add_page` deep-clones the page (`_writer.py:271`), so the cloned objects never go through the hash comparison in `_resolve_indirect_object` (`:1157` returns early on `data.pdf == self`). What actually bit is the translation table `_id_translated[id(self.pdf)][idnum]` (`_base.py:241-245`): the old loop rebound `input_pdf = PdfReader(...)` per file, the previous reader was freed, a new reader could get the same `id()`, and objects with an already-mapped idnum were resolved to the previous file's objects. That explains the intermittency and the `cannot find XObject/ExtGState resource` errors. Worth correcting in the body or in #1136 so the record is accurate.
- `PyPDF2` is unpinned in `gnrpy/pyproject.toml:46` and `sys/requirements.txt:3`; I did not verify that `PdfWriter.append` exists on the 2.x releases the pin still admits.

Issue #1141, opened 2026-08-24
[FEATURE] Run "Export to Sphinx" on the handbook grid selection


dgpaci:
### Summary

Make the `export_to_sphinx` action of `docu.handbook` work on the current grid selection, so that several handbooks can be exported in a single run.

### Problem or Need

The action appears in the view gear menu (it is `tags = '_DEV_'`, so only for developers), but it cannot actually be launched from there, and it can only ever process one handbook.

Today the only working entry point is the form toolbar button `Exp.To Sphinx` (`projects/gnrcore/packages/docu/resources/tables/handbook/th_handbook.py:133`), which passes the pkey of the record on screen:

```python
genro.publish("table_script_run",{table:"docu.handbook", res_type:'action',
                                  resource:'export_to_sphinx',
                                  handbook_id: pkey, ...})
```

and `pre_process` reads exactly that single id (`.../handbook/action/export_to_sphinx.py:40`):

```python
self.handbook_id = self.batch_parameters['extra_parameters']['handbook_id']
```

The gear menu goes through a different path: `th_slotbar_resourceActions` (`resources/common/th/th_view.py:1072`) fires `th_batch_run`, and `th_view_batch_caller` (`resources/common/th/th.js:27`) merges `grid.currentSelectionPars()` — i.e. `selectionName` / `selectedPkeys` — into the batch call. It never sets `extra_parameters`, and `table_script_run` only forwards that key when it is truthy, so the batch launched from the gear dies immediately on `KeyError: 'extra_parameters'`.

So the framework already hands the batch the selection, and the batch throws it away:

* the gear entry is dead weight — it looks runnable but is not;
* exporting N manuals means opening N forms and pressing the button N times, waiting for each full sphinx build.

### Proposed Solution

Turn the batch into a selection-aware one, the way other gnrcore actions already do:

* `pre_process` resolves the list of handbooks from the selection the framework provides (`self.get_selection()` / `get_selection_pkeys()`, see `gnrpy/gnr/web/batch/btcbase.py:239`), falling back to `extra_parameters['handbook_id']` so the form button keeps working on the current record;
[...]

Issue #1140, opened 2026-08-24
[BUG] docu: handbook logo preference uploads to home: instead of documentation: storage


dgpaci:
### Problem Description

The handbook logo uploaded from the `docu` application preferences is stored on the **`home:`** storage instead of the **`documentation:`** storage.

In `projects/gnrcore/packages/docu/resources/preference.py:47-51` the image field is declared as:

```python
fb.img(src='^.logo', lbl='!![en]Handbook Logo', width='200px', height='100px',
                    crop_width='200px', crop_height='100px',
                    edit=True, placeholder=True,
                    upload_filename='docu_logo.png',
                    upload_folder='home:documentation/img')
```

Every other docu asset already lives on the `documentation:` storage:

- handbook OGP image → `upload_folder='documentation:handbooks_images'` (`th_handbook.py:86`)
- documentation attachments → `documentation:attachments/<pkey>` (`model/documentation.py:156`)

**Observed behavior:** the logo ends up under `home:documentation/img/docu_logo.png`, i.e. on the instance home storage, which is normally a local filesystem folder.

**Expected behavior:** the logo must be stored on the documentation storage, e.g. `documentation:img/docu_logo.png`, so that it shares the same storage service as all the other documentation media. When `documentation:` is mounted on a remote service (e.g. an S3 bucket dedicated to documentation), the `img` folder is created inside that bucket, and the logo can be published/served together with the rest of the documentation assets (public URLs, exported handbooks) instead of being stranded on the instance's local home.

### Reproduction Steps

1. On a sandbox instance with the `docu` package installed, configure a `documentation` storage service pointing somewhere distinct from `home` (a different local path is enough to observe the problem, an S3 bucket makes it evident).
2. Open the application preferences → `Documentation` → tab `Handbooks theme`.
3. Upload an image in the `Handbook Logo` field.
4. Inspect the storages.

Observed result: the file is written to `home:documentation/img/docu_logo.png`; nothing is created on the `documentation:` storage.
Expected result: the file is written to `documentation:img/docu_logo.png`.

### If It Cannot Be Reproduced Externally

- Module involved: `docu` package, application preferences pane (`prefpane_docu` → `handbooksTheme`).
- The stored preference value is the storage path itself, so any change must consider preferences already saved with the old `home:documentation/img/...` path (the value is resolved through `site.externalUrl()` in `export_to_sphinx.py:89-90` when generating `html_logo` for the Sphinx build).

### Environment

[...]

PR #1142, opened 2026-08-24
docu: handbook logo on the documentation storage, export to sphinx on the grid selection

dgpaci:
Two docu issues that meet in the same publishing area, plus the removal of a
notification channel the batch no longer needs to own.

> **Heads-up:** the third commit removes the Telegram notification of the export
> (and the docu preferences gating it). Notifications are configured through
> [gnrxadm](https://github.com/softwellsrl/gnrxadm) now. Nothing else in the tree
> referenced it.

## #1140 — the handbook logo lives on the documentation storage

The logo was the only docu asset still uploaded on the instance home
(`home:documentation/img`), while the attachments of the pages, their images and
the preview image of the handbook all live on `documentation:`. Pointing that
storage at a bucket of its own left the logo behind on the instance filesystem.

- `handbooksTheme` now uploads to `documentation:img`.
- the sphinx conf resolves `html_logo` through `logoUrl`, which reuses the
  public-media gate the images of the build already go through
  (`documentation.publicMediaUrl`, from #1085): the stable public url when the
  service publishes on a public base of its own, the instance-served url
  otherwise.

**No migration needed**: the preference holds the url of the uploaded file, not a
name resolved against the upload folder, so logos already sitting under
`home:documentation/img` keep resolving where they are.

## #1141 — Export to Sphinx runs on the grid selection

The gear menu entry looked runnable but died on `KeyError: 'extra_parameters'`:
`th_view_batch_caller` hands the batch `selectionName`/`selectedPkeys`, never
`extra_parameters`, and the batch read only
`batch_parameters['extra_parameters']['handbook_id']`.

- `handbookIds()` resolves the handbooks from the selection the framework already
  delivers (`get_selection_pkeys`), falling back to `handbook_id` so the form
  toolbar button — and any other caller passing it — keeps working;
- the per-handbook state moves from `pre_process` into `prepareHandbook`, rebuilt
  for each manual;
- `call_steps` wraps the base step machinery in the loop over the selection
  (`super().call_steps(offset=offset + 1)`, so the base line bookkeeping starts
  below the new handbook thermo line) and `post_process` becomes
  `publishHandbook`, run once per handbook;
- a handbook that fails is logged, rolled back (`rollbackAll`, or an aborted
  transaction would take the remaining manuals down with it on PostgreSQL) and
  reported in the batch result; the run continues with the others.
  `GnrBatchStoppedException` still propagates, so stopping the batch stops it.
  `prepareHandbook` is inside that guard together with the steps: the reachable
  failures of an export are in the setup - a handbook whose tree has nothing
  published, the storage node erased before the build, a pkey that no longer
  resolves - and `handbookFailed` names the handbook from the pkey the loop is
  iterating on, the only thing it still has when the setup is what failed.

An empty run (neither a selection nor `handbook_id`) now answers
`No handbook to export` instead of raising.

## Telegram notification removed

`sendHandbookNotification`/`sendNotification`, the parameters dialog asking for
bot and message, and the `telegram_notification`/`bot_token` docu preferences are
gone, along with their four orphaned entries in `localization.xml`. The batch has
no parameters of its own left, so `hasParameters` is now false and the dialog only
carries the scheduling section.

## Test case

`gnr dev tests gnrdevelop docu` — **74 passed** (53 before, 21 new).

New — `packages/docu/tests/test_docu_handbook_logo.py` (5):
- the preference pane declares `upload_folder='documentation:img'` (built through
  a real `GnrDomSrc` structure);
- a documentation storage with a public base is linked by its unsigned public url;
- a plain storage and a private bucket keep the instance-served url;
- a logo already on `home:documentation/img` still resolves.

New — `packages/docu/tests/test_export_to_sphinx_selection.py` (16): selection vs
`handbook_id` precedence, every handbook of a selection published with its own
state, thermo captions and the step line surviving each step, the zip of a local
handbook, failure isolation and reporting - injected both in a step and in
`prepareHandbook`, first handbook of the selection included - stop propagation. Real database, real
local storage services, real framework `call_steps`/`thermo_wrapper`/
`recordToUpdate`; only the sphinx build is replaced by steps writing a
recognizable file.

Shared test infra in `packages/docu/tests/sitestub.py` grew `PageStub`/
`PageStoreStub` (the hierarchical resolvers ask the current page for a db, a uuid
and a store), `zipFiles`, and the two signing services `test_docu_media_export.py`
used to define on its own.

`cd gnrpy && pytest` — 2131 passed, 69 skipped.

### Manual check to run

1. Configure `documentation` as a storage distinct from `home`, upload a logo in
   preferences → `Documentation` → `Handbooks theme`, verify it lands in
   `documentation:img/docu_logo.png`.
2. Select several handbooks in the view and run `Export to Sphinx` from the gear
   menu: every manual is built, the thermometer names each one, and the toolbar
   button on a single form still behaves as before.

### Notes for the reviewer

- Erasing the published folder before the build is pre-existing behaviour
  (`prepareHandbookFolder`): a failing export still empties that handbook's
  published folder — now per handbook rather than for the whole run.
- `dialog_height='450px'`/`dialog_width='650px'` were sized for the notification
  form and are now oversized for a dialog carrying only the scheduling section. I
  left them alone rather than guess at a size I cannot verify here.
- The `is_published` formula the page-tree lookup filters on is written in
  PostgreSQL (`ILIKE` over an `EXISTS` subquery) and cannot run on the sqlite test
  database: the selection tests wrap the batch's `doctable` in `PortableDocTable`,
  which reads the tree without that condition. Production code is untouched.


fporcari (changes requested):
The diagnosis of #1141 is the part worth keeping: `th_view_batch_caller` hands the batch
`selectionName`/`selectedPkeys` and never `extra_parameters`, so the gear menu entry
could only ever raise — and letting the selection win with `handbook_id` as the fallback
keeps the form toolbar button working without a second entry point. Three details are
right in a way that is easy to get wrong:

- `super().call_steps(offset=offset + 1)`. The base at `gnrpy/gnr/web/batch/btcbase.py:141`
  deletes `self.btc.line_codes[offset:]` after each step, so without the shift the new
  handbook thermo line would be torn down on the first step. Verified against the base,
  not taken from the comment.
- `rollbackAll()` rather than a plain rollback: on PostgreSQL an aborted transaction
  would take the remaining manuals of the selection down with it, which is exactly the
  failure the loop is trying to contain.
- committing each handbook as soon as it is published, so a later failure leaves what is
  already online registered on its record.

Verified rather than trusted: `get_selection_pkeys` (btcbase.py:320) returns a pkey list
or `[]`; `thermo_wrapper` calls `cb_message(item, progress, maximum, **kwargs)`, which is
the signature `handbookCaption` declares; `btc.exception_stopped` is
`GnrBatchStoppedException`; `hasParameters` is derived from
`hasattr(self, 'table_script_parameters_pane')` in `batch_handler.py:101`, so dropping
the method really does turn it off — it is not a flag anybody has to remember to flip;
`self.handbook_url` moved into `prepareHandbook` and is rebuilt per handbook. The three
deletion justifications hold: no Telegram identifier survives anywhere in the docu
package, and none of the four removed lockeys is referenced in any `.py`, `.xml` or `.js`
in the tree. flake8 clean, `pytest projects/gnrcore/packages/docu/tests` gives 72 passed
here too.

## The failure isolation does not cover its most likely trigger

`resources/tables/handbook/action/export_to_sphinx.py:68` — `prepareHandbook` is called
*outside* the `try`:

```python
for handbook_id in self.btc.thermo_wrapper(self.handbook_ids, ...):
    self.prepareHandbook(handbook_id)      # <- outside
    try:
        super().call_steps(offset=offset + 1)
        self.publishHandbook()
    except self.btc.exception_stopped:
        raise
    except Exception as e:
        self.handbookFailed(e)
```

So anything raised while a handbook's state is being set up escapes `call_steps` and ends
the whole run — the handbooks left in the selection are never exported, which is the
opposite of what the body promises. I put it under test rather than reasoning about it: a
probe that makes `prepareHandbook` fail for `beta` on a selection of
`alpha, beta, gamma` ends with `RuntimeError` propagating out of the batch and `gamma`
never built. The three existing isolation tests all inject the failure in
`step_fakeSphinxBuild`, i.e. inside the `try`, so none of them can see this.

It matters because `prepareHandbook` is where the reachable failures are, not the steps:

- `getHierarchicalData(root_id=..., condition='$is_published IS TRUE')['root']['#0']` on a
  handbook whose tree has nothing published — a `Bag` miss returns `None` and `None['#0']`
  raises;
- `prepareHandbookFolder()`, which resolves the handbook storage node and, as you note
  yourself, **erases the published folder before the build**. On a selection that means
  one handbook can lose what was online *and* take the rest of the run with it;
- a pkey that no longer resolves.

Two lines to close it, and the second is not optional:

```python
for handbook_id in self.btc.thermo_wrapper(self.handbook_ids, 'btc_handbooks',
                                           message=self.handbookCaption):
    try:
        self.prepareHandbook(handbook_id)
        super().call_steps(offset=offset + 1)
        self.publishHandbook()
    except self.btc.exception_stopped:
        raise
    except Exception as e:
        self.handbookFailed(handbook_id, e)
```

`handbookFailed` currently opens with `name = self.handbook_record['name']`, so with
`prepareHandbook` moved inside the `try` a failure on the *first* handbook would raise
`AttributeError` there instead of being reported. Take the name from `handbook_id` the way
`handbookCaption` already does:

```python
def handbookFailed(self, handbook_id, error):
    name = self.tblobj.readColumns(columns='$name', pkey=handbook_id) or handbook_id
```

A test injecting the failure in `prepareHandbook` rather than in a step would pin the
half of the guarantee that is currently unpinned.

Not verified: I did not run the two manual checks — no documentation storage on a public
base and no real sphinx build here, so the logo resolution is covered only by the five
new tests, and the multidomain path is untouched by them.

dgpaci:
Both lines taken, in 6a1bca3ef — the second one was needed exactly where you said it would be.

`prepareHandbook` is now inside the `try`, and `handbookFailed(handbook_id, error)` reads the
name from the pkey through a `handbookName` helper that `handbookCaption` shares, since it was
already resolving the name the same way:

```python
    def handbookName(self, handbook_id):
        return self.tblobj.readColumns(columns='$name', pkey=handbook_id) or handbook_id
```

Two tests pin the half that was unpinned, injecting the failure in `prepareHandbook` instead of
in a step. `_batch` grew a `broken_setup` argument next to `broken`, and the injection reads the
name with `readColumns` rather than through the new helper, so the tests do not lean on the code
they are pinning:

- `test_a_handbook_failing_its_setup_does_not_abort_the_others` — `beta` broken in the middle of
  `alpha, beta, gamma`: `alpha` and `gamma` come out with a fresh `last_exp_ts`, `beta` keeps its
  own, and the result reads `Exported 2 of 3 handbooks, failed: beta`;
- `test_the_first_handbook_failing_its_setup_is_reported_by_name` — the same failure on the
  *first* handbook of the selection, where there is no export state at all yet.

Both were run against the reviewed code before the fix: they fail with the `RuntimeError`
propagating out of the batch, which is your probe. With only the first line of the fix applied
(setup inside the `try`, `handbookFailed` still on `self.handbook_record['name']`) the second one
fails with `AttributeError: 'SelectionExport' object has no attribute 'handbook_record'`, so it
holds down the naming half on its own.

The assertions of the two new tests go through `last_exp_ts` rather than the published folders:
the temp storages are shared by the whole class, and `test_the_user_stopping_the_batch_stops_the_run`
reads `not self._published('beta').exists`, which only holds because nothing else in the file
publishes `beta`. Neither new test publishes it.

flake8 clean on both files, `cd gnrpy && pytest` still 2131 passed / 69 skipped.

One difference from your run worth flagging, unrelated to this PR: here
`pytest projects/gnrcore/packages/docu/tests` gives **66 passed, 8 errors**, the errors all being
`TestDocuExamplesSync` failing its `setup_class` with `Exception: Missing genro configuration`.
`test_examples_sync.py` passes on its own (8 passed) and errors as soon as any other docu test
file that builds a `GnrApp` runs before it — including `test_docu_media_export.py` alone, and
including on an untouched `origin/develop` worktree, with or without the per-worktree
`GENRO_GNRFOLDER`. So it is a pre-existing interference in the docu suite on this machine, not
something this branch introduces; the 74 in the body is 72 + the 2 new tests, matching your count
rather than mine. Happy to open a separate issue for it.

fporcari (approved):
Both lines are in, and the second one is doing exactly the work it was asked to do.

`export_to_sphinx.py:68-79` — `prepareHandbook` is now the first statement inside the `try`,
next to `super().call_steps()` and `publishHandbook()`, so a handbook that fails resolving its
tree, its storage node or its pkey is contained like a handbook that fails its build. The
`except self.btc.exception_stopped: raise` clause still sits above the generic one, so a stop
is not swallowed as a per-handbook failure — and it now covers a stop raised from the setup too,
for free.

`export_to_sphinx.py:92-101` — `handbookFailed(handbook_id, error)` names the handbook through
`handbookName(handbook_id)` (`:85-90`), a `readColumns` on the pkey. The helper is genuinely
shared: `handbookCaption` at `:81-83` delegates to it and keeps its
`(handbook_id, progress, maximum, **kwargs)` signature, which is the one
`thermo_wrapper` calls `cb_message` with — the refactor did not touch the callback contract.

## The two new tests do pin the fix — verified by reverting, not by reading

`test_export_to_sphinx_selection.py:311-337`. The injection goes through a `broken_setup`
argument of `_batch` that overrides `prepareHandbook` and reads the name with its own
`readColumns` rather than through `handbookName`, so the test does not lean on the code it pins.
Both run on the real database, the real `call_steps`/`thermo_wrapper` and the real per-handbook
state.

Reverted the production hunk of `6a1bca3ef` in a worktree, kept the tests:

- `test_a_handbook_failing_its_setup_does_not_abort_the_others` — `RuntimeError: the setup of
  beta failed` propagating out of `call_steps` at `export_to_sphinx.py:68`, `gamma` never built;
- `test_the_first_handbook_failing_its_setup_is_reported_by_name` — same.

Then re-applied the fix but put `handbookFailed` back on `self.handbook_record['name']`, to check
the naming half is pinned on its own:

- the first-handbook test fails with `AttributeError: 'SelectionExport' object has no attribute
  'handbook_record'` at `:97`, as expected;
- the other one fails too, on `assert ['alpha'] == ['beta']` — the middle-of-the-selection case
  reports the *previous* handbook's name when the naming half is missing. So the pair pins the
  fix more tightly than claimed: one catches the misnaming, the other the crash.

Restored, tree clean.

## Numbers on this machine

`pytest projects/gnrcore/packages/docu/tests` — **80 passed** (head `4d7d9fbe6`, with `develop`
merged in, per-worktree `GENRO_GNRFOLDER`). No `TestDocuExamplesSync` errors here, the whole
directory runs green in one process; that interference looks local to your setup, and a separate
issue for it is the right place if it persists.

`flake8` on the seven touched `.py` files — clean, zero findings.

Not verified: no documentation storage on a public base and no real sphinx build here, so the
logo resolution and the multidomain path are still covered only by the five new logo tests, and
the two manual checks in the body remain to be run on a real instance.

Approving.

PR #1143, opened 2026-08-24
fix(security): sanitize_js switch was unreadable and stripping was silent

fporcari:
## Problem

The XSS sanitization introduced in #774 was meant to be **opt-in** (`sanitize_js` switch in instanceconfig, default OFF), but the client check never worked:

- `genro.getData('gnr.switches?sanitize_js')` ignores the `?attr` suffix and returns the **node value**, which is always a truthy Bag (set unconditionally by `gnrwebpage.py`).
- Result: sanitization was **always active** in every instance, and any configured switch value was ignored (both directions).
- The stripping was also **silent**: applications relying on inline `on*` handlers in grid-cell HTML broke with no diagnostic (found as a production regression in a client app: action buttons inside grid cells stopped working, the click only selected the row).

The feature's own test page passed only because of this accident (it set the switch as a node *attribute*, which was equally unreadable).

## Fix

- The switch is now a real item, `gnr.switches.sanitize_js`, fed from **both** instanceconfig shapes: `<switches sanitize_js="f"/>` (attributes) and `<switches><sanitize_js>f</sanitize_js></switches>` (child items).
- Sanitization is **ON by default** (deliberately, not by accident). A falsy value (`f`/`false`/`F`/`0`/false) is an explicit, temporary opt-out for legacy applications that still need inline handlers while they migrate.
- When active sanitization strips content it now emits a **console warning** — once per distinct content, with migration guidance (format_isbutton/format_onclick, delegated onCellClick, whitelisted template/js cells, or the opt-out) — plus a toast/error entry in developer mode (HTML-escaped, so the warning itself cannot re-inject the stripped payload).

## Tests

- `test_sanitize_js.py` updated to the new switch shape; new `test_4` section covers switch handling, every falsy opt-out value, default-ON with missing switch, and warning deduplication (7/7 pass in sandbox).
- Verified end-to-end in a sandbox instance: default → inline `onclick` stripped **with** console warning + dev toast; `<switches sanitize_js="f"/>` in instanceconfig → content untouched and handlers working.
- Full gnrpy suite green (2159 passed, 10 skipped).


cgabriel (commented):
It's looking good to me, but the test page confuses me.

The test 4 runs as:

Image

But I get this:

Image

Also, on the first opening of the page page, I get:

Image


It's not clear to me this is a feature and intended behaviour or what.

fporcari:
Yes — this is intended developer-mode behavior. The complete fixture deliberately renders nine distinct active-HTML payloads, so each one is sanitized and reported visibly. A console warning alone would be too easy to miss when sanitization has just disabled behavior produced by a `dataController`; the red developer errors are meant to expose that regression immediately and include the stripped content needed to locate it.

I reproduced the same full-page result on `sandboxpg`, while test 4 in isolation passes 7/7. The isolated test verifies switch semantics and deduplication; the full page verifies that real rendered payloads are both stripped and made conspicuous. I therefore do not plan a code change for this screenshot.

Not verified: the opt-out path in a separate legacy application during this pass; the PR body records that earlier end-to-end check.

cgabriel:
> Yes — this is intended developer-mode behavior. The complete fixture deliberately renders nine distinct active-HTML payloads, so each one is sanitized and reported visibly. A console warning alone would be too easy to miss when sanitization has just disabled behavior produced by a `dataController`; the red developer errors are meant to expose that regression immediately and include the stripped content needed to locate it.

Ok, thank you for providing the details.

cgabriel (approved):

genro (approved):
Reading both shapes of `<switches>` (children plus attributes, `gnrwebpage.py:2380-2382`) is the right fix, and the default stays ON when the node is absent, so instances without the switch keep today's behaviour. The dedup on the first 200 chars with a cap is a sensible anti-flood. Test page is a real testhandler page.

Two non-blocking notes:

- The opt-out at `genro.js:328` recognises only `false, 0, 'f', 'false', 'F'`. `"False"`, `"no"`, `"off"` leave sanitization ON with no warning. `String(sw).toLowerCase()` against `('f','false','0','n','no','off')` would close that.
- After 100 distinct payloads (`genro.js:344`) stripping goes silent again. A grid whose `onclick` embeds the row id hits the cap quickly. Acceptable as anti-flood, worth knowing.

PR #1145, opened 2026-08-24
fix(datastore): trigger a set with attributes only on a real change

### LINKED ISSUE #1144: [BUG] Grid selectedId always fires the ^ trigger, even when id and row attributes are unchanged
## Problem Description

Selecting a row in a grid always fires the trigger of any `^` datapath bound to `selectedId`, even when the very same row is clicked again and nothing about it has changed.

`selectedId` is written by `gnrUpdateSelect()` with the whole row as node attributes, so that `^grid.selectedId?somefield` works:

```js
this.sourceNode.setAttributeInDatasource('selectedId', selectedId, null, row, true);
```

The last argument is `forceChanges`: `setAttributeInDatasource()` compares the *value* only, so the only way to keep the attributes fresh (a row edited in place must republish its fields) was to force the write, and with it the trigger. Consequence: every click on the already selected row re-runs every `dataController`, `dataRpc` and `dataFormula` listening on that path — repeated server calls, forms reloaded under the user's hands, flickering panes.

The same blind spot affects the whole "set a value together with its attributes" family, in the opposite direction: `genro.setData()`, `sourceNode.setRelativeData()` and `setItem(..., {lazySet:true})` skip the write when the value is equal *ignoring the attributes*, so an attribute-only change is silently dropped and its trigger never fires.

Expected behaviour, in one rule: a set carrying attributes must trigger when the value **or** the attributes change, and must stay silent when neither does.

## Reproduction Steps

1. Put a grid on a page with `selectedId='.selected_id'` and a `dataController` on `^.selected_id` (an `alert`, or a counter).
2. Click a row: the controller runs (correct).
3. Click the same row again, or anything that re-publishes the selection (`updatedSelectedRow`): the controller runs again, with the identical id and identical row data.
4. Symmetric case: `genro.setData('.x', sameValue, {label:'changed'})` — the new attributes are dropped and no trigger fires at all.

## Environment

- Version: `develop` (client side, `gnrjs/gnr_d11`)
- OS: any
- Browser: any

## Additional Information

Involved code:
- `gnrjs/gnr_d11/js/gnrdomsource.js` — `setAttributeInDatasource()`, the `forceChanges` flag
- `gnrjs/gnr_d11/js/gnrbag.js` — `GnrBag.set()`, the `lazySet` value-only comparison
- `gnrjs/gnr_d11/js/genro_grid.js` — `gnrUpdateSelect()` and the `updatedSelectedRow` subscription

Note that some grids hand out the live store row (`node.attr`) as `rowByIndex()`, so the attributes stored in the datastore node can be the *same object* as the row: any comparison is meaningless unless the datastore keeps its own copy.

fporcari:
Fixes #1144

## What was wrong

`selectedId` is published with the whole row as node attributes, so `^grid.selectedId?field` works. `setAttributeInDatasource()` compared the value alone, so the only way to keep those attributes fresh was `forceChanges`, which wrote (and triggered) unconditionally: clicking the already selected row, or any re-publication of the same selection, re-ran every controller bound to that path.

The same value-only comparison hurt in the opposite direction everywhere else: `genro.setData()`, `setRelativeData()` and `setItem(..., {lazySet:true})` skipped the write when the value was equal, silently **dropping** an attribute-only change and its trigger.

## The rule now implemented

A set carrying attributes has exactly three outcomes:

| value | attributes | result |
|---|---|---|
| same | same | nothing written, no trigger |
| same | changed | attributes updated, only the attribute listeners notified (`^path?attr`) |
| changed | any | value and attributes updated, everybody notified |

The middle row is what the annoying case needed: after an inline edit of the selected row the attribute watchers get the fresh row, while a form bound to `^grid.selectedId` does not reload for an id that never changed.

## Changes

- `gnrlang.js`: `attrIsUnchanged(currattr, newattr, updattr)` — tells whether applying `newattr` would change anything, honouring both the replace and the update semantics of `setAttr()`, and comparing values with `isEqual` (Date and Bag included).
- `gnrbag.js` — `GnrBag.set()`: the `lazySet` comparison covers the attributes; same value with changed attributes goes through `setAttr()`, which notifies the attribute listeners without faking a value change.
- `gnrdomsource.js` — `setAttributeInDatasource()`: `forceChanges` is gone, the write is lazy, and the attributes are copied before being stored, because a grid can hand out the live store row (`node.attr`) and an aliased object makes any comparison blind. An attribute that was never published no longer materializes an empty datanode.
- `genro_grid.js`: the two `selectedId` publications drop the `forceChanges` argument.

## Test case

New test page `test/datastore/attribute_trigger` (no database needed):

- **test_0** — a `bagGrid` with `selectedId`, two counters (`^...selectedId` and `^...selectedId?name`), plus buttons that republish the very same selection and that rename the selected row.
- **test_1** — `genro.setData()` with attributes: same value/same attribute, same value/new attribute, new value.

Verified on a running instance:

| action | value triggers | attribute triggers |
|---|---|---|
| first click on a row | 1 | 1 |
| same row clicked again / selection republished | 1 | 1 |
| selected row renamed, then republished | 1 | 2 |
| `setData` same value same tag (twice) | 1 | 1 |
| `setData` same value new tag | 1 | 2 |
| `setData` new value | 2 | 3 |

Regression pass on the same instance: `plainTableHandler` and `borderTableHandler` (selection loads the form, switching rows loads the related rows), inline editing on a `bagGrid`, formatted/masked values in `dojo/displayed_value`. `pytest core web sql app xtnd`: 2159 passed, 10 skipped.

**On the test counts across runs.** Nothing in this diff touches Python, so the executed/skipped split follows whichever services are up when the suite runs, not the diff: `2159 passed, 10 skipped` and `2100 passed, 69 skipped` have the same total and differ by 59 tests moving from executed to skipped. Neither is a regression against the other.

### DIFF
--- a/gnrjs/gnr_d11/js/genro.js
+++ b/gnrjs/gnr_d11/js/genro.js
@@ -1581,7 +1581,7 @@ dojo.declare('gnr.GenroClient', null, {
                     var inner=dpath.slice(registered_path.length);
                     if(!inner || inner[0]=='.'){
                         genro._serverstore_changes = genro._serverstore_changes || {};
-                        genro._serverstore_changes[genro._serverstore_paths[registered_path]+inner] = asTypedTxt(kw.value);
+                        genro._serverstore_changes[genro._serverstore_paths[registered_path]+inner] = asTypedTxt(kw.node._value);
                         break;
                     }
                 }
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -889,7 +889,7 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
         sourceNode.subscribe('updatedSelectedRow',function(){
             var selectedIndex = this.widget.selection.selectedIndex;
             this.widget.sourceNode.setAttributeInDatasource('selectedId', this.widget.rowIdByIndex(selectedIndex), 
-                                                                null, this.widget.rowByIndex(selectedIndex), true);
+                                                                null, this.widget.rowByIndex(selectedIndex));
         });
         sourceNode.subscribe('configuratorPalette',function(){
             this.widget.configuratorPalette();
@@ -1504,7 +1504,7 @@ dojo.declare("gnr.widgets.DojoGrid", gnr.widgets.baseDojo, {
                             selNodes, {'count':selNodes?selNodes.len():0});
         }
         if(this.sourceNode.attr.selectedId) {
-            this.sourceNode.setAttributeInDatasource('selectedId', selectedId, null, row, true);
+            this.sourceNode.setAttributeInDatasource('selectedId', selectedId, null, row);
         }
         if(this.sourceNode.attr.selectedRowData){
             var rowDataBag = new gnr.GnrBag(row);
--- a/gnrjs/gnr_d11/js/gnrbag.js
+++ b/gnrjs/gnr_d11/js/gnrbag.js
@@ -1747,11 +1747,11 @@ dojo.declare("gnr.GnrBag", null, {
                     label = splittedlabel[0];
                     var attr = splittedlabel[1];
                     var node = obj.getNode(label, false, true);
-                    if(kwargs.lazySet && isEqual(node.attr[attr],value)){  
-                        return;
-                    }
-                    var auxattr = {};
+                    var auxattr = objectUpdate({}, _attributes);
                     auxattr[attr] = value;
+                    if(kwargs.lazySet && !changedAttrKeys(node.attr,auxattr,'*').length){
+                        return node;
+                    }
                     var _doTrigger = true;
                     if ('doTrigger' in kwargs) {
                         _doTrigger = kwargs.doTrigger;
@@ -1816,11 +1816,15 @@ dojo.declare("gnr.GnrBag", null, {
                 node.setResolver(resolver);
             }
             if(kwargs.lazySet && isEqual(node._value,value)){
-                if(attrIsUnchanged(node.attr,_attributes,_updattr)){
+                var changedAttrs = changedAttrKeys(node.attr,_attributes,_updattr);
+                if(!changedAttrs.length){
                     return node;
                 }
-                //same value, changed attributes: only the attribute listeners have to be notified
-                node.setAttr(_attributes, _doTrigger, _updattr);
+                //same value, changed attributes: only the attribute listeners have to be notified.
+                //changedAttr comes from here because setAttr cannot default it without changing the
+                //callers that read its absence as a whole-attribute change
+                node.setAttr(_attributes, _doTrigger, _updattr,
+                             changedAttrs.length==1?changedAttrs[0]:null);
                 return node;
             }
             node.setValue(value, _doTrigger, _attributes, _updattr,kwargs.fired);
--- a/gnrjs/gnr_d11/js/gnrbag.js
+++ b/gnrjs/gnr_d11/js/gnrbag.js
@@ -1815,10 +1815,13 @@ dojo.declare("gnr.GnrBag", null, {
             if (resolver) {
                 node.setResolver(resolver);
             }
-            if(kwargs.lazySet){
-                if(isEqual(node._value,value)){
+            if(kwargs.lazySet && isEqual(node._value,value)){
+                if(attrIsUnchanged(node.attr,_attributes,_updattr)){
                     return node;
                 }
+                //same value, changed attributes: only the attribute listeners have to be notified
+                node.setAttr(_attributes, _doTrigger, _updattr);
+                return node;
             }
             node.setValue(value, _doTrigger, _attributes, _updattr,kwargs.fired);
             return node;
--- a/gnrjs/gnr_d11/js/gnrbag.js
+++ b/gnrjs/gnr_d11/js/gnrbag.js
@@ -1749,14 +1749,15 @@ dojo.declare("gnr.GnrBag", null, {
                     var node = obj.getNode(label, false, true);
                     var auxattr = objectUpdate({}, _attributes);
                     auxattr[attr] = value;
-                    if(kwargs.lazySet && !changedAttrKeys(node.attr,auxattr,'*').length){
+                    var changedAttrs = changedAttrKeys(node.attr,auxattr,'*');
+                    if(kwargs.lazySet && !changedAttrs.length){
                         return node;
                     }
                     var _doTrigger = true;
                     if ('doTrigger' in kwargs) {
                         _doTrigger = kwargs.doTrigger;
                     }
-                    node.setAttr(auxattr, _doTrigger, '*', attr);
+                    node.setAttr(auxattr, _doTrigger, '*', changedAttrs.length>1 ? null : attr);
                     return node;
                 }
                 else {
--- a/gnrjs/gnr_d11/js/gnrbag.js
+++ b/gnrjs/gnr_d11/js/gnrbag.js
@@ -1757,7 +1757,8 @@ dojo.declare("gnr.GnrBag", null, {
                     if ('doTrigger' in kwargs) {
                         _doTrigger = kwargs.doTrigger;
                     }
-                    node.setAttr(auxattr, _doTrigger, '*', changedAttrs.length>1 ? null : attr);
+                    node.setAttr(auxattr, _doTrigger, '*',
+                                 changedAttrs.length>1 ? null : (changedAttrs.length ? changedAttrs[0] : attr));
                     return node;
                 }
                 else {
--- a/gnrjs/gnr_d11/js/gnrdomsource.js
+++ b/gnrjs/gnr_d11/js/gnrdomsource.js
@@ -510,13 +510,26 @@ dojo.declare("gnr.GnrDomSourceNode", gnr.GnrBagNode, {
     setAttributeInDatasource: function(attrname, value, doTrigger, attributes) {
         doTrigger = (doTrigger == null) ? this:doTrigger;
         var path = this.attrDatapath(attrname);
-        if (path == null || (value == null && !attributes && !genro._data.getNode(path))) {
+        if (isBag(attributes)) {
+            attributes = attributes.asDict();
+        }
+        if (path == null || (value == null && !objectNotEmpty(attributes) && !genro._data.getNode(path))) {
             return; //nothing to publish: an attribute never set needs no datanode
         }
-        if (attributes) {
-            // attributes can be a live store row: without a copy the datastore node
-            // would alias it and no comparison could ever detect a change
-            attributes = objectUpdate({}, attributes);
+        if (objectNotEmpty(attributes)) {
+            // attributes can be a live store row: without a copy of its container values too
+            // the datastore node would alias them and no comparison could detect a change
+            var snapshot = {};
+            for (var k in attributes) {
+                var v = attributes[k];
+                if (isBag(v)) {
+                    v = v.deepCopy();
+                } else if (v != null && v.constructor === Object) {
+                    v = objectUpdate({}, v);
+                }
+                snapshot[k] = v;
+            }
+            attributes = snapshot;
         }
         genro._data.setItem(path, value, attributes, {'doTrigger':doTrigger,'lazySet':true});
     },
--- a/gnrjs/gnr_d11/js/gnrdomsource.js
+++ b/gnrjs/gnr_d11/js/gnrdomsource.js
@@ -507,16 +507,18 @@ dojo.declare("gnr.GnrDomSourceNode", gnr.GnrBagNode, {
         }
         
     },
-    setAttributeInDatasource: function(attrname, value, doTrigger, attributes, forceChanges) {
+    setAttributeInDatasource: function(attrname, value, doTrigger, attributes) {
         doTrigger = (doTrigger == null) ? this:doTrigger;
         var path = this.attrDatapath(attrname);
-        var old_value = genro._data.getItem(path);
-        //if (forceChanges){
-        //    genro._data.setItem(path,v,null,{'doTrigger':false});
-        //}
-        if (!isEqual(value,old_value) || (forceChanges && value != null)) {
-            genro._data.setItem(path, value, attributes, {'doTrigger':doTrigger});
+        if (path == null || (value == null && !attributes && !genro._data.getNode(path))) {
+            return; //nothing to publish: an attribute never set needs no datanode
         }
+        if (attributes) {
+            // attributes can be a live store row: without a copy the datastore node
+            // would alias it and no comparison could ever detect a change
+            attributes = objectUpdate({}, attributes);
+        }
+        genro._data.setItem(path, value, attributes, {'doTrigger':doTrigger,'lazySet':true});
     },
     defineForm: function(formId, formDatapath, controllerPath, pkeyPath,kw) {
         this.form = new gnr.GnrFrmHandler(this, formId, formDatapath, controllerPath, pkeyPath,kw);
--- a/gnrjs/gnr_d11/js/gnrlang.js
+++ b/gnrjs/gnr_d11/js/gnrlang.js
@@ -547,24 +547,29 @@ function objectIsEqual(obj1, obj2) {
     }
 }
 
-function attrIsUnchanged(currattr, newattr, updattr) {
+function changedAttrKeys(currattr, newattr, updattr) {
     if (newattr == null) {
-        return true; //a setter without attributes leaves them untouched
+        return []; //a setter without attributes leaves them untouched
     }
     currattr = currattr || {};
+    var changed = [];
     for (var k in newattr) {
-        if (!isEqual(newattr[k], currattr[k])) {
-            return false;
+        if (updattr == '*' && newattr[k] == null) {
+            if (k in currattr) {
+                changed.push(k); //update with '*': a null deletes the attribute
+            }
+        } else if (!(k in currattr) || !isEqual(newattr[k], currattr[k])) {
+            changed.push(k);
         }
     }
     if (!updattr) {
         for (var k in currattr) {
             if (!(k in newattr)) {
-                return false;
+                changed.push(k); //replace semantics: a dropped attribute changes its own path
             }
         }
     }
-    return true;
+    return changed;
 }
 
 function isNullOrBlank(elem){
--- a/gnrjs/gnr_d11/js/gnrlang.js
+++ b/gnrjs/gnr_d11/js/gnrlang.js
@@ -547,6 +547,26 @@ function objectIsEqual(obj1, obj2) {
     }
 }
 
+function attrIsUnchanged(currattr, newattr, updattr) {
+    if (newattr == null) {
+        return true; //a setter without attributes leaves them untouched
+    }
+    currattr = currattr || {};
+    for (var k in newattr) {
+        if (!isEqual(newattr[k], currattr[k])) {
+            return false;
+        }
+    }
+    if (!updattr) {
+        for (var k in currattr) {
+            if (!(k in newattr)) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
+
 function isNullOrBlank(elem){
     return elem === null || elem===undefined || elem === '';
 }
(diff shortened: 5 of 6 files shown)


cgabriel (dismissed):
I've verified the changes using the test page, I don't see any issue with the proposed changes. It would be nice to introduce a unit test suite for the javascript portion of the framework, though.

dgpaci (changes requested):
The rule is the right one and it lands on a mechanism that already existed: `getTriggerReason` (gnrdomsource.js:141) already suppresses value-path subscribers when `updattr && !updvalue`, so the targeted `setAttr()` fits in cleanly. I verified the behaviour on a real th page of a complex project rather than only on the test page.

### Measured

Instrumented the datastore root (`genro._dataroot.subscribe`) on an `anc_base.socio` th view, same actions on both branches:

| action | develop | this branch |
|---|---|---|
| click a row | 1 trigger, `updvalue:true` | 1 trigger, `updvalue:true` |
| click the **same** row again | 1 trigger, `oldvalue === value` | **0 triggers** |
| store reload (selection restored to same pkey) | `id → null` then `null → id` | identical |
| deselect | 1 trigger | identical |
| `setData(path, sameValue, {newAttr})` | **nothing at all** (attribute silently dropped) | attribute written + 1 attribute-only trigger |

So both halves of the fix are confirmed: the spurious re-notification is gone, and the previously dropped attribute-only change now goes through.

### Blocking: the attribute-only trigger has no `changedAttr`

`gnrbag.js:1823` calls `node.setAttr(_attributes, _doTrigger, _updattr)` without the 4th argument, so the emitted kw is:

```
evt:'upd'  updvalue: undefined  updattr: true  changedAttr: undefined  changedAttributes: {_probe: true}
```

The kw knows which attribute changed, but the field every consumer reads is undefined:

- `dataChangeLogger` (genro_frm.js:1780) enters the `kw.value == kw.oldvalue` branch (both undefined) and then `if(kw.updattr && kw.changedAttr && ...)` is false, so it returns: the attribute change is **never recorded in the form's changes logger** and `updateStatus()` never runs. The datastore changed while the form still reports itself clean.
- `doUpdateAttrBuiltObj` (gnrdomsource.js:1383) resolves `oldvalue` to null for a `_class` bound to an attribute path, so `removeClass` is skipped and the stale CSS class stays on the DOM node.
- `framegrid.py:708`/`780` and `tag_matrix_grid.py:211` bail out on `!changedAttr`.

Reproduced from the console on a th page, with a row selected:

```js
var t=null; genro._data.walk(function(n){ if(n.label=='selectedId' && n.getStaticValue()!=null){ t=n } }, 'static');
genro.setData(t.getFullpath(null,true), t.getStaticValue(), objectUpdate(objectUpdate({}, t.attr), {_probe: Date.now()}));
```

Please do **not** fix it by defaulting `changedAttr` inside `setAttr()`: that would also change `updAttributes()` (gnrbag.js:453), and some consumers branch on the field being absent — `_evlg_loader` (framegrid.py:783) does `if(kw.changedAttr){ return; }`. It has to come from the new call site.

Suggestion that also removes some duplication: have the helper return the list of changed keys instead of a boolean (empty array = unchanged), so the call site can pass the name when exactly one attribute changed. As a side effect the negative-logic name goes away, and so does the overlap with the two comparison helpers already in `gnrlang.js` — `attrIsUnchanged`'s `!updattr` branch is `objectIsEqual(currattr, newattr)` and its `updattr` branch is `objectIsContained(newattr, currattr)`, the only difference being `isEqual` instead of `===` per value. Three shallow-comparison primitives in one file is one too many.

### Also worth fixing

**`dataTrigger` records None into the serverStore delta.** The attribute-only kw has no `value` key, and `genro.js:1584` stores `asTypedTxt(kw.value)`. Since `convertToText(undefined)` returns `['NN','']`, the recorded delta is `'::NN'`, so an attribute-only set on a serverStore-registered path (`serverpath` attribute, registered at `genro_src.js:513`) ships None to the server and overwrites the stored value. The branch immediately below already reads `kw.node._value` — this one should do the same. Pre-existing for `setData('a.b?attr', v)`, but this PR makes it reachable from every `setData`/`setRelativeData`/websocket push that carries attributes.

**`attrIsUnchanged` misses a new null attribute under update semantics.** With `_updattr:true` (e.g. `setRelativeData(..., {_updattr:true})` at `genro_components.js:6064`), adding a key whose value is null to a node that does not have that key evaluates `isEqual(null, undefined)` → true, so the set is skipped. The non-lazy path would have added the key, computed `changedAttributes[k] = true` and notified `^path?k`. Compare key presence as well as values when `updattr` is set.

### Minor

**The `?attr` branch of `setItem` was not brought in line.** `gnrbag.js:1750` still compares only `node.attr[attr]` and ignores `_attributes` entirely, so `setData('a.b?tag', v, {other:1})` silently drops `{other:1}` — the same bug this PR fixes on the value path — and it returns `undefined` instead of the node. The documented three-outcome rule holds on one of the two paths through the function.

**The defensive copy in `setAttributeInDatasource` is one level deep.** For a paged grid, `mixin_rowByIndex` (genro_grid.js:2570) returns the live `node.attr`; if a column holds a Bag or object, `objectUpdate({}, attributes)` copies the reference, so a later in-place mutation still aliases the datanode and `isEqual` on that key compares an object with itself. Also note `objectUpdate` converts a Bag source via `asDict()`, so a caller passing a Bag as `attributes` now stores a plain dict.

**`gnrdomsource.js:513`**: `!attributes` never fires for an empty object, and `_gnrUpdateSelect` passes `row = {}` on deselection. No observable effect — the `selectedId` datanode is autocreated by the `^...selectedId` bindings before any selection on both branches, so nothing extra is materialized — but `objectNotEmpty(attributes)` would match the comment.

fporcari:
Pushed in f7322d4 — thanks, the blocking point was real and I reproduced every consumer you named.

**changedAttr from the call site.** `attrIsUnchanged()` is now `changedAttrKeys(currattr, newattr, updattr)` returning the changed keys, and `set()` passes the name when exactly one changed, `null` otherwise. Nothing was defaulted inside `setAttr()`, so `updAttributes()` and `_evlg_loader` keep reading the absence of the field as a whole-attribute change. Verified against the real `gnrbag.js`: same value and same attributes emits nothing at all, one changed key emits `updattr:true, changedAttr:'name', changedAttributes:{name}` with no `value` key, a changed value still goes through `setValue()` untouched.

The multi-key case stays where it was — `changedAttr: null`, `changedAttributes` carrying both — and no call site can invent a single name for two. Since the four consumers give up on the singular field while the plural one has the answer, and the fix belongs on the reader side you ruled out touching from `setAttr()`, I opened #1169 for it rather than growing this PR.

**Also fixed, same commit:**

- `dataTrigger` (genro.js:1584) reads `kw.node._value`, like the branch below it. Confirmed `convertToText(undefined)` returns `['NN','']`, so an attribute-only set on a `serverpath` really did ship `'::NN'`.
- key presence is compared as well as values, so a new attribute at null under `_updattr` is no longer swallowed by `isEqual(null, undefined)`. The `'*'` semantics are honoured too: a null for a key the node does not have is a no-op, not a change.
- the `?attr` branch of `setItem` merges `_attributes` and returns the node instead of `undefined`, so the three-outcome rule now holds on both roads through the function.
- `setAttributeInDatasource` snapshots container values too (`deepCopy()` for a Bag, a shallow copy for a plain object), converts a Bag argument explicitly rather than through `objectUpdate`'s side effect, and uses `objectNotEmpty()`.

On the duplication: `objectIsEqual` and `objectIsContained` are left alone. Expressing them through the new helper would swap their per-value `===` for `isEqual` for all their other callers, which is a change of a different size — `changedAttrKeys` is at least not a third boolean comparator.

`test/datastore/attribute_trigger` now shows `changedAttr` of the last trigger in test_0, and test_1 has a button writing an attribute by path while carrying a second one. `pytest core web sql app xtnd`: 2100 passed, 69 skipped.

dgpaci (changes requested):
Everything from the first round is addressed, and the `changedAttrKeys` shape is better than what I asked for: returning the key list instead of a boolean gives the call site what it needs without defaulting anything inside `setAttr()`, so `updAttributes()` and `_evlg_loader` keep their current reading of an absent field. Opening #1169 for the reader side rather than growing this PR is the right call — the multi-key case genuinely cannot be solved by inventing a single name at the writer.

One thing did not make it across, and it is the same defect on the other road.

### Blocking: `gnrbag.js:1759` still hardcodes `changedAttr` on the `?attr` branch

```js
node.setAttr(auxattr, _doTrigger, '*', attr);
```

That was correct while `_attributes` was ignored — only one attribute could ever change, so naming it was always true. Now that `auxattr` merges `_attributes`, this line can declare one name while two attributes changed, which is the opposite of the rule you just applied to the value branch.

Your own new test button is exactly this case:

```js
genro.setData(this.absDatapath('.target?tag'), 'bypath_1', {other:'other_2'})
```

`auxattr` becomes `{other:'other_2', tag:'bypath_1'}`, both keys change, and the trigger carries `changedAttr:'tag'` with `changedAttributes:{tag:true, other:true}`. `dataChangeLogger` (genro_frm.js:1780) logs `tag` and silently drops `other` — the same loss I blocked on last round.

And it is worse here than the honest `null` of the value branch: a reader-side fix along the lines of #1169 will use `changedAttr` when it is present and fall back to `changedAttributes` only when it is null. So this road would keep hiding the second attribute even after #1169 lands, and the bug would look fixed everywhere else.

The list is already computed one line above. I would keep `attr` as the fallback rather than deriving the name from the list, so nothing changes for non-lazy sets — there `changedAttr` has always been `attr` even when the attribute was rewritten with the same value, and `_evlg_saver` (framegrid.py:709) gives up on `!changedAttr`:

```js
var changedAttrs = changedAttrKeys(node.attr, auxattr, '*');
if (kwargs.lazySet && !changedAttrs.length) {
    return node;
}
...
node.setAttr(auxattr, _doTrigger, '*', changedAttrs.length > 1 ? null : attr);
```

One line moved and one line changed, and the two roads through `setItem()` finally say the same thing.

### Two notes, not blocking

**`changedAttributes` never reports a deleted attribute** (gnrbag.js:468). The loop iterates `this.attr` after the update, so a key that was removed cannot appear in it, while `changedAttrKeys` correctly counts it as changed. Drop one attribute under replace semantics and you get `changedAttr:'droppedKey'` with `changedAttributes:{}`; same under `'*'` with a null value, where `objectUpdate(..., removeNulls)` deletes the key. Nothing reads that field today so there is no bug to fix here, but #1169 is going to make four consumers depend on it, and deletions will be invisible to them. Cheapest fix while you are in there: build it from `oldattr` as well as `this.attr`.

**Arrays escape the snapshot** (gnrdomsource.js:527). `v.constructor === Object` covers plain objects and `isBag(v)` covers Bags, but an array value is stored by reference. `isEqual` compares arrays element-wise (gnrlang.js:519) — which is itself the evidence that array attribute values are an expected case — so after an in-place mutation the comparison walks the same object twice and reports 'unchanged', the exact blindness the snapshot exists to prevent. `else if (v instanceof Array) { v = v.slice(); }` closes it.

### One question

The first run reported `2159 passed, 10 skipped`, this one `2100 passed, 69 skipped`. The total is identical, so 59 tests moved from executed to skipped rather than disappearing — I assume services that were up for one run and not the other, since nothing in this diff touches Python. Worth confirming so the numbers are not read as a regression later.

dgpaci (changes requested):
The multi-key case is right now, and computing the list once is what I was after. But the expression I handed you in the last round was wrong, and you implemented it faithfully — my mistake, so let me correct it precisely.

### Blocking: `gnrbag.js:1760` names `attr` whichever key actually changed

```js
node.setAttr(auxattr, _doTrigger, '*', changedAttrs.length>1 ? null : attr);
```

The `length > 1` test covers two-or-more, and the zero case correctly falls back to `attr`. What neither of us accounted for is **one changed key that is not `attr`**:

```js
genro.setData('a.b?tag', currentTagValue, {other:'new'})
```

`auxattr` is `{other:'new', tag:currentTagValue}`. `tag` compares equal, so `changedAttrKeys` returns `['other']` — length 1, so the call passes `attr`, i.e. `'tag'`. The trigger then carries `changedAttr:'tag'` alongside `changedAttributes:{other:true}`, and the two fields contradict each other outright.

`dataChangeLogger` (genro_frm.js:1780) takes `cattr = 'tag'` and reads `kw.oldattr['tag']` and `kw.node.attr['tag']`, which are identical, so it registers a changes-logger entry with `from == to`: the form is marked dirty for an attribute that never moved, and `other`, the one that did, is never logged. Both halves are wrong, and it is reachable from any re-set of an unchanged attribute value that carries a second attribute.

The zero-key fallback has to stay — non-lazy sets have always named `attr` even when rewriting the same value, and `_evlg_saver` (framegrid.py:709) gives up on `!changedAttr`. Only the one-key branch needs the real key:

```js
node.setAttr(auxattr, _doTrigger, '*',
             changedAttrs.length > 1 ? null : (changedAttrs.length ? changedAttrs[0] : attr));
```

With that, all four cases hold: nothing changed names `attr` as before, one changed key names that key, two or more name nothing, and `changedAttr` can no longer disagree with `changedAttributes`.

### Still open from the second round

Neither note was addressed or mentioned, which is fine for non-blocking items, but I would like an explicit disposition — fixed here, or folded into #1169 — so they do not fall through the gap:

- **`changedAttributes` never reports a deleted attribute** (gnrbag.js:468). The loop iterates `this.attr` after the update, so a removed key cannot appear, while `changedAttrKeys` counts it as changed. This one matters more now: #1169 is going to make the four consumers read exactly that field, and deletions will be invisible to them.
- **Arrays escape the container snapshot** (gnrdomsource.js:527). `v.constructor === Object` covers plain objects, `isBag(v)` covers Bags, an array value is stored by reference. `isEqual` compares arrays element-wise (gnrlang.js:519), which is itself the evidence that array attribute values are expected here.

### On the test counts

Fair enough that you cannot reconstruct the earlier run. Since nothing in this diff touches Python, the executed/skipped split is almost certainly environmental; a one-line note to that effect in the PR body would stop `2100 passed` being read as a regression by whoever looks at this later.

fporcari:
> The `length > 1` test covers two-or-more, and the zero case correctly falls back to `attr`. What neither of us accounted for is **one changed key that is not `attr`**: […] The trigger then carries `changedAttr:'tag'` alongside `changedAttributes:{other:true}`, and the two fields contradict each other outright.
> — [review](https://github.com/genropy/genropy/pull/1145#pullrequestreview-5040699272)

Applied verbatim in dc35ccda2, `gnrbag.js:1760`. The premise checks out: `changedAttrKeys` (`gnrlang.js:550`) iterates `newattr`, so with `auxattr = {other:'new', tag:<unchanged>}` it returns `['other']`, and the old `length>1 ? null : attr` handed `'tag'` to a trigger whose `changedAttributes` said `other`. `dataChangeLogger` (`genro_frm.js:1783-1786`) then reads `kw.oldattr['tag']` and `kw.node.attr['tag']`, identical, and `changes.setItem` records `from == to` — the entry is created on the first set, since the `n.attr.from == newvalue` pop only fires when one already exists.

Worth recording that the other road through `setItem()` already said it correctly: the value branch at `gnrbag.js:1829` has been `changedAttrs.length==1?changedAttrs[0]:null` all along. The zero-key case cannot be reached there — it returns early on `!changedAttrs.length` — which is why only the `?attr` branch needed the three-way form, and the two roads now agree.

## Disposition of the two notes

Both folded into #1169, neither fixed here — on your own reasoning from the second round, that the reader side belongs there rather than growing this PR:

- **`changedAttributes` never reports a deleted attribute** (`gnrbag.js:468`). Building it from `oldattr` as well as `this.attr` changes what four consumers will read, and #1169 is where those consumers are defined. Fixing it here would land the shape without the readers that depend on it.
- **Arrays escape the container snapshot** (`gnrdomsource.js:527`). This one is independent of the trigger contract and independent of #1169 too; I would rather it did not ride in on a PR about `changedAttr`. If you prefer it as its own issue instead of inside #1169, say so and I will split it.

## Test counts

Added to the PR body: nothing in this diff touches Python, so the executed/skipped split moves with the services that happen to be up, and `2100 passed` is not a regression against `2159 passed`.

## Verification

- `node --check gnrjs/gnr_d11/js/gnrbag.js` — clean.
- The four cases read off the new expression: zero changed keys names `attr`, one names that key, two or more name nothing, and `changedAttr` can no longer disagree with `changedAttributes`.

Not verified: I did not re-run `test/datastore/attribute_trigger` in a browser for this commit. The case your example describes — re-setting an unchanged attribute value alongside a second attribute — is not one of the existing test buttons, so it would need a new one to be exercised rather than reasoned.

`develop` dismisses stale approvals on push; this commit touches only the line you named, and nothing previously approved has moved.

dgpaci (approved):
Approving. Verified `dc35ccda2` from the commit blob: both roads through `setItem()` now say the same thing, and the four cases hold — nothing changed names `attr`, one changed key names that key, two or more name nothing, and `changedAttr` can no longer disagree with `changedAttributes`. Your point about the value branch is right too: the zero-key case is unreachable there because it returns early on `!changedAttrs.length`, which is why only the `?attr` road needed the three-way form. `node --check` clean.

The mechanism underneath is sound and I checked it on a real project rather than only on the test page: on an `anc_base.socio` th view, a re-click on the selected row went from one spurious value trigger to none, and `setData(path, sameValue, {newAttr})` went from silently dropping the attribute to writing it with a single attribute-only trigger.

Two things recorded, neither blocking the merge.

**Please add the test button before merging if you can.** The case that produced the third round has no coverage: the existing 'Tag by path, carrying another attribute' always changes the path value (`'bypath_'+genro.getCounter()`), so it only exercises the multi-key branch. Nothing writes a path attribute at its current value alongside a changed second attribute — the `changedAttrs.length == 1 && changedAttrs[0] != attr` branch, the one that actually broke. A button with a fixed tag value plus a counter-based second attribute pins all four outcomes. Three rounds all landed on this one expression and each missing case was found by reading rather than by running; that button is what breaks the pattern.

**Split the array snapshot into its own issue**, not #1169. It is independent of the trigger contract and of the reader-side work, so it has no reason to live inside an issue about consumers of `changedAttr`. Your reasoning for folding `changedAttributes` into #1169 I agree with as it stands: landing the shape there, next to the readers that consume it, is better than landing it here without them.

Thanks for the three quick turnarounds, and for flagging what you had not verified each time rather than rounding it up — that is what made it possible to review the reasoning instead of re-deriving it.

genro (approved):
Approving `dc35ccda2`. This changes the write path of the whole datastore, so I read it for the way it can fail — a set that is silently dropped — rather than for the case it is meant to fix.

`changedAttrKeys` is the hinge and it handles both semantics rather than the common one. Replace (`!updattr`) also walks `currattr` for keys the new attributes drop, since losing an attribute changes its own path; update with `'*'` treats a null as a deletion and counts it only when the key was actually there. A setter with no attributes at all returns `[]` and leaves them untouched, which is what keeps every existing value-only caller behaving exactly as before. `isEqual` rather than `==` is what makes Dates and Bags compare by content.

The snapshot in `setAttributeInDatasource` is the detail that shows this was met in practice and not designed on paper: a grid hands out its live store row, and storing it by reference makes every later comparison blind — the datanode and the row are the same object, so nothing can ever look changed. Copying before storing is the only way the lazy comparison means anything.

**I checked the one thing that removing a parameter can break silently.** `forceChanges` is gone from the signature, and a caller still passing a fifth argument would hand it to nothing. Across `gnrjs/`, `resources/` and `projects/` there are exactly two five-argument call sites, and they are the two `selectedId` publications in `genro_grid.js` that this PR updates. Every other caller passes three or four. Nothing is left behind.

`node --check` passes on all four touched files at this head.

**Two things about the description rather than the code.**

The body describes `attrIsUnchanged(currattr, newattr, updattr)` returning whether anything changed. The code is `changedAttrKeys` returning the list of changed keys, and that list feeds `changedAttr` on the trigger so a consumer can branch on *which* attribute moved — which is more than the body promises and is exercised by the test page. Three rounds of review moved the design and the description did not follow; worth a pass before merge, since the body is what the next reader will trust.

The snapshot is one level deep — Bags are `deepCopy`'d, plain objects get a shallow `objectUpdate({}, v)`. For row fields that is proportionate and I would not change it, but the comment presents it as protection against aliasing without saying where the guarantee stops. A nested plain object inside an attribute still aliases.

Not verified by me: the browser behaviour and the trigger counts in your table. @dgpaci went through three rounds on this one and approved the result, which is the verification that counts here.

Issue #1151, opened 2026-08-24
[FEATURE] Remaining client build margins: dijit share, JSON wire format, source-level nodeId registry


fporcari:
## Feature
After the optimizations of #1146 #1147 #1148 #1149 the build of a very large page breaks down roughly as:

- ~45% dijit widget instantiation (dojo.declare machinery, templates, DOM): not addressable without touching dojo 1.1 or building fewer widgets (wider use of `_lazyBuild` for out-of-view tabs would help immediately);
- ~15% XML parsing of the main page source (`fromXmlDoc`): a JSON wire format for the main call would cut most of it;
- ~10% residual full-source scans on `genro.nodeById` misses for ids that legitimately do not exist (grid searchbox probe, `pyref`): needs a source-level nodeId registry maintained on node insertion, which is invasive on the Bag core (a per-batch pre-index was tried and reverted: components pop and re-create nodes with the same nodeId during the build, so pre-indexed entries go stale - see docs/architecture/client_performance.md);
- the rest is spread over attribute extraction (`objectExtract` chains, ~21 passes per node) and form/validation bookkeeping.

This issue tracks those future margins. Numbers measured on `test/perf/test_heavy_build`.

PR #1152, opened 2026-08-24
perf(gnrjs): path-indexed trigger dispatch, per-handler caches and bag hot paths

### LINKED ISSUE #1146: [BUG] Every datastore change is dispatched to every dynamic attribute subscription
## Problem Description
Every dynamic attribute (`value='^.foo'`, `_class='^...'`, `storepath='^...'`, ...) subscribes to the single dojo topic `_trigger_data`. `genro.setData` publishes on that topic, so every subscription on the page runs `trigger_data` -> `attrDatapath` -> `absDatapath` -> `getTriggerReason` (with fresh path-string building each time) just to find out the event is not for them.

On a large page this makes every data change O(number of dynamic attributes on the page): profiling `test/perf/test_heavy_build` (~7.900 subscriptions) shows 2.55 million `trigger_data` invocations for 600 `setData` calls (~4.200 subscribers scanned per event), 384ms for the burst. This cost is paid on every data change for the whole life of the page and is the main reason very large pages feel sluggish after they are built.

The event path string (`kw.pathlist.slice(1).join('.')`) is also recomputed inside every subscriber for the same event.

## Reproduction Steps
1. Open `test/perf/test_heavy_build` (test package)
2. Press 'Data storm (600 set)' in test_3
3. Observed: ~380ms for 600 setData; a single setData costs ~0.6ms regardless of how many nodes actually match

## Expected
An event should reach only the subscriptions whose datapath can match it (same path, ancestor or descendant), not every subscription on the page.

### LINKED ISSUE #1147: [BUG] Full source-tree walk on every del/upd trigger to refresh index and subscribers
## Problem Description
`GnrSrcHandler.refreshSourceIndexAndSubscribers` is called by `_trigger_del` and `_trigger_upd`: it walks the entire source tree to rebuild the nodeId index and drop the subscriptions of removed nodes. On a big page each call visits thousands of bags; during the initial build of `test/perf/test_heavy_build` it accounts for ~450.000 `Bag.walk` recursions (48 refresh calls), and every later dom update (grid rebuilds, remote content) pays a full-tree walk again.

Related: `genro.nodeById` misses during the build fall back to `getNodeByAttr('nodeId', ...)`, another full-tree scan per probe; the FramePane 'existing frame' assert performs one such probe (always a miss) per frame.

## Reproduction Steps
1. Open `test/perf/test_heavy_build?_profile=1`
2. `gnrProfiler.report()` in console
3. Observed: Bag.walk dominates with hundreds of thousands of calls, mostly under refreshSourceIndexAndSubscribers and nodeById fallbacks

## Expected
A del/upd should clean up only the removed subtree (its subscriptions and its nodeId index entries); uniqueness probes for ids that always self-register (frames) should use the index directly.

### LINKED ISSUE #1148: [TASK] Cache widget factory resolution and handler mixin plans
## Task
Two per-widget costs in `genro_wdg.js` repeat work that never changes:

- `getWidgetFactory` re-resolves the widget class on every creation: `dojo.require` calls plus a window class-path walk (~1.400 times on the profiling page, ~130ms plus GC pressure). The factory can be cached keyed by the resolved dojo tag. The key must be the resolved `_dojotag`, not the genro tag: handlers like `menuline` switch `_dojotag` per instance (MenuItem/MenuSeparator/PopupMenuItem).
- `doMixin` scans every property of the handler for every widget instance, with per-property string work, to find the applicable `mixin_/patch_/versionpatch_/nodemixin_/validatemixin_` props. The applicable set never changes per handler: scan once, apply a precomputed plan per instance.

Measured on `test/perf/test_heavy_build`: these two caches are the largest single build win (startUp median goes from ~1.6s to ~0.5s together with the bag hot-path work) and remove most of the build-time GC noise (run-to-run variance from ±20% to ±3%).

### LINKED ISSUE #1149: [TASK] Bag hot-path cleanups: htraverse fast path, static walk, fromXmlDoc conversions
## Task
Three hot paths in `gnrbag.js` do avoidable work on every call:

- `htraverse` regex-matches and splits every path even when it is a single plain label - the most common case (every setItem/getNode of a direct child, every addItem from fromXmlDoc). Single plain labels can return immediately.
- `walk(cb, 'static')` calls `getValue('static')` per visited node; in static mode the value is just `_value`.
- `fromXmlDoc` runs every attribute value through `convertFromText` and then `mapConvertFromText` again; values without a `'::T'` suffix are plain strings and need no conversion at all.

Also worth documenting in code: `_pop` must keep rebuilding the `_nodes` array instead of splicing it in place - several callers (slot resolution in slotbars, `mergeRemoteContent`, `replaceContent`) iterate `_nodes` while popping and rely on the original array staying untouched. An in-place splice breaks every tablehandler page.

### LINKED ISSUE #1150: [FEATURE] Heavy test page and client-side profiler for build/trigger profiling
## Feature
A reproducible way to measure client engine performance beyond guesses:

- `test/perf/test_heavy_build`: a deliberately oversized testhandler page (8 tabs x 120 bound fields, 8 bag grids 200x20, 3000 divs with dynamic class binding, two tablehandlers, a data-storm/rebuild test) - ~8.200 source nodes, ~1.400 dojo widgets.
- `heavy_profiler.js` (test package resource) with three instruments:
  - `?_profile=1`: wrappers on the hot engine methods with call counts and self/total timing (`gnrProfiler.report()`)
  - `?_sample=1`: native JS Self-Profiling API across the build (the page sends `Document-Policy: js-profiling`), with self and inclusive per-frame reports accumulated across reloads
  - `?_bt_sample=N`: N reloads collecting the startUp build time in localStorage, because single build measurements vary ±20% with GC/JIT state

Documented in `docs/architecture/client_performance.md` together with the findings.

fporcari:
## What

Profiling-driven optimization of the client engine (gnrjs) for very large pages, plus the profiling infrastructure used to find the real bottlenecks and a document with the findings.

Measured on the new `test/perf/test_heavy_build` page (~8.200 source nodes, ~1.400 dojo widgets), median of 6 runs:

| metric | before | after |
|---|---|---|
| `startUp` build | 1609 ms | 514 ms (3.1x) |
| 600 `setData` on bound fields | 384 ms | 30 ms (13x) |

## How

- **`gnr.GnrTriggerIndex`** (genro_src.js): dynamic `^attr` subscriptions with a stable absolute datapath are indexed in a trie by path segments; an event reaches only the nodes on its path, its ancestors or its subtree instead of every subscription on the page (2.55M `trigger_data` calls for 600 setData before). Subscriptions with pointer/symbolic datapaths stay in a floating set checked on every event as before; `trigger_data` keeps its full semantic check, so routing false positives are harmless. The `_trigger_data` dojo topic is still published for the direct subscribers (grid cellpars, tree).
- **Incremental subscriber cleanup**: `_trigger_del`/`_trigger_upd` clean only the removed subtree instead of rebuilding index and subscriptions with a full source-tree walk per trigger.
- **`getFormHandler` negative caching**: nodes outside any form stop at their parent instead of walking to the root per node.
- **Widget factory cache** keyed by the resolved dojo tag (handlers like menuline switch `_dojotag` per instance) and **per-handler mixin plan** replacing the per-widget scan of every handler property. Largest single build win; also removes most build-time GC noise (variance ±20% → ±3%).
- **Bag hot paths**: `htraverse` fast path for single plain labels, `walk` static mode without `getValue`, `fromXmlDoc` single conversion (plain strings untouched), `_pop` array-replacement behavior documented (callers iterate `_nodes` while popping).
- **FramePane** uniqueness assert on the nodeId index instead of a full-tree `nodeById` probe.

Details, methodology and remaining margins in `docs/architecture/client_performance.md`. Remaining margins tracked in #1151.

## Verification

- Real-browser regression pass: `plainTableHandler` (data, searchbox filter, autoSelect), `test_flex_minwidth` (bagGrid, columnsets), sandbox app th pages (Clienti/Fatture) with menus/dialogs, heavy page (bound fields receive values, validations, 8 grids with 200 rows, 3000 dynamic divs, data storm, tab rebuild) - no new console errors.
- Two approaches were tried and reverted after breaking tablehandler pages in the browser (batch nodeId pre-index, in-place `_pop` splice); both are documented in the doc so they are not retried blindly.
- Full python test suite green (pre-commit hook), flake8 clean on the new test page.

Fixes #1146
Fixes #1147
Fixes #1148
Fixes #1149
Fixes #1150


gnrjs/gnr_d11/js/genro_src.js
```diff
@@ -207,6 +286,14 @@ dojo.declare("gnr.GnrSrcHandler", null, {
         }
     },
 
+    unsubscribeHandle:function(handle){
+        if (handle && handle._gnrTriggerIndex) {
+            this.triggerIndex.remove(handle);
+        } else {
+            dojo.unsubscribe(handle);
```
cgabriel: so dojo.unsubscribe is called only for null handles? What's the point?

gnrjs/gnr_d11/js/gnrbag.js
```diff
@@ -1925,11 +1930,12 @@ dojo.declare("gnr.GnrBag", null, {
 
     walk: function (callback, mode, kw, notRecursive) {
         var result;
-        var bagnodes = this.getNodes();
+        var isStatic = mode && mode.indexOf('static') >= 0;
+        var bagnodes = this._nodes;
```
cgabriel: Why the change of not using the abstract getNodes() method?

cgabriel (changes requested):
I get the following try to use the test_heavy_build demonstrator:

```
  File "/home/cgabriel/genropy/genropy/gnrpy/gnr/web/gnrasync.py", line 773, in websocket_handler
    await session.on_message(msg.data)
  File "/home/cgabriel/genropy/genropy/gnrpy/gnr/web/gnrasync.py", line 680, in on_message
    result = handler(_time_start=time.time(), **kwargs)
  File "/home/cgabriel/genropy/genropy/gnrpy/gnr/web/gnrasync.py", line 730, in do_connected
    self.server.registerPage(page_id=page_id)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/cgabriel/genropy/genropy/gnrpy/gnr/web/gnrasync.py", line 897, in registerPage
    page = self.gnrsite.resource_loader.get_page_by_id(page_id)
  File "/home/cgabriel/genropy/genropy/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py", line 170, in get_page_by_id
    return self.instantiate_page(page_id=page_id,class_info=class_info, init_info=init_info, page_info=page_info)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cgabriel/genropy/genropy/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py", line 177, in instantiate_page
    page = page_class(site=self.site, page_id=page_id,page_info=page_info, **init_info)
  File "/home/cgabriel/genropy/genropy/gnrpy/gnr/web/gnrsimplepage.py", line 87, in __init__
    self.onIniting(request_args, request_kwargs)
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/cgabriel/genropy/genropy/projects/gnrcore/packages/test/webpages/perf/test_heavy_build.py", line 29, in onIniting
    self.response.add_header('Document-Policy', 'js-profiling')
    ^^^^^^^^^^^^^
AttributeError: 'GnrCustomWebPage' object has no attribute 'response'
```

Having the data storm working (blazing fast!), the rebuild fails with the following error:

Image

gnrjs/gnr_d11/js/genro_src.js
fporcari: Not only for null handles — the branch split is by subscription kind. Every `^path` data trigger now carries `_gnrTriggerIndex` and is removed through the trigger index; what falls through to `dojo.unsubscribe` are the genuine dojo topic subscriptions (`subscribe_*` attributes), which are collected in the same `subscriptions[reason]` lists this function is called on (see the `dojo.forEach(subscriptions[reason], ...)` caller). A null handle reaching it is a harmless no-op, same as before.

gnrjs/gnr_d11/js/gnrbag.js
fporcari: No good reason — `getNodes()` with no condition returns `this._nodes` as-is, so the abstraction costs nothing. Restored in e285659e4. The measured win in `walk` is the `static` branch reading `_value` without going through `getValue(mode)`, not the direct field access.

fporcari:
Thanks for running the demonstrator — both findings addressed in e285659e4.

> I get the following try to use the test_heavy_build demonstrator: `AttributeError: 'GnrCustomWebPage' object has no attribute 'response'`

Confirmed and fixed. `onIniting` also runs on the websocket registration path (`gnrasync.registerPage → gnrsimplepage.__init__`), where `self.response` does not exist — it is only set in `GnrWebPage.__init__` on the WSGI path. The header is now set only when a response is there (it is meaningless on the websocket registration anyway).

> Having the data storm working (blazing fast!), the rebuild fails

Root cause understood, and it is pre-existing behavior this branch does not touch: `node.rebuild()` is `setValue(this._value)`, i.e. it re-expands the already-expanded source. Component nodes do not survive that — `_beforeCreation` moves their attributes into `_saved_attributes` on first expansion and nothing ever restores them, so the second expansion finds the framepane without its `frameCode` (your exact stack: `createContent` via `walk`). The demo button was asking the framework for an operation it has never supported over expanded components (bagGrid, formbuilder). The button now rebuilds the bound-divs subtree (3000 nodes with dynamic bindings, no components), which is the part that actually demonstrates the incremental subscriber cleanup; the limit is documented in the page docstring. Making components survive a source rebuild (restoring `_saved_attributes` and the original children) is real design work — if you think it is worth it, I would open a dedicated issue rather than widen this PR.

Not verified live in this round: the rebuild button on a served instance (the fix removes components from the rebuilt subtree by construction). Full python suite green locally (2100 passed, 69 skipped).

Both inline threads answered in place.

dgpaci (approved):
Reviewed the code and tested the branch end to end against a large customer project in production, whose pages are considerably heavier than the synthetic benchmark. Approving: behaviour is unchanged and the win is **bigger on real pages than on the test page**.

## Measured on the customer project

Same measurement on both branches: 600 `genro.setData` on a scratch path with no subscribers, run per page (each page lives in its own iframe with its own `genro.src`), same browser session.

| page | subscriptions | develop | this PR | speedup |
|---|---|---|---|---|
| application shell | 154 | 99.4 ms | 31.8 ms | 3.1x |
| th page, single table | 466 | 377.2 ms | 92.6 ms | 4.1x |
| configuration page | 431 | 369.5 ms | 58.2 ms | 6.3x |
| accounting dashboard | 843 | 783.2 ms | 119.5 ms | 6.6x |
| master/detail page A | 1359 | 1647.3 ms | 286.4 ms | 5.8x |
| master/detail page B | 1508 | 2448.2 ms | 403.5 ms | 6.1x |

The speedup grows with page size - 3.1x on the shell, around 6x on pages with 1300+ subscriptions - which is the signature of the per-subscriber scan being gone. In absolute terms the heaviest page went from **4.1 ms to 0.67 ms per `setData`**: on develop that page burns about half a second of CPU every 120 values written, which is exactly the sluggishness users report on big tableHandler pages.

## Functional pass: no regressions

Exercised on the same project: menus and context menus including the `menuline` `_dojotag` switch (separators still render as separators), forms, validations, save/lock and the unsaved-changes guard, `#FORM` / `#ROW` / `#WORKSPACE` paths, dynamic storepath switching (collection stores and grouplets), editable grid, remote panes, dialogs, and repeated open/close cycles of the same pages. No subscription leak, no `duplicate nodeId`, no `existing frame`, no new console errors.

Worth recording: the per-page subscription counts are identical to develop, so the new dispatch does not lose subscriptions on the way.

## What is left of the old cost

On those pages **18-28% of the subscriptions stay in the floating set** (25% on the heaviest one) and are still scanned on every event. That share, not the trie, dominates what remains. Two small additions would make it visible and keep the index honest:

- a `triggerIndex.stats()` reporting the indexed/floating split and `targets.length` per event: it is the metric to drive #1151, and right now the floating share is invisible without an ad-hoc snippet;
- ship the index self-check as a debug helper: walk the trie and compare `attrDatapath(sub.attr)` with the key each subscription is indexed under. I wrote a throwaway version of it for this review and it immediately surfaced a latent bug unrelated to this PR (filed as #1160), which is a good argument for having it in the box.

## Two non-blocking points on the indexing probe

1. `_stableTriggerPath` rests on the invariant *a literal `datapath` in the ancestor chain never changes after build*. Four places assign `attr.datapath` at runtime: `genro_wdg.js:43`, `applyStorepath` at `genro_wdg.js:957` (re-called by `setNewStorepath`), `genro_wdg.js:1702`, `genro_components.js:30`. I could not make any of them produce a stale key on the customer project, across ~3500 indexed subscriptions, so the invariant appears to hold - but it deserves to be stated explicitly in the comment, since the whole trie rests on it. Worth noting in the same comment that the failure mode is benign by construction: `trigger_data` always re-resolves the path, so a stale key can only cause a *missed* notification, never a delivery to the wrong node.

2. `_stableTriggerPath` moves an `absDatapath()` resolution from trigger time to build time. On a node whose relative path cannot be resolved - detached fragment, menus built from a data Bag - `absDatapath` logs `unresolved relativepath` and runs `debugger` under `genro.isDeveloper`, so a developer with devtools open now stops on a breakpoint while merely building a page. The probe is best-effort by design: a failed resolution should fall back to the floating set silently.

## On the two open comments

- `unsubscribeHandle`: the `dojo.unsubscribe` branch is not dead. Handles registered through `registerSubscription` are still real dojo handles - grid cellpars at `genro_wdg.js:2181`, tree at `genro_tree.js:249`, plus every other reason - and only `_setDynAttributes` produces indexed ones. A one-line comment naming those two callers would save the next reader the same question.
- `walk` using `_nodes` instead of `getNodes()`: `getNodes()` is `condition ? this._nodes.filter(condition) : this._nodes` and has no overrides, so the two are equivalent; the real win in that hunk is `isStatic ? _value : getValue(mode)`. Restoring `getNodes()` keeps the accessor at no measurable cost.

## Verified equivalent while reviewing

Recording these so nobody has to redo the analysis: the `htraverse` single-label fast path (`#parent` and `#N` behave exactly as before); the `fromXmlDoc` simplification (a DOM attribute value is always a string, so without `::` the old chain ended in `convertFromText` with `t='T'`, which is the identity, and `convertFromText` already handles JSON/JS recursively); `_pop` (unchanged, only documented); the factory cache (never caches a negative result, so lazily loaded widgets still resolve); the mixin plan (no runtime mutation of `mixin_*` properties anywhere in the engine); `publishDataTrigger` as the only remaining publisher of `_trigger_data`; and the FramePane assert, since frames register their nodeId immediately after the check.

Good work - the documentation of the two reverted approaches in `docs/architecture/client_performance.md` is as valuable as the optimization itself.

cgabriel (approved):
LGTM, reviewed after change request fixes.

genro (approved):
Approving `05e45c1f8`. I read the part where this change can only fail one way — a subscription that stops being reached — and it is built to fail toward the old behaviour.

`_stableTriggerPath` is the whole hinge, and every uncertain case falls back to the floating set rather than into the trie: a non-string or non-`^` value, a `#parent` anywhere in the path, a symbolic head that is not `#DATA`/`#FORM*`/`#ANCHOR*`, a relative path whose ancestor chain carries a pointer datapath. `_stableDatapathChain` stopping at an absolute datapath — "shields the upper chain" — is right, since an absolute path does not depend on what is above it.

The second net is what makes it safe rather than merely careful: the key actually indexed is `attrDatapath(attr)`, the resolved absolute path, and it is rejected if it still contains a `#`. So even a wrong "stable" verdict cannot put an unresolved symbolic path in the trie. And on the read side `publish` snapshots the floating set and collects the trie targets before notifying anything, then skips `sub._removed` — so a node that rebuilds while being notified cannot corrupt the batch it is in.

Routing false positives being harmless is the right claim to lean on, because `trigger_data` keeps its full semantic check. The risk was never the extra call; it was the missing one, and the design puts every doubt on the side of calling too much.

**Where the confidence in this actually comes from, and it is not from me.** I cannot verify client behaviour without a browser. @dgpaci ran it end to end against a large customer project in production, on pages heavier than the synthetic benchmark, and measured 3.1x–4.1x with behaviour unchanged; @cgabriel blocked on a real traceback from the test page and approved after the fix. That is a better verification than this PR would normally get, and it is what makes a change to the hot path of the entire client reviewable at all.

Keeping the `_trigger_data` dojo topic published for the direct subscribers is the detail that stops this from being a breaking change for code outside the framework.

**The most valuable file here is the one that ships no behaviour.** `docs/architecture/client_performance.md` records the two approaches that were tried and reverted — pre-indexing the nodeIds of a build batch, and `_pop` via `splice` — with the reason each broke the tablehandler pages. Six months from now that section is what stops somebody retrying them from scratch, and it is the part of this PR I would least want dropped in a rebase.

Not verified by me: the browser regression pass, the measurements, and the 19-page test infrastructure. All three are attested by reviewers who ran them.

Issue #1153, opened 2026-08-25
Service freshness check reads the register on every call, even for services that can never expire


genro:
## Problem

`BaseServiceType.__call__` (`gnrpy/gnr/lib/services/__init__.py:229-237`) caches service
instances in-process but re-reads `globalServices_lastChangedConfigTS.<type>_<name>`
from the register global store on every call, to detect configuration changes made in
other processes.

Measured on a legacy stack (gunicorn + separate sitedaemon over Pyro), a single login
performs 384 register calls / 386 wire round-trips / 175 ms inside the register.
242 of those calls (63%, 112.7 ms) come from this single freshness check:
121 service requests × 2 round-trips each.

Two distinct defects:

1. **`ServerStore.data` evaluates `register_item` twice**
   (`gnrpy/gnr/web/daemon/siteregister_client.py:394-397`): `if self.register_item:`
   followed by `self.register_item['data']`. Each evaluation is a Pyro round-trip,
   so every `getItem` on any store costs 2 round-trips instead of 1.

2. **The freshness check is unconditional.** The timestamp key is written only by the
   `sys.service` table triggers (`projects/gnrcore/packages/sys/model/service.py:63-67`).
   Services whose configuration is resolved from `conf_<name>` methods or
   `DEFAULT_STORAGE_CONFIGS` (e.g. storage `rsrc`, `gnr`, `dojo`, `_raw_`) have no
   `sys.service` record: the key is never written and the check reads `None` forever.
   In the measured login those are 118 of the 121 service requests.

## Proposed fix

1. Evaluate `register_item` once in `ServerStore.data` (242 → 121 round-trips).
2. Mark on the service instance whether its configuration was resolved from the
   database. Skip the freshness check for instances whose configuration did not come
   from `sys.service` — the timestamp can never concern them.

Combined effect on the measured login: register round-trips for the freshness check
drop from 242 to 3.

## Trade-off

Creating a **new** `sys.service` record that overrides a name previously resolved from
built-in configuration (e.g. redefining storage `dojo`) will no longer be picked up by
running workers holding the cached instance: it requires a restart. Overriding a
built-in volume is a deploy-time operation, so the restart is considered implicit.

genro:
Merged as `a1c0a8dd0` via #1154 — the closing reference did not form (the PR title carries `fix(#1153)`, a mention rather than a closing keyword), so closing by hand.

PR #1154, opened 2026-08-25
fix(#1153): skip register freshness check for services not configured in db

genro:
## Problem

`BaseServiceType.__call__` caches service instances in-process but re-read
`globalServices_lastChangedConfigTS.<type>_<name>` from the register global store on
every call, to detect configuration changes made in other processes.

Measured on a legacy stack (gunicorn + separate sitedaemon over Pyro), a single login
performed 384 register calls / 386 wire round-trips / 175 ms inside the register.
242 of those calls (63%, 112.7 ms) came from this single freshness check:
121 service requests × 2 round-trips each.

Two distinct defects:

1. **`ServerStore.data` evaluated `register_item` twice** (`if self.register_item:`
   followed by `self.register_item['data']`). Each evaluation is a Pyro round-trip,
   so every `getItem` on any store cost 2 round-trips instead of 1.
2. **The freshness check was unconditional.** The timestamp key is written only by the
   `sys.service` table triggers. Services whose configuration is resolved from
   `conf_<name>` methods, `DEFAULT_STORAGE_CONFIGS` or explicit kwargs without a
   `sys.service` record (e.g. storage `rsrc`, `gnr`, `dojo`, `_raw_`) never get the
   key written: the check read `None` forever. In the measured login those were
   118 of the 121 service requests.

## Fix

- `ServerStore.data` evaluates `register_item` once (242 → 121 round-trips).
- `addService` resolves the database configuration first and marks the instance with
  `_config_from_db`; `__call__` returns a cached instance immediately when its
  configuration was not resolved from `sys.service` — the timestamp can never
  concern it.

Combined effect on the measured login: register round-trips for the freshness check
drop from 242 to 3.

## Trade-off

Creating a **new** `sys.service` record that overrides a name previously resolved from
built-in configuration (e.g. redefining storage `dojo`) is no longer picked up by
running workers holding the cached instance: it requires a restart. Overriding a
built-in volume is a deploy-time operation, so the restart is considered implicit.
Updates to services that already have a `sys.service` record keep the current
behavior: the freshness check still runs for them on every call.

## Tests

- `tests/core/gnrservices_test.py`: two new tests — a cached non-db service performs
  no register reads; a db-configured service keeps the per-call freshness check.
- Test site facades gained a `gnrapp.packages` attribute, now always traversed by
  `addService`.
- flake8 clean on modified files; full suite green except pre-existing failures
  (`flatfiles_test`, `test_gnrbaseadapter`).

Ref #1153

### DIFF
--- a/gnrpy/gnr/lib/services/__init__.py
+++ b/gnrpy/gnr/lib/services/__init__.py
@@ -89,7 +89,10 @@ class BaseServiceType(object):
         self._implementations_lock = threading.RLock()
 
     def addService(self, service_name=None, **kwargs):
-        service_conf = kwargs or self.getConfiguration(service_name)
+        db_conf = self.getServiceConfigurationFromDb(service_name)
+        service_conf = kwargs or db_conf or \
+                self.getServiceConfigurationFromSiteConfig(service_name) or \
+                self.getServiceConfigurationFromSelf(service_name)
         if not service_conf:
             return
         implementation = service_conf.pop('implementation',None) or service_conf.pop('resource',None) #resource is the oldname for implementation
@@ -103,6 +106,7 @@ class BaseServiceType(object):
         service.service_type = self.service_type
         service.service_implementation = implementation
         service._service_creation_ts = datetime.now()
+        service._config_from_db = db_conf is not None
         self.service_instances[service_name] = service
         return service
 
@@ -229,6 +233,11 @@ class BaseServiceType(object):
     def __call__(self, service_name=None, **kwargs):
         service_name = service_name or self.default_service_name
         service = self.service_instances.get(service_name)
+        if service is not None and not service._config_from_db:
+            # The expiration timestamp is written only by the sys.service table
+            # triggers: a service whose configuration was not resolved from the
+            # database can never expire, so skip the register read entirely.
+            return service
         gs = self.site.register.globalStore()
         cache_key = 'globalServices_lastChangedConfigTS.%s_%s' % (self.service_type, service_name)
         lastChangedConfigurationTS = gs.getItem(cache_key)
--- a/gnrpy/gnr/web/daemon/siteregister_client.py
+++ b/gnrpy/gnr/web/daemon/siteregister_client.py
@@ -393,8 +393,10 @@ class ServerStore(object):
 
     @property
     def data(self):
-        if self.register_item:
-            return self.register_item['data']
+        # register_item is a remote call: evaluate it once
+        register_item = self.register_item
+        if register_item:
+            return register_item['data']
 
     @property
     def datachanges(self):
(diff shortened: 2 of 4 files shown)


fporcari (approved):
The diagnosis is the valuable part here: the freshness check was reading a key that
nothing ever writes for services configured outside `sys.service`, so 118 of the 121
service requests in the measured login were paying two round-trips for a value that
could only ever be `None`. Marking the instance at creation time instead of guessing at
call time is the right place to put the decision, and stating the trade-off — a new
`sys.service` record shadowing a built-in name now needs a restart — in the body rather
than leaving it to be discovered is exactly right.

Verified, not taken on trust:

- the inlined chain in `addService` is call-for-call identical to the `getConfiguration`
  it replaces, same order, same short-circuits;
- `addService` is the only writer of `service_instances` in the whole repository, so the
  early return at `gnrpy/gnr/lib/services/__init__.py:236` can never meet an instance
  without `_config_from_db`;
- `__call__` already ignored `kwargs` on a cache hit before this change, so the early
  return does not silently drop a reconfiguration that used to be honoured;
- both new tests fail on `develop` with `AttributeError: 'Service' object has no
  attribute '_config_from_db'`, and pass on the branch both in isolation and in file
  order. flake8 clean on the four files, 9 passed in the two touched test modules.

Three notes, none of them a reason to hold this up.

**1. `sys.service` is now queried on the `kwargs` path, where it was not before.**
`gnrpy/gnr/lib/services/__init__.py:92` computes `db_conf` unconditionally, while the old
`kwargs or self.getConfiguration(service_name)` short-circuited to zero database work
whenever kwargs were supplied. `addService` only runs on a cache miss, so this is one
extra query per service per process rather than a per-call cost — but it is a database
touch on a path that had none, and it is what forced `gnrapp.packages` onto the two test
site facades. Worth a line in the body so the next reader does not have to work out why
those facades grew an attribute.

**2. `_config_from_db` is set to `True` when kwargs won and a database record merely
happens to exist.** At `:109` the flag is `db_conf is not None`, but `service_conf` is
`kwargs or db_conf or ...`: with kwargs present the configuration did not come from the
database at all, yet the instance is marked as if it had and keeps paying the freshness
check for the rest of the process's life. The predicate that matches the resolution
actually used is

```python
service._config_from_db = not kwargs and db_conf is not None
```

The error is in the safe direction — a lost optimisation, never a stale instance — which
is why this is a note and not a change request.

**3. The `ServerStore.data` half has no test.** `gnrpy/gnr/web/daemon/siteregister_client.py:397`
is half of the measured win (242 to 121 round-trips) and nothing goes red if somebody
folds it back into `if self.register_item:` during a later cleanup. The comment is the
only thing guarding it. A test asserting one `register_item` evaluation per `data` access
would pin the half of the fix that is currently held together by a comment.

Not verified: I did not reproduce the 384-call measurement on a real Pyro plus gunicorn
stack, and the new tests do not exercise the multidomain `storename` branch of
`getServiceConfigurationFromDb`.

PR #205, opened 2025-07-25
Feature/docu theme

dgpaci:
### **User description**
Instead of building a theme, which generally is not needed, possibility to specify instance logo, colors, css in docu preferences and build conf file and html consequently.

___

### **PR Type**
Enhancement

___

### **Description**
- Enhanced documentation preferences with theme customization options

- Added support for handbook logos, copyright, and styling

- Improved Sphinx configuration generation with dynamic settings

- Added publish date display and localization support

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["Preferences UI"] --> B["Theme Settings"]
  B --> C["Sphinx Config"]
  C --> D["HTML Output"]
  A --> E["CSS/JS/HTML"]
  E --> D
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>preference.py</strong><dd><code>Enhanced documentation preferences interface</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

projects/gnrcore/packages/docu/resources/preference.py

<ul><li>Restructured preferences UI with tabbed interface<br> <li> Added handbook theme settings (logo, copyright, version display)<br> <li> Added code editors for custom CSS, HTML, and JS<br> <li> Enhanced form layout with better organization</ul>

</details>

  </td>
  <td><a href="https://github.com/genropy/genropy/pull/205/files#diff-2e29296d347381dab9a7b82db855001c205bd57a3ca893b409d3b8dc372e974f">+39/-3</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>export_to_sphinx.py</strong><dd><code>Enhanced Sphinx export with dynamic configuration</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

projects/gnrcore/packages/docu/resources/tables/handbook/action/export_to_sphinx.py

[...]

PR #1157, opened 2026-08-25
fix(docu): register the sphinx extensions the docu preferences depend on

### LINKED ISSUE #1156: [BUG] Sitemap is never generated by the export to sphinx even when the preference is enabled
### Problem Description

With the docu preference **Enable sitemap** turned on, the handbooks exported by the *Export to sphinx* batch are published **without any `sitemap.xml`**.

The preference is read correctly, but the only thing the batch writes into the generated `conf.py` is the `sitemap_url_scheme` config value. The sitemap file is produced by the `sphinx_sitemap` extension, and that extension is never listed in `extensions`, so nothing ever reads that config value and no sitemap is built.

The regression comes from 55f97e8a9 ("Feature/docu theme", #205), which emptied the extension list in `packages/docu/resources/sphinx_env/default_conf.py`:

```diff
-extensions = ['sphinx_sitemap','sphinxext.opengraph']
+extensions = [] #E.g. 'sphinx_sitemap','sphinxext.opengraph'
```

with the intent of enabling the extensions from the batch depending on the preferences, but `export_to_sphinx.py` only ever appends the url scheme (the `#DP Enabled extensions = [...]` comment on that line stayed a note, it was never implemented).

**Same root cause, second symptom:** `sphinxext.opengraph` was dropped too, so the `-D ogp_image=...` override passed by `step_buildHtmlDocs` is an unknown config value for sphinx and is silently ignored: handbooks with an OG preview image get no `og:` meta tags either.

Expected behavior: with the preference on, the exported handbook contains a `sitemap.xml` listing its pages under the configured base url.

### Reproduction Steps

1. In the docu preferences, enable **Enable sitemap** (and optionally set a Sphinx baseurl).
2. Run the *Export to sphinx* action on any handbook.
3. Open the published handbook folder.

Observed result: no `sitemap.xml` in the published folder. Expected: `sitemap.xml` with the page urls.

The core of it can be reproduced without an instance, running sphinx on the shipped default configuration:

```
cp packages/docu/resources/sphinx_env/default_conf.py /tmp/src/conf.py
cat >> /tmp/src/conf.py <<'CONF'
html_theme = 'alabaster'
html_baseurl = 'http://example.org/_documentation/'
sitemap_url_scheme = 'myhandbook/{link}'
CONF
printf 'Title\n=====\n\nhello\n' > /tmp/src/index.rst
python -m sphinx /tmp/src /tmp/build
```

Observed result: build succeeds, `/tmp/build/sitemap.xml` is not created. Adding `extensions = ['sphinx_sitemap']` to the same `conf.py` makes the sitemap appear.

### Environment

genropy 26.08.19.1 (develop), production deployment; sphinx 8.2.3, sphinx-sitemap installed (both listed in `packages/docu/requirements.txt`).

### Additional Material

Affected code: `projects/gnrcore/packages/docu/resources/tables/handbook/action/export_to_sphinx.py` (`step_prepareConfFile`).

Note for the rollout: the fix acts on the generation of `conf.py`, so already published handbooks stay without a sitemap until they are exported again.

dgpaci:
Closes #1156

## Problem

With the docu preference **Enable sitemap** on, exported handbooks are published without any `sitemap.xml`.

The preference is read correctly, but `step_prepareConfFile` only appended `sitemap_url_scheme` to the generated `conf.py`. The sitemap file is written by the `sphinx_sitemap` extension, and 55f97e8a9 ("Feature/docu theme", #205) had emptied the extension list of `default_conf.py`:

```diff
-extensions = ['sphinx_sitemap','sphinxext.opengraph']
+extensions = [] #E.g. 'sphinx_sitemap','sphinxext.opengraph'
```

meaning to enable them from the batch depending on the preferences, but the batch never did (the `#DP Enabled extensions = [...]` comment on that line stayed a note). Without the extension loaded, `sitemap_url_scheme` is a variable nothing reads.

The same commit dropped `sphinxext.opengraph` too, so the `-D ogp_image=...` override that `step_buildHtmlDocs` passes to sphinx-build is an unknown config value and is silently ignored: handbooks with an OG preview image get no `og:` meta tags either.

## Solution

The batch now builds the extension list next to the config values that depend on it and writes it into `conf.py`. The assignment comes after the empty one shipped in `default_conf.py`, so it is the one sphinx reads:

- `sphinx_sitemap` when the sitemap preference is on (together with `sitemap_url_scheme`)
- `sphinxext.opengraph` when the handbook declares an `ogp_image`
- `installedSphinxExtensions()` filters out anything missing from the environment, logging a warning: a missing package now costs its own feature instead of aborting the whole export with an `ExtensionError`

## Test case

New `projects/gnrcore/packages/docu/tests/test_docu_sphinx_conf.py`, 6 tests on a real app, database and local storage services (the shipped `default_conf.py` is the file being extended, no mocks):

- the sitemap preference adds `sphinx_sitemap` to `extensions` and sets the url scheme and base url
- with the preference off neither appears
- `ogp_image` adds `sphinxext.opengraph`, and nothing adds it without one
- an extension missing from the environment is skipped, not fatal
- **end to end**: sphinx runs on the generated `conf.py` and writes a `sitemap.xml` containing `https://docs.example.org/_documentation/testhandbook/index.html`

Reverting the source change makes 4 of the 6 fail, the end-to-end one included.

```
gnr dev tests gnrdevelop docu
```

51 passed (45 before this PR). The 8 errors reported by `test_examples_sync.py` are pre-existing and unrelated: `BaseGnrTest` points `GENRO_GNRFOLDER` at a temporary dir and does not restore it, so that file (which builds its own `GnrApp` instead of using `BaseGnrAppTest`) cannot find the configuration when it runs after the others. It passes 8/8 on its own, on develop as well as here.

## Manual check

The fix acts on the generation of `conf.py`, so **already published handbooks stay without a sitemap until they are exported again**: after deploying, re-run *Export to sphinx* on the handbooks that need one. If `sphinx-sitemap` is not installed in the target environment (it is in `packages/docu/requirements.txt`), that is now a warning in the log instead of silence.


fporcari (approved):
The diagnosis is right and the archaeology is the useful part: `55f97e8a9` emptied the list in `default_conf.py` meaning to fill it from the batch, the batch never did, and the `#DP Enabled extensions = [...]` comment sitting on the `enable_sitemap` line is the trace of the intention that was never carried out. Writing the list next to the config values that depend on it is the right place for it — the two travel together and neither means anything alone.

Verified against the checkout rather than the description:

- `default_conf.py:38` really is `extensions = []`, and the batch copies that file and appends with `open('a')` (`export_to_sphinx.py:82-83`, `:139`), so the new assignment lands after the empty one and is the one sphinx executes. The file ends with newlines, so the append starts on its own line.
- `ogp_image` is a real column (`model/handbook.py:32`, `dtype='P'`) and `step_buildHtmlDocs:181-192` does pass it as `-D ogp_image=...`, so the second half of the fix is a real gap and not a speculative one.
- Both packages are already in `packages/docu/requirements.txt`, so `installedSphinxExtensions()` degrades a provisioning mistake instead of an ordinary install.
- flake8 clean on both files.

Approving: the production change is correct and the end-to-end test is the right shape — it runs sphinx on the generated `conf.py` and asserts the `sitemap.xml` content, which is the only assertion that could have caught the original bug.

Two notes on the tests, neither worth blocking on.

**Three of the six fail, rather than skip, without the extensions installed.** `_batch()` guards with `pytest.importorskip('sphinx')`, but `test_sitemap_preference_enables_the_extension`, `test_ogp_image_enables_the_opengraph_extension` and `test_missing_extension_is_skipped` assert that an extension *is* in the generated list — which holds only where that extension is installed, because `installedSphinxExtensions()` filters it out otherwise. So in an environment with `sphinx` but without `sphinx-sitemap` — precisely the environment the new method exists for — those three go red, and they go red on the one behaviour the PR got right. `test_sitemap_file_is_built` is guarded correctly and shows the pattern.

The cheap fix is the same `importorskip` on the three. The better one is to separate the two things being tested: the batch's *intent* (which extensions it selects for a given preference) is environment-independent and worth pinning always, while the filtering is one test that can skip. Right now the filter is threaded through every assertion and drags the environment in with it.

Worth stating for severity: CI runs `cd gnrpy && pytest`, so it never runs this file at all. The cost falls on whoever runs `gnr dev tests` in a partially provisioned checkout, who sees three failures that say nothing about their change.

**The site stub is the third copy.** `tests/sitestub.py` exists for exactly this and `test_docu_handbook_storage.py:21` and `test_docu_media_export.py:26` both import it; `test_examples_sync.py:26` already forks its own, and this file makes three. The one substantive difference here — the `isinstance(fullpath, StorageNode)` passthrough in `storageNode` — is an improvement that belongs in the shared stub, where the other two tests would get it too. The rest is the shared class minus `addService`, `pathListFromUrl`, `storageType` and `storageNodeFromPathList`.

**Not verified:** I did not run the suite. There is no `gnrdevelop` instance on this machine, so the `51 passed (45 before)` and the "reverting makes 4 of 6 fail" are your measurements, not reproductions of mine. Everything above is read from the code.

Issue #1158, opened 2026-08-25
Carry _lastUserEventTs/_lastRpc on every RPC, not only on the ping


genro:
## Context

The three activity clocks of a site register item — `last_refresh_ts`,
`last_user_ts`, `last_rpc_ts` — are written by one method only,
`SiteRegister.refresh()` (`gnr/web/daemon/siteregister.py:206`):

```python
def refresh(self, register_item_id, last_user_ts=None, last_rpc_ts=None, refresh_ts=None):
    item = self.registerItems.get(register_item_id)
    if not item:
        return
    item['last_user_ts'] = max(...) if item.get('last_user_ts') else last_user_ts
    ...
```

Its only entry point is `handle_ping()` (`siteregister.py:904-910`), which reads
the clocks the CLIENT reports in the ping payload (`_lastUserEventTs`,
`_lastRpc`) and cascades page → connection → user (`siteregister.py:802-811`).

So between two pings a page can serve any number of RPCs and its activity
clocks do not move. A row that has never been pinged carries none of the three.

## Why it matters

Whoever judges idleness or expiry has to read the real activity clocks, not
`last_refresh_ts` — a beat alone keeps that one warm forever. With per-ping
delivery, "last activity" is only ever as fresh as the last beat.

Visible today in `connected_users_bag`
(`gnr/web/gnrwebpage_proxy/connection.py:195-198`):

```python
last_refresh_ts = arguments.get('last_refresh_ts') or arguments['start_ts']
last_user_ts = arguments.get('last_user_ts') or arguments['start_ts']
last_refresh_age = (now - last_refresh_ts).seconds
last_event_age = (now - last_user_ts).seconds
```

A user actively working between two beats reads as idle for up to the beat
interval, and `user_disconnected` is applied on `last_refresh_age > 60`.

The same gap is why the daemon-less SPA worker in genropy-asgi stamps
`last_rpc_ts` server-side on every HTTP request instead of relying on the page
protocol. Its own docstring names the workaround: *"`last_rpc_ts` is a real
call, the surrogate metre until the page protocol carries the human event of
its own"*.

## Proposal

Carry `_lastUserEventTs` and `_lastRpc` in the payload of every RPC, not only
in the ping. Each register then decides whether to write them on every call or
keep the current per-ping policy.

## Cost

- **Sending them**: payload only. The RPC leaves anyway, so there is no extra
  round trip and no extra request.
- **Writing them per RPC in the daemon**: one extra Pyro round trip per RPC.
[...]

PR #1159, opened 2026-08-25
feat(#1158): carry _lastUserEventTs/_lastRpc on every RPC

genro:
## Summary

`_lastUserEventTs` and `_lastRpc` were sent only in the ping payload
(`genro_rpc.js` `ping`). Between two beats a page could serve any number of
RPCs without its activity clocks moving, so `last_user_ts` and `last_rpc_ts`
on the register item were only ever as fresh as the last beat.

This PR is the "send only" half of #1158: two lines in `_serverCall`, the
single funnel every RPC goes through. No Python change.

The server already accepts the two keys on every call — `_rpcDispatcher`
pops them into `self._lastUserEventTs` / `self._lastRpc`
(`gnrwebpage.py:641-642`). Today those attributes are always `None` on a
non-ping RPC and nothing in the repo reads them. After this PR they carry a
value.

Position matters: the two lines sit **after** `genro._lastRpc = new Date()`.
A user-driven call reports its own timestamp; a sysrpc, which does not touch
`genro._lastRpc`, reports the last real RPC.

## Out of scope

Writing the clocks to the register on every RPC — that is the extra Pyro
round trip discussed in #1158, and the write policy stays a separate
decision. This PR leaves the daemon cost unchanged.

## Cost

~100 bytes per RPC (two `yyyy-MM-dd HH:mm:ss::DH` strings plus key names,
url-encoded) inside a POST that leaves anyway. No extra round trip, no extra
request, no daemon call added.

## Checks

- `node --check gnrjs/gnr_d11/js/genro_rpc.js` clean
- No client-side cache is keyed on the payload: `content.callcounter`
  already changes on every call (`register_call`)
- The other two routes of `_get_call_handler`, `_plugin` and external `rpc`,
  are not reachable from `_serverCall`: `pluginCommand` is client-side only
  and the `rpc` route is not generated by the JS
- `genro._lastUserEventTs` and `genro._lastRpc` are set in the `genro`
  constructor, before `genro.rpc` exists — no init-order hazard
- Cross-repo search on both keys: all hits are in this repo
- No test asserts on the RPC payload

## Test plan

- Open a page, run RPCs without waiting for a ping, verify the register item
  shows `last_user_ts` / `last_rpc_ts` advancing on the next beat rather than
  only at beat time
- Verify the ping still works: same two keys, unchanged path (`_ping`)
- Verify a sysrpc does not overwrite `last_rpc_ts` with its own timestamp

### DIFF
--- a/gnrjs/gnr_d11/js/genro_rpc.js
+++ b/gnrjs/gnr_d11/js/genro_rpc.js
@@ -363,6 +363,8 @@ dojo.declare("gnr.GnrRpcHandler", null, {
         if(!sysrpc){
             genro._lastRpc = new Date();
         }
+        content._lastUserEventTs = genro.getServerLastTs();
+        content._lastRpc = genro.getServerLastRpc();
         if (genro.debugRpc) {
             this.debugRpc(kw);
         }


fporcari (approved):
Read the whole diff and walked the server side. The change is correct and safe, and I could confirm each load-bearing claim:

- `_rpcDispatcher` pops both keys unconditionally (`gnrwebpage.py:641-642`) before `self.rpc(method=..., **kwargs)`, so nothing leaks into a public method's signature.
- `handle_ping` (`siteregister.py:905-910`) is the only caller of `refresh()`, so the daemon cost really is unchanged.
- The ping sets `sysrpc: true` (`genro_rpc.js:846`), so `genro._lastRpc` is not touched and the two new lines recompute exactly the values `pingKw` already carried — ping semantics unchanged.
- Init order is fine: `start_ts = new Date()` (`genro.js:83`), the two attributes at `genro.js:112/115`, `this.rpc = new gnr.GnrRpcHandler` only at `genro.js:217`; both are only ever assigned a `Date`.
- `kw.content = content` keeps the reference and `register_call` mutates it in place, so the two keys do reach the wire even though they are set after `register_call`.
- Only the `pageCall` route of `get_call_handler` (`gnrwebpage.py:345` — the description calls it `_get_call_handler`) is reachable: `_plugin` is never sent as an RPC kwarg from the JS and the `rpc` route is not generated.

One thing to fix in the description rather than in the code. The first test plan item says:

> Open a page, run RPCs without waiting for a ping, verify the register item shows `last_user_ts` / `last_rpc_ts` advancing on the next beat rather than only at beat time

That cannot be observed with this diff. Nothing writes the register outside `handle_ping`, and at beat time the ping already sends the current `getServerLastTs()` / `getServerLastRpc()` — so the register item is exactly as fresh after this PR as before it. Consistent with the Out of scope section, but the test plan reads as if it were not.

What this PR actually makes true is narrower: `page._lastUserEventTs` and `page._lastRpc` stop being `None` on a non-ping RPC. Nothing in the repo reads them yet (grep gives only the assignments), so the observable behaviour is unchanged — which is the point, it is groundwork for the write-policy half of #1158. Worth saying that in the test plan instead.

Minor, non-blocking: `developer.py:228` strips `_lastUserEventTs` from the debug `call_kwargs` but not `_lastRpc`; after this PR the asymmetry shows on every debug entry. Harmless today since `call_kwargs` is only used for `method` and `_debug_info` (the `_method_parameters` line is commented out).

Approving.

PR #1161, opened 2026-08-25
feat(serve): --collect records requests and site register calls

fporcari:
Adds `--collect` to `gnr web serve`: the development server records every
request and every call the site makes to its site register, into one archive
per run.

A request header joins the two, so **every register call is attributable to the
exchange that caused it** — which is what neither an HTTP log nor a SQL trace
can show.

## Why it might be worth having

Pointed at `sandboxpg` for the first time (login plus three pages, 88 exchanges,
1144 register calls):

- opening the application index costs **211 register calls** and 716 ms, and
  none of it is SQL or XML;
- a test page with **nothing on it** costs **167**;
- across the run, `getItem` on a store was called **527 times to read 13
  distinct keys**. In one empty page render a single key —
  `globalServices_lastChangedConfigTS.storage_gnr` — was read **54 times**, and
  every store read costs two Pyro round trips: **108 IPC round trips for one
  value that cannot change during a request.**

I am not proposing a fix for that here. This PR is only the flag that makes it
visible.

## What the diff does

Three points in `serverwsgi.py`, 39 lines added and none changed:

- `--collect` beside the other `add_argument` calls;
- `start_probe()`, called **before** `GnrWsgiSite` is built — the site's
  `__init__` forces its register into existence (under genropy's own comment
  *"this is needed, don't remove"*), so a recorder assigned afterwards would
  patch a name the site has already read;
- `probe.wrap(gnrServer)` **after** the debugger has wrapped the site, so the
  recorder is outermost and the exchange id is in the environ before anything
  reads the request.

## genropy takes no dependency

The import lives inside the branch the flag opens. Without the flag genropy
imports nothing; without the package installed it logs one line and serves
normally.

The recorders themselves live in [`genropy/gnrprobe`](https://github.com/genropy/gnrprobe),
which was extracted from the legacy/bridge comparison bench of `genropy-asgi`.
It also carries a `gnrprobe dev` verb that installs the same recorders on an
**unmodified** genropy, by assigning over werkzeug's `make_server` — so this
flag is the tidy way, not the only way, and nothing depends on it landing.

## Verified

On `sandboxpg`, three ways:

| | result |
|---|---|
| without the flag | nothing written, nothing imported |
| with `--collect` | 3 requests → 3 HTTP lines + 100 register lines, one archive |
| with `gnrprobe` unimportable | `--collect requires gnrprobe` logged, site answers 200 |

Full genropy suite on the branch: **2161 passed, 10 skipped**. flake8 clean.

## Worth knowing before merging

`X-GnrSqlTime` and `X-GnrSqlCount` need site debug **and** the page's own
`debug_sql` (`GnrWsgiSite.sqlDebugger`), so they arrive as zero in a normal
`--collect` run. That is genropy's behaviour, not the flag's; the reports say so
rather than letting a zero read as a fast query.


cgabriel (approved):
Ok for me, tested with latest gnrprobe, and report data are consistent. Due to the goal and context of the tool, I've skipped the forked enviroment analysis.

genro (approved):
Approving. Thirty-nine lines added and none changed, which is the right shape for a development-only flag.

I checked the constraint the docstring states about itself, since it is the kind that is easy to write and easy to break later: `probe = self.start_probe()` sits before `gnrServer` is assigned, so it really does run before `GnrWsgiSite` is built. That matters exactly as described — the site's `__init__` forces its register into existence, so a recorder attached afterwards would patch a name the site has already read. `is_running_from_reloader` is imported at module level (`serverwsgi.py:9`) and the parent-under-reload case returns `None` in the same branch that builds `FakeApp`, so the two agree.

The import living inside the branch the flag opens is what keeps this honest: a server started without `--collect` imports nothing, and genropy takes no dependency on `gnrprobe`. The `ImportError` path logs what to install instead of dying, which is the right outcome for a tool flag.

**The numbers in the description are the reason to want this merged, not the flag itself.** 211 register calls to open an application index, 167 for a page with nothing on it, and one key — `globalServices_lastChangedConfigTS.storage_gnr` — read 54 times in a single empty render, at two Pyro round trips each. That is 108 IPC round trips for a value that cannot change during a request. Not proposing a fix here is the right call for this PR, but that finding should not live only in a PR description: it deserves its own issue, otherwise the tool that found it merges and the thing it found does not get tracked.

Worth saying too that this is the same family as #1135 (`ServerStore.data` doing two round trips where one would do) — same shape, same path, and now there is an instrument that can measure whether fixing them moves anything.

PR #1163, opened 2026-08-25
fix(grid): do not select an unresolved pkey in NewIncludedView

### LINKED ISSUE #1162: [BUG] NewIncludedView.setSelectedId overwrites the subscribed selectedId with null for an unknown pkey
## Problem Description

`NewIncludedView.mixin_setSelectedId` (`gnrjs/gnr_d11/js/genro_grid.js:4448` on `develop`) resolves the row with `collectionStore().getIdxFromPkey(pkey)` and feeds the result straight to `selection.select(idx)`, guarding only the upper bound:

```js
var idx = this.collectionStore().getIdxFromPkey(pkey);
if (idx >= nrow) {
    idx = nrow - 1;
}
this.selection.select(idx);
this.scrollToRow(idx);
```

`getIdxFromPkey` returns **-1** when the pkey is not in the store (`gnrjs/gnr_d11/js/genro_components.js:6895` and `:8094`), and nothing catches it.

`selection.select(-1)` is not inert. It calls `addToSelection(-1)`, whose `onCanSelect` returns `true` unconditionally, so it sets `selectedIndex = -1` and fires `grid.onSelected(-1)`. That reaches `_gnrUpdateSelect(-1)` (`genro_grid.js:1455`), where `selectedId` is computed as `null` because of the `idx >= 0` guard and then written back:

```js
if(this.sourceNode.attr.selectedId) {
    this.sourceNode.setAttributeInDatasource('selectedId', selectedId, null, row, true);
}
```

So a grid subscribed with `selectedId='^.selectedId'` **overwrites the caller's value with `null`** whenever the pkey it was handed is not in the current store. The write is silent: no exception, no log. `scrollToRow(-1)` runs too, moving the viewport.

The expected behaviour is the one `selectByRowAttr` already implements for the same operation: a pkey that does not resolve leaves both the selection and the subscribed datapath untouched.

This is not a corner case. `mixin_newDataStore` applies the subscribed `selectedId` on every whole-store replacement, and at that moment the subscribed value still holds a pkey from the *previous* store. Any replacement with different keys walks straight into it.

## Reproduction Steps

The test page below is the one added by #1130; these steps apply once that PR is merged.

1. Open the testhandler `/test15/gnrwdg/includedview_bagstore/1`.
2. Click **Load rows**. Both grids show 40 rows.
3. Click **Select last**. Both grids select `row_039` and scroll to it.
4. Click **Select missing**, which sets `selectedId` to `row_999` on both grids.

Observed result, read from the datastore and from the widget after step 4:

| grid | selection | subscribed `selectedId` |
|---|---|---|
| legacy `IncludedView` | kept on row 39 | still `row_999` |
| `NewIncludedView` | cleared | **overwritten with `null`** |

Expected: the `NewIncludedView` row behaves like the legacy one — the value the caller wrote survives.

It reproduces in the neutral testhandler above, so no customer context is needed. Without that page, any `bagGrid` with `grid_selectedId='^.selectedId'` shows it: set the subscribed path to a pkey absent from the store and watch the path go back to `null`.

## Environment

`develop`, `gnrjs/gnr_d11/js/genro_grid.js`, dojo 11 client. Reproduced locally on a sqlite gnrcore instance, Chrome.

## Additional Material

The legacy half of the table above is green because #1130 implements `setSelectedId` on `VirtualStaticGrid` on top of `selectByRowAttr`, which guards the missing row. The `NewIncludedView` override was deliberately left untouched there, since it uses the optimized `getIdxFromPkey` path and fixing it was out of that PR's scope.

The minimal fix is the same guard, keeping the optimized lookup:

```js
mixin_setSelectedId: function(pkey) {
    var nrow = this.rowCount;
    if (nrow == 0 || isNullOrBlank(pkey)) {
        this.selection.unselectAll();
    } else {
        var idx = this.collectionStore().getIdxFromPkey(pkey);
        if (idx >= nrow) {
            idx = nrow - 1;
        }
        if (idx >= 0) {
            this.selection.select(idx);
            this.scrollToRow(idx);
        }
    }
},
```

Worth deciding as part of it whether a pkey that does not resolve should leave the previous selection alone (what the legacy path now does) or clear it. The two differ, and only the write-back of `null` is unambiguously wrong.

fporcari:
Fixes #1162

## Problem

`NewIncludedView.mixin_setSelectedId` resolves the row with `collectionStore().getIdxFromPkey(pkey)` and guards only the upper bound. `getIdxFromPkey` returns **-1** when the pkey is not in the store (`genro_components.js:6895` and `:8094`), and `selection.select(-1)` is not inert: `addToSelection(-1)` finds `onCanSelect` returning `true` unconditionally, so it fires `grid.onSelected(-1)` → `_gnrUpdateSelect(-1)` (`genro_grid.js:1455`), where `selectedId` is computed as `null` and written back to the datastore.

A grid subscribed with `selectedId='^.selectedId'` therefore **overwrites the caller's value with `null`** whenever the pkey it was handed does not resolve. Silently: no exception, no log. `scrollToRow(-1)` runs too.

`mixin_newDataStore` applies the subscribed `selectedId` on every whole-store replacement, and at that moment the subscribed value still holds a pkey from the *previous* store — so any replacement with different keys walks into it.

## Change

Keep the optimized `getIdxFromPkey` lookup and skip the selection when it resolves to nothing, which is what `selectByRowAttr` already does on the `VirtualStaticGrid` side. Four lines, one class, no other behaviour touched.

## What is deliberately left out

`NewIncludedView.mixin_selectByRowAttr` does not resolve immediately: it waits on `pendingStore` until `collectionStore()` reports neither `runningQuery` nor `loadingData` (`genro_grid.js:4814`). `setSelectedId` has no such wait, so on a store that is still loading `getIdxFromPkey` returns -1 for a **valid** pkey as well.

This patch changes that case from "the subscribed value is destroyed" to "the selection is not applied", which is strictly better but not complete. Adding the wait means sharing the `pendingStore` block between the two methods, and it is a different defect from the one #1162 reports — I have not reproduced it live, so I am not folding an unverified change into this one. Worth its own issue if someone hits it.

## Verification

- `node --check gnrjs/gnr_d11/js/genro_grid.js`
- Repository suite (`core web app sql xtnd`): 2076 passed, 69 skipped
- Live browser test on the testhandler `/test15/gnrwdg/includedview_bagstore/1`, reading widget state and datastore after each action. That page comes from #1130, so the run was done on `develop` + this patch + #1130 applied locally — i.e. the state both branches produce once merged:

| action | legacy `IncludedView` | `NewIncludedView` |
|---|---|---|
| Load rows (whole-store replacement) | 40 rows, no selection | 40 rows, no selection |
| Select last (`row_039`) | row 39 selected, scrolled to it | row 39 selected, scrolled to it |
| Select missing (`row_999`) | selection kept, `selectedId` still `row_999` | selection kept, `selectedId` still `row_999` |
| Clear selection (`null`) | selection cleared, rows kept | selection cleared, rows kept |
| Load rows again | 40 rows, no error | 40 rows, no error |

Before the patch the third row read *selection cleared, `selectedId` overwritten with `null`* for `NewIncludedView`. The two columns are now identical at every step.

168 local requests, all 200, `_ping` included. No console errors from the application (the only console entries are refused Google Fonts stylesheets, external network being unavailable in this environment).

## Note on #1130

Independent of #1130 and mergeable in either order — that PR touches `VirtualStaticGrid` and `selectByRowAttrDo`, this one only the `NewIncludedView` override. They touch the same file, so whichever lands second may need a trivial rebase.

### DIFF
--- a/gnrjs/gnr_d11/js/genro_grid.js
+++ b/gnrjs/gnr_d11/js/genro_grid.js
@@ -4454,8 +4454,10 @@ dojo.declare("gnr.widgets.NewIncludedView", gnr.widgets.IncludedView, {
             if (idx >= nrow) {
                 idx = nrow - 1;
             }
-            this.selection.select(idx);
-            this.scrollToRow(idx);
+            if (idx >= 0) {
+                this.selection.select(idx);
+                this.scrollToRow(idx);
+            }
         }
     },
 


genro (approved):
The `idx >= 0` guard covers the `-1` from both `getIdxFromPkey` implementations and an undefined `rowidx`. One note for the record: on `IncludedView`/`NewIncludedView` the `onSelected` connect is commented out (`genro_grid.js:4286`), so the `selectedId=null` write reached `_gnrUpdateSelect` through `patch_onSelectionChanged` → `getFirstSelected()` returning -1. Same outcome as the body describes. Merges cleanly with #1130.

PR #1164, opened 2026-08-25
fix(#1160): anchor the save_failed controller to #FORM so nested forms show the refused-save message

### LINKED ISSUE #1160: [BUG] Nested form: a refused save shows no message because the save_failed controller watches the wrong datapath
## Problem Description

`th_form` declares on the form node a controller that warns the user when a save is refused (`resources/common/th/th_form.py:236-243`):

```python
form.dataController(""" if(reason=='nochange'){return;}
                        genro.dlg.alert(msg+' '+this.form.getRecordCaption()+': '+(...));""",
                    reason="^.controller.save_failed", _if='reason', ...)
```

The form handler raises that flag through `fireControllerData('save_failed', ...)` -> `getControllerData()` -> `this.sourceNode.getRelativeData(this.controllerPath)` (`gnrjs/gnr_d11/js/genro_frm.js:1524`), with `controllerPath = '.controller'` and `sourceNode` = the form node. The value is therefore written to:

```
<form node datapath>.controller.save_failed
```

On **nested forms** the `reason` attribute does not resolve to that path. It resolves one level deeper:

```
<form node datapath>.record.controller.save_failed
```

so the controller never fires: a refused save produces **no feedback at all** - no "Invalid record", no "No change to save" - and the record silently stays unsaved. On a root form the same controller works.

**Cause.** `form` is a gnrwdg: `gnr.widgets.FrameForm.createContent` (`gnrjs/gnr_d11/js/genro_components.js:958`) rebuilds the subtree at build time and re-parents the children declared in Python under the FramePane structure, whose content node carries the `.record` datapath (`formDatapath:'.record'`). The `dataController` declared on `form` ends up inside that node, so its relative `^.controller.save_failed` gains a `.record` segment. The path is correct at the moment the subscription is registered (before the re-parenting), which is why the mismatch is invisible when reading the source.

**Expected**: a refused save on a nested form shows the same alert as on a root form.

## Reproduction Steps

1. Open a th page whose form contains a nested form (a sub-form on a related table: attachments, child records, a configuration block).
2. Enter the nested form and press Save without changing anything (`nochange` case), or leave a mandatory field empty and save (`invalid` case).
3. Observed result: no dialog appears and the record is not saved. Repeating the same action on the root form of the same page shows the dialog as expected.
4. Confirmation from the console, on the affected page:

```js
// written by the form handler
genro.getData('<form node datapath>.controller.save_failed')
// actually watched by the dataController
genro.getData('<form node datapath>.record.controller.save_failed')
```

Found while reviewing #1152 on a large production project: 12 occurrences across 5 pages, all with the same shape - key `<...>.form.controller.save_failed` versus resolved path `<...>.form.record.controller.save_failed`. Affected forms were sub-forms on related tables (attachments, child collections, configuration blocks) and the effect is identical on `develop` and on the PR branch, so it is not related to that work.

The divergence can be detected on any page by comparing, for each dynamic-attribute subscription, the path resolved now with the one resolved at registration time:

```js
// list every ^attr subscription whose datapath resolves differently than at build time
(function(){var out=[];
 for(var id in genro.src._subscribedNodes){}
 genro.src._main.walk(function(n){
   if(!n._dynattr){return}
   for(var a in n._dynattr){
     var p=null; try{p=n.attrDatapath(a)}catch(e){p='<error>'}
     out.push({node:n.attr.tag,attr:a,value:n.attr[a],resolved:p});
   }
 },'static');
 console.table(out.filter(function(r){return r.attr=='reason'}));})()
```

## Environment

genropy 26.08.19.1, branch `develop` (0acf37fa3), Chrome on macOS 15

## Additional Material

Fix direction: either make the controller path explicit instead of relative (the same file already uses the nodeId-anchored form elsewhere for the grid context menu, `resources/common/th/th_view.py:488-492`), or have `FrameForm.createContent` keep the declared controllers outside the `.record` content node so that a relative path keeps pointing at the form node.

fporcari:
## Problem / Root cause

`th_form` declares on the form node a dataController that alerts the user when a save is refused, watching `^.controller.save_failed` — a relative path (`resources/common/th/th_form.py:239`).

At build time `FrameForm.createContent` re-parents the children declared on the form node into the FramePane center content node (`gnrjs/gnr_d11/js/genro_components.js`, `content.concat(children)` in `gnrwdg._beforeCreation`, then again in `FramePane.createContent` which returns the center node's bag). When the form resource builds its center with the `form.record` idiom — a contentPane with `datapath='.record'` (`gnrpy/gnr/web/gnrwebstruct/base.py:318`), the typical shape of sub-forms on related tables — the relative path gains a `.record` segment and resolves to `<form>.record.controller.save_failed`.

The form handler however fires the flag at `<form>.controller.save_failed` (`gnrjs/gnr_d11/js/genro_frm.js:1523-1531`, `getControllerData` resolves `.controller` against the node `defineForm` was called on). The controller never triggers: a refused save (invalid record, no change) gives no feedback and the record silently stays unsaved.

The discriminating axis is the resource layout, not root vs nested as such: root forms usually place a borderContainer (no datapath) at center, so the relative path happens to resolve correctly there — which is why the defect surfaces on nested/sub-forms.

## Change

One line: anchor the watch to the form — `reason="^#FORM.controller.save_failed"`. `#FORM` resolves against the ancestor node owning `formId` (`gnrjs/gnr_d11/js/genro.js`, `nodeById`), the same base `fireControllerData` writes to, so the path is correct in every layout. Same idiom already used a few lines above in the same function (`_fired='^#FORM.controller.loaded'`, `th_form.py:183`). Where the relative path happened to work the resolved path is unchanged, so no regression.

The alternative fix — keeping declared controllers outside the `.record` content node in the FrameForm/FramePane re-parenting — would touch the core widget building for every form and was discarded as riskier.

## Verification

- flake8 on the touched file: clean.
- `pytest gnrpy/tests/ --ignore=gnrpy/tests/sql`: 2102 passed, 69 skipped.
- Path resolution verified statically against the resolver: `#FORM` → `attributeOwnerNode('formId')` → `absDatapath('.controller.save_failed')` matches the write base of `fireControllerData('save_failed', ...)`. Not verified on a live page in this session: the repo test packages carry no nested-form th page reproducing the issue's setup.

## Related issue

Fixes #1160

### DIFF
--- a/resources/common/th/th_form.py
+++ b/resources/common/th/th_form.py
@@ -236,7 +236,7 @@ class TableHandlerForm(BaseComponent):
 
         form.dataController(""" if(reason=='nochange'){return;}
                                 genro.dlg.alert(msg+' '+this.form.getRecordCaption()+': '+(reason=='invalid'?invalid:nochange),titledialog);""",
-                            reason="^.controller.save_failed",_if='reason',
+                            reason="^#FORM.controller.save_failed",_if='reason',
                             titledialog='!!Save failed',
                             msg='!!You cannot save',
                             invalid='!!Invalid record',


genro (approved):
The diagnosis holds, and the one-line fix is the right one.

Verified the two paths coincide by construction, not by luck. `#FORM` resolves through `symbolicDatapath` (`gnrdomsource.js:666`) to `nodeById`, which for `FORM` returns `scope.attributeOwnerNode('formId,_fakeform')` (`genro.js:2089-2090`), then calls `absDatapath('.controller.save_failed')` on that node. The write side goes `fireControllerData` → `getControllerData` → `this.sourceNode.getRelativeData(this.controllerPath)` (`genro_frm.js:1523-1532`), where `controllerPath` is `'.controller'`, set by `FrameForm.createContent` (`genro_components.js:956`), and `sourceNode` is the node `defineForm` was called on — the same node that carries `formId` (`genro_widgets.js:191-196`). Same base node, same relative path: the watch now lands exactly where `fireControllerData('save_failed', ...)` writes, whatever the resource puts at center.

The re-parenting that causes the defect is confirmed too: `content.concat(children)` in `gnrwdg._beforeCreation` (`genro_components.js:48`) moves the declared `dataController` under the frame content, so with the `form.record` idiom the old `^.controller.save_failed` picked up the `.record` segment. On a root form whose center carries no datapath the resolved path was already `<form>.controller.save_failed`, so this is not a behaviour change there.

Approving. Not exercised on a live nested form here either — the static resolution is what I checked.

PR #1168, opened 2026-08-25
fix(#1167): attachmentPane grid drop uploads the file but never creates the record

### LINKED ISSUE #722: [BUG] AttachManager creates DB record before file upload — orphan records on storage failure

### LINKED ISSUE #1167: [BUG] attachmentPane: drop on grid uploads the file but creates no record, silently
## Problem

Dropping a file onto the grid of `attachmentPane()` uploads the file to storage but creates **no attachment record**: the grid stays empty and nothing is reported to the user — no toast, no console error, no server traceback.

Reproduced on a form using the old pane:

```python
bc.contentPane(title='Allegati').attachmentPane(pbl_classes=False, nodeId='allegatiSegnalazione_#')
```

## Root cause

Regression introduced by the fix for #722.

`at_attachmentPane` arms the grid as a `Files` drop target passing **only** the pre-upload callback (`resources/common/gnrcomponents/attachmanager/attachmanager.py`):

```python
th.view.grid.attributes.update(dropTarget_grid='Files', onDrop='AttachManager.onDropFiles(this,files);',
                               dropTypes='Files', _uploader_fkey='=#FORM.pkey',
                               _uploader_onUploadingMethod=self.onUploadingAttachment)
```

and `AttachManager.onDropFiles` (`attachmanager.js`) forwards only `onUploadingMethod` in the RPC params.

Before 5618b5d6f this was enough: `onUploadingAttachment` did the `insert` + `commit` itself. That commit moved record creation to `onUploadedAttachment`, which `rpc_upload_file` invokes **only if `onUploadedMethod` is present in the kwargs** (`gnrpy/gnr/web/gnrwebpage_proxy/rpc.py`). On this path it never is, so the file is stored and the record is never written.

56a30b45e fixed the same omission for the three `dropUploader` instances in the views, but the grid-drop path of `at_attachmentPane` was left behind.

## Impact

Every project still using `attachmentPane()` — the drop-on-grid gesture is silently dead. Present in `master` (26.08.19.1) and in `develop`.

Projects that wire the same drop by hand on their own grid (`dropTarget_grid='Files'` + `AttachManager.onDropFiles` + `_uploader_onUploadingMethod`) are affected too and need `_uploader_onUploadedMethod` added on their side; the JS change here makes that possible.

## Fix

- `at_attachmentPane`: add `_uploader_onUploadedMethod=self.onUploadedAttachment`
- `AttachManager.onDropFiles`: forward `onUploadedMethod` into the RPC params

Note: this restores the drop. Layer 2 of #722 (no error handling around `site.uploadFile`, failures still not surfaced to the client) remains open.

fporcari:
Hotfix on `master`: dropping a file onto the grid of `attachmentPane()` uploads it to storage but writes no record — grid empty, no toast, no console error, no traceback.

## Cause

Regression left behind by the fix for #722.

`at_attachmentPane` arms the grid as a `Files` drop target passing only the **pre-upload** callback, and `AttachManager.onDropFiles` forwards only that one in the RPC params. Until 5618b5d6f that was enough, because `onUploadingAttachment` did the insert + commit itself. That commit moved record creation into `onUploadedAttachment`, which `rpc_upload_file` invokes **only if `onUploadedMethod` is in the kwargs** — on this path it never is.

56a30b45e closed the same hole for the three view `dropUploader`s; the grid-drop path of `at_attachmentPane` was missed, and so was `at_slotbar_screenshot`.

## Change

- `at_attachmentPane`: add `_uploader_onUploadedMethod=self.onUploadedAttachment`
- `AttachManager.onDropFiles`: forward `onUploadedMethod` into the RPC params
- `at_slotbar_screenshot`: same omission, third site of the regression. It builds its upload kwargs without `onUploadedMethod`, so `genro.dev.takePicture` → `htmlToCanvas` → `uploadMultipart_oneFile` stores the snapshot and never reaches `onUploadedAttachment`. Reachable in production through `at_attachmentGrid(screenshot=True)`.
- `at_attachmentPane` grid drop now passes `onResult='this.form.reload();'`, the same idiom the sibling `dropUploader` in `_atc_viewerStack` already uses. Without it the record is created but the grid stays visually empty until a manual reload — `clientPublish('inserted_attachment', …)` in `onUploadedAttachment` has no subscriber anywhere in the repo.

No behaviour change beyond restoring the insert and the refresh. The JS side also lets projects that wire the same drop by hand on their own grid opt in, by declaring `_uploader_onUploadedMethod` — they are affected identically and cannot fix it without this.

## Scope

Does **not** close #722: Layer 2 of that issue (no error handling around `site.uploadFile`, storage failures still not surfaced to the client) stays open.

The screenshot path is fixed for the record creation only, not for the grid refresh: it goes through `fm.callOnCurrentIframe('dev','takePicture',[kw])`, and passing a callback across that iframe boundary needs its own verification. Worth a separate ticket together with the dead `inserted_attachment` publish.

## Verification

flake8 clean on the touched file, both files syntax-checked, and the two generated JS snippets (the `onDrop` attribute and the `dataController` body) parsed with `node --check`. No test infrastructure covers this flow: `gnrpy/tests/web/` has nothing on upload or attachmanager, and a real test would need multipart plus storage. Manual browser check on a form with `attachmentPane()` and with `attachmentGrid(screenshot=True)` still to be done.

The suite's only failure, `gnrbag_test.py::TestBasicBag::test_fillFromUrl`, is unrelated and pre-existing on `master`: it fetches `https://www.genropy.org/feed/`, which now 404s after the site migration. Already fixed on `develop` by aa7818667, which repoints it at `sitemap.xml`.

## Related issue

Fixes #1167

Note for the merger: since this PR targets `master` and not the default branch `develop`, GitHub does not register the closing reference (`closingIssuesReferences` is empty), so the merge will not auto-close #1167 — it needs closing by hand.


fporcari:
Nota per il merge: la base è `master`, non il default branch, quindi GitHub non registra il link di chiusura (`closingIssuesReferences` vuoto nonostante il `Fixes #1167` nel corpo). #1167 va chiuso a mano dopo il rilascio.

cgabriel (approved):

genro (approved):
Correct diagnosis: `rpc_upload_file` invokes `onUploadedMethod` only when present, `56a30b45ee` added it to the three `dropUploader` views but not to the grid drop nor the snapshot, and `AttachManager.onDropFiles` forwarded only `onUploadingMethod`. `onResult='this.form.reload()'` resolves `this` correctly through `funcCreate`'s hitch. Branch is cut from master (merge-bases coincide), carries only the two files. After merge, develop needs the back-merge: `_uploader_onUploadedMethod` is absent there. Multi-file drop reloads the form once per file, harmless.

fporcari:
@cgabriel this one is ready and waiting on you: approved by you and by @genro on
the current head (`6eb17f2b2`), `mergeStateStatus` `CLEAN`, no open threads.
`master` is yours — push restricted and `enforce_admins` on — so the merge and
the release are your call, not something I can do from here.

Two things to carry along with it.

**The back-merge into `develop` is not optional.** The fix does not exist there:
`resources/common/gnrcomponents/attachmanager/attachmanager.py` on `develop` has
no `_uploader_onUploadedMethod` on the `at_attachmentPane` grid drop, and
`attachmanager.js` there does not mention `onUploadedMethod` at all, so
`onDropFiles` still forwards only the pre-upload callback. Without the realign
the same drop stays broken on `develop` after this ships.

**#1167 will not close by itself.** The PR body describes the fix but carries no
`Fixes #1167`, so GitHub links nothing — `closingIssuesReferences` is empty and
the issue is still open. I am leaving the body as it was reviewed; the issue needs
closing by hand after the merge.

PR #1178, opened 2026-08-26
added commodity methods to awsmanager's Route53, and mapped methods

cgabriel:
added commodity methods to awsmanager's Route53, and mapped methods in the cloudmanager/awsmanager service accordingly.

The methods allows simple low-level operations on route53, and higher logic method are added, like ensure_cname_record. This method will be used in a deployer context, when the user enters the requested FQDN for the applicatio, it can:

1/ check if the zone is managed - if not, just report the matter, leaving to the user update the non-manager zone accordingly
2/ check if the record already exists - if so, reports an error, to avoid disruption
 with existing services
3/ create the record with the provided value (the ingress/endpoint server, up to the caller)

The method supports dryrun to verify all the above during a wizard, and then execute the change at commit time.

Added relevant tests using mocked boto3

### DIFF
--- a/gnrpy/gnr/utils/awsmanager.py
+++ b/gnrpy/gnr/utils/awsmanager.py
@@ -478,22 +478,84 @@ class Route53Manager(BaseAwsService):
             HostedZoneId=hosted_zone_id)['ResourceRecordSets']
         return dict([(r['Name'].lower(), record_to_dict(r)) for r in resource_records])
 
-    def update_record(self, hosted_zone_id=None, name=None, 
-        type=None, target=None, ttl=360):
+    def update_record(self, hosted_zone_id=None, name=None,
+        type=None, target=None, ttl=300):
         r53 = self.client
-        parameters = {'Comment': 'add %s -> %s' % (name, target),
-                    'Changes': [
-                            {
-                             'Action': 'UPSTERT',
-                             'ResourceRecordSet': {
-                                 'Name': name,
-                                 'Type': type,
-                                 'TTL': ttl,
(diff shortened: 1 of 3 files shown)


mbertoldi (approved):

fporcari (approved):
The interesting part of this PR is not the new methods, it is that `update_record` was **completely non-functional** before it and now works. Three independent defects in one method:

- `'Action': 'UPSTERT'` — not a valid Route53 action, so the API rejected the batch;
- `r53.client.change_resource_record_sets(...)` where `r53 = self.client` already — an `AttributeError` before any request went out;
- the parameters passed flat instead of wrapped in `HostedZoneId` + `ChangeBatch`, which is the shape the API requires.

Any one of those makes every call fail, so this is a real fix and not just an addition. No caller exists in this repo (`git grep update_record` finds only the definition), so there is no regression surface either.

The rest reads well. `find_managed_zone` walking the label suffixes and stopping before the TLD is the right way to resolve `a.b.example.com` to `example.com`, `ensure_cname_record` raising on a value mismatch rather than silently overwriting is the correct default for DNS, and the `dryrun` split — validate during setup, write after commit — is a sensible contract for a caller inside a transaction. The three `route53_*` pass-throughs in the cloudmanager service add no logic, which is what a service mapping should look like.

Mocking `boto3` here is the right call: it is an external network API, which is precisely the case where a mock isolates rather than hides.

Non-blocking: `ensure_cname_record`'s docstring says "Four possible outcomes" and then lists five bullets.

Not verified: no call was made against a real Route53 zone; I read the request shapes against the API contract rather than executing them.

PR #1179, opened 2026-08-26
hotfix(adm): align notification audience on both paths, with static/dynamic lists

dgpaci:
Supersedes #964, now closed: this PR is the sole carrier of the change, and it targets `master` as a hotfix.

### The bug: one audience, two answers
The audience of `adm.notification` was evaluated in two divergent ways:
- **bulk snapshot** (triggers → `updateUserNotificationsFromQuery`): considered `tag_rule`, `group_code`, `linked_query`;
- **per-user alignment at login** (`updateGenericNotification`): considered only `all_users` and `tag_rule`.

So a notification restricted by `group_code` reached the users already enrolled and silently not the new ones. Both paths now share one evaluation, in two halves that both entry points pair:
- `audienceWhere()` builds the SQL pre-selection, with the criteria SQL can express: `group_code` and `linked_query`;
- `tagRuleMatches()` applies the tag rule on the candidate rows, delegating to `checkResourcePermission`.

The entry points are `audienceUserIds()` for the snapshot and `userMatchesAudience()` for the login.

**Tag rule.** Evaluated on `$all_tags`, the column `adm.main.authenticate` reads to build the avatar tags. A bare `$auth_tags LIKE '%tag%'` reaches a user tagged `superadmin` from a rule on `admin`, honors none of the rule syntax the permission engine understands (`%` wildcards, `NOT`, `;`), and misses the tags inherited from the group.

**Group code.** Matched on delimited codes, `',' || $all_groups || ','` against `%,code,%`. `all_groups` is a comma-join with no delimiter at the string boundaries, so the previous unanchored LIKE reached the users in `superadmin`, `admin_ro` or `nonadmin` from a rule on `admin` — a false positive, i.e. a group-restricted notification delivered outside its audience. This matters more now than it did before: that LIKE used to govern the snapshot alone, while the login ignored `group_code` entirely, so unifying the two paths applies the criterion at login as well and it has to be applied correctly.

Also fixed: a notification carrying only a `tag_rule` (no `linked_query`) was wrongly flagged `all_users`. `all_users` is now implied only when no explicit audience criterion is set.

### Dynamic vs static
New `dynamic_list` column, **dynamic by default**.
- **Dynamic** (default): newcomers matching the criteria are enrolled incrementally at login — what notifications have always done. The login path tests `dynamic_list IS NOT FALSE`, so the rows that predate the column (NULL) keep being delivered: **no migration needed**.
- **Static** (explicit opt-out): the audience is photographed once, at save time. If 1000 users match, those 1000 are frozen — newcomers are never added.

The list is re-photographed only when the audience criteria change (`tag_rule`, `group_code`, `linked_query`, `all_users`). The snapshot drops the pending rows and rebuilds membership from the current population, so firing it on the date window would enrol exactly the newcomers a static list exists to exclude.

### Activity window
New optional `start_date` / `end_date` columns (NULL = no limit). Outside the window a notification neither enrolls new users nor is surfaced to the user (`nextUserNotification` is date-gated).

### UI
Form/embed expose `dynamic_list`, `start_date`, `end_date`; the list view also shows `confirmed_count` / `pending_count` (idiomatic `formulaColumn(select=...)`).

## Deployment notes

- **This is not hotfix-shaped, and the decision is conscious.** It carries a schema change — three new columns on `adm.notification`, so instances need a `gnr migrate` — plus new UI and the static/dynamic feature. Whether it goes out in a hotfix release or a normal one is the maintainer's call, not something to inherit from the branch prefix.
- **No data migration**: `default=True` feeds `newrecord()` only and never becomes a SQL `DEFAULT`, so existing rows stay NULL — and `IS NOT FALSE` is what keeps them dynamic, i.e. behaving exactly as before. An installation that upgrades and never touches the new fields keeps the behavior it had.
- **Save-time cost**: saving a notification photographs its audience, so with `all_users` it writes one `adm.user_notification` row per user inside the insert transaction, where previously the eager insert only happened with a `linked_query`. When a `tag_rule` is set, `$all_tags` is a pyColumn and costs a query per candidate — on an operation that is already one insert per user. Worth timing against a realistic `adm.user` count before cutting the release; **not** measured here.

## Test plan

`projects/gnrcore/packages/adm/tests/test_notification_audience.py` — 19 tests on a throwaway SQLite application, run with `gnr dev tests <instance> adm`. Users, tags and groups are created through `adm.htag`, `adm.user_tag`, `adm.group` and `adm.user_group` the way the framework does, so the records go through the actual triggers.

Static / dynamic:
- static list frozen after the snapshot (newcomer not added);
- dynamic list grows at login;
- a new notification is dynamic unless opted out;
- a notification that predates the column (`dynamic_list` NULL, no snapshot of its own) still enrolls a user logging in.

Tag rule, each on both the snapshot and the login path:
- `tag_rule` audience coherent on both paths;
- a rule on `ZZADMIN` does not reach a user tagged `SUPERZZADMIN` (what a bare `LIKE '%tag%'` does);
- wildcards (`ZZADM%`), exclusions (`A NOT B`) and alternative rules (`A;B`) are honored;
- a tag inherited from the group matches on both paths;
- `all_users` implied only without explicit rules.

Group code — the criterion that previously had no audience coverage at all:
- `group_code` audience coherent on both paths;
- a notification restricted to `ZZADMG` does not reach a user in `SUPERZZADMG`;
- a group held through `adm.user_group` rather than through the user's own `group_code` still matches, which is where a delimited match on a comma-join could plausibly break;
- `group_code` alone does not imply `all_users`.

Re-snapshot and window:
- changing an audience criterion rebuilds the membership;
- editing only the date window does not re-photograph a static list (pending rows survive, newcomers stay out);
- the date window gates visibility (`nextUserNotification`).

Verified against `master`'s framework code: 19/19 passing, `gnrpy/tests/sql` green (1163 passed, 32 skipped), flake8 clean on the modified files. Mutation-checked: restoring the substring tag match fails 4 tests, restoring the unanchored group LIKE fails 1, restoring `IS TRUE` fails the legacy-notification one, restoring the date fields in the trigger fails the re-snapshot one.

## Known CI failure, unrelated to this change

`tests/core/gnrbag_test.py::TestBasicBag::test_fillFromUrl` fails on `master` with `HTTP Error 404`, and it fails for **any** PR opened against `master` today: master's copy of the test fetches `https://www.genropy.org/feed/`, which the site migration removed. `develop` already carries cgabriel's fix (`aa7818667d`, moving it to `/sitemap.xml`), one of the 134 commits master does not have yet.

That one-line fix is deliberately **not** included here, so this branch carries the hotfix and nothing else. It also would not fix the underlying problem: the test still reaches the live network either way, which the project's own guidelines rule out for tests — the durable fix is a local fixture, in its own PR against `develop`.

### DIFF
--- a/gnrpy/tests/core/gnrbag_test.py
+++ b/gnrpy/tests/core/gnrbag_test.py
@@ -27,8 +27,8 @@ class TestBasicBag(object):
         assert b == self.mybag
         
     def test_fillFromUrl(self):
-        b = Bag('https://www.genropy.org/feed/')
-        assert b['rss.channel.title'] == 'Genropy'
+        b = Bag('https://www.genropy.org/sitemap.xml')
+        assert b['urlset.url']
 
     def test_fillFromXml(self):
         b = Bag("<name>John</name>")
--- a/projects/gnrcore/packages/adm/localization.xml
+++ b/projects/gnrcore/packages/adm/localization.xml
@@ -204,6 +204,12 @@
 				<en_for_all_users fr="Pour tous les utilisateurs" de="Für alle Benutzer" en="For all users" it="Per tutti gli utenti" base="For all users"/>
 				<en_letterhead fr="Tête" de="Briefkopf" en="Letterhead" it="Carta intestata" base="Letterhead" zh="信头"/>
 				<en_linked_query base="Linked query" it="Query collegata" en="Linked query" zh="关联查询"/>
+				<en_dynamic_list base="Dynamic list" en="Dynamic list" it="Lista dinamica"/>
+				<en_dynamic base="Dynamic" en="Dynamic" it="Dinamica"/>
+				<en_confirmed_users base="Confirmed users" en="Confirmed users" it="Utenti confermati"/>
+				<en_conf_users base="Conf.users" en="Conf.users" it="Utenti conf."/>
+				<en_pending_users base="Pending users" en="Pending users" it="Utenti da confermare"/>
+				<en_pend_users base="Pend.users" en="Pend.users" it="Utenti pend."/>
 		</notification>
 		<pkginfo path="model/pkginfo.py" ext="py">
 			<en_package_info base="Package info" en="Package info" it="Informazioni sul pacchetto"/>
--- a/projects/gnrcore/packages/adm/model/notification.py
+++ b/projects/gnrcore/packages/adm/model/notification.py
@@ -16,31 +16,78 @@ class Table(object):
         tbl.column('all_users','B',name_long='!!For all users')
         tbl.column('letterhead_id',size='22',group='_',name_long='!!Letterhead').relation('htmltemplate.id',relation_name='notifications',mode='foreignkey')
         tbl.column('linked_query', dtype='X', name_long='!![en]Linked query',_sendback=True)
+        # Dynamic by default: this is how notifications behaved before the
+        # column existed, so the rows that predate it (NULL) and the new ones
+        # (True) keep enrolling users at login. Static is an explicit opt-out.
+        tbl.column('dynamic_list','B',default=True,
+                   name_long='!!Dynamic list', name_short='!!Dynamic')
+        tbl.column('start_date','D',name_long='!!Start date')
+        tbl.column('end_date','D',name_long='!!End date')
         
         tbl.formulaColumn('existing_for_current_user',"""
                 (EXISTS(SELECT * FROM adm.adm_user_notification AS un WHERE un.user_id=:env_user_id AND un.notification_id=#THIS.id))
             """,dtype='B')
 
+        tbl.formulaColumn('confirmed_count',select=dict(table='adm.user_notification',
+                                                        columns='COUNT(*)',
+                                                        where='$notification_id=#THIS.id AND $confirmed IS TRUE'),
+                                            dtype='L',name_long='!!Confirmed users',
+                                            name_short='!!Conf.users')
+
+        tbl.formulaColumn('pending_count',select=dict(table='adm.user_notification',
+                                                      columns='COUNT(*)',
+                                                      where='$notification_id=#THIS.id AND $confirmed IS NOT TRUE'),
+                                          dtype='L',name_long='!!Pending users',
+                                          name_short='!!Pend.users')
+
     def trigger_onInserting(self, record):
-        if not record['linked_query']:
+        if self._hasNoAudienceRule(record):
             record['all_users'] = True
-    
+
     def trigger_onInserted(self, record):
-        if record['linked_query']:
-            self.updateUserNotificationsFromQuery(record)
-            
+        # The audience is photographed as soon as the notification is saved:
+        # for a static list this snapshot *is* the list, and a dynamic one
+        # still has to reach the users matching right now. With `all_users`
+        # that means one adm.user_notification row per user, written inside
+        # the insert transaction: saving a notification costs one insert per
+        # user of the installation. That is the intended price of a snapshot.
+        self.updateUserNotificationsFromQuery(record)
+
     def trigger_onUpdating(self, record, old_record=None):
-        if not record['linked_query']:
+        if self._hasNoAudienceRule(record):
             record['all_users'] = True
-            
+
     def trigger_onUpdated(self, record, old_record=None):
-        if record['linked_query'] and self.fieldsChanged('linked_query,tag_rule,all_users,group_code', record, old_record):
+        # Only a change to the audience criteria re-photographs the list: the
+        # snapshot drops the pending rows and rebuilds membership from the
+        # current population, so firing it on `dynamic_list` or on the date
+        # window would silently re-photograph a static list -- extending the
+        # window of a static notification would enrol exactly the newcomers
+        # it is meant to exclude, and drop the users who no longer match.
+        if self.fieldsChanged('linked_query,tag_rule,all_users,group_code', record, old_record):
             self.updateUserNotificationsFromQuery(record)
-            
-    def updateUserNotificationsFromQuery(self, notification_record):
+
+    def _hasNoAudienceRule(self, record):
+        # all_users is implied only when no explicit audience criterion is set
+        return not (record['linked_query'] or record['tag_rule'] or record['group_code'])
+
+    def audienceWhere(self, notification_record):
+        """Build the (where, kwargs) pre-selecting the adm.user rows that are
+        candidates for this notification audience.
+
+        Only the criteria SQL can express live here, `group_code` and
+        `linked_query`. The tag rule is applied by `tagRuleMatches` on the
+        candidate rows instead: it is a permission-engine expression (exact
+        tags, `%` wildcards, `AND`/`NOT`, `;`) that a LIKE cannot reproduce.
+        Callers must therefore pair this pre-selection with the tag check --
+        `audienceUserIds` and `userMatchesAudience` are the two entry points
+        that do, and that is what keeps the bulk snapshot and the per-user
+        login alignment on one single rule.
+
+        Returns (None, {}) when the notification targets all users."""
+        if notification_record['all_users']:
+            return None, {}
         where = []
-        user_tbl = self.db.table('adm.user')
-        user_notification_tbl = self.db.table('adm.user_notification')
         selection_kwargs = {}
 
         # Check if user has at least one of the specified groups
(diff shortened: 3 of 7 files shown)


fporcari (approved):
Approving for the release, with the open point stated so it is not lost.

**What is right here.** The core finding is real and the cure is the correct one: the audience was genuinely evaluated two divergent ways — the bulk snapshot honoring `tag_rule`+`group_code`+`linked_query`, the login alignment only `all_users`+`tag_rule` — so a notification restricted by `group_code` reached the already-enrolled users and silently not the new ones. Collapsing both onto `audienceWhere()` + `tagRuleMatches()` is the fix that makes the static/dynamic distinction meaningful at all.

Two design calls worth naming because a naive fix would have got them wrong:

- **`dynamic_list IS NOT FALSE`** on the login path means rows predating the column (NULL) keep enrolling users, so no data migration is needed and an installation that upgrades without touching the new fields keeps the behavior it had.
- **The tag rule leaves SQL** and goes through `checkResourcePermission` on `$all_tags` — the column the login itself reads to build the avatar tags — so wildcards, `NOT`, `;` and tags inherited from the group all work, where a bare `$auth_tags LIKE '%tag%'` honored none of them.

I verified both against the code, not the description, and the branch is correctly cut from `master` (the fork points against `master` and `develop` are the same commit). The six `adm` files are byte-for-byte identical to #964's current head, so this carries the resolved version.

## Open point, deliberately deferred — please read before cutting the release

The tag rule was made strict; **`group_code` was not**. `audienceWhere` still matches groups with an unanchored LIKE:

```python
group_conditions.append(f"$all_groups LIKE :group_{i}")
selection_kwargs[f'group_{i}'] = f'%{g}%'
```

and `all_groups` is a plain comma-join of group codes with no delimiter at the string boundaries. So a notification restricted to group `admin` also reaches users in `superadmin`, `admin_ro` or `nonadmin`. The error direction is the unsafe one — a **false positive**, so the notification is delivered to more users than intended, which for a group-restricted notification is disclosure outside the audience rather than a missed delivery.

This is accepted here rather than fixed because the change is still a net improvement over `master` today, and because the proper fix wants a test alongside it. It is **blocking on #964**, where I have requested changes for exactly this, so it lands on `develop` with coverage and reaches `master` in the next release. The boundary match is two lines:

```python
group_conditions.append(f"',' || $all_groups || ',' LIKE :group_{i}")
selection_kwargs[f'group_{i}'] = f'%,{g},%'
```

Note that this means the `group_code` bug named in the PR summary is only partly closed by this PR: new users now enter the evaluation, but which users match is still approximate.

## Two things for whoever cuts the release

- **This is not hotfix-shaped.** It carries a schema change — three new columns on `adm.notification`, so instances need `gnr migrate` — plus new UI and the static/dynamic feature. Worth deciding consciously whether it belongs in a hotfix release or a normal one, rather than inheriting the `hotfix` label.
- **The PR is assigned to its author.** Hotfixes on `master` are the maintainer's to review, merge and release; @cgabriel should hold this one if it is going out.
- `gnrpy/tests/core/gnrbag_test.py` is still in this branch and unrelated to notifications (see my earlier comment) — it should come out, but it does not block the release.

Not verified: I did not run the suite on this branch, and I did not measure the save-time snapshot cost against a realistic `adm.user` count — the timing check the deployment notes ask for is still owed.

genro (commented):
Read the branch at `68ad40424` rather than the description, and checked the three things this PR's shape depends on.

**The hotfix routing is correct.** The fork points of the branch against `master` and against `develop` are the same commit (`d17fedff9`), which is the only reliable test — `git merge-base --is-ancestor origin/master HEAD` passes for a branch grown on `develop` too, since `develop` contains `master`. It would carry two commits into `master` and nothing unreleased. Assigned to @cgabriel, which is the right assignee for a hotfix.

**Both of @fporcari's asks are already closed on this head.** The seventh file is gone — the PR now touches six, all under `adm`. And the `group_code` matching is delimited on both sides, `',' || $all_groups || ','` against `%,{g},%`, so a rule on `admin` no longer reaches `superadmin`, `admin_ro` or `nonadmin`. The comment above it explains why, which is worth keeping.

One thing I probed and found sound, so it does not become a review round: the `linked_query` branch rebinds `selection_kwargs` from `sqlWhereFromBag`. That would drop the `group_*` keys if the callee built a fresh dict, and the WHERE clause still names `:group_0`. It does not — `sqlWhereFromBag` (`gnr/sql/gnrsqltable/query.py:95-105`) returns the same dict it was handed, substituting one only when `None` is passed.

## The red build, and why it is not what it looks like

CI fails on 3.11 with `tests/core/gnrbag_test.py::TestBasicBag::test_fillFromUrl` — `HTTPError: 404` on `https://www.genropy.org/feed/`. The other three versions are cancelled by the first failure.

@fporcari asked for the `gnrbag_test.py` change to be dropped and was right that it does not belong in a hotfix. But his third argument — that it does not fix the underlying problem and the durable fix belongs in its own PR against `develop` — describes work that has already happened: `aa7818667` ("update bag test to use an existing remote URL, the previous one doesn't exists anymore due to site migration") is on `develop` and points the test at `sitemap.xml`. `master` still carries `feed/` at line 30.

So the removed hunk was not an improvised workaround. It was `develop`'s existing fix, carried along because **this branch is cut from `master`, which does not have that commit yet** — and that is normal, `develop` being ahead of `master` is what `develop` is for. The anomaly is only that a hotfix branch runs the whole suite and inherits a test that is already fixed on the other side.

@cgabriel, this is your call as the one who cuts releases and realigns:

- **accept the red build**, noting in the PR that the failure is `test_fillFromUrl` and that it is already fixed on `develop` by `aa7818667`. `master` has `required_status_checks: []`, so nothing is actually blocked;
- **cherry-pick `aa7818667` onto `master`** before merging this. One commit, and every hotfix branch cut from `master` afterwards starts green.

The second costs a minute and removes the problem for the next hotfix too, but the choice is yours.

Not verified: I did not run the suite on this branch, and I did not review the audience logic itself beyond the `group_code` point — @fporcari has already approved that half and these six files are identical by construction to #964, which is where that review lives.

genro (approved):
Approving `68ad40424`, having now read the audience logic itself rather than only the `group_code` point I checked earlier.

The unification is real and it is the right shape: both entry points go through the same `audienceWhere()` + `tagRuleMatches()` pair, `audienceUserIds()` for the snapshot and `userMatchesAudience()` for the login, so the two paths cannot drift again the way they had. Delegating the tag rule to `checkResourcePermission` instead of a `LIKE` is what makes the rule syntax work at all — `%` wildcards, `NOT`, `;` — and it picks up the tags inherited from the group, which the old `$auth_tags LIKE` never saw.

Evaluating on `$all_tags` rather than `$auth_tags` is the detail that makes it correct, since that is the column the avatar is actually built from. And `dynamic_list IS NOT FALSE` rather than `IS TRUE` is what makes the upgrade free: rows predating the column stay NULL and keep behaving exactly as before, so an installation that never touches the new fields notices nothing.

**One asymmetry worth a follow-up, not a block.** `audienceUserIds` guards the expensive column — `if not tag_rule:` returns the plain pkeylist and never asks for `$all_tags`. `userMatchesAudience` does not: it always fetches `columns='*,$all_tags'`, so a notification with only a `group_code` or a `linked_query` still pays the pyColumn. That lands on the login path, which now also fetches strictly more rows than before — the old query filtered `($all_users IS TRUE OR $tag_rule IS NOT NULL)` and the new one fetches every dynamic notification not yet delivered to that user, by design, since ignoring `group_code` at login was the bug. So the per-login cost grows on two axes at once. Mirroring the same early return in `userMatchesAudience` costs three lines.

**A question rather than a finding**: the date window is compared against `self.db.workdate`, not the real date. That is the framework's convention for business dates and may well be intended, but a user working with a back-dated workdate then sees a different set of notifications. Worth being deliberate about, since a notification window feels like wall-clock rather than business time.

On the deployment note: I read it and agree with putting it there rather than in the branch prefix. Three new columns on `adm.notification` means instances need `gnr migrate`, and whether that rides a hotfix release is @cgabriel's call. The save-time cost you flag as unmeasured is the one I would want a number for before cutting the release, since with `all_users` it is one `adm.user_notification` insert per user inside the save transaction.

Not verified: I did not run the suite on this branch, and the 19 tests are described but not executed here.

fporcari (approved):
Approved on the current head.

One follow-up is worth tracking, but it does not block this PR: an existing notification that was implicitly all_users cannot currently be narrowed by adding a tag_rule, group_code, or linked_query. trigger_onUpdating() sets all_users to True when no audience rule exists, but it does not clear a previous implied True when a restriction is added. audienceWhere() and audienceTagRule() then give all_users precedence, so the new restriction is ignored.

This behavior predates the PR, and the audience unification here is still a clear improvement. Please track the all-users-to-restricted transition separately, with a real-database regression test.

PR #1182, opened 2026-08-26
fix: make the relative storage service usable

### LINKED ISSUE #834: [BUG] Storage relative service
### Problem Description

Stack is as follows, when adding a relative storage service:

  File "/home/ubuntu/genropy/gnrpy/gnr/web/gnrwebpage.py", line 623, in _rpcDispatcher
    result = self.rpc(method=method, _auth=auth, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/web/gnrwebpage_proxy/rpc.py", line 38, in __call__
    result = handler(**kwargs)
             ^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/resources/common/gnrcomponents/storagetree.py", line 109, in st_getStorageRes
    result.setItem('root',StorageResolver(storagepath,cacheTime=2,
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/core/gnrbag.py", line 2715, in __call__
    result = self.load()
             ^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/lib/services/storage.py", line 839, in load
    self.storageNode = self._page.site.storageNode(self.storageNode)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/web/gnrwsgisite.py", line 734, in storageNode
    return self.storage_handler.storageNode(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py", line 417, in storageNode
    return self.makeNode(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py", line 523, in makeNode
    service = self.storage(service_name)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py", line 462, in storage
    return self.site.getService(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/web/gnrwsgisite.py", line 713, in getService
    return self.services_handler.getService(service_type=service_type,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/lib/services/__init__.py", line 74, in getService
    return self(service_type)(service_name, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/lib/services/__init__.py", line 237, in __call__
    service = self.addService(service_name, **kwargs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/gnrpy/gnr/lib/services/__init__.py", line 94, in addService
    service = service_factory(self.site, **service_conf)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/genropy/resources/common/services/storage/relative.py", line 14, in __init__
    delattr(self, attr)
AttributeError: 'Service' object has no attribute '_call'

### Reproduction Steps

Just add a "relative" storage service to raise Error

_No response_

### Environment

genropy@latest

### Additional Material

_No response_

dgpaci:
## Problem

Adding a `relative` storage service always failed with `AttributeError: 'Service' object has no attribute '_call'`: the implementation tried to strip the `StorageService` methods from a fresh instance with `delattr` (they are class attributes, so the first name already raised) and then to morph the instance into the parent service class, losing the `base_path` property that was its whole purpose. The implementation has never worked.

## Solution

A relative service is built as a subclass of the parent service class, `Service` first in the mro:

```
Relative_local -> Service -> local.Service -> BaseLocalService -> StorageService
```

`internal_path`, `open`, `children`, `url`, `serve`... are the ones of the parent implementation, whatever it is (local, aws_s3, sftp), while `Service` overrides only `base_path` (parent base path + relative path). The parameters are not copied: they are read from the live parent service, so credentials and buckets have a single source. A relative service can also be the parent of another one.

A service record is created before its parameters are filled in, so a relative service with no usable parent is an ordinary state, not an error: it behaves as an empty storage (no base path, no children, `exists()` false) instead of raising, otherwise the storage tree of the very form where the parent gets chosen fails on the service it configures. Writing on it is still refused with an explicit message: nothing can be written where there is no path.

Two fixes on the parameters form:

- the parent can now be any configured storage service (built-in and siteconfig ones included), instead of the sole services registered in `sys.service`; the configured parent stays an option even if it is not available any more;
- the pane is rebuilt on every record load and the form kept the validations of the widgets destroyed by the rebuild, so on a new record the status stayed invalid whatever was chosen. `th_service` now forgets them the same way the `bagField` and `grouplet` dynamic contents already do (`form.checkInvalidFields()` on `_onRemote`).

## Test cases

`gnrpy/tests/web/gnrstoragehandler_test.py`, on the real site infrastructure, with files written and read back on disk:

- `test_relative_storage_creation`: the service is created and is an instance of the parent implementation (fails on the old code with the `AttributeError` of the issue);
- `test_relative_storage_base_path` / `test_relative_storage_without_relative_path`: base path composition, an empty relative path is an alias of the parent;
- `test_relative_storage_write_and_read`: a file written through the relative service is the same file the parent sees under the relative path;
- `test_relative_storage_children`: listing is limited to the subtree;
- `test_relative_storage_inherits_parent_parameters`: parameters and `location_identifier` come from the parent (so copies stay native);
- `test_relative_storage_nested`: relative service over a relative service;
- `test_relative_storage_missing_parent_service` / `test_relative_storage_unsupported_parent_service` / `test_relative_storage_tree_without_parent_service`: an unconfigured service is an empty, non-writable storage and its tree loads empty;
- `test_relative_storage_from_service_record`: end to end from a real `sys.service` record, through the triggers, the `storage_params` registry, the service factory, a written file and the `StorageResolver` used by the tree;
- `test_relative_storage_parent_options`: the parents offered by the form.

Manual test on a running instance: create a storage service with implementation `relative`, choose a parent and a relative path, save; the tree of the chosen folder shows up and files written on the service land in the parent subfolder.

Fixes #834

### DIFF
--- a/gnrpy/tests/web/gnrstoragehandler_test.py
+++ b/gnrpy/tests/web/gnrstoragehandler_test.py
@@ -4,6 +4,9 @@ import tempfile
 import shutil
 
 from gnr.core.gnrbag import Bag
+from gnr.core.gnrconfig import getGenroRoot
+from gnr.core.gnrlang import GnrException, gnrImport
+from gnr.lib.services.storage import StorageResolver
 
 from webcommon import BaseGnrDaemonTest
 
@@ -67,7 +70,7 @@ class TestStorageHandler(BaseGnrDaemonTest):
             assert 'implementation' in params, \
                 f"Parameters for '{storage_name}' should have 'implementation' key"
             impl = params['implementation']
-            assert impl in ['local', 'symbolic', 'raw', 'http', 'aws_s3'], \
+            assert impl in ['local', 'symbolic', 'raw', 'http', 'aws_s3', 'relative'], \
                 f"Unknown implementation '{impl}' for storage '{storage_name}'"
 
     # ========================================================================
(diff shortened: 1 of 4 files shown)


fporcari (changes requested):
The design choice here is the right one, and it is worth saying why before the objections. Building a dynamic subclass `type('Relative_<impl>', (Service, parent_class), {})` instead of mutating an instance gives an MRO of `Relative_local -> Service (relative) -> Service (local) -> BaseLocalService -> StorageService -> GnrBaseService`, which I confirmed on a live site. `Service` genuinely lands first, so `base_path` is a read-only property that wins, and everything else is the parent implementation untouched.

Two consequences make it right rather than merely clever. The parent backend's `__init__` never runs, so nothing tries to assign `base_path` — and `base_path` is only ever assigned inside each backend's own `__init__` (`storage.py:737`, `aws_s3.py:117`, `sftp.py:60`, `compressed.py:15`, `symbolic.py:13`), never from outside, so there is no caller this breaks. And delegating to the *live* parent means the child shares its mutable state rather than a copy: with an sftp parent it shares `_thclient` and `lock`, i.e. the same connection pool; with s3 it shares the boto session once built; and `location_identifier` comes from the parent, so parent-to-child copies take the native fast path.

The old-failure claim in the description checks out verbatim. `dir(StorageService) - dir(GnrBaseService)` sorts `_call` first, and `delattr` on it raises exactly the `AttributeError` quoted. The implementation had never worked.

The tests are real — no mocks at all, a real site, real files written and read back from disk, a real `sys.service` record through the insert triggers, the real `StorageResolver`. I ran them: 47 passed on `develop`, 59 on this head, so 12 new tests that all exercise the mechanism.

Two things I would like fixed before this lands. Both are inside `relative.py:58-69` and both were reproduced on a live site.

**1. No cycle guard: a relative service pointing at itself, or two pointing at each other, gives `RecursionError`.**

`relative.py:63` calls `self.parent.storage(parent_service)` from inside `__init__`, but `BaseServiceType.addService` only caches the instance *after* the factory returns (`gnr/lib/services/__init__.py:104-110`). The lookup therefore never hits the cache and recurses until the stack ends.

This is reachable from the very form this PR fixes: `relativeStorageParents` (`relative.py:152-154`) excludes only the service itself plus symbolic and http, so other relative services are offered as parents. Create `rel_a` with parent `rel_b`, then edit `rel_b` to point at `rel_a`, and any access to either raises — including the storage tree in that same form. Self-parent is reachable by direct DB edit or siteconfig. I confirmed both cases.

A set of names currently being resolved, or a walk of `storage_params` for the parent chain before instantiating, returning `None` on a cycle so it falls into `UnconfiguredService`, would close it.

**2. `_resolveParentService` does not return `None` for a missing parent, so the storage silently re-roots under `site_static_dir`.**

The docstring at `relative.py:59` says "None if it is missing or unusable", but that only happens for an *empty* `parent_service` and for symbolic/http. For a name that does not exist, `LegacyStorageHandler.storage` (`gnrstoragehandler.py:450-458`) falls back to creating a `local` service rooted at `site_static_dir/<name>`. Verified live:

```
parent_service='definitely_not_a_service', relative_path='docs'
-> class Relative_local
-> base_path .../instances/gnrdevelop/site/definitely_not_a_service/docs
```

And after deleting a real parent and clearing the child from `service_instances`, the child silently moved from `/tmp/.../pp/sub` to `.../site/p_par2/sub`. So the case the description presents as "an ordinary state, not an error" is, for a deleted parent, documents written into a freshly created directory somewhere else — with no error and no log. `test_relative_storage_parent_options` even asserts that a removed parent stays selectable "even if it is not available any more", which is reasonable in the form but means the runtime path has to handle it.

Checking `self.parent.storage_handler.getStorageParameters(parent_service)` before calling `storage()`, and returning `None` when it is absent, is one line. The `site_static_dir` fallback itself is pre-existing framework behaviour — this PR is just where it turns into a data-location hazard.

Three smaller notes, none blocking:

- **A `local` parent with an empty `base_path` makes the child write relative to the process CWD.** `BaseLocalService.__init__` (`storage.py:737`) sets `base_path = None` when the parameter is falsy, and the framework acknowledges that state (`isdir` guards on it at `:799`). `relative.py:74-75` then returns the bare `relative_path`, and `internal_path` does `os.path.join('docs/invoices', 'x.txt')`. The `not parent_base_path` branch is correct for s3 and sftp, where `''` means bucket root; distinguishing `is None` from `''` would cover local too.
- **`UnconfiguredService` refuses less than its docstring claims.** `relative.py:88-131` overrides `open`, `local_path`, `mkdir` and `makedirs`, but not `delete`, `size`, `mtime`, `md5hash`, `serve`, `copy`, `move`, `renameNode` or `duplicateNode`. On a symbolic-parent service I saw `delete('a.txt')` return `None` — a silent no-op — `size` return `None`, and `serve` raise `KeyError: 'REQUEST_METHOD'` instead of the class's own message. `delete` is the one that matters.
- **The warning at `relative.py:61` does not name the service.** `service_name` is not set until after `__init__` returns (`services/__init__.py:105`), so with more than one relative service the line is not actionable. The other branch already logs `parent_service_name`; this one could log `relative_path`.

The plumbing all connects, for the record: `buildServiceParameters` already passed `service_name` on `develop`, `_onRemote` is the existing idiom (`genro_components.js:3126`, `dynamicform.py:369`) and `checkInvalidFields` exists at `genro_frm.js:2049`; the localization keys resolve as `en_parent_service` through the flat lookup, and the root `localization.xml` is the right file since the `core` slot has `resources/common` among its roots. The `dbSelect` to `filteringSelect` change is a real improvement — siteconfig and built-in storages like `site`, `home` and `mail` become selectable, which the `sys.service`-only `dbSelect` could never offer. Existing records keep an orphan `parent_service_id` key that `**kwargs` absorbs harmlessly.

Not verified: no real aws_s3 or sftp parent — no credentials — so those conclusions come from reading `aws_s3.py` and `sftp.py` and the MRO, not from a run. I did not open the `sys.service` form in a browser, so the `_onRemote` fix and the new options list are verified as code paths only. I ran `tests/web/gnrstoragehandler_test.py`, not the full suite, and did not check multidomain resolution of the parent.

dgpaci:
Both blocking points and the three notes are addressed in ef0920a95, on top of a rebase on current `develop`. Every finding is now a test that fails on the previous head — I ran them against it before fixing, and the failures reproduce your observations verbatim, including the re-rooting under `site_static_dir` and the two `RecursionError`.

**1. Loop in the parent chain.** `_loopingParentChain` walks the chain on the stored parameters before the parent gets built, and returns `None` on a loop, so the service falls into `UnconfiguredService`. The walk starts from the parent name rather than from the service itself on purpose: `service_name` is assigned only after the factory returns (`services/__init__.py:105`), so at `__init__` time the service does not know its own name — but revisiting any name in the chain is enough to catch both a self-parent and two services pointing at each other. `test_relative_storage_looping_parent_services` and `test_relative_storage_parent_of_itself` cover them; both raised `RecursionError` before.

**2. Missing parent.** The parent is now looked up with `getStorageParameters` first, and an absent one leaves the service unconfigured, so the legacy `site_static_dir` fallback is never reached. `test_relative_storage_missing_parent_service_name` asserts the base path is `None` and that no folder appeared under the site static dir; on the previous head it resolved to `.../site/rel_gone_parent_storage/docs`. `test_relative_storage_deleted_parent_service` reproduces your live scenario — a working relative service, its parent removed, the child re-resolved — and it moved to `.../site/rel_deleted_parent/sub` before, base path `None` now.

**3. Local parent with an empty base path.** Fixed, but at resolution time rather than in the property: a parent whose `base_path` is `None` is refused, so the service is unconfigured instead of exposing a `None` base path that `internal_path` would turn into `os.path.join(None, ...)`. That keeps the `not parent_base_path` branch correct for the bucket-root `''` of s3 and sftp — `BaseLocalService` is the only implementation that yields `None` (`storage.py:737`; sftp and s3 both normalise to `''`). The same check ends the chain on a relative parent that is unconfigured itself, which had the same effect: `test_relative_storage_parent_without_base_path` resolved to `docs` before, `test_relative_storage_on_unconfigured_relative_parent` to `2026` — both relative to the process CWD, as you said.

**4. `UnconfiguredService` refusing less than its docstring.** `delete` now raises, and so do `renameNode` and `duplicateNode`, so every write entry point either raises the message or is unreachable. `size`, `mtime` and `md5hash` I left returning `None`, consistent with `exists()` being false — they are the abstract stubs of `StorageService`, and an empty storage answering "nothing there" reads better than an exception on a read. On `serve`: the base implementation already returns `not_found_exception` when `internal_path` is falsy (`storage.py:665-668`), and `not_found_exception` reads the environ through `Request(environ).url` (`gnrwsgisite.py:1506`), so the `KeyError: 'REQUEST_METHOD'` comes from the synthetic environ rather than from the service — with a real request it is a 404, which is what an empty storage should answer.

**5. Log line without the service name.** It now logs the relative path, as you suggested.

`tests/web/gnrstoragehandler_test.py` is at 65 tests (59 before, 6 new). Full suite on the rebased head: 2380 passed, 10 skipped. flake8 clean on the three modified files.

Still not verified, same as your review: no real aws_s3 or sftp parent, and the form changes are verified as code paths only.

fporcari (approved):
Both blocking points are closed, and closed in the code rather than in the description. Approving. Below is what I verified myself on ef0920a9, not what I took from your reply.

**1. Cycle guard.** `_loopingParentChain` walks `getStorageParameters` before anything is instantiated, so the recursion is never entered. I traced the three shapes: self-parent (`X -> X`) revisits `X` on the second pass and returns `bool('X')`; the two-service cycle revisits the entry name; a longer cycle `X -> Y -> Z -> X` revisits `Y`. Starting the walk at the parent rather than at the service is sound precisely because any reachable cycle gets revisited, and it sidesteps the fact that `service_name` is not assigned until `services/__init__.py:105`. A chain that ends on a non-relative implementation returns False at the first non-relative link, and one that ends on an unconfigured relative exits with a falsy name — also False, and then refused one check later by the `base_path is None` test, not by the guard. Both correct.

**2. Missing parent.** `getStorageParameters` first, `None` when absent, so `LegacyStorageHandler.storage`'s `site_static_dir` fallback is unreachable. Worth recording that the guard does not cost the feature it was added next to: `storage_params` is populated from `DEFAULT_STORAGE_CONFIGS`, the dynamic `home`/`site`/`mail`, siteconfig `services`, the legacy `volumes` section and the `sys.service` rows, and `relativeStorageParents` reads the same registry through `getAllStorageParameters`. Form and runtime therefore agree on exactly one set of names — a built-in or siteconfig parent stays usable, which is what the `dbSelect` -> `filteringSelect` change was for.

**3-5.** Refusing a parent whose `base_path` is `None` at resolution time is better than distinguishing `None` from `''` inside the property, and it ends the chain on an unconfigured relative parent with the same one check. On `UnconfiguredService`: `delete`, `renameNode` and `duplicateNode` raising does close the write surface — I checked the two entry points that are not overridden, and both are covered. `copy`/`move` into an unconfigured destination reach either `duplicateNode` (same `location_identifier`) or `copyNodeContent` -> `open` (different), so both raise the message; with an unconfigured *source*, `isfile` and `isdir` are false and the call is a no-op, which is right for an empty storage. `size`/`mtime`/`md5hash` returning `None` matches `StorageService`, where they are bare `pass` — agreed, that reads better than raising on a read. And `serve` at storage.py:665-668 does return `not_found_exception` on a falsy `internal_path`, so the `KeyError: 'REQUEST_METHOD'` I hit was my synthetic environ, not the service.

**The ratchet**, since it decides whether the tests are worth anything. `tests/web/gnrstoragehandler_test.py`: 65 passed on this head. Restoring only `resources/common/services/storage/relative.py` from 31983f2b and re-running the six new cases, all six fail, and they fail with my own observations verbatim — `test_relative_storage_missing_parent_service_name` resolves to `.../instances/gnrdevelop/site/rel_gone_parent_storage/docs` instead of `None`, and `test_relative_storage_looping_parent_services` and `test_relative_storage_parent_of_itself` raise `RecursionError: maximum recursion depth exceeded`. The tests pin the fixes rather than passing either way.

Two things left, neither an objection:

- The checks are still running as I write this — build 3.11 through 3.14 and Analyze (javascript-typescript) are IN_PROGRESS. Nothing to do, just not green yet.
- @genro is still a requested reviewer. My approval satisfies the CODEOWNERS requirement on `gnrpy/`, `resources/` and `projects/`, so this is mergeable without waiting for him — but the merge is yours, and whether to wait for his review is your call too.

Still not verified, same as my previous review and your reply: no real `aws_s3` or `sftp` parent (no credentials), so those conclusions are from reading the implementations and the MRO; the `th_service` `_onRemote` fix and the new parent list are verified as code paths, not in a browser; and I did not check multidomain resolution of the parent.

PR #1183, opened 2026-08-26
Dns Manager Service

cgabriel:
As discussed with @mbertoldi, I've added an initial DnsManager service, with the first implementation using awsmanager provided by the framework.

The reason to provide a subset of the cloudmanager/awsmanager service in a separate service is to provide:

1/ a common class for dns management among different providers
2/ avoid using cloudmanager/awsmanager with a IAM account with a limited policy. For example, if only dns management is needed by the application (like a deployer), it's best practice to use an IAM account with access limited to Route53 services. So having a dedicated service avoid to have a service which most methods won't work, and it's discovered at runtime.

More provider to come, initial implementation is related to AWS's Route53.

The service provide a minimal configuration in the parameters section to test and verify its functionality.


fporcari (commented):
The least-privilege argument is the right reason to split this out, and the code honours it rather than just asserting it: `Service.__init__` takes only the two Route53 keys and builds an `AWSManager` that is never handed an IAM, EC2 or S3 surface, so a deployer can hold a Route53-only policy. With `cloudmanager/awsmanager` it would need a key whose other methods fail at runtime, discovered one call at a time. Dropping `region_name` is also correct and not an omission — Route53 is partition-global and resolves to `aws-global` with no region set.

The registration follows the convention exactly: `lib/services/dnsmanager.py` plus `resources/services/dnsmanager/` is what `ServiceHandler.__init__` scans, `_buildImplementations` picks up `route53.py` by its `Service` class, and `getAvailableServiceTree` makes `dnsmanager - route53` appear in the New service picker with no extra wiring. And every method delegates to `gnr.utils.awsmanager.Route53Manager`, which already carries its own tests — no logic is reimplemented here.

One thing I would like fixed in this PR, then two questions.

**`r53_check_record_exists` answers the wrong thing when the zone is not managed** — `route53.py:76-79`, and `r53_verify_record` at `:87-89` has the same shape. `Route53Manager.record_exists` returns `None` in two different situations: the record is genuinely absent, and `find_managed_zone` found no zone matching that name (`gnr/utils/awsmanager.py:490-495`, pinned by the existing `test_record_exists_returns_none_when_no_managed_zone`). The RPC collapses both into "record does not exist". That is the distinction 70661305b deliberately kept in `ensure_cname_record`, which raises `ValueError('No managed Route53 zone found for %s')` rather than reporting absence. In a panel whose whole job is to tell an operator whether their configuration works, "record does not exist" for a domain we do not manage sends them looking in the wrong place. `find_managed_zone` is already there and unexposed; surfacing it fixes both methods in about three lines.

**Does the parameters panel actually render?** `route53.py:33` opens a `borderContainer` on the pane it is given, but `th_service.py:36-41` hands it a `center.contentPane().remote(...)` — a plain ContentPane nested in a ContentPane, which is not a layout widget and will not size its children. Every other `service_parameters` in the repo builds a flat `formbuilder`/`formlet` on the given pane (`sourcerer.py:37`, `mailproxy.py:481`, `awsmanager.py:115`, `aws_s3.py:445` and ten more), and every bare `borderContainer` I found under `packages/sys/resources` hangs off a sized region — `form.center` or `frame.center` — never a nested contentPane. If it collapses to zero height the PR's headline feature is invisible. I have not opened it in a browser, so this is inference from the surrounding idiom, not observed behaviour — a screenshot would settle it.

**Do `cloudmanager`'s four `route53_*` methods now get deprecated?** `cloudmanager/awsmanager.py:59-70` exposes `get_hosted_zones`, `route53_record_exists`, `route53_verify_record` and `route53_ensure_cname_record` — the same four operations against the same `Route53Manager`, added yesterday in 70661305b. Nothing is duplicated in the sense of reimplemented logic; both sides are one-line passthroughs. What is duplicated is the wiring, so a future `Route53Manager` method has to be added in two places. Given the IAM rationale that seems a fair trade, but the two sets being one day apart is worth a decision rather than an accident.

Smaller things, none of which need to hold the PR up:

- The test buttons call `getService('dnsmanager', service_name)` (`route53.py:67,75,84`), which instantiates from the saved `sys.service` record — so they test the last saved credentials and silently ignore keys just typed into the form. `sourcerer.py:50-52` saves first for exactly this reason.
- No `try`/`except` anywhere in either file. A revoked key surfaces as a raw botocore `ClientError` in the generic server-error dialog instead of the `genro.dlg.alert` the buttons already set up; `datarpc_adderrback` sits right next to the `addCallback` in use.
- No tests. The service class is pure delegation and the delegated logic is covered, so I would not ask for much — but the three `@public_method` RPCs hold the only new logic in the PR, and that is exactly where the finding above lives.
- `DnsManager` (`dnsmanager.py:4-6`) declares no methods, so the "common class across providers" rationale has nothing behind it yet. It matches `CloudManager`'s existing shape, so this is a note rather than an objection: if more providers are coming, naming the four methods here is what would make the abstraction real.
- The new user-facing strings carry no `!!` translation markers, unlike `sourcerer.py:38-52` and `mailproxy.py:483+`. Its direct sibling `cloudmanager/awsmanager.py:117-119` omits them too, so this is not a regression against the file it was copied from.

Not verified: I made no live AWS call, so the Route53 behaviour is read from `Route53Manager` and its tests rather than exercised; and I did not render the parameters panel, so the layout question above is static inference. flake8 with the repo's own selection passes clean on both files.

cgabriel:
> One thing I would like fixed in this PR, then two questions.
> 
> **`r53_check_record_exists` answers the wrong thing when the zone is not managed** — `route53.py:76-79`, and `r53_verify_record` at `:87-89` has the same shape. `Route53Manager.record_exists` returns `None` in two different situations: the record is genuinely absent, and `find_managed_zone` found no zone matching that name (`gnr/utils/awsmanager.py:490-495`, pinned by the existing `test_record_exists_returns_none_when_no_managed_zone`). The RPC collapses both into "record does not exist". That is the distinction [7066130](https://github.com/genropy/genropy/commit/70661305b313bf32b8dc789d92c205a2051d005b) deliberately kept in `ensure_cname_record`, which raises `ValueError('No managed Route53 zone found for %s')` rather than reporting absence. In a panel whose whole job is to tell an operator whether their configuration works, "record does not exist" for a domain we do not manage sends them looking in the wrong place. `find_managed_zone` is already there and unexposed; surfacing it fixes both methods in about three lines.

I'll rearrange the API accordingly, it's a good point. The service was born with only a requirement (ensure_cname_record), so the rest of the API is a WIP. That's also why the base DnsManager doesn't define the interface yet.


> **Does the parameters panel actually render?** `route53.py:33` opens a `borderContainer` on the pane it is given, but `th_service.py:36-41` hands it a `center.contentPane().remote(...)` — a plain ContentPane nested in a ContentPane, which is not a layout widget and will not size its children. Every other `service_parameters` in the repo builds a flat `formbuilder`/`formlet` on the given pane (`sourcerer.py:37`, `mailproxy.py:481`, `awsmanager.py:115`, `aws_s3.py:445` and ten more), and every bare `borderContainer` I found under `packages/sys/resources` hangs off a sized region — `form.center` or `frame.center` — never a nested contentPane. If it collapses to zero height the PR's headline feature is invisible. I have not opened it in a browser, so this is inference from the surrounding idiom, not observed behaviour — a screenshot would settle it.

It's working fine:

image



> **Do `cloudmanager`'s four `route53_*` methods now get deprecated?** `cloudmanager/awsmanager.py:59-70` exposes `get_hosted_zones`, `route53_record_exists`, `route53_verify_record` and `route53_ensure_cname_record` — the same four operations against the same `Route53Manager`, added yesterday in [7066130](https://github.com/genropy/genropy/commit/70661305b313bf32b8dc789d92c205a2051d005b). Nothing is duplicated in the sense of reimplemented logic; both sides are one-line passthroughs. What is duplicated is the wiring, so a future `Route53Manager` method has to be added in two places. Given the IAM rationale that seems a fair trade, but the two sets being one day apart is worth a decision rather than an accident.

cloudmanager/awsmanager service will keep its own low level method, if they're needed. The Dns manager service is a higher level interface.

> Smaller things, none of which need to hold the PR up:
> 
> * The test buttons call `getService('dnsmanager', service_name)` (`route53.py:67,75,84`), which instantiates from the saved `sys.service` record — so they test the last saved credentials and silently ignore keys just typed into the form. `sourcerer.py:50-52` saves first for exactly this reason.

Looking at sourcerer code, it seems to me that it save the token received in the service configuration, since it's a needed entry required for the service to be operational, it gather the data in the onResult which obtain the response from the callback url. It's a completely different motivation. Anyway, the parameters form is just a simple test bed for the service, it's not related to the configuration itself. I compile the form, SAVE, and check if it's working or not.

> * No `try`/`except` anywhere in either file. A revoked key surfaces as a raw botocore `ClientError` in the generic server-error dialog instead of the `genro.dlg.alert` the buttons already set up; `datarpc_adderrback` sits right next to the `addCallback` in use.

Still regarding the parameters forms?

> * No tests. The service class is pure delegation and the delegated logic is covered, so I would not ask for much — but the three `@public_method` RPCs hold the only new logic in the PR, and that is exactly where the finding above lives.


> * `DnsManager` (`dnsmanager.py:4-6`) declares no methods, so the "common class across providers" rationale has nothing behind it yet. It matches `CloudManager`'s existing shape, so this is a note rather than an objection: if more providers are coming, naming the four methods here is what would make the abstraction real.

As stated above, the API is still a WIP, that is why there is not structured defined yet. When more provider are to come, we'll be able to draft a common ground.

> * The new user-facing strings carry no `!!` translation markers, unlike `sourcerer.py:38-52` and `mailproxy.py:483+`. Its direct sibling `cloudmanager/awsmanager.py:117-119` omits them too, so this is not a regression against the file it was copied from.

ok

> Not verified: I made no live AWS call, so the Route53 behaviour is read from `Route53Manager` and its tests rather than exercised; and I did not render the parameters panel, so the layout question above is static inference. flake8 with the repo's own selection passes clean on both files.

They both work fine.

cgabriel:
In addition to the comments above, I've changed the service API to provide better error information to high level consumer, without the necessity to import custom Exceptions.

fporcari (changes requested):
Solid rationale for splitting DNS out of `cloudmanager`: a Route53-only IAM account is exactly the case where a fat service fails half its methods at runtime, and a dedicated service type makes that visible at configuration time instead. The `lib/services/dnsmanager.py` + `resources/services/dnsmanager/route53.py` split follows the `cloudmanager`/`awsmanager` pair line for line, and the `_ask` dialogs in `ServiceParameters` make the service testable from the service form, which is the right place for it.

Two things about the `gnr/utils/awsmanager.py` half, which is the part with reach beyond this PR.

**1. The sibling `cloudmanager` wrappers now leak the new dict through a boolean-shaped API.**
`record_exists` and `verify_record` change from `record|None` and `bool` to a `dict(value=..., status=..., description=...)`. The new `dnsmanager` callers read `result['status']` correctly, but `projects/gnrcore/packages/sys/resources/services/cloudmanager/awsmanager.py:62-67` forwards both verbatim:

```python
def route53_verify_record(self, name, record_type, value, hosted_zone_id=None):
    return self.aws_manager.Route53.verify_record(...)
```

Every failure path now returns `dict(value=False, status='error', ...)` — a truthy object. Any `if service.route53_verify_record(...)` that used to take the false branch on a type or value mismatch now takes the true branch, silently, and a caller verifying a DNS record before writing it would believe every record verifies. Same shape for `route53_record_exists`, which used to return `None` when the record was absent.

Either update the two wrappers to unwrap (`return result['value']` / `result['status'] == 'ok'`), or leave `verify_record` returning a bool and add the detailed variant under its own name.

**2. `verify_record`'s missing-record path returns `value=None`, its three siblings return `value=False`.**
`gnrpy/gnr/utils/awsmanager.py:524-525` returns the `record_exists` error dict unchanged, so `value` is `None` there while the type-mismatch and value-mismatch paths both give `False`. `test_verify_record_missing` only asserts `status == 'error'`, so the inconsistency is not pinned either way. One line:

```python
if result['status'] == 'error':
    return dict(result, value=False)
```

Two smaller notes, neither blocking: the PR closes no issue (`Fixes #<n>` in the body would make it traceable), and it has no assignee — on our workflow the assignee of a PR is its author.

Not verified: no live AWS run, and no visual pass on the `ServiceParameters` pane — the `borderContainer` there is a different shape from `cloudmanager`'s flat `formbuilder` and I did not check it renders with a height in the service form. CI is green on 3.11–3.14. A cross-repository search for `route53_verify_record` and `route53_record_exists` found no consumer outside this repo, so the blast radius above is latent rather than already breaking a known caller.

mbertoldi (dismissed):

cgabriel:
addressed change requests in de92ba0a15a8c932b4640acbf0808f602c93a62e 

    record_exists/verify_record on Route53Manager now return a
    dict(value=, status=, description=) so the dnsmanager service form can
    show why a check failed. The cloudmanager wrappers were forwarding that
    dict verbatim, turning the old record|None / bool contract into an
    always-truthy object: `if service.route53_verify_record(...)` would take
    the true branch on every type or value mismatch, and route53_record_exists
    never returned None for an absent record.
    
    - cloudmanager route53_record_exists unwraps to result['value']
      (record dict when found, None when absent or the zone is unmanaged).
    - cloudmanager route53_verify_record unwraps to result['status'] == 'ok'
      (True only on a real match). The dnsmanager service keeps the full dict.
    - verify_record's missing-record path now returns value=False like its
      type-mismatch and value-mismatch siblings, instead of leaking value=None
      from record_exists; test_verify_record_missing pins it.
    - dnsmanager ServiceParameters uses a flat formbuilder like every other
      service_parameters pane, dropping the borderContainer that had no
      guaranteed height in the service form.

fporcari (approved):
Approved on de92ba0.

Both requests are addressed and verified against the head: `route53_record_exists` unwraps to `result['value']` (the record dict, or `None` when absent or the zone is unmanaged) and `route53_verify_record` to `result['status'] == 'ok'`, so the two cloudmanager wrappers keep their `record|None` / `bool` contract; `verify_record` now returns `value=False` on all three error paths, and `test_verify_record_missing` pins it. The dnsmanager RPCs keep the full dict and surface `description`, so an operator sees "Zone not managed" rather than "Record not found" for a domain outside the managed zones, which was the first-round question. The parameters pane is a flat formbuilder like every other `service_parameters`. `tests/utils/awsmanager_test.py` 18 passed, flake8 clean on the five files, merge with develop clean, CI green.

Two nits, neither blocking: `lib/services/dnsmanager.py` ends with a whitespace-only line (line 8), and the PR closes no issue, so the feature has no ticket to trace back to.

Not verified: no live Route53 call, no browser pass on the pane.

genro (approved):
The service resolves correctly through the real lookup — I checked the three things registration needs rather than assuming the directory layout was enough: the type scan over `<pkg>/lib/services/*.py`, the matching `resources/services/dnsmanager/` folder, and `getattr(module,'Service')`. The two declared parameters survive `filterServiceConf` and match `Service.__init__`. With no credentials, registration still succeeds and the failure arrives on the first call, since `AWSManager` builds its boto3 client lazily — which is the behaviour I would want.

The parameters pane with the test methods is the right touch and follows what sourcerer and supervisord already do.

One thing I want to flag before this goes in, because it is outside the new service. `Route53Manager.record_exists` and `verify_record` change their return type from `record|None` and `bool` to a dict. Both in-repo callers are updated, but that is a framework-public utility, and the dict is always truthy: an application doing `if mgr.verify_record(...)` now takes the true branch even when the value does not match, silently. There is no shim and no note. Either a release-note line or a differently-named method would close it — your call, but somebody outside this repo will hit it otherwise.

Two smaller ones. A record that is simply absent now reports `status='error'` with `'Record not found'`, which at the status level is indistinguishable from `'Zone not managed'` — a normal outcome and a configuration fault reading the same. And `Service.ensure_cname_record` is the only write operation exposed, with no caller anywhere in the repo, no test, and no way to reach it from the pane.

`DnsManager` itself declares no method, so a second provider has nothing to implement against. Same shape as the existing `cloudmanager`, so I am noting it rather than asking you to change it here.

PR #1187, opened 2026-08-27
initial security policy

cgabriel:
Initial proposal for security policy. This is an hotfix to have the document land in a release ASAP.


genro (changes requested):
The document itself is good: the two reporting channels are the right ones and in the right order, the "what to expect" section commits to concrete windows instead of the usual vagueness, and the Scope section is the part most policies leave out — saying plainly that dependencies and downstream applications are not covered is what stops a reporter from filing an upstream CVE against us. Worth having.

I also checked the branch hygiene, since this is on `master`: correctly cut from `master` (the merge-base against `master` and against `develop` coincide at `a2f7878c8`) and it carries exactly one commit, `923fc55e5`. Nothing unreleased is dragged into `master`.

## Blocking: on `master` alone the policy is invisible

`SECURITY.md` is one of the files GitHub reads **from the default branch only** — root, `.github/` or `docs/`. This repository's default branch is `develop`, and this PR targets `master`.

So after merging, none of what the file is for actually happens: no Security policy entry, no "Report a vulnerability" link surfaced from the policy, no banner pointing a reporter here when they open an issue. I checked there is no fallback either — no `SECURITY.md` on `develop` or `master` today, and there is no organisation-level `genropy/.github` repository to inherit one from.

It would eventually work, once `master` is realigned into `develop` after the next release, but that makes the visibility of the security policy depend on a realignment nobody is tracking for this purpose.

The other half of it: a new document is not a hotfix. There is nothing broken on `master` to fix, so the release-branch route buys nothing and costs a `master`-only commit that has to be merged back. Retargeting this to `develop` puts the file where GitHub reads it, and it reaches `master` on its own at the next release, like everything else.

## Confirmed: the preferred channel exists

Since the file names private reporting as preferred, I verified the button is really there rather than assuming it: private vulnerability reporting is enabled on the repository (`/private-vulnerability-reporting` → `{"enabled": true}`). Good — a policy whose preferred channel is a disabled feature is the classic way these documents go stale on day one.

I could not verify that `security@softwell.it` is a live, monitored mailbox. Worth confirming before this lands, since it is the fallback channel for anyone who cannot use the Security tab.

## The release format does not match what we publish

```
Released versions are using the format 'YY.MM.DD.xxx'
```

Published releases are `release-26.08.19`, `release-26.08.19.1`, `release-26.07.31`: the `release-` prefix is missing from the description, and `.xxx` reads as a mandatory three-digit serial when it is actually an optional single increment, present only on same-day re-releases. Suggested:

```
Released versions use the format 'release-YY.MM.DD', with an optional
trailing increment for same-day re-releases (e.g. release-26.08.19.1).
```

The claim that release tags are based on `master` is correct — checked `release-26.08.19.1` and `release-26.07.31`, both reachable from `master`.

## Supported Versions answers a different question than a reporter asks

The table lists branches, while Scope excludes "Issues in unsupported versions (see table above)". A reporter holding `release-26.02.14` has no way to read their answer off that table: the six-month window is stated in prose above it and never becomes a rule. Either fold the window into the table, or make Scope point at the six-month sentence rather than at a table that only lists branch names.

## Typos

- `develpment` → `development`
- `Released packages tag are based` → `Released package tags are based`
- trailing whitespace after `security@softwell.it`

## Assignee

The PR has no assignee. For a hotfix on `master` that should be the maintainer, who reviews, merges, cuts the release and realigns `develop` afterwards — worth setting so someone formally answers for it, rather than leaving it to whoever notices.

cgabriel:
> The document itself is good: the two reporting channels are the right ones and in the right order, the "what to expect" section commits to concrete windows instead of the usual vagueness, and the Scope section is the part most policies leave out — saying plainly that dependencies and downstream applications are not covered is what stops a reporter from filing an upstream CVE against us. Worth having.
>

Thank you, that is what was needed.
 
> ## Blocking: on `master` alone the policy is invisible
> `SECURITY.md` is one of the files GitHub reads **from the default branch only** — root, `.github/` or `docs/`. This repository's default branch is `develop`, and this PR targets `master`.
> 
> So after merging, none of what the file is for actually happens: no Security policy entry, no "Report a vulnerability" link surfaced from the policy, no banner pointing a reporter here when they open an issue. I checked there is no fallback either — no `SECURITY.md` on `develop` or `master` today, and there is no organisation-level `genropy/.github` repository to inherit one from.
> 
> It would eventually work, once `master` is realigned into `develop` after the next release, but that makes the visibility of the security policy depend on a realignment nobody is tracking for this purpose.
> 
> The other half of it: a new document is not a hotfix. There is nothing broken on `master` to fix, so the release-branch route buys nothing and costs a `master`-only commit that has to be merged back. Retargeting this to `develop` puts the file where GitHub reads it, and it reaches `master` on its own at the next release, like everything else.

This is an hotfix because we need to release the file ASAP, and as per policy, all hotfix are immediately merged back to master. I'm pretty aware that the file is not visible to Github until it lands on the default branch, but I'm not worried this due to the policy itself.
 
> I could not verify that `security@softwell.it` is a live, monitored mailbox. Worth confirming before this lands, since it is the fallback channel for anyone who cannot use the Security tab.

Live and monitored.
 
> ## The release format does not match what we publish
> ```
> Released versions are using the format 'YY.MM.DD.xxx'
> ```
> 
> Published releases are `release-26.08.19`, `release-26.08.19.1`, `release-26.07.31`: the `release-` prefix is missing from the description, and `.xxx` reads as a mandatory three-digit serial when it is actually an optional single increment, present only on same-day re-releases. Suggested:
> 
> ```
> Released versions use the format 'release-YY.MM.DD', with an optional
> trailing increment for same-day re-releases (e.g. release-26.08.19.1).
> ```
> 
> The claim that release tags are based on `master` is correct — checked `release-26.08.19.1` and `release-26.07.31`, both reachable from `master`.

The document states "Release *VERSIONS* format", and not "relese TAGS". The version format is YY.MM.DD.xxx, the tag is release-YY.MM.DD.xxx. Since the document is mostly relevant for installation via packages, the VERSION is the relevant part.

> ## Supported Versions answers a different question than a reporter asks
> The table lists branches, while Scope excludes "Issues in unsupported versions (see table above)". A reporter holding `release-26.02.14` has no way to read their answer off that table: the six-month window is stated in prose above it and never becomes a rule. Either fold the window into the table, or make Scope point at the six-month sentence rather than at a table that only lists branch names.

Since it's written that releases are based on master, and since the release holds the release date, what's complicated about finding if I have a release older than 6 months?

> ## Typos
> * `develpment` → `development`
> * `Released packages tag are based` → `Released package tags are based`
> * trailing whitespace after `security@softwell.it`

Thank you, fixed in f43b9d3

> ## Assignee
> The PR has no assignee. For a hotfix on `master` that should be the maintainer, who reviews, merges, cuts the release and realigns `develop` afterwards — worth setting so someone formally answers for it, rather than leaving it to whoever notices.

The PR was assigned well before this review, looks like an information gathering lag. And yes, develop get realigned afterwards, that's why the first comment about file visibility is irrelevant.

fporcari (approved):
Content-wise this is good to have, and the reply round addressed the review point by point. Approving on content.

On the one standing dispute (master vs develop): the hotfix route ships the file in a release ASAP, and develop is realigned right after, so the GitHub-side visibility follows within the same cycle. I can live with either sequencing - that call sits between author and owner, and the standing CHANGES_REQUESTED is genro's to lift, not mine.

Two small non-blocking nits on the same line:
- the installed package reports version `26.4.22`, so the trailing `.xxx` is optional rather than part of every version - "format 'YY.MM.DD', with an optional trailing increment" would match what we actually publish;
- "Released packages tags are based" -> "Released package tags are based".

mbertoldi (approved):

genro (approved):
Withdrawing the block. My objection assumed the realignment of `develop` after a hotfix was something nobody was tracking, and you have answered that it is the policy — hotfixes are merged back immediately. That was the whole load-bearing part of the argument: if the file reaches `develop` as a matter of process rather than of somebody remembering, then GitHub picks the policy up right away and nothing I was worried about happens. You cut the releases and you know the policy; I was reasoning from the outside about a process I could not see.

The other three answers land as well. `security@softwell.it` being live and monitored was the one thing I could not check myself. On the Supported Versions table you are right that a reporter holding a dated release can work out whether it is inside six months — I was asking the document to do arithmetic it does not need to do. And the assignee was there before my review; my data was stale, which is on me.

**On the release format you are right about the distinction, and there is a smaller correction underneath it.** VERSION is not TAG, and for somebody who installed a package the version is what they can read — that is the right framing and I had missed it.

But the version a reporter actually sees is not `YY.MM.DD`. `gnr.VERSION` is `"26.08.26"`, and PEP 440 normalisation strips the leading zero of each component, so `pip show genropy` reports **`26.8.26`**. I checked it on my own installation: the source said `26.01.15`, pip says `26.1.15`. Same for a same-day re-release — `26.08.19.1` becomes `26.8.19.1`.

So a reporter comparing what `pip show` prints against a document that says `YY.MM.DD.xxx` is looking at two different strings. One clause fixes it:

```
Released versions use the format 'YY.MM.DD', with an optional trailing
increment for same-day re-releases. Package metadata normalises the leading
zeros, so version 26.08.26 is reported by pip as 26.8.26.
```

Not blocking — it is a documentation refinement on a document that is already worth having, and the substance of your point stands.

Everything I said in favour of the file the first time still holds: two reporting channels in the right order, concrete response windows instead of the usual vagueness, and a Scope section that says plainly what is not covered. Private vulnerability reporting is enabled on the repository, so the preferred channel is real.

PR #1189, opened 2026-08-27
fix(email): isolate each received email in its own transaction

### LINKED ISSUE #1188: IMAP reception: one unreadable email blocks the whole mailbox (deferred FK on email_attachment)
## Summary

A single unreadable incoming email permanently blocks IMAP reception for the whole
account: every subsequent run fails at commit with

```
insert or update on table "email_attachment" violates foreign key constraint "fk_email_attachment_message_id"
DETAIL:  Key (message_id)=(...) is not present in table "email_message".
```

The key in the DETAIL line is a fresh uuid on every run, so this is not stale data in
the table: the orphan is created and re-created by each reception cycle.

## Root cause: children are written before the parent

`email.message.newReceivedMessage()`
(`projects/gnrcore/packages/email/model/message.py:192`) builds the message **in memory
only** — `newrecord(assignId=True)` mints the uuid without inserting anything — but
`parseAttachment()` (same file, line 256) performs a **real** insert into
`email.attachment` using that `message_id`.

So every reception writes the children before the parent exists. It only works because
the FK is declared `deferred=True` in `model/attachment.py`, i.e.
`DEFERRABLE INITIALLY DEFERRED`, which postpones the check to COMMIT.

The parent is inserted later by the caller, `ImapReceiver.receive()`
(`projects/gnrcore/packages/email/lib/imap.py:65`):

```python
try:
    msgrec = self.createMessageRecord(email_id, mailbox_id)   # attachments already inserted here
except Exception as e:
    msgrec = None
    print('Error in email', str(e))
if not msgrec:
    continue                                                  # message will never be inserted
self.messages_table.insert(msgrec)
...
self.db.commit()                                              # deferred FK blows up here
```

If any email raises **after** at least one attachment has been inserted — a second
attachment with `filename` None, a corrupt base64 payload, a storage write that fails —
the orphan stays in the transaction. The `except` swallows the error, the loop moves on,
and the bill is paid by the final `commit()`.

## Why it never recovers on its own

1. The commit fails, so the **entire cycle** is lost, including the healthy messages
   already inserted.
2. `self.account_table.update(dict(id=..., last_uid=items[-1]))` runs **before** the
   commit, so `last_uid` never advances.
3. The next run asks IMAP for the same UIDs, rebuilds the same orphan, fails the same
   way. Forever.

## Aggravating factor in the batch

`projects/gnrcore/packages/email/resources/tables/account/action/receive_mail.py:22`:

```python
except Exception as e:
    raise
    print('ERROR IN RECEIVING IMAP receive_email batch', str(e))
```

The `raise` makes the `except` decorative and the `print` unreachable: the first failing
account also stops every other scheduled account.

On top of that, the `print` in `imap.py` is invisible on instances that do not write logs
to file, so there is currently no way to tell **which** email is poisoning the cycle.

## Proposed fix

Inverting the order (message first, attachments after) is **not** an option: receivers
read the attachments from `trigger_onInserted`. On a real project this drives PEC content
extraction, voicemail recognition and the attachments copied onto the generated event, so
the current order is a contract, not an accident. It is now stated in a comment instead of
being silently load-bearing.

What changes instead is the transaction boundary:

1. `lib/imap.py` — one transaction per message: `commit()` after each successful email,
   `rollbackAll()` on failure. The orphan dies with its own transaction and never reaches
   the commit of the healthy ones.
2. `lib/imap.py` — `last_uid` advances per message, and past a failing one too, so a
   single unreadable email is skipped instead of stalling the mailbox forever.
3. `lib/imap.py` — `logger.exception` with account and uid replaces the `print`, so the
   skipped email stays recoverable on instances that do not write logs to file. In debug
   the exception is re-raised after the rollback, as before.
4. `model/message.py` — `parseAttachment` tolerates an attachment with no filename
   (`os.path.splitext(None)` raised `TypeError`, a likely trigger: inline signature
   images arriving after a real attachment already inserted).
5. `resources/tables/account/action/receive_mail.py` — drop the dead `raise`, log the
   failing account, keep going with the others.

There are no savepoints in `gnr.sql`, so per-message isolation has to be done with
incremental commits; there is no finer-grained alternative.

fporcari:
Fixes #1188

## The failure

A single unreadable incoming email permanently blocked IMAP reception for the whole
account, with every run failing at commit:

```
insert or update on table "email_attachment" violates foreign key constraint "fk_email_attachment_message_id"
DETAIL:  Key (message_id)=(...) is not present in table "email_message".
```

`newReceivedMessage()` builds the message **in memory only** (`newrecord(assignId=True)`
mints the uuid without inserting) but `parseAttachment()` performs a **real** insert into
`email.attachment` using that `message_id`. Children are written before the parent, and
only `DEFERRABLE INITIALLY DEFERRED` on the FK holds it together until commit.

In `ImapReceiver.receive()` the `except` swallowed any parse failure and `continue`d, so
an email failing **after** one attachment had already been inserted left an orphan in the
transaction. The final `commit()` then took down the entire cycle, healthy messages
included — and since `last_uid` was updated *before* that commit, it never advanced: the
next run fetched the same UIDs and failed identically. Forever.

## The fix

Not by reversing the order. Receivers read the attachments from `trigger_onInserted` —
on iCond that drives PEC content extraction, voicemail recognition and the attachments
copied onto the generated event — so attachments-before-message is a contract, now stated
in a comment instead of being silently load-bearing.

What changes is the transaction boundary:

- **One transaction per message** — `commit()` per successful email, `rollbackAll()` on
  failure. The orphan dies with its own transaction and never reaches anyone else's commit.
- **`last_uid` advances per message**, past a failing one too, so an unreadable email is
  skipped rather than stalling the mailbox forever.
- **`logger.exception` with account and uid** replaces the `print`, which was invisible on
  instances that do not write logs to file — there was no way to tell *which* email was at
  fault. In debug the exception is re-raised after the rollback, as before.
- **The attachment decoding tolerates a lost padding** — this is the trigger actually
  confirmed in production, see below.
- **`parseAttachment` tolerates a missing filename** — `os.path.splitext(None)` raised
  `TypeError`, another way in.
- **`receive_mail` batch**: the `raise` sat *before* the `print`, making the `except`
  decorative and stopping every other scheduled account at the first failure.

## The trigger, confirmed in production

```
SALTATA uid=24695  Mon, 24 Aug 2026 15:53:05 +0200
   da: "Per conto di: gestionecredito@pec.gruppoiren.it" <posta-certificata@legalmail.it>
   oggetto: POSTA CERTIFICATA: Sollecito Pagamento
   errore: binascii.Error: Incorrect padding
```

A Legalmail PEC — exactly the shape that produces the orphan, since a PEC always carries
several attachments (`postacert.eml`, `daticert.xml`, `smime.p7s`): the first is inserted,
the next raises, the message never is.

mail-parser hands over the base64 exactly as it travels in the MIME part
(`get_payload(decode=False)`), so it arrives wrapped at 76 columns and, on this message,
without its trailing `=`. `b64decode` skips the newlines but will not tolerate the
missing padding.

The padding is recomputed after stripping **everything outside the base64 alphabet**, not
just whitespace: `b64decode` ignores those characters anyway, so counting them would
compute the padding against the wrong length. And when a payload cannot be decoded at
all, the attachment is stored raw under a `.b64` name and the message is imported
regardless — losing one attachment is recoverable by hand, losing the email is not.

## Verification

`tests/test_imap_receive.py` drives `receive()` on fakes (no socket, no DB) over three
UIDs, poisoning the middle one. Checked both ways:

`tests/test_imap_receive.py` also covers the decoding: wrapped base64, base64 with the
padding stripped, stray characters between the lines, and a payload that cannot be padded
into anything valid.

- with the fix: `2176 passed, 10 skipped`, flake8 clean
- reverting only `lib/imap.py`: `test_a_failing_message_does_not_stop_the_others` fails —
  messages 1 and 3 are lost with 2, and nothing is rolled back

## Trade-off worth a reviewer's eye

`last_uid` advancing past a failing email means a genuinely unreadable message is skipped
rather than retried. That is deliberate — the current blockage is exactly the cost of the
opposite choice — and the `logger.exception` carries account and uid so it stays
recoverable by hand. Say the word if you would rather have it retried and only the
blocking behaviour removed.


cgabriel (dismissed):
Nice catch, sync issues are the worst. Thank you!

fporcari:
@cgabriel sorry, your approval got dismissed by a second commit I pushed after it — the
production instance reported the actual failing email while the PR was already open, and
it turned out the trigger was not what I had guessed.

Nothing changed in `b11d1ee98`, the one you already approved. The new commit is
`90b34f522` alone:

- `binascii.Error: Incorrect padding` on a Legalmail PEC was the real trigger, not the
  missing filename I had assumed. mail-parser hands over the base64 exactly as it travels
  in the MIME part, so it arrives wrapped at 76 columns and, on this message, without its
  trailing `=`.
- `decodeAttachmentPayload` recomputes the padding after stripping everything outside the
  base64 alphabet — `b64decode` ignores those characters anyway, so counting them computes
  the padding against the wrong length. A test I wrote failed on exactly this before I
  narrowed the regex, so it is covered.
- When a payload cannot be decoded at all, the attachment is stored raw under a `.b64`
  name and the message is imported regardless: losing one attachment is recoverable by
  hand, losing the email is not.

Four tests added for the decoding, `2176 passed, 10 skipped`, flake8 clean.

One thing worth your judgement: I kept this in the same PR because it is one incident —
the padding is the cause, the deferred-FK orphan is what turns it into a mailbox-wide
outage. Happy to split it into its own PR if you would rather review them separately.

cgabriel (approved):
re-approved, no significant changes.

genro (approved):
Approving. The diagnosis holds at every step and the fix is placed where the problem actually is.

`newReceivedMessage` does insert the attachments before the message exists, and only the deferred FK holds them together until commit. The old `except` swallowed and continued, so an email failing *after* one attachment had landed left an orphan inside the transaction, and the final `commit()` took down the whole cycle. Because `last_uid` was written before that commit it never advanced, so the next run fetched the same UIDs and failed identically — "forever" in the description is not rhetoric, it is what the code does.

Not reversing the order is the right call, and the reason is worth having in a comment rather than in somebody's memory: receivers read the attachments from `trigger_onInserted`, so attachments-before-message is a contract, not an accident. Moving the boundary instead of the order leaves that contract intact and kills the orphan inside its own transaction.

`logger.exception` with account and uid replacing the `print` is the part that turns this from a fix into something operable — the original failure was invisible on any instance not writing logs to file, and there was no way to tell which email was at fault.

I also checked the tail: `items[-1]` after the loop cannot raise, since `if not items: return` guards it upstream.

**One thing I would change, and it is three lines.**

```python
msgrec = self.createMessageRecord(email_id, mailbox_id)
if msgrec:
    self.messages_table.insert(msgrec)
```

If `msgrec` were ever falsy, the attachments are already inserted and the message is not — which is precisely the state this PR exists to prevent. The `commit()` two lines below would then violate the FK, the `except` would catch it, and the log would carry a foreign-key error instead of the real cause.

It cannot happen today: `createMessageRecord` returns `newReceivedMessage`, which always returns `new_mail` or raises. So the guard defends against an unreachable case and, in doing so, hides the one way the invariant can break. Either drop it, or make it explicit:

```python
msgrec = self.createMessageRecord(email_id, mailbox_id)
if not msgrec:
    raise GnrImapException('no message record built for uid %s' % email_id)
self.messages_table.insert(msgrec)
```

The `raise` costs nothing — the `except` right below already turns it into a skipped email with a log line naming the uid, which is the correct outcome — and it stops a silent orphan from ever reaching a commit.

**One note, not a request.** In debug the `raise` fires after the `rollbackAll` and before `last_uid` advances, so a debug instance still stalls on the unreadable email exactly as production did before this PR. That is stated in the description and is defensible — a developer wants to see the failure — but it is worth being deliberate about, because debug is now the one configuration where the original bug survives intact.

The base64 padding tolerance is right as written: stripping everything outside the alphabet before recomputing the padding is necessary, since `b64decode` discards those characters anyway and counting them would compute the padding against the wrong length. Good that the docstring says so.

PR #1191, opened 2026-08-27
feat(email): show the composed body in the message ui

### LINKED ISSUE #1190: [TASK] Show the rendered message body in the email UI (compiled_body)
## Summary

Show the rendered message body (`compiled_body`) in the email UI, so a message whose stored body is a placeholder is never displayed verbatim.

## Objective

`email.message.getBody()` lets a project store a lightweight body and compose the real content only when the message is sent (`model/message.py:395`, used by `sendMessage()` at `message.py:369` and, since #1122, by the mail proxy at `webpages/mailproxy/mp_endpoint.py:532`). The `compiled_body` pyColumn already exposes that composition to the UI (`tbl.pyColumn('compiled_body', py_method='getBody')`, `model/message.py:89`).

No UI in the framework requests it. Every place that displays a message body reads `$body` directly, so on an instance that customizes `getBody` the user sees the placeholder — `$testo_fisso` on ANACI circolari today, `$campagna_body` on DEM campaigns as soon as softwellsrl/gnrcommunication#41 lands.

Projects work around this one by one: ANACI renders the body through a database template (`msg_preview`, a `templateChunk` whose `virtual_columns` attribute carries `$compiled_body`). That workaround is not portable — the template is a record of that instance. The framework should ask for the pyColumn directly.

## Scope

In scope — the three places that display a message body:

- `packages/email/resources/tables/message/th_message.py` — `Form.th_form`: the Body tab is `simpleTextArea(value='^.record.body', editor=True)`, i.e. the editor used to write a draft. It stays as it is; a read-only **Preview** tab on `^.record.compiled_body` is added next to it.
- `packages/email/resources/tables/message/th_message.py` — `FormFromDashboard.th_form`: `pane.div('^.body')` is a read-only display, it moves to `^.compiled_body`.
- `packages/email/webpages/mail_dashboard.py` — `div('^.current.record.body')` is a read-only display, it moves to `^.current.record.compiled_body`.

Each of those loads must request the virtual column: `form.store.handler('load', virtual_columns='$compiled_body')` for the forms (same pattern as `resources/common/gnrcomponents/attachmanager/attachmanager.py:550`), `virtual_columns='$compiled_body'` on the `dataRecord` for the dashboard.

Out of scope:

- The `plain_text` and `abstract` formula columns (`model/message.py:66-67`) are SQL expressions over `$body`, so grids and list previews keep showing the placeholder. Changing that means storing text per message and is a separate decision.
- The default `getBody` is unchanged: it returns `message['body']`, so on an instance that does not customize it the Preview tab shows exactly what the Body tab shows.

## Technical Details / Notes

`getBody` is reached by two different callers with two different record shapes: a Bag when a single record is loaded (`gnrpy/gnr/sql/gnrsqldata/record.py:541`) and a plain dict when it is a query row or a proxy payload (`gnrpy/gnr/sql/gnrsqldata/query.py:279`). A dict that did not select the column a `getBody` override needs raises `KeyError`, while a Bag returns `None`. Overrides must therefore read defensively — worth stating in the hook's docstring, since the framework now invites projects to use it from the UI too.

Rendering a body moves work to display time, so a failing `getBody` would break a message form or the dashboard. The hook should never raise.

## Dependencies

- #1122 — the mail proxy must honour `getBody` (already merged, #1129)
- softwellsrl/gnrcommunication#41 — the DEM change that needs this UI fix to ship

## Acceptance Criteria

- [ ] The message form has a read-only Preview tab showing the composed body; the Body tab still edits `$body` on a draft
- [ ] `FormFromDashboard` and the mail dashboard show the composed body
- [ ] On an instance with no `getBody` customization, Preview and Body show the same content
- [ ] On ANACI, a circolare message shows the `testo_fisso` of its circolare instead of `$testo_fisso`

dgpaci:
Fixes #1190

## What

`email.message.getBody()` lets a project keep a lightweight placeholder in `$body` and compose the real content only when the message is sent — that is how ANACI circolari stopped storing the same HTML once per recipient, and it is what softwellsrl/gnrcommunication#41 is about to do for DEM campaigns. The `compiled_body` pyColumn already exposed that composition to the UI, but no UI in the framework asked for it, so every place that displays a message body showed the placeholder verbatim (`$testo_fisso`, and soon `$campagna_body`).

Projects were working around it one at a time: ANACI renders the body through `msg_preview`, a database template whose `virtual_columns` attribute carries `$compiled_body`. That workaround cannot be moved into the framework, because the template is a record of that instance.

## How

- **Message form** — the Body tab is the editor used to write a draft (`simpleTextArea(editor=True)`), so it keeps editing `$body` untouched. A read-only **Preview** tab is added beside it on `^.record.compiled_body`.
- **`FormFromDashboard`** and **`mail_dashboard`** are read-only displays, so they now read `compiled_body` directly.
- Each load asks for the column explicitly (`virtual_columns='$compiled_body'` on the form load handler and on the dashboard's `dataRecord`): a pyColumn is not evaluated unless requested.
- `getBody`'s docstring now states the two contracts an override has to respect — it is called with a Bag on a record load and with a plain dict on query rows and mail-proxy payloads (where an unselected column raises `KeyError` instead of returning `None`), and it must never raise, since it now runs while a message is merely being displayed.

On an instance that does not customize `getBody`, the default returns `message['body']`, so Preview shows exactly what the Body tab shows — no behaviour change.

## Test case

New `projects/gnrcore/packages/email/tests/test_message_compiled_body.py`, which installs a package extension overriding `getBody` (same harness as `test_mailproxy_body.py`) and stores a message whose body is `$shared_content, this is the message.`:

- `test_record_load_resolves_compiled_body` — what the forms do: `record(pkey, virtual_columns='$compiled_body')` returns the composed body while `$body` still holds the placeholder.
- `test_query_row_resolves_compiled_body` — what a grid does: the pyColumn is evaluated on the fetched row too, and the query has to select the columns the override reads.

```
cd projects/gnrcore/packages/email/tests
python -m pytest test_message_compiled_body.py test_mailproxy_body.py
```

3 passed. `flake8 --select=F401,E9,F63,F7,F82 --max-line-length=100` clean on the changed files.

Manual check to run on an instance with a `getBody` customization (ANACI): open a circolare message and confirm the Preview tab shows the `testo_fisso` of its circolare instead of `$testo_fisso`, and that the mail dashboard shows the same.

## Out of scope

`plain_text` and `abstract` (`model/message.py:66-67`) are SQL expressions over `$body`, so grids and list previews keep showing the placeholder. Changing that means storing text per message and is a separate decision, noted in #1190.


fporcari (approved):
The design is right: the Body tab keeps editing $body while Preview, FormFromDashboard and the mail dashboard ask for $compiled_body explicitly - a pyColumn is only evaluated when requested - so instances without a getBody override see no behaviour change. The docstring stating the Bag-vs-dict contract for overrides is worth as much as the feature. The tests install a real package extension overriding getBody and cover both the record-load and the query-row paths.

Not verified here: the manual check the PR body names on an instance with a getBody customization (ANACI circolari Preview and dashboard).

PR #1192, opened 2026-08-27
fix(email): offer the proxy sending states only where a proxy dispatches

dgpaci:
## Problem

The mail proxy merge added two sending status sections to the `email.message` table handler without asking whether a proxy is actually there, and gave one of them a caption that collides with another package's translation.

**Two sections that can never match without the proxy**

- `dispatched_to_proxy` filters on `$proxy_ts IS NOT NULL`. Only `mp_endpoint` writes that column, so on a plain smtp instance the section is always empty.
- `queue_issues` filters on `$queue_mismatch <> 0`. The discrepancy is real without a proxy too, but the Sending Dashboard already reports it as a `Mismatch` column in its no-proxy layout, so the message upperbar was carrying a second, redundant entry point.

The same merge already gates its proxy ui elsewhere — `th_account.py` wraps the `proxy_*` fields and `sending_dashboard.py` picks its whole layout — both through `Package.getMailProxy(raise_if_missing=False)`. The message table handler was the one place left out.

**"Inviato" twice in the Italian upperbar**

The caption was `!!Dispatched`. A localization key is `flatten(caption)` in a single flat dictionary shared by every loaded package (`GnrLocalizer.updateLocalizationDict` does `localizationDict.update(...)`), so `en_dispatched` belongs to no package in particular. Both `gnrsupport/helpdesk` and `gnrxadm/dbwatcher` ship `<en_dispatched base="Dispatched" it="Inviato">`, and on an instance loading either of them whichever slot loaded last won: the upperbar showed **Inviato** for `Sent` and **Inviato** again for `Dispatched`.

## Solution

Both `View.th_sections_sendingstatus` and `ViewOutOnly.th_sections_sendingstatus` build their list against `getMailProxy(raise_if_missing=False)`. With a proxy the sections are unchanged; without one, `View` drops both proxy sections and `ViewOutOnly` drops `queue_issues` (it never had `dispatched_to_proxy`, and this PR does not add it).

The caption becomes `!!Dispatched to proxy`, which names the state it describes and moves the key to `en_dispatched_to_proxy` — verified unused anywhere else. `localization.xml` gets the Italian for it and for `Queue issues`, which had none either.

## Impact and scope

Only the sending status sections of `email.message`. An instance with an activated mailproxy service sees exactly what it saw before, except that Dispatched now reads *Consegnato al proxy* in Italian instead of *Inviato*. The `queue_mismatch` diagnostic is not lost on no-proxy instances: it stays in the Sending Dashboard, where the no-proxy layout has always shown it.

## Test case

`projects/gnrcore/packages/email/tests/test_message_sections.py` — two classes, each on its own real instance (`BaseGnrTest` + `GnrDummySite` + sqlite), asking the real resource for its sections:

- `TestSectionsWithoutMailProxy` — no service configured: `getMailProxy` returns `None`, and neither view offers `dispatched_to_proxy` or `queue_issues` while `to_send` and `sent` stay put.
- `TestSectionsWithMailProxy` — an activated proxy registered through `sys.service.addService(...)` with a `proxy_url` and `tenant_registered`: both sections come back.

```
GENRO_GNRFOLDER=<gnrenv>/etc/gnr python3 -m pytest projects/gnrcore/packages/email/tests -q
7 passed
```

Checked against the pre-fix resource as well: the two no-proxy view tests fail there and pass here, so they are not vacuous. flake8 clean on the touched python files, `localization.xml` parses.


fporcari (approved):
Verified the mechanism, not just the prose: `proxy_ts` is written only by mp_endpoint (model/message.py merely resets it to None on requeue), so on a no-proxy instance `dispatched_to_proxy` can never match; and the gating mirrors what th_account.py:93 and sending_dashboard.py:12 already do with `getMailProxy(raise_if_missing=False)`. The caption collision is real - `en_dispatched` ships from other packages with it="Inviato" - and moving to `en_dispatched_to_proxy` with its own localization entry is the right fix. The tests ask the real resource on a real instance for both shapes (with and without the service).

Not verified here: a browser pass on a proxy instance.

PR #1194, opened 2026-08-28
fix(migrate): keep column commands when the table also needs a pkey rebuild

### LINKED ISSUE #1193: [BUG] Migration drops ADD COLUMN when the same table also needs a primary key rebuild
### Problem Description

`SqlMigrator` silently drops all `ADD COLUMN` / `ALTER COLUMN` statements for any table that also needs a primary key rebuild in the same migration. The migration then emits the indexes and foreign keys that depend on those columns, and PostgreSQL aborts with `column "<name>" does not exist`.

The bug is in `ExecutorMixin.sqlCommandsForTable` (`gnrpy/gnr/sql/gnrsqlmigration/executor.py`):

```python
table_command = tbl_item.get('command')
if table_command:
    # New table: the command is the complete CREATE TABLE
    command_list.append(table_command)
elif col_commands:            # <-- column commands lost when table_command exists
    # Existing table: ALTER TABLE with all columns
    command_list.append(f"{alter_table_command}\n{col_commands};")
```

The comment assumes `tbl_item['command']` is only set for a new table (`added_table`, the full `CREATE TABLE`). That is not true: `changed_table` in `command_builder.py` writes into the *same* key when a primary key must be dropped and recreated:

```python
if changed_attribute == 'pkeys':
    ...
    self.schema_tables(schema_name)[table_name]['command'] = (
        f"{drop_pk_sql}\n{add_pk_sql}"
    )
```

So for an existing table that needs both a PK rebuild and a new column, the `elif` branch is never reached and the column commands are discarded. `added_index` and `added_relation` write to separate keys (`indexes`, `relations`), so they *are* emitted — producing SQL that references a column the migration never created.

**Observed behavior**: migration fails with `column "xgroup" does not exist`. Since the whole build script is sent to the adapter as a single `execute()`, nothing at all is applied.

**Expected behavior**: the `ADD COLUMN` is emitted before the PK rebuild and the migration completes.

### Reproduction Steps

1. Create an instance whose ORM contains a table with a primary key and a column that does not yet exist in the database (any `adm`-based instance on a recent `develop` works: `adm.user.xgroup` was added together with the `adm.xgroup` table).
2. Bring the database into a state where that table has no primary key, so the migrator produces a `changed_table/pkeys` event:
   ```sql
   ALTER TABLE "adm"."adm_user" DROP CONSTRAINT adm_user_pkey;
   ALTER TABLE "adm"."adm_user" DROP COLUMN IF EXISTS "xgroup";
   ```
3. Run `gnr db migrate <instance>`.

Observed result:

```
INFO: gnr.db/gnrmigrate - STRUCTURE NEEDS CHANGES
INFO: gnr.db/gnrmigrate - APPLYING CHANGES TO DATABASE...
ERROR: gnr.db/gnrmigrate - store main: column "xgroup" does not exist
ERROR: gnr.db/gnrmigrate - Migration completed with errors on 1 store(s): main
```

`gnr db migrate <instance> -c -v` shows the mismatch directly — there is no `ADD COLUMN "xgroup"` anywhere in the script, but both of these are present:

```sql
CREATE INDEX idx_5fd0c3cf ON "adm"."adm_user" USING btree ("xgroup");
ALTER TABLE "adm"."adm_user"
 ADD CONSTRAINT "fk_8b2a513d" FOREIGN KEY ("xgroup") REFERENCES "adm"."adm_xgroup" ("code") ON UPDATE CASCADE DEFERRABLE INITIALLY DEFERRED;
```

The condition is easy to hit on a database restored from a schema-less or data-only dump, where *no* table has a primary key: every table then produces a `changed_table/pkeys` event and the migrator drops the column commands for all of them at once.

Dumping `migrator.commands` after `prepareMigrationCommands()` on such a database shows the discarded statements clearly. On the instance where this was found (103 tables, 0 primary keys, 0 indexes) the following were all silently lost:

```
adm.adm_user         ADD COLUMN "xgroup" character varying(10)
dem.dem_campagna     ADD COLUMN "cliente_tpl_id", "contenuto_fisso", "contenuto", "contenuto_snapshot"
dem.dem_cliente_tpl  ADD COLUMN "contenuto_fisso"
+ 25 tables with ALTER COLUMN "id" SET NOT NULL
```

Note that the second failure mode is worse than the visible one: when a lost column has no index and no foreign key, the migration reports success and the column is simply never created.

### Environment

genropy `develop` @ 9a3ca4c32, Python 3.12.9, PostgreSQL 16, macOS 15 (Darwin 25.1.0)

### Additional Material

Suggested fix — emit column changes first (a PK rebuild may target a column added in the same migration), then the table command. `added_table` builds its columns inline into the `CREATE TABLE` and never populates `tbl_item['columns']`, so for a new table `col_commands` is empty and the two branches cannot overlap:

```python
table_command = tbl_item.get('command')
# Column changes go first: a PK rebuild (changed_table) may target a column
# added in this same migration. New tables have no column commands here —
# their columns are inline in the CREATE TABLE built by added_table.
if col_commands:
    command_list.append(f"{alter_table_command}\n{col_commands};")
if table_command:
    command_list.append(table_command)
```

dgpaci:
Fixes #1193

## Problem

`ExecutorMixin.sqlCommandsForTable` treated the table `command` and the column commands as mutually exclusive:

```python
table_command = tbl_item.get('command')
if table_command:
    # New table: the command is the complete CREATE TABLE
    command_list.append(table_command)
elif col_commands:
    command_list.append(f"{alter_table_command}\n{col_commands};")
```

The comment assumes `tbl_item['command']` is only set by `added_table` for a `CREATE TABLE`. It is also set by `changed_table` when a primary key must be dropped and recreated, so any existing table needing **both** a PK rebuild and a column change lost every one of its `ADD COLUMN` / `ALTER COLUMN` statements.

`added_index` and `added_relation` write to separate keys, so they were still emitted — producing SQL that references a column the migration never created:

```
ERROR: gnr.db/gnrmigrate - store main: column "xgroup" does not exist
```

The silent variant is worse: when a lost column had neither an index nor a foreign key, the migration reported success and the column was simply never created.

## Fix

Emit the column commands **before** the table command instead of as an alternative to it. The order matters: a PK rebuild may target a column added by the same migration.

New tables are unaffected — `added_table` inlines their columns into the `CREATE TABLE` and never populates `tbl_item['columns']`, so `col_commands` is empty for them and the two branches cannot overlap.

The docstrings describing the assembly order were updated accordingly.

## Test Cases

New `TestSqlCommandsForTable` class in `gnrpy/tests/sql/test_gnrsqlmigration.py`, three unit tests, no database required:

- `test_added_column_kept_with_pkey_rebuild` — a table with a PK rebuild + a new column + an index and an FK on that column: asserts the `ADD COLUMN` is emitted, and that it precedes both the PK rebuild and the index. **Fails on `develop`**, passes with this change.
- `test_new_table_emits_only_create_table` — a new table emits its `CREATE TABLE` and nothing else (guards against a duplicate `ALTER TABLE`).
- `test_changed_column_without_pkey_rebuild` — with no table command, column changes are still merged into a single `ALTER TABLE`.

`pytest tests/sql` → 1213 passed, 10 skipped.

## Verification on a real instance

Reproduced on a database restored without its post-data section (103 tables, 0 primary keys, 0 indexes), which makes every table produce a `changed_table/pkeys` event. Before the fix, six `ADD COLUMN` and 27 `ALTER COLUMN ... SET NOT NULL` were discarded; after it the generated SQL is correctly ordered:

```sql
ALTER TABLE "adm"."adm_user"
ALTER COLUMN "id" SET NOT NULL,
ADD COLUMN "xgroup" character varying(10);
ALTER TABLE "adm"."adm_user" DROP CONSTRAINT IF EXISTS adm_user_pkey;
ALTER TABLE "adm"."adm_user" ADD PRIMARY KEY (id);
...
CREATE INDEX idx_5fd0c3cf ON "adm"."adm_user" USING btree ("xgroup");
```

and `gnr db migrate` completes, followed by `STRUCTURE OK` on a re-check.


fporcari (approved):
Approved. The failure mechanism is confirmed in `sqlCommandsForTable`: a primary-key rebuild and column commands can coexist on an existing table, so treating `table_command` and `col_commands` as alternatives drops valid migration SQL. Emitting the column `ALTER TABLE` first is the required order when the rebuilt key may use a newly added column, while new tables remain unaffected because their columns are already inline in `CREATE TABLE`.

The isolated focused tests pass (`3 passed, 122 deselected`), flake8 is clean, and all GitHub checks are green. #1193 remains linked and assigned to the PR author.

Not verified: the restored-database migration described in the PR body; my local verification covered the SQL assembly path and its ordering invariants.

PR #1196, opened 2026-08-28
fix(glbl): add the ISO 3166-1 codes assigned since the country list was compiled

dgpaci:
Fixes #1195

## Problem

The country list shipped by `gnr_it:glbl` is the ISO 3166-1 alpha-2 list as it stood before 2006. It holds 240 entries and is missing 11 codes that have been officially assigned for 14 to 20 years.

`glbl.nazione` is a lookup other packages point foreign keys at, so this is not merely cosmetic: one contact carrying such a code makes the foreign key uncreatable and aborts the migration.

```
ERROR: gnr.db/gnrmigrate - store main: insert or update on table "dem_contatto"
violates foreign key constraint "fk_b050c38b"
DETAIL:  Key (nazione)=(RS) is not present in table "glbl_nazione".
```

## Changes

The 11 missing codes, with their ISO 3166-1 alpha-3 and numeric codes:

| code | alpha-3 | num | country | assigned |
|---|---|---|---|---|
| `RS` | SRB | 688 | Serbia | 2006 |
| `ME` | MNE | 499 | Montenegro | 2006 |
| `GG` | GGY | 831 | Guernsey | 2006 |
| `IM` | IMN | 833 | Isle of Man | 2006 |
| `JE` | JEY | 832 | Jersey | 2006 |
| `BL` | BLM | 652 | Saint Barthélemy | 2007 |
| `MF` | MAF | 663 | Saint Martin (French part) | 2007 |
| `BQ` | BES | 535 | Bonaire, Sint Eustatius and Saba | 2010 |
| `CW` | CUW | 531 | Curaçao | 2010 |
| `SX` | SXM | 534 | Sint Maarten (Dutch part) | 2010 |
| `SS` | SSD | 728 | South Sudan | 2011 |

They were added to **both** places the dataset lives:

- **`startup_data.gz`** — the file `Package.loadStartupData()` actually reads, and therefore what the `glbl|0001_ensure_data` upgrade, `nazione.onDbUpgrade()` and the *Restore GLBL* preference button install. Regenerated in place by appending to its `nazione` branch, so the other six tables (`cap`, `nuts`, `provincia`, `regione`, `comune`, `localita`) are carried over untouched — verified row by row against the previous archive.
- **`populate()`** in `model/nazione.py` — the fixed-width seed list. Nothing calls it today, but it is the readable record of the same data and letting it drift from the archive is how this gap goes unnoticed again. The 11 lines slot into the existing alphabetical order; the diff is +11 with no reflow.

[...]

PR #1199, opened 2026-08-28
hotfix(glbl): add the ISO 3166-1 codes assigned since the country list was compiled

### LINKED ISSUE #1195: [BUG] glbl country list is the pre-2006 ISO 3166-1 list: 11 valid codes missing, breaks foreign keys
### Problem Description

The country list shipped by `gnr_it:glbl` is the ISO 3166-1 alpha-2 list as it stood **before 2006**. It still contains codes that were withdrawn (`CS` Serbia and Montenegro, `AN` Netherlands Antilles) and is missing 11 codes that have been officially assigned for 14 to 20 years:

| code | country | assigned |
|---|---|---|
| `RS` | Serbia | 2006 |
| `ME` | Montenegro | 2006 |
| `GG` | Guernsey | 2006 |
| `IM` | Isle of Man | 2006 |
| `JE` | Jersey | 2006 |
| `BL` | Saint Barthélemy | 2007 |
| `MF` | Saint Martin (French part) | 2007 |
| `BQ` | Bonaire, Sint Eustatius and Saba | 2010 |
| `CW` | Curaçao | 2010 |
| `SX` | Sint Maarten (Dutch part) | 2010 |
| `SS` | South Sudan | 2011 |

`glbl.nazione` is a lookup referenced by foreign key from application tables (for example `dem.contatto.nazione` in `gnrcommunication:dem`). Any record carrying one of these codes makes the foreign key uncreatable, so `gnr db migrate` aborts on a database that has real contact data:

```
ERROR: gnr.db/gnrmigrate - store main: insert or update on table "dem_contatto"
violates foreign key constraint "fk_b050c38b"
DETAIL:  Key (nazione)=(RS) is not present in table "glbl_nazione".
```

Outside of migration the same gap shows up as a country that simply cannot be selected in the UI, and as imports silently rejecting valid addresses.

The stale list appears in two places:

- `projects/gnr_it/packages/glbl/startup_data.gz` — the file actually loaded by `Package.loadStartupData()`, which is what the `glbl|0001_ensure_data` upgrade, `nazione.onDbUpgrade()` and the *Restore GLBL* preference button all call. It holds 240 countries.
- `projects/gnr_it/packages/glbl/model/nazione.py` — the fixed-width `populate()` seed list, same vintage. Worth noting that nothing in the framework or in the package calls `populate()` today; it survives as the readable record of where the data came from.

### Reproduction Steps

1. Take any instance including `gnr_it:glbl` and a package with a foreign key onto `glbl.nazione` (`gnrcommunication:dem` has one on `dem.contatto.nazione`).
2. Load the lookup data: `gnr db migrate <instance> -U` (runs `glbl|0001_ensure_data`).
3. Insert a contact with `nazione = 'RS'` — a valid ISO 3166-1 alpha-2 code.
4. Run `gnr db migrate <instance>`.

Observed result: the migration fails with the foreign key violation above. `select count(*) from glbl.glbl_nazione` returns 240, and `select * from glbl.glbl_nazione where code = 'RS'` returns no rows, while `code = 'CS'` returns `SERBIA AND MONTENEGRO`.

Checking the shipped dataset directly is enough — no instance required:

```python
import gzip, shutil
from gnr.core.gnrbag import Bag
with gzip.open('projects/gnr_it/packages/glbl/startup_data.gz','rb') as f_in, open('/tmp/sd.pik','wb') as f_out:
    shutil.copyfileobj(f_in, f_out)
codes = {r['code'] for r in Bag('/tmp/sd.pik')['nazione']}
print(len(codes), 'RS' in codes, 'CS' in codes)   # -> 240 False True
```

### Environment

genropy `develop` @ 9a3ca4c32, Python 3.12.9, PostgreSQL 16, macOS 15 (Darwin 25.1.0)

### Additional Material

Proposed fix: add the 11 missing codes to both `startup_data.gz` and the `populate()` list, with their ISO 3166-1 alpha-3 and numeric codes:

```
BL BLM 652   BQ BES 535   CW CUW 531   GG GGY 831   IM IMN 833   JE JEY 832
ME MNE 499   MF MAF 663   RS SRB 688   SS SSD 728   SX SXM 534
```

`CS` and `AN` should be **kept**, not removed: existing databases may already reference them, and both are recorded in ISO 3166-3 as formerly-used codes. Dropping them would turn a stale-data problem into a data-loss one.

dgpaci:
Backport of #1196 (merged into `develop`) onto `master`.

Fixes #1195

## Problem

The country list shipped by `gnr_it:glbl` is the ISO 3166-1 alpha-2 list as it stood before 2006. It holds 240 entries and is missing 11 codes that have been officially assigned for 14 to 20 years.

`glbl.nazione` is a lookup other packages point foreign keys at, so this is not merely cosmetic: one contact carrying such a code makes the foreign key uncreatable and aborts the migration.

```
ERROR: gnr.db/gnrmigrate - store main: insert or update on table "dem_contatto"
violates foreign key constraint "fk_b050c38b"
DETAIL:  Key (nazione)=(RS) is not present in table "glbl_nazione".
```

## Changes

The 11 missing codes, with their ISO 3166-1 alpha-3 and numeric codes:

| code | alpha-3 | num | country | assigned |
|---|---|---|---|---|
| `RS` | SRB | 688 | Serbia | 2006 |
| `ME` | MNE | 499 | Montenegro | 2006 |
| `GG` | GGY | 831 | Guernsey | 2006 |
| `IM` | IMN | 833 | Isle of Man | 2006 |
| `JE` | JEY | 832 | Jersey | 2006 |
| `BL` | BLM | 652 | Saint Barthélemy | 2007 |
| `MF` | MAF | 663 | Saint Martin (French part) | 2007 |
| `BQ` | BES | 535 | Bonaire, Sint Eustatius and Saba | 2010 |
| `CW` | CUW | 531 | Curaçao | 2010 |
| `SX` | SXM | 534 | Sint Maarten (Dutch part) | 2010 |
| `SS` | SSD | 728 | South Sudan | 2011 |

They were added to **both** places the dataset lives:

- **`startup_data.gz`** — the file `Package.loadStartupData()` actually reads, and therefore what the `glbl|0001_ensure_data` upgrade, `nazione.onDbUpgrade()` and the *Restore GLBL* preference button install. Regenerated in place by appending to its `nazione` branch, so the other six tables (`cap`, `nuts`, `provincia`, `regione`, `comune`, `localita`) are carried over untouched — verified row by row against the previous archive.
- **`populate()`** in `model/nazione.py` — the fixed-width seed list. Nothing calls it today, but it is the readable record of the same data and letting it drift from the archive is how this gap goes unnoticed again. The 11 lines slot into the existing alphabetical order; the diff is +11 with no reflow.

`CS` (Serbia and Montenegro) and `AN` (Netherlands Antilles) are **kept on purpose**. Existing databases may already reference them and ISO 3166-3 records both as formerly used, so removing them would turn a stale-data problem into a data-loss one.

The pre-fix `startup_data.gz` and `model/nazione.py` on `master` were byte-identical to the ones this fix was written against, so the backported files are exactly the reviewed ones.

## Test Cases

New `gnrpy/tests/projects/test_glbl_nazione_data.py`, 7 tests, no database required. It reads the archive the way the loader does and drives the real `populate()` (subclassing `Table` and capturing `insertOrUpdate`) rather than reimplementing its fixed-width parsing:

- `test_startup_data_and_populate_agree` — the two sources must describe the same set of countries; catches the drift that let this gap persist.
- `test_codes_added_since_2006_are_present` (both sources) — the 11 codes are present with the right alpha-3. **Fails on `master` without this change**, listing all 11 as missing.
- `test_withdrawn_codes_are_kept` (both sources) — guards `CS` and `AN` against a well-meaning cleanup.
- `test_records_are_well_formed` (both sources) — no duplicate codes; alpha-2/alpha-3/numeric widths and character classes hold.

`pytest tests/projects tests/sql` on this branch → 1170 passed, 32 skipped. flake8 clean on both modified Python files.

## Verification on a real instance

On a `gnrcommunication:dem` instance with ~5000 contacts, one of them Serbian, `gnr db migrate` failed on `fk_b050c38b`. With the countries in place all 98 foreign keys of that migration can be created and the migration completes, followed by `STRUCTURE OK` on a re-check.


fporcari (changes requested):
The dataset update itself is correct and narrowly scoped: the two shipped sources agree, the six non-country data branches are unchanged row for row, existing country records are unchanged, and the country branch adds exactly the eleven intended codes. The focused tests and flake8 pass, and all GitHub checks are green.

However, this does not update the already-populated databases described in #1195. I reproduced the upgrade path on a temporary PostgreSQL database using the actual `glbl` package from this head:

- loading the `master` archive produced 240 countries with no `RS` record;
- running the real `onDbUpgrade` broadcast and `0001_ensure_data.py` left the database at 240 countries with no `RS` record;
- only a manual `loadStartupData(empty_before=False)` added the eleven rows, producing 251 countries with `RS` present.

Both `nazione.onDbUpgrade()` and `0001_ensure_data.py` load startup data only when the table is empty, so an installation with the existing 240 rows skips the new archive. That is the affected state in the issue reproduction.

Please add an idempotent upgrade path for populated databases and a real-database regression covering 240 existing rows becoming 251. The rollout also needs to account for command order: a normal migration applies structural changes before package upgrades, so an upgrade-only step may be required before recreating the failing foreign key.

fporcari (approved):
Approved after correcting my previous review.

My change request conflated two separate questions: whether the existing upgrade hooks invoke the loader automatically, and whether this change can add the missing rows to an already-populated database. The latter is the contract of this dataset change, and it works.

I copied all 240 real `glbl.nazione` rows from `sandboxpg` into a temporary PostgreSQL database, then ran `loadStartupData()` with the archive from this head and `empty_before=False`. The result was 251 rows, with all eleven new codes present and all existing rows retained. This exercises the same loader used by Restore GLBL.

The current head only adds the expected merge from `master`; the PR diff remains limited to the three reviewed files. The focused tests and flake8 passed during the original review, and every GitHub check on the current head is green.

Operationally, populated installations must invoke the loader before recreating a foreign key that depends on one of the new codes. The existing `onDbUpgrade`/`0001_ensure_data` guards do not do that automatically, but that is distinct from the correctness of this PR.

Not verified: the production-instance migration described in the PR body.

Issue #1200, opened 2026-08-31
[BUG] multidb getStorePreference fails when the current storename is a multi-store selection


fporcari:
### Problem Description

With the multidb package, when the connection env carries a multi-store selection (`storename` is a comma-separated list, e.g. `store_a,store_b`), any preference read crashes the request.

`adm.preference.getStorePreferences()` calls `multidb.getStorePreference()`, which does:

```python
store_record = self.db.table(self.db.storetable).record(dbstore=storename).output('record')
```

Two problems:

1. The lookup runs inside the current env, so on a storetable with `use_dbstores` truthy (e.g. `multidb='one'`) the SELECT is dispatched to every store connection as a multi-cursor execution instead of hitting the root store (compare with `setStorePreference`, which correctly wraps the lookup in `tempEnv(storename=self.db.rootstore)`).
2. Even if it reached the root store, no record can have `dbstore` equal to the comma-joined string, so the semantics are wrong for multi-store selections: each store's local preferences should be read and merged.

The observed failure is an `AttributeError: 'list' object has no attribute 'rollback'` raised while handling the SQL error of the multi-cursor execution (see companion core issue), which masks the original exception. Net effect: any page dispatch with a multi-store selection dies, because the framework itself reads `sys.experimental.page_class_cache` on every request.

### Reproduction Steps

1. Instance with the multidb package and a storetable whose table has `use_dbstores` truthy (e.g. `multidb='one'`).
2. Set a multi-store env: `db.currentEnv['storename'] = 'store_a,store_b'` (in the UI: select more than one store).
3. Call `db.table('adm.preference').getPreference('experimental.page_class_cache', pkg='sys')` — or just load any page.

Observed result: `AttributeError("'list' object has no attribute 'rollback'")` from `gnr/sql/gnrsql/transactions.py`.

### Proposed fix

In `multidb.getStorePreference()`: split the comma-separated storename (`'*'` expands to all stores), read each store record with `tempEnv(storename=rootstore)` and `ignoreMissing=True`, and merge the preference bags in selection order (later stores override earlier ones on conflicting paths).

### Environment

Production, genropy develop (`gnrresourceloader` page_class_cache path), PostgreSQL.

fporcari:
[...]

PR #1202, opened 2026-08-31
fix(sql): rollback rolls back every connection of a multi-store selection

### LINKED ISSUE #1201: [BUG] GnrSqlDb.rollback crashes on multi-store connections, masking the original SQL error
### Problem Description

`GnrSqlDb.rollback()` (gnr/sql/gnrsql/transactions.py) does:

```python
def rollback(self):
    self.connection.rollback()
```

But the `connection` property returns a **list** of connections when the current storename is `'*'` or a comma-separated multi-store selection (`connections.py::_get_connection`). In that case `rollback()` raises `AttributeError: 'list' object has no attribute 'rollback'`.

The practical damage: `execute()` calls `self.rollback()` in its `except` clause, so whenever a multi-cursor execution fails, the AttributeError **replaces the original SQL error** in the traceback — the real cause never reaches the logs. Found in production while diagnosing #1200: three stack traces, all ending on the AttributeError instead of the actual failure.

### Reproduction Steps

1. `db.currentEnv['storename'] = 'store_a,store_b'` (any multidb instance with two stores).
2. Run any statement that fails on at least one store, e.g. `db.execute('SELECT no_such_column FROM some_table')`.

Observed result: `AttributeError("'list' object has no attribute 'rollback'")`; the original ProgrammingError is lost.
Expected: each connection in the list is rolled back and the original exception propagates.

### Proposed fix

```python
def rollback(self):
    connection = self.connection
    if isinstance(connection, list):
        for c in connection:
            c.rollback()
    else:
        connection.rollback()
```

(`commit()` already walks all the thread's connections; `rollback` is the only transaction method assuming a single connection.)

### Environment

Production, genropy develop, PostgreSQL.

fporcari:
Fixes #1201

`GnrSqlDb.rollback()` assumed a single connection, but with a multi-store storename (`'*'` or comma-separated) the `connection` property returns a list, so the call raised `AttributeError: 'list' object has no attribute 'rollback'` — replacing, inside `execute()`'s except clause, the original SQL error of the failed multi-cursor execution.

Now each connection in the list is rolled back, so the real exception propagates. Found in production while diagnosing #1200.

Full gnrpy test suite green locally (2202 passed, 10 skipped).

### DIFF
--- a/gnrpy/gnr/sql/gnrsql/transactions.py
+++ b/gnrpy/gnr/sql/gnrsql/transactions.py
@@ -250,8 +250,18 @@ class TransactionMixin(GnrSqlDbBaseMixin):
             cursor.setConstraintsDeferred()
 
     def rollback(self) -> None:
-        """Roll back the current connection's transaction."""
-        self.connection.rollback()
+        """Roll back the current connection's transaction.
+
+        With a multi-store storename (``'*'`` or comma-separated) the
+        ``connection`` property returns a list of connections: each one
+        is rolled back.
+        """
+        connection = self.connection
+        if isinstance(connection, list):
+            for c in connection:
+                c.rollback()
+        else:
+            connection.rollback()
 
     def rollbackAll(self) -> None:
         """Roll back every uncommitted connection of the current thread.


cgabriel (approved):
ok

genro (approved):
Approving. Twelve lines, and the guard is exact rather than merely defensive: `_get_connection` returns a list comprehension at `connections.py:117` when the storename addresses more than one store, so `isinstance(connection, list)` matches the only shape that can occur — not a tuple, not a generator, nothing that would slip past it.

The part worth naming is what the bug did rather than what it was. `rollback()` raising `AttributeError` **inside `execute()`'s except clause** meant the new exception replaced the original SQL error, so every multi-store failure reported the rollback instead of its own cause. That is why #1200 was hard to read at all, and why this fix is a prerequisite for diagnosing anything in that area rather than a tidy-up: it does not make failures rarer, it makes them legible.

Same story on the other side: #1204's description says the preference crash was masked by this very bug. Two reports that could not be understood until the rollback stopped eating their exceptions.

PR #1204, opened 2026-08-31
fix(adm): with a multi-store selection the main store preferences apply

### LINKED ISSUE #1203: [BUG] Preference lookup with a multi-store selection crashes; main store preferences must apply
### Problem Description

With the multidb package, when the connection env carries a multi-store selection (`storename` = `'*'` or a comma-separated list, e.g. `store_a,store_b`), any preference read crashes the request: `adm.preference.getStorePreferences()` hands the multi-store storename to `multidb.getStorePreference()`, which looks up a single storetable record by `dbstore=storename` — a lookup that cannot succeed for a comma-joined string and, on storetables with `use_dbstores` truthy (e.g. `multidb='one'`), is even dispatched as a multi-cursor SELECT to every store connection, where it fails (until #1201 the failure was masked by the rollback AttributeError; three production stack traces attached there).

Since the framework itself reads `sys.experimental.page_class_cache` on every dispatch, **every page load with a multi-store selection dies**.

### Design choice

Store-local preferences are meaningful only when exactly one store is active. With a multi-store selection there is no correct single answer: merging the stores' preferences would be arbitrary (values from different stores would mix, or the last one would silently win). Therefore:

- **multi-store env → the main store preferences apply, with no per-store overlay**;
- code that needs a store-local preference must read it within a single-store env (`tempEnv(storename=...)`) — which is where such reads already happen in practice (store-scoped operations);
- `multidb.getStorePreference()` keeps its contract: it is only ever called with a single storename.

### Proposed fix

Short-circuit upstream, in `adm.preference.getStorePreferences()`: when the current storename addresses more than one store, return `getMainStorePreference()` directly. A small `db.usingMultiStore()` helper alongside `usingRootstore()` keeps the check readable.

### Reproduction Steps

1. Instance with the multidb package and at least two stores.
2. `db.currentEnv['storename'] = 'store_a,store_b'` (in the UI: select more than one store).
3. Load any page, or call `db.table('adm.preference').getPreference('experimental.page_class_cache', pkg='sys')`.

Observed: request dies (AttributeError masking a multi-cursor SQL failure). Expected: main store preferences are returned.

### Environment

Production, genropy develop, PostgreSQL. Supersedes #1200.

fporcari:
Fixes #1203

With a multi-store selection (`storename` = `'*'` or comma-separated), `adm.preference.getStorePreferences()` handed the multi-store storename to `multidb.getStorePreference()`, whose single-record lookup on the storetable cannot succeed for a comma-joined string and — on storetables with `use_dbstores` truthy — was dispatched as a failing multi-cursor SELECT. Since the framework reads `sys.experimental.page_class_cache` on every dispatch, every page load with a multi-store selection died (the failure itself was masked by the rollback bug fixed in #1202).

**Design choice** (see #1203): store-local preferences are meaningful only with exactly one active store — merging several stores' preferences would be arbitrary (values mix, or the last store silently wins). So with a multi-store selection the main store preferences apply, with no per-store overlay; code needing a store-local preference must read it within a single-store `tempEnv`.

The short-circuit lives upstream in `getStorePreferences()`, via a new `db.usingMultiStore()` helper alongside `usingRootstore()`, so `multidb.getStorePreference()` keeps its single-store contract and is never called with a multi-store storename. The two runtime imports of `MAIN_CONNECTION_NAME` in `env.py` move to module level (style hook).

Full gnrpy test suite green locally (pre-push hook).

### DIFF
--- a/gnrpy/gnr/sql/gnrsql/env.py
+++ b/gnrpy/gnr/sql/gnrsql/env.py
@@ -35,7 +35,7 @@ from typing import Any
 
 from gnr.core.gnrlocale import defaultLocale
 from gnr.sql._typing import GnrSqlDbBaseMixin
-from gnr.sql.gnrsql.helpers import TempEnv
+from gnr.sql.gnrsql.helpers import MAIN_CONNECTION_NAME, TempEnv
 
 
 class EnvMixin(GnrSqlDbBaseMixin):
@@ -192,9 +192,13 @@ class EnvMixin(GnrSqlDbBaseMixin):
 
     def usingMainConnection(self) -> bool:
         """Return ``True`` if the current connection is the main connection."""
-        from gnr.sql.gnrsql.helpers import MAIN_CONNECTION_NAME
-
         return self.currentConnectionName == MAIN_CONNECTION_NAME
+
+    def usingMultiStore(self) -> bool:
+        """Return ``True`` if the current storename addresses more than one
+        store (``'*'`` or a comma-separated selection)."""
+        storename = self.currentStorename
+        return storename == '*' or ',' in storename
 
     @property
     def currentStorename(self) -> str:
@@ -204,8 +208,6 @@ class EnvMixin(GnrSqlDbBaseMixin):
     @property
     def currentConnectionName(self) -> str:
         """The name of the currently active connection."""
-        from gnr.sql.gnrsql.helpers import MAIN_CONNECTION_NAME
-
         return self.currentEnv.get('connectionName') or MAIN_CONNECTION_NAME
 
     def setLocale(self) -> None:
--- a/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py
+++ b/gnrpy/gnr/web/gnrwsgisite_proxy/gnrresourceloader.py
@@ -59,16 +59,19 @@ class ResourceLoader(object):
     def page_class_cache_enabled(self):
         """The experimental flag, read through a small TTL (kill switch stays live).
 
-        ``getPreference`` resolves per dbstore: on a multidb instance the value
-        that holds for the next TTL is the one read by whichever store refreshed
-        it. That is fine for a kill switch - it is not a per-store setting.
+        The read is pinned to the root store: it runs at dispatch time, before
+        the page applies its own env, so the thread env still carries whatever
+        the previous request left there (possibly a multi-store selection).
+        A ``sys`` preference lives in the main store anyway.
         Turning the flag off also drops what is already cached, so the switch
         frees the classes instead of merely stopping new ones.
         """
         value, read_ts = self._page_class_cache_flag
         now = time.time()
         if now - read_ts > PAGE_CLASS_CACHE_FLAG_TTL:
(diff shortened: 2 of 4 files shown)


fporcari:
Added a second commit that fixes the exact origin of the production stack traces: `page_class_cache_enabled` reads the `sys.experimental.page_class_cache` preference **at dispatch time, before the page applies its own env**, so it ran under whatever env the previous request left on the worker thread — with a leftover multi-store selection, the lookup crashed the dispatch. The read is now pinned to the root store with `tempEnv` (a `sys` preference lives in the main store by nature). The guard in `getStorePreferences` from the first commit remains as the semantic rule for any other preference read under a multi-store env.

cgabriel (approved):
Verified with @fporcari the nature of the change in adm/model/preference.py, so it's fine for me.

genro (approved):
Approving. The diagnosis holds and the fix is placed where it belongs.

Putting the short circuit in `getStorePreferences()` rather than inside `multidb.getStorePreference()` is the right call: that method keeps its single-store contract and is simply never handed a comma-joined name, instead of learning to defend itself against one. And the design choice behind it is the part worth having written down — store-local preferences only mean something with exactly one active store, so merging several would have to pick between mixing values and letting the last store win, and both are arbitrary. Falling back to the main store is the only answer that is explainable to whoever hits it.

I checked the one thing `usingMultiStore` could trip on, since it runs on every call: `currentStorename` is `self.currentEnv.get('storename') or self.rootstore`, and `rootstore` is a class attribute defaulting to `'_main_db'` (`gnrsql/db.py:90`), so it can never be `None` and `',' in storename` cannot raise. Moving the two `MAIN_CONNECTION_NAME` imports to module level is safe as well — `helpers` does not import `env`, so there is no cycle to reintroduce.

You are right about the comment you replaced, and it was mine. It argued that a per-dbstore read was fine for a kill switch, which was true about *which* store answers and blind to the fact that the current store may not be a single store at all. The new reasoning — the read happens at dispatch time, before the page applies its own env, so the thread still carries whatever the previous request left — is the accurate one.

**One thing I would like added, and it is only the test file.** The `tempEnv(storename=db.rootstore)` in `page_class_cache_enabled` is the line that actually fixes the crash on the dispatch path, and it is the one thing the new tests do not exercise: `_FakeTempEnv` is a silent context manager, so every assertion passes identically with the `with` removed. Three lines close it:

```python
class _FakeDb:
    rootstore = '_main_db'

    def __init__(self):
        self.tempenv_calls = []

    def tempEnv(self, **kwargs):
        self.tempenv_calls.append(kwargs)
        return _FakeTempEnv()
```

```python
def test_flag_is_read_pinned_to_the_root_store():
    loader, site = _loader(preference=True)
    loader.page_class_cache_enabled()
    assert site.db.tempenv_calls == [{'storename': '_main_db'}]
```

Not blocking, because the production change is right and I would rather it shipped. But without it a later refactor of that method can drop the `with`, keep the suite green, and put the bug back exactly as it was — silent, on every page load with a multi-store selection. The other two changes do not need this: `usingMultiStore` is trivial and the short circuit is verifiable by reading.

PR #1207, opened 2026-08-31
fix(gnrjs): keep subtree subscriptions on same-bag rebuilds

### LINKED ISSUE #1206: documentFrame never reloads after the first render when its pkey changes (regression window: #1152 trigger dispatch rework)
## Symptom (field report, icond production)

A page shows a `documentFrame` whose `pkey` follows a grid selection. The first document renders correctly; from then on, **changing the selected record never updates the frame** — the previous record's PDF stays on screen.

Two hard facts from the server, which rule out the print side:

- since the last restart, each page session's `page:pdf` folder holds **exactly one** file per print resource. With record B selected, only record A's PDF exists on disk. The print is never re-run, so no rpc reaches the server.
- a sibling tab keyed on the very same datapath (`condition_anag_id='^main.<th>.view.grid.selectedId'` on a tablehandler) **does** refresh. So the selection value changes and normal `^attr` subscriptions on that path react.

The user reports it started "after the last updates". The instance tracks `develop`.

## Where the reload should come from

`gnr.widgets.DocumentFrame` (`gnrjs/gnr_d11/js/genro_components.js:2129`) builds a framePane with `_workspace: true` and wires:

- the iframe with `_reloader = '^#WORKSPACE.reload_iframe'` and `_if = '^#WORKSPACE.enabled'`
- a `dataController` with the print parameters as `^` triggers, whose script is
  `SET #WORKSPACE.enabled = true; FIRE #WORKSPACE.reload_iframe;`

So the first render can be driven by `set_if` / node creation (`created:` calls `setSrc`, `genro_widgets.js:999`), while **every later render depends on the `FIRE` reaching the iframe's `_reloader` subscription** — a subscription on a `#WORKSPACE` path.

## Hypothesis, and why this window

`b4ba4d799` (#1152, 2026-08-26) replaced the global `_trigger_data` topic with a path-indexed dispatch. Two things in it touch this path:

1. `#WORKSPACE` heads are deliberately **not** indexable (`_stableSymbolicStart`, `gnrdomsource.js:1229`), so those subscriptions live in the floating set and are checked on every event. That set is always included in `GnrTriggerIndex.publish` (`genro_src.js:74`), so registration alone looks correct.

2. `getTriggerReason` now memoizes the event path **on the event object** (`gnrdomsource.js:132`):

```js
var eventpath = kw._eventpath !== undefined ? kw._eventpath :
                (kw._eventpath = kw.pathlist.slice(1).join('.'));
```

while `Bag.onNodeTrigger` re-publishes **the same `kw`** at every level of the bag hierarchy, replacing `pathlist` each time (`gnrbag.js:1994`):

```js
this.runTrigger(kw);
if (this._parent != null) {
    kw.pathlist = [this._parentnode.label].concat(kw.pathlist);
    this._parent.onNodeTrigger(kw);
}
```

A value memoized from a field that is then replaced is wrong by construction: any consumer reached at a later level compares subscriber paths against the earlier, shorter path, and the match silently fails. That is a latent defect independent of this symptom and should be fixed on its own.

## What is NOT verified

- I did **not** establish that a single event reaches `getTriggerReason` at two different levels. The DOM dispatch is subscribed on the datastore root only (`genro.js:690`), and the two bag-level subscribers I read (`rowLogger`, `genro_wdg.js:2018`; `dataLogger`, `genro_frm.js:517`) do not call `getTriggerReason` on the paths I followed. So the memo may be latent-but-harmless here, and the documentFrame failure may have another cause inside the same dispatch rework.
- The symptom itself is **not reproduced locally** yet. All the above is code reading plus the server-side evidence.

## Experiment that settles it

`projects/gnrcore/packages/test/webpages/components/document_frame.py` is already the minimal case: a `documentFrame` whose `pkey` comes from a `dbselect`. Locally, no login needed.

1. On today's `develop`: change the record in the select — does the document reload? Expected failure.
2. On `b4ba4d799^`: same page, same action. If it reloads there, the regression is confirmed and bisected to #1152.
3. If confirmed, instrument `GnrTriggerIndex.publish` / `getTriggerReason` for the `FIRE` on `#WORKSPACE.reload_iframe`: does the iframe subscription get called at all (registration/floating-set problem), or is it called and rejected (event-path matching problem)?

Worth checking with **two** documentFrames on the page as well: the field case has two, the test page one.

## Candidate fixes

- event-path matching: invalidate the memo where the pathlist is replaced (`delete kw._eventpath` next to the reassignment in `onNodeTrigger`), or compute it once per publish in `genro.publishDataTrigger` — both keep the per-level saving the memo was added for, which was avoiding one `join` per subscriber of the same event.
- registration: if the floating subscription is the problem, `#WORKSPACE` subs need either a stable index key or a guaranteed floating registration at build time.

## Impact

Every project aligned to `develop` since 2026-08-26. `documentFrame` is the visible casualty because its refresh path goes through a `#WORKSPACE` `FIRE`, but if the event-path memo really can be poisoned, the blast radius is any `^attr` subscription on an event that crosses more than one bag level.

fporcari:
## Problem

Since #1152, a `documentFrame` whose `pkey` follows a selection renders only the first document: later changes never reach the server (#1206). Field report from icond production; reproduced locally on `test/components/document_frame`.

## Root cause

`rebuild()` re-sets the **same** content bag (`setValue(this._value)`, `gnrdomsource.js`). The incremental cleanup introduced by #1152 runs in `_trigger_upd` on `kw.oldvalue` — which on a rebuild IS the kept content — and unsubscribes every node of the subtree from the trigger index. Widget nodes re-register when they are rebuilt, but data nodes (`dataController`, `dataRpc`, `dataFormula`) register only once in `moveData`, behind the `_alreadyStripped` guard: they never resubscribe and go silent.

`documentFrame` is the visible casualty because its iframe `setSrc` calls `rebuild()` on every reload, killing the sibling dataController that drives the reload itself. The blast radius is wider: any data node inside a subtree that rebuilds after first build (`_if` toggles, `unfreeze`, dynamic `storepath`, the generic dynamic-attribute rebuild fallback).

The `_eventpath` memo hypothesis raised in #1206 was verified NOT to be the cause: both `getTriggerReason` callers run on the root-final pathlist. It remains a latent hardening, out of scope here.

## Fix

Skip the content-subscription cleanup when the node value is the same bag: the content is kept, not discarded — the case where the pre-#1152 `refreshSourceIndexAndSubscribers` preserved subscriptions of nodes still in the tree. Replaced content (a different bag) is still cleaned up as introduced by #1152.

## Verification

New test page `test/datastore/trigger_rebuild.py` covers the doubtful cases with exact-step counters:

- dataController + dataFormula surviving explicit rebuilds of their box
- dataController under an `_if`-governed container across visibility toggles
- remote content replacement: exactly one fire per trigger after multiple reloads (guards against stale duplicate subscriptions)
- sibling isolation: rebuilding one box never affects the other

All four verified in the browser (counters step by exactly 1). Also verified live on the sandbox instance:

- `test/components/document_frame`: each record change produces one `callTableScript`, with two documentFrames active on the same page
- `test15/gnrwdg/formhandler_baggrid`: grid selection reloads the linked form on every selection
- `test/perf/test_heavy_build` (the #1152 profiling page): builds and switches tabs with no console errors

flake8 clean on touched files; full gnrpy suite green (2202 passed, 10 skipped).

Fixes #1206

### DIFF
--- a/gnrjs/gnr_d11/js/genro_src.js
+++ b/gnrjs/gnr_d11/js/genro_src.js
@@ -223,7 +223,11 @@ dojo.declare("gnr.GnrSrcHandler", null, {
             console.log('missing destination in rebuild');
         }
         this._onDeletingContent(kw.oldvalue);
-        this.cleanupContentSubscriptions(kw.oldvalue);
+        if (kw.oldvalue !== kw.node._value) {
+            //on a same-bag rebuild the content is kept, not discarded: its
+            //data* nodes are already stripped and would never resubscribe
+            this.cleanupContentSubscriptions(kw.oldvalue);
+        }
         var domNode = kw.node.getDomNode();//get the domnode
         var newNode = document.createElement('div');
         var widget = kw.node.widget;
--- a/projects/gnrcore/packages/test/webpages/datastore/trigger_rebuild.py
+++ b/projects/gnrcore/packages/test/webpages/datastore/trigger_rebuild.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+
+"""Data node subscriptions across subtree rebuilds (#1206)"""
+
+from gnr.core.gnrdecorator import public_method
+
+
+class GnrCustomWebPage(object):
+    py_requires = "gnrcomponents/testhandler:TestHandlerFull"
+
+    def test_0_datacontroller_survives_rebuild(self, pane):
+        """Type in Input: Fired increments by 1 every time, Formula doubles it.
+        Click Rebuild box (even many times), then type again: both must still react."""
+        fb = pane.formbuilder(cols=4, border_spacing='3px')
+        fb.textbox(value='^.input', lbl='Input')
+        fb.button('Rebuild box').dataController("genro.nodeById('rb_box_0').rebuild();")
+        fb.numberTextBox(value='^.counter', lbl='Fired', readOnly=True)
+        fb.numberTextBox(value='^.doubled', lbl='Formula', readOnly=True)
+        box = pane.contentPane(nodeId='rb_box_0', border='1px solid silver',
+                               padding='5px', margin='5px')
+        box.div('The dataController and dataFormula of this box must survive its rebuild')
+        box.dataController("SET .counter = (GET .counter || 0) + 1;", _fired='^.input')
+        box.dataFormula('.doubled', 'inp*2', inp='^.counter')
+
+    def test_1_datacontroller_under_if(self, pane):
+        """Type in Input: Fired increments. Uncheck and recheck Visible (the box
+        rebuilds, like documentFrame does on every reload), then type again:
+        Fired must keep incrementing."""
+        pane.data('.flag', True)
+        fb = pane.formbuilder(cols=3, border_spacing='3px')
+        fb.textbox(value='^.input', lbl='Input')
+        fb.checkbox(value='^.flag', label='Visible')
+        fb.numberTextBox(value='^.counter', lbl='Fired', readOnly=True)
+        box = pane.contentPane(_if='^.flag', border='1px solid silver',
+                               padding='5px', margin='5px')
+        box.div('Box governed by _if')
+        box.dataController("SET .counter = (GET .counter || 0) + 1;", _fired='^.input')
+
+    def test_2_remote_content_replaced(self, pane):
+        """Change Reload token: the remote content is replaced with a fresh one.
+        Then type in Probe: Fired must increment by EXACTLY 1 per keystroke
+        (a higher step means stale subscriptions of the discarded content)."""
+        fb = pane.formbuilder(cols=3, border_spacing='3px')
+        fb.textbox(value='^.token', lbl='Reload token')
+        fb.textbox(value='^.probe', lbl='Probe')
+        fb.numberTextBox(value='^.counter', lbl='Fired', readOnly=True)
+        pane.contentPane(border='1px solid silver', padding='5px',
+                         margin='5px').remote(self.remoteInner, token='^.token')
+
+    @public_method
+    def remoteInner(self, pane, token=None, **kwargs):
+        pane.div('Remote content for token: %s' % (token or ''))
+        pane.dataController("SET .counter = (GET .counter || 0) + 1;", _fired='^.probe')
+
+    def test_3_sibling_isolation(self, pane):
+        """Rebuild box A many times, then type in both inputs: each counter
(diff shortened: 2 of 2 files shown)


genro (approved):
Approving the code. The fix is right and the guard is precise: on a same-bag rebuild the content is kept rather than discarded, so unsubscribing it leaves `data*` nodes that are already stripped and will never resubscribe. `kw.oldvalue !== kw.node._value` is exactly the condition that separates the two cases, and the comment says why in the two lines it takes.

The test page is the right shape too — a counter and a formula that must both keep reacting after the box is rebuilt any number of times, which is a property a reader can check by hand rather than a description of one.

**Two things I would not want to lose when this merges, and both are about #1206 rather than about this diff.**

**The event-path memo is a real defect and this PR does not touch it.** I verified it on `develop`: `getTriggerReason` memoizes `kw._eventpath` from `kw.pathlist` (`gnrdomsource.js:132-133`), and `onNodeTrigger` replaces `kw.pathlist` **on that same `kw`** while walking up the hierarchy (`gnrbag.js:2000`). A value memoized from a field that is later reassigned is wrong by construction: anything reached at a higher level compares subscriber paths against the shorter path from the level below, and the match fails silently.

Your own issue says it — "a latent defect independent of this symptom and should be fixed on its own" — and carries the one-line fix, `delete kw._eventpath` next to the reassignment. But `Fixes #1206` closes the only place where it is written down, and the blast radius you state there is much wider than `documentFrame`: any `^attr` subscription on an event crossing more than one bag level. Please open it as its own issue before this merges, so the closure does not take it with it.

**The production symptom is not established as fixed.** #1206 says plainly that it was not reproduced locally and lists the experiment that would settle it — the same `document_frame` page on `b4ba4d799^` versus today's `develop`, then instrumenting whether the `#WORKSPACE` subscription is never called or called and rejected. Nothing in this PR reports that experiment having been run.

So this closes a real defect found while looking for the reported one, which is a good outcome, but "fixes the icond documentFrame" and "fixes a bug in the same area" are different claims. If the experiment has been run since, say so in the PR and I have nothing left to ask. If it has not, I would rather #1206 stay open against the symptom and this PR close a new issue describing what it actually fixes — otherwise the field report is marked resolved on a hypothesis.

Neither point is a reason to hold the code: the change is correct on its own terms and the sooner it is on `develop` the better, since every project tracking it since 2026-08-26 is exposed.

cgabriel (approved):
Fine for me in the scope of the related PR/Issue.

PR #1209, opened 2026-08-31
fix(sql): type X columns in the selection output

### LINKED ISSUE #1197: [BUG] X columns returned by app.dbSelect reach the client as raw XML text instead of a Bag
### Problem Description

A column with `dtype='X'` that reaches the client through a **dbSelect** is delivered as a raw XML string instead of a Bag, because `out_selection` does not apply the `::X` type suffix that the grid path applies.

The same data on the grid path is typed: `gridSelectionData` rewrites every `X` column as `"%s::X" % v` before serialization (`gnrpy/gnr/web/gnrwebpage_proxy/apphandler/misc.py:519`), and the client rebuilds a `GnrBag` out of it (`gnrjs/gnr_d11/js/gnrlang.js:1023`). `out_selection` (`gnrpy/gnr/sql/gnrsqldata/selection.py:1272`) — which is used **only** by `app.dbSelect` — stores the raw column values as node attributes with no typing, so the client receives the serialized XML text of the Bag.

Consequence: **`defaultFrom` on an `X` column is always broken client-side**, in any project. The framework itself wires the failure up:

1. For a column declared as `tbl.column('mybag', dtype='X', defaultFrom='@fkey_id.mybag')`, `_selected_defaultFrom` (`gnrpy/gnr/web/gnrwebstruct/_helpers.py:28`, called from `gnrpy/gnr/web/gnrwebstruct/dojo11.py:1040`) automatically adds `selected_mybag='.mybag'` to the dbSelect built for `fkey_id`. No application code asks for it.
2. The client adds `mybag` to the dbSelect `hiddenColumns` (`gnrjs/gnr_d11/js/genro_widgets.js:4805`).
3. The server returns it untyped (the bug).
4. `_updateSelect` writes that **string** into `.mybag` in the record (`gnrjs/gnr_d11/js/genro_widgets.js:4468`).

From there every consumer that expects a Bag gets a string. Server-side the same `defaultFrom` works correctly: `extendDefaultValues` converts with `Bag(v)` (`gnrpy/gnr/sql/gnrsqltable/record.py:594`).

**Observed failure with `templateChunk`.** When the target column feeds a `templateChunk(template='^.mybag')`, the JS `loadTemplate` checks `kw.template instanceof gnr.GnrBag`; with a string the check fails, so the value is treated as a *template name* and concatenated into an address (`gnrjs/gnr_d11/js/genro_components.js:3898`):

```
template_address = table + ':' + <the whole serialized XML>
```

The server then raises, because `loadTemplate` splits on `':'` with no `maxsplit` (`gnrpy/gnr/web/gnrwebpage.py:869`):

```
File "gnrpy/gnr/web/gnrwebpage.py", line 869, in loadTemplate
    segments,pkey = template_address.split(':')
ValueError: too many values to unpack (expected 2)

>>> template_address
":<?xml version='1.0' encoding='UTF-8'?>\n<GenRoBag><content>..."
```

**Expected:** an `X` column returned by `app.dbSelect` should reach the client as a Bag, exactly as it does on the grid path.

### Reproduction Steps

In a sandbox instance:

1. Add two tables, `tpl` and `doc`:
   - `tpl`: `tbl.column('tplbag', dtype='X')`
   - `doc`: `tbl.column('tpl_id').relation('tpl.id')` and `tbl.column('tplbag', dtype='X', defaultFrom='@tpl_id.tplbag')`
2. Create a `tpl` record whose `tplbag` holds a non-empty Bag (any nested content, so the serialized XML contains at least two `:` characters — an `http://` link or an inline `style="color:red"` is enough).
3. In the `doc` form, add `fb.field('tpl_id')` and a `templateChunk(template='^.tplbag', table='<some table>')`.
4. Open a `doc` record and pick a template in the `tpl_id` dbselect.

Observed result: `ValueError: too many values to unpack (expected 2)` from `gnrwebpage.loadTemplate`.

Even without a `templateChunk`, the defect is visible on its own: after selecting in the dbselect, `#FORM.record.tplbag` holds a plain string rather than a `GnrBag` (inspectable from the dev tools data view).

Not applicable — the reproduction above is generic. Originally found in a `gnrcommunication`/`dem` campaign form, but no application code is involved in the failing path: the `selected_*` attribute that carries the `X` column to the client is generated by the framework from the model's `defaultFrom`.

### Environment

genropy 26.08.26, `develop` @ 8f1f91830, dojo 11 client, macOS 15 / Chrome.

### Additional Material

Suggested fix, in two independent parts:

1. **Root cause.** Apply the `X` typing in `out_selection` (`gnrpy/gnr/sql/gnrsqldata/selection.py:1272`), or in `app.dbSelect` right after `selection.output('selection', ...)` (`gnrpy/gnr/web/gnrwebpage_proxy/apphandler/db_select.py:189`), mirroring what `gridSelectionData` already does with `selection.colAttrs[k]['dataType'] == 'X'`. `out_selection` has no other caller, so the blast radius is limited to the dbSelect path. Worth factoring the typing into a single shared helper instead of duplicating the loop a second time.

2. **Diagnosability.** `loadTemplate` and `saveTemplate` (`gnrpy/gnr/web/gnrwebpage.py:869` and `:906`) should use `split(':', 1)`. That does not fix anything by itself, but it turns an opaque `ValueError` into a normal "missing template" path instead of a traceback.

fporcari:
## Problem

A column with `dtype='X'` returned to the client through **app.dbSelect** arrives as a raw XML string instead of a Bag. Every `defaultFrom` on an X column is therefore broken client-side, in any project: `_selected_defaultFrom` wires the column into the dbSelect automatically, so no application code is involved. When the target column feeds a `templateChunk`, the string fails the `instanceof gnr.GnrBag` check, is taken for a template name and concatenated into a template address, and `loadTemplate` dies with `ValueError: too many values to unpack (expected 2)`.

## Root cause

`out_selection` (`gnrpy/gnr/sql/gnrsqldata/selection.py`) — the only output mode `app.dbSelect` uses — stored the row values as node attributes with no type suffix. Node attributes reach the client as plain strings, and without the `::` suffix `gnrbag.js` does not convert them, so the client keeps the serialized XML text.

The grid path did it right but on its own: `gridSelectionData` rewrote every `X` column as `"%s::X" % v` inline.

A real Bag as a node attribute is not an option: RPC serialization runs with `omitUnknownTypes=True` (`gnrpy/gnr/web/gnrwebpage_proxy/rpc.py:83`) and would drop the attribute silently.

## Change

Variant (a), the shared helper:

- `SqlSelection.typedAttributes(attributes)` reads `colAttrs[...]['dataType']` and applies the suffix.
- `out_selection` calls it, and `gridSelectionData` reuses it instead of its own inline loop, so the two paths can no longer drift apart.
- `loadTemplate` and `saveTemplate` (`gnrpy/gnr/web/gnrwebpage.py`) split the template address with `split(':', 1)`. That fixes nothing by itself, but an unexpected address now takes the normal "missing template" path instead of raising.

Rejected alternative: typing inside `dbSelect` right after `output('selection')` — it stays in the web layer, but leaves the duplication in place and rewrites nodes that were just built.

## Verification

New `gnrpy/tests/sql/test_selection_x_typing.py`, on the real `db_sqlite` fixture and the `invc.product.details` X column of `test_invoice`: a product is inserted with a Bag whose serialization holds more than one `:`, and `output('selection')` is checked to carry `details` with the `::X` suffix and to round-trip back into a Bag; a plain text column is checked to be left alone.

- Narrow test: `pytest gnrpy/tests/sql/test_selection_x_typing.py -q` -> 3 passed. Reverted the `out_selection` call locally to confirm it fails without the fix (the attribute came back as the raw `<?xml ...?><GenRoBag>...` text).
- `pytest gnrpy/tests/sql/f_selection_test.py -q -k sqlite` -> 4 passed.
- `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` -> 1017 passed, 59 skipped, 1 failed: `gnrpy/tests/web/gnrasync_test.py::test_wsproxy_push` (TimeoutError), which passes on its own — timing under parallel load, unrelated to this change.
- flake8 clean on all four touched files.

Fixes #1197

### DIFF
--- a/gnrpy/gnr/sql/gnrsqldata/selection.py
+++ b/gnrpy/gnr/sql/gnrsqldata/selection.py
@@ -1227,6 +1227,24 @@ class SqlSelection(object):
             result.addItem(label,content , _pkey=pkey)
         return result
 
+    def typedAttributes(self, attributes):
+        """Add the ``::X`` suffix every ``X`` column needs to reach the client.
+
+        Node attributes travel to the client as plain strings: without the
+        suffix an ``X`` column arrives as its serialized text instead of
+        being rebuilt as a Bag.
+
+        Args:
+            attributes: The row ``dict`` to type, modified in place.
+
+        Returns:
+            dict: The same *attributes*.
+        """
+        for k, v in list(attributes.items()):
+            if v and self.colAttrs.get(k, {}).get('dataType') == 'X':
+                attributes[k] = '%s::X' % v
+        return attributes
+
     def out_selection(self, outsource, recordResolver=False, caption=False):
         """Return a Bag where each item represents a row, keyed by pkey.
 
@@ -1270,7 +1288,8 @@ class SqlSelection(object):
                     rowcaption = None
                 row['caption'] = self.dbtable.recordCaption(row, rowcaption=rowcaption)
             result.addItem('%s' % spkey, content,
-                           _pkey=pkey, _attributes=row, _removeNullAttributes=False)
+                           _pkey=pkey, _attributes=self.typedAttributes(row),
+                           _removeNullAttributes=False)
         return result
 
     def out_grid(self, outsource, recordResolver=True, **kwargs):
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -866,7 +866,7 @@ class GnrWebPage(GnrBaseWebPage):
         missingMessage = missingMessage or '<div class="chunkeditor_emptytemplate">Missing Template</div>'
         dataInfo = dict()
         if ':' in template_address:
-            segments,pkey = template_address.split(':')
+            segments,pkey = template_address.split(':', 1)
             if segments:
                 segments = segments.split('.')
         else:
@@ -905,7 +905,7 @@ class GnrWebPage(GnrBaseWebPage):
         #pkg.table:resource_module
         #pkg.table:resource_module,custom
         if ':' in template_address:
-            segments,pkey = template_address.split(':')
+            segments,pkey = template_address.split(':', 1)
             if segments:
                 segments = segments.split('.')
         else:
--- a/gnrpy/gnr/web/gnrwebpage_proxy/apphandler/misc.py
+++ b/gnrpy/gnr/web/gnrwebpage_proxy/apphandler/misc.py
@@ -513,11 +513,7 @@ class MiscMixin:
                           _relation_value=pkey,
                           _resolver_name='relOneResolver')
             value = None
-            attributes = kw.get('_attributes')
-            colAttrs = selection.colAttrs
-            for k, v in list(attributes.items()):
-                if v and colAttrs.get(k, {}).get('dataType') == 'X':
-                    attributes[k] = "%s::X" % v
+            attributes = selection.typedAttributes(kw.get('_attributes'))
             if attributes and '__value__' in attributes:
                 value = attributes.pop('__value__')
             result.appendNode(row_key, value, **kw)
--- a/gnrpy/tests/sql/test_selection_x_typing.py
+++ b/gnrpy/tests/sql/test_selection_x_typing.py
@@ -1,0 +1,65 @@
+"""Typing of X columns in the selection output (issue #1197).
+
+``out_selection`` feeds app.dbSelect: node attributes travel to the client as
+plain strings, so an X column has to carry the ``::X`` suffix or the client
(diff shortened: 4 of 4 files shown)


genro (approved):
`typedAttributes` is the right seam: it touches only keys whose `dataType` is `X` and a truthy value, so callers without X columns get byte-identical output, and `gridSelectionData` now shares the same code path instead of its inline loop. Test runs on the real sqlite fixture against `invc.product`.

PR #1210, opened 2026-08-31
fix(services): filter service parameters against the factory signature

### LINKED ISSUE #1181: [BUG] sys.service parameters bag is exploded into kwargs unfiltered: one stray parameter makes getService unusable
## Problem Description

`BaseServiceType.addService()` explodes the whole `parameters` bag stored on the `sys.service` record into keyword arguments of the implementation constructor, without filtering it against what that constructor actually accepts:

https://github.com/genropy/genropy/blob/develop/gnrpy/gnr/lib/services/__init__.py#L94

```python
service = service_factory(self.site, **service_conf)
```

As a consequence **a single parameter that the implementation does not declare makes the service permanently unusable**: every `site.getService(<type>, <name>)` raises `TypeError` before the service is even built. There is no way to recover from the UI, because the service configuration form itself needs `getService` on some pages, and saving the record again does not remove the offending key.

Observed behaviour — a `dbadmin`/`postgres` service record whose bag had accumulated three parameters belonging to a *different* service type (`k8s`):

```
TypeError: Service.__init__() got an unexpected keyword argument 'in_cluster'
```

Expected behaviour — an unknown or obsolete parameter in the stored bag should be ignored (ideally with a warning naming it), not turn into a hard failure of the whole service.

### How the foreign parameters get in

The parameters bag is shared by every implementation of every service type. In `buildServiceParameters()`:

https://github.com/genropy/genropy/blob/develop/projects/gnrcore/packages/sys/resources/tables/service/th_service.py#L46

```python
mixinpath = '/'.join(['services', service_type, implementation])
self.mixinComponent('%s:ServiceParameters' % mixinpath, safeMode=True)
if hasattr(self, 'service_parameters'):
    self.service_parameters(pane, datapath='.parameters', ...)
```

every implementation writes its widgets on the same `.record.parameters` datapath, and nothing prunes keys that do not belong to the implementation being edited. A record can therefore be saved carrying another implementation's fields — in the case observed, as empty `_T="NN"` nodes:

```xml
<GenRoBag>
  <dbadmin_host>...</dbadmin_host>
  <dbadmin_port>5432</dbadmin_port>
  <dbadmin_internal_host>...</dbadmin_internal_host>
  <dbadmin_internal_port>5432</dbadmin_internal_port>
  <dbadmin_user>postgres</dbadmin_user>
  <dbadmin_password>...</dbadmin_password>
  <in_cluster _T="NN"></in_cluster>          <!-- k8s field -->
  <kubeconfig_path _T="NN"></kubeconfig_path> <!-- k8s field -->
  <kubedashboard_url _T="NN"></kubedashboard_url> <!-- k8s field -->
  <test_mode _T="B">True</test_mode>
</GenRoBag>
```

The two halves are independent problems: the form should not persist fields foreign to the implementation, and `addService` should not break when it finds them. Fixing only the form still leaves every existing dirty record broken, and leaves a service unusable after a parameter is renamed or removed from an implementation.

## Reproduction Steps

Deterministic, independent of the form (this is the core of the bug):

1. Configure any service from `sys.service` — for example `mail`, `storage` or `shortener` — and check that `site.getService(<type>, <name>)` works.
2. Add one key that the implementation does not declare to the record's `parameters` bag:
   ```sql
   -- illustrative: any extra node in the bag has the same effect
   update sys.sys_service
      set parameters = replace(parameters, '</GenRoBag>', '<bogus_param _T="NN"></bogus_param></GenRoBag>')
    where service_identifier = '<type>_<name>';
   ```
3. Call `site.getService('<type>', '<name>')`.

Observed result: `TypeError: Service.__init__() got an unexpected keyword argument 'bogus_param'`, raised from `addService` at `gnr/lib/services/__init__.py:94`.
Expected result: the service is built, `bogus_param` is ignored and logged.

The same is reached without touching the database by editing service records of two different implementations from the `sys.service` form in the same page session, then reopening the second one — see the section below.

The database-level repro above is fully deterministic. The *form* side — how the foreign keys got into the bag in the first place — was observed but the exact click sequence was not isolated, so it is reported as a clue rather than as a recipe:

- Module/feature involved: `sys.service` form (`projects/gnrcore/packages/sys/resources/tables/service/th_service.py`), `gnr/lib/services/__init__.py`.
- The k8s and dbadmin service records of a `gnrdeploy` instance were both edited in the same session to tick a newly added `test_mode` checkbox. Afterwards the dbadmin record carried the three k8s fields as null nodes, while its own fields and labels were intact (the `test_mode` node kept the dbadmin `_valuelabel`, so the dbadmin form itself had rendered correctly).
- Possibly relevant: `mixinComponent(..., safeMode=True)` installs `service_parameters` on the **page instance**, and `buildServiceParameters` is a `remote()` fired on `^#FORM.controller.loaded`; nothing resets that attribute between two records opened in the same page instance, and `hasattr(self,'service_parameters')` is already true the second time.

Full traceback of the failure as it reaches the application:

```
File ".../packages/deploy/model/deployment.py", line 202, in get_k8s_definition
    db_service = self.db.application.site.getService(*record['database_service'].split("_"))
File ".../gnr/web/gnrwsgisite.py", line 719, in getService
    return self.services_handler.getService(service_type=service_type, ...)
File ".../gnr/lib/services/__init__.py", line 74, in getService
    return self(service_type)(service_name, **kwargs)
File ".../gnr/lib/services/__init__.py", line 237, in __call__
    service = self.addService(service_name, **kwargs)
File ".../gnr/lib/services/__init__.py", line 94, in addService
    service = service_factory(self.site, **service_conf)
TypeError: Service.__init__() got an unexpected keyword argument 'in_cluster'
```

## Environment

genropy develop @ 68ad40424 (2026-08-26), Python 3.12.9, PostgreSQL 16, macOS (Darwin 25.1.0), local `gnrdeploy` instance

## Additional Material

Suggested direction for the `addService` half — pass only what the factory accepts, leave factories that declare `**kwargs` untouched, and make the discarded keys visible instead of silent:

```python
service_conf = service_conf or {}
sig = inspect.signature(service_factory.__init__)
if not any(p.kind is p.VAR_KEYWORD for p in sig.parameters.values()):
    accepted = set(sig.parameters) - {'self'}
    ignored = sorted(k for k in service_conf if k not in accepted)
    if ignored:
        logger.warning("service %s/%s: ignoring unknown parameters %s",
                       self.service_type, service_name, ", ".join(ignored))
        service_conf = {k: v for k, v in service_conf.items() if k in accepted}
service = service_factory(self.site, **service_conf)
```

For the form half, `buildServiceParameters` could drop from `.record.parameters` the keys that the mixed-in `ServiceParameters` did not declare, so a record never persists another implementation's fields.

Encountered while testing softwellsrl/gnrdeploy#23, which adds a `test_mode` checkbox to two service implementations and therefore invites editing several service records in a row.

fporcari:
## Problem

The `parameters` bag of a `sys.service` record is shared by every implementation of every service type, so it can accumulate keys belonging to a different implementation (or keys that an implementation later renamed or dropped). Every such key was exploded into the implementation constructor as a keyword argument, and most implementations do not declare `**kwargs`: a single stray key made the service permanently unusable, with `TypeError: Service.__init__() got an unexpected keyword argument 'in_cluster'` raised before the instance was even built. There was no way out from the UI, because the service configuration form itself needs `getService` on some pages.

## Root cause

`gnrpy/gnr/lib/services/__init__.py`, `BaseServiceType.addService`:

```python
service_conf = service_conf or {}
service = service_factory(self.site, **service_conf)
```

The stored configuration was passed through unfiltered, with no check against what the resolved factory actually accepts.

## Change

Server-side filter only. `addService` now passes the configuration through `filterServiceConf`, which inspects `inspect.signature(service_factory)`:

- a factory declaring `**kwargs` receives everything, untouched;
- otherwise, only the parameters it declares are passed, and the discarded keys are named in a `logger.warning` following the idiom already used in `_buildImplementations`.

Deliberately **not** in scope:

- pruning the bag in the `buildServiceParameters` meta form (`projects/gnrcore/packages/sys/resources/tables/service/th_service.py:46`). The server-side filter is the real fix: it also repairs every record already dirty in the database, and it covers the case of a parameter renamed or removed from an implementation, which no form-side prune would. The UI prune remains a separate, optional improvement.
- the related form defect at `th_service.py:49` — `mixinComponent(..., safeMode=True)` sets `service_parameters` on the page instance, so `hasattr(self, 'service_parameters')` stays true when a second record whose implementation has no `ServiceParameters` component is opened in the same page. It gets its own issue.

## Verification

- `gnrpy/tests/core/services_addservice_test.py` extended with two tests over real repository implementations, no mocking of the stack:
  - `authentication/dummy` (`Main.__init__(self, parent=None, url=None, template=None, case=None)`, no `**kwargs`) built with a stray `bogus_param`: the service is created, the parameter is dropped and named in the log. This test fails with `TypeError: Main.__init__() got an unexpected keyword argument 'bogus_param'` when the filter is neutralized.
  - `storage/symbolic` (declares `**kwargs`) built with an extra parameter: nothing is filtered, nothing is logged.
- `pytest gnrpy/tests/core/services_addservice_test.py -q` → 4 passed.
- `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` → 1020 passed, 59 skipped.
- Full suite including `gnrpy/tests/sql` → 2177 passed, 69 skipped.
- `flake8` on both touched files → zero errors.

Fixes #1181

### DIFF
--- a/gnrpy/gnr/lib/services/__init__.py
+++ b/gnrpy/gnr/lib/services/__init__.py
@@ -24,6 +24,7 @@
 #Copyright (c) 2007 Softwell. All rights reserved.
 
 
+import inspect
 import os
 import threading
 from datetime import datetime
@@ -100,7 +101,8 @@ class BaseServiceType(object):
         if service_factory is None:
             raise GnrException('no implementation %r for service type %r (service %r)'
                                % (implementation, self.service_type, service_name))
-        service_conf = service_conf or {}
+        service_conf = self.filterServiceConf(service_factory, service_conf or {},
+                                              service_name=service_name, implementation=implementation)
         service = service_factory(self.site, **service_conf)
         service.service_name = service_name
         service.service_type = self.service_type
@@ -109,6 +111,21 @@ class BaseServiceType(object):
         service._config_from_db = db_conf is not None
         self.service_instances[service_name] = service
         return service
+
+    def filterServiceConf(self, service_factory, service_conf, service_name=None, implementation=None):
+        # The parameters bag is shared by every implementation of every service type,
+        # so it can carry keys a given factory never declared (see issue #1181).
+        try:
+            parameters = inspect.signature(service_factory).parameters
+        except (TypeError, ValueError):
+            return service_conf
+        if any(p.kind is p.VAR_KEYWORD for p in parameters.values()):
+            return service_conf
+        unknown = sorted(k for k in service_conf if k not in parameters)
+        if unknown:
+            logger.warning("Service %s/%s: implementation %s does not accept parameters %s: ignored",
+                           self.service_type, service_name, implementation, ', '.join(unknown))
+        return {k: v for k, v in service_conf.items() if k in parameters}
 
     def getConfiguration(self, service_name):
         return self.getServiceConfigurationFromDb(service_name) or \
--- a/gnrpy/tests/core/services_addservice_test.py
+++ b/gnrpy/tests/core/services_addservice_test.py
@@ -1,4 +1,5 @@
-"""Regression tests for BaseServiceType.addService factory resolution.
+"""Regression tests for BaseServiceType.addService factory resolution and
+parameter filtering.
 
 A missing service implementation used to crash with a bare
 ``TypeError: 'NoneType' object is not callable`` at the factory call site
@@ -7,11 +8,18 @@ error a cold-start race on ``site.resources_dirs`` produced in production
 (see issue #984). ``addService`` must instead raise a clear ``GnrException``
 naming the implementation and the service type.
 
+A stray key in the stored ``parameters`` bag used to crash the same call site
+with ``TypeError: got an unexpected keyword argument`` for every implementation
+not declaring ``**kwargs``, making the service permanently unusable
+(see issue #1181). ``addService`` must drop what the factory cannot accept and
+log it, while factories declaring ``**kwargs`` keep receiving everything.
+
 The harness exercises the real ``BaseServiceType`` and the real
 ``ResourceLoader.getResourceList`` over the repository's own ``resources/common``
 tree — no site instance, no register daemon needed.
 """
 
+import logging
 import os
 from types import SimpleNamespace
 
@@ -54,3 +62,24 @@ def test_add_service_missing_implementation_raises_clear_error():
     handler = BaseServiceType(site=make_site(), service_type='faketype')
     with pytest.raises(GnrException, match='faketype'):
         handler.addService('x', implementation='any')
+
+
+def test_add_service_ignores_parameters_the_factory_does_not_accept(caplog):
+    """authentication/dummy declares no **kwargs: a foreign key must not break it."""
+    handler = BaseServiceType(site=make_site(), service_type='authentication')
+    with caplog.at_level(logging.WARNING, logger='gnr.lib'):
+        service = handler.addService('auth', implementation='dummy',
+                                     url='www.dummy.com', bogus_param=None)
+    assert service is not None
+    assert service.url == 'www.dummy.com'
+    assert not hasattr(service, 'bogus_param')
+    assert 'bogus_param' in caplog.text
+
+
+def test_add_service_keeps_every_parameter_for_var_keyword_factory(caplog):
+    """storage/symbolic declares **kwargs: nothing may be filtered out."""
+    handler = BaseServiceType(site=make_site(), service_type='storage')
+    with caplog.at_level(logging.WARNING, logger='gnr.lib'):
+        service = handler.addService('sym', implementation='symbolic', extra_param=7)
+    assert service is not None
+    assert 'extra_param' not in caplog.text


genro (approved):
Server-side filtering is the right answer to #1181: it repairs every record already dirty in the database and it covers a parameter renamed or dropped from an implementation, which no form-side prune would. `filterServiceConf` gets the details right — full passthrough when the factory declares `**kwargs`, `TypeError`/`ValueError` from `inspect.signature` falls back to no filtering, and the discarded keys are named in the warning. Tests build real implementations (`authentication/dummy` without `**kwargs`, `storage/symbolic` with it) through the real `BaseServiceType`, and the caplog is on the right logger. Merges cleanly with #1236 on the shared test file.

One non-blocking note on the breadth of the filter. #1181 is about the `parameters` bag stored on `sys.service`, but the filter is applied to `service_conf` after the whole `kwargs or db_conf or siteconfig or self` chain (:93-95), so it also covers a hand-written `siteconfig.xml` and programmatic `addService(name, **kw)` calls. On those two sources a misspelled *required* parameter used to fail loudly with `TypeError` at the first `getService`; now it produces a warning and the service is built with the default. Storage `local` with `base_paht` instead of `base_path` comes up as a service that exists with `base_path=None` and fails on first use, with the cause only in the log. Restricting the filter to the `db_conf` branch would keep the fix for the case the issue describes and leave a typo in a hand-written config loud. Your call — the current breadth is defensible, it is just wider than the issue and the body does not distinguish the sources.

fporcari:
On the breadth of the filter ([review](https://github.com/genropy/genropy/pull/1210#pullrequestreview-5098026614)):

> Restricting the filter to the `db_conf` branch would keep the fix for the case the issue describes and leave a typo in a hand-written config loud. Your call — the current breadth is defensible, it is just wider than the issue and the body does not distinguish the sources.

Merged with the current breadth, deliberately. The reason is the repair, not the
symmetry: the records already stored in `sys.service` are the ones nobody will go
back and clean, and a parameter renamed or dropped from an implementation keeps
producing them. A filter that only ran on `db_conf` would cover exactly that —
but the same `addService` call is also how a project wires a service in code, and
splitting the behaviour there needs the source threaded through
`filterServiceConf` plus tests on all four branches of the chain at `:93-96`.

Your point about the hand-written sources stands and is not lost: an unknown key
from `siteconfig.xml` or from explicit `addService` kwargs should raise, not warn,
and `getServiceConfigurationFromSelf` needs its own look before it is grouped with
either. That is #1241, with the `base_paht` case written down as the example.

PR #1211, opened 2026-08-31
fix(gnrjs): cache a translation only when its status is OK

### LINKED ISSUE #1177: [BUG] Failed translations are cached client-side as valid, and the unlocalized marker is dead code
## Problem Description

On an `it-IT` instance, the delete/unlink confirmation dialog can show up in English even though the Italian translations exist in `localization.xml`.

Reported by an end user of an Italian instance: deleting a record from a picker grid raises a confirmation dialog written in English. The report covers two different grids, so it is not resource-specific.

While investigating I found two distinct defects in the localization path. The first is certainly a bug; the second is the most likely explanation of the user report, but it has not been reproduced in a browser yet.

### 1. Failed translations are cached as if they were valid, and the diagnostic marker is dead code

`gnrjs/gnr_d11/js/gnrlang.js:57` — in `_T()`:

```js
var localizedString = result['translation'];
if(result.status!='OK'){
    localsdict[str] ='<span class="unlocalized">'+localizedString+'</span>';
}
localsdict[str] = localizedString;
genro.setInStorage('local',localekey,localsdict);
```

The first assignment is immediately overwritten by the second. The `unlocalized` span never reaches the DOM, so a missing translation is visually indistinguishable from a real one and nobody ever notices it needs writing.

Worse, the value is written to `localsdict` and persisted to local storage **regardless of `result.status`**. A `NOKEY` or `NOLANG` answer — i.e. the English fallback — gets cached in the browser exactly like a successful translation.

### 2. The client-side translation dictionary is never invalidated

Still in `_T()`, `localsdict_<language>` is read from local storage and a key is only ever resolved on a miss. There is no version, no expiry, no invalidation hook.

Consequence: once a string is cached, adding or fixing its translation in `localization.xml` has no effect for any user who already visited that page. The English fallback sticks in that browser indefinitely.

Combined with defect 1, this is self-reinforcing: an English fallback served once (locale not yet configured, translation not yet written, a `NOKEY` during development) is cached permanently and silently.

### 3. Minor, related: one string in that same dialog has no Italian at all

`localization.xml:249`:

```xml
<en_records_to_delete base="Records to delete"></en_records_to_delete>
```

No `it`, and no `en` either — only `base`. It is the label of the confirmation-count field rendered by `deleteAsk` when more than one row is selected (`gnrjs/gnr_d11/js/genro_components.js:6810`), so on a multi-row delete an Italian user gets an English label next to Italian text.

All the other strings of that dialog do have `it`: `delete_one`, `delete_many`, `delete_logical_one`, `delete_logical_many`, `unlink_one`, `unlink_many`, plus `Cancel` and `Confirm`.

## Reproduction Steps

Defect 3, deterministic:

1. Italian instance (`client_locale='it-IT'`).
2. Open any TableHandler grid with `delrow` enabled.
3. Select **two or more** rows and delete.
4. Observed: the dialog body is in Italian but the count field is labelled `Records to delete`.

Defect 1, deterministic by inspection: any `_T()` call whose `getRemoteTranslation` returns `status != 'OK'` caches the fallback and drops the `unlocalized` marker. The two lines contradict each other, so the branch is unreachable by construction.

Defect 2, to reproduce the stale cache:

1. On an Italian instance, visit a page containing a string with no `it` translation, so the English fallback is cached.
2. Add the `it` translation to `localization.xml` and restart.
3. Reload the page in the **same browser**.
4. Observed: still English. Expected: the new translation. Clearing local storage fixes it, which is the confirmation.

Origin of the report: an Italian production instance, `client_locale='it-IT'` and `server_locale='it-IT'` set in `instanceconfig.xml`. The user sees the confirmation in English when removing a row from two different picker grids (`plainTableHandler` with `picker=`, so the `unlink_*` branch of `deleteAsk`).

The Italian translations for those `unlink_*` keys are present and correct in `localization.xml`, which is why I believe the cause is the stale client-side dictionary (defect 2) rather than a missing translation — the user has been on that instance for years, so a dictionary cached in an earlier state is plausible.

Not yet verified in a browser: whether clearing that user's local storage restores the Italian. Worth doing before any fix, since it would confirm or kill the hypothesis.

Modules involved:

- `gnrjs/gnr_d11/js/gnrlang.js` — `_T()`
- `gnrjs/gnr_d11/js/genro_components.js` — `gnr.stores._Collection.messages`, `deleteAsk()`
- `gnrpy/gnr/app/gnrlocalization.py` — `AppLocalizer.getTranslation()`, which correctly returns `status` `NOKEY`/`NOLANG`; it is the client that throws it away

## Environment

genropy `develop` @ `0acf37fa3` (release-26.08.19.1-126), reported on an it-IT instance, browser not yet identified

## Additional Material

Worth deciding together, beyond the three fixes:

- whether the `unlocalized` marker should be restored as a visible signal (it was clearly intended) or whether missing translations should be reported some other way, e.g. logged server-side on `NOKEY`
- whether a fallback should be cached at all: not caching it would cost one server call per miss but would let a translation added later take effect without touching the browser

fporcari:
## Problem

On an `it-IT` instance the delete/unlink confirmation dialog can show up in English even though the Italian translations exist in `localization.xml`, and clearing local storage is the only thing that fixes it.

Two separate defects behind it: a failed translation is cached client-side as if it were valid, and the `unlocalized` marker meant to signal it is unreachable code.

## Root cause

`gnrjs/gnr_d11/js/gnrlang.js`, in `_T()`:

```js
var localizedString = result['translation'];
if(result.status!='OK'){
    localsdict[str] ='<span class="unlocalized">'+localizedString+'</span>';
}
localsdict[str] = localizedString;
genro.setInStorage('local',localekey,localsdict);
```

The second assignment overwrites the first unconditionally, so:

- the `unlocalized` marker never reaches the DOM — and no CSS ever defined the class (`grep -rn unlocalized` over the repo now returns nothing but this branch's removal), so it was dead in two ways;
- the English fallback of a `NOKEY`/`NOLANG` answer is persisted to `localsdict_<lang>` exactly like a real translation. That dictionary has no version and no expiry, so the fallback stays in that browser forever and a translation added to `localization.xml` later never reaches a user who already visited the page.

The server already distinguishes the cases: `AppLocalizer.getTranslation()` (`gnrpy/gnr/app/gnrlocalization.py`) returns `status` `OK` / `NOLANG` / `NOKEY`. It was only the client throwing it away.

## Change

Minimal variant, deliberately:

1. `_T()` persists to local storage **only** when `status == 'OK'`. A non-OK answer goes into a page-level memo, `genro._pageLocals`, keyed by the same `localsdict_<lang>` key — one server call per string per page, exactly as before, and the next page load picks up a translation written meanwhile. The memo is declared in the `gnr.GenroClient` constructor next to `this._debugPaths` / `this._aliasDatapaths` (`gnrjs/gnr_d11/js/genro.js`), the existing idiom for page-scoped dicts.
2. The dead `unlocalized` line is removed rather than resurrected as HTML: no CSS backs it, and it would land as literal markup inside labels and captions. In its place, `getRemoteTranslation()` (`gnrpy/gnr/web/gnrwebpage.py`) logs a `debug` line on a non-OK status, where the status is actually known.
3. `en_records_to_delete` in `localization.xml` gains its missing `it="Record da eliminare"` — the count label of the multi-row delete confirmation, which had only `base`.

### Not in scope

A version token on the client dictionary keyed to `gnr.vendoredMtime` (`gnrwebpage.py:2421`), which would also expire the translations already cached wrong in existing browsers. That is a separate improvement; the reported defect is fully fixed by never persisting a non-OK status, and existing bad entries still need one Clear LS.

## Verification

- `pytest gnrpy/tests/app/gnrlocalization_test.py -q` → **4 passed**. Adds `test_missing_language_is_reported_as_nolang`, which pins the server contract the client now gates on: a key present but missing the requested language answers `NOLANG` with the English fallback, both in the plain-text and in the `GnrLocString` branch, while a real translation answers `OK`.
- `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql` → **1014 passed, 59 skipped, 5 failed**. The 5 are `gnrpy/tests/web/gnrasync_test.py` failing on `OSError: [Errno 48] Address already in use`; re-run alone that file is **7 passed**. `gnrpy/tests/sql` was excluded because the postgres fixture could not `initdb` in this environment (`could not create shared memory segment`), unrelated to this change — no SQL code is touched.
- `flake8` on `gnrpy/gnr/web/gnrwebpage.py` and `gnrpy/tests/app/gnrlocalization_test.py` → zero errors.
- **There is no JS test runner in this repo, so the `gnrlang.js` change has no automated test here.** It was verified by extracting `_T()` and running it under node against a stubbed `genro` (storage, `serverCall`, `locale`), asserting: an `OK` answer is written to storage and served from it on the second call; a `NOLANG` or `NOKEY` answer is never written to storage, is served from the page memo on the second call (still one server call), and carries no `<span>`; once the server starts answering `OK` for that same string, a fresh page gets the new translation and persists it; the blank and `lazy` short-circuits are unchanged. `node --check` passes on both edited JS files. Not verified in a browser on a live it-IT instance.

Fixes #1177

### DIFF
--- a/gnrjs/gnr_d11/js/genro.js
+++ b/gnrjs/gnr_d11/js/genro.js
@@ -106,6 +106,7 @@ dojo.declare('gnr.GenroClient', null, {
         this.lastPing = start_ts;
         this._debugPaths = {};
         this._aliasDatapaths = {};
+        this._pageLocals = {};
         this.sendAllEvents=false;
         this._lastMouseEvent = {};
         this._longClickDuration = 1500;
--- a/gnrjs/gnr_d11/js/gnrlang.js
+++ b/gnrjs/gnr_d11/js/gnrlang.js
@@ -66,20 +66,23 @@ function _T(str,lazy){
         return str;
     }
     var localsdict = genro.getFromStorage('local',localekey) || {};
-    if(isNullOrBlank(localsdict[str])){
-        var toTranslate = noLocMarker?'!!'+str:str;
-        var result = genro.serverCall('getRemoteTranslation',{txt:toTranslate,language:language}) || {};
-        var localizedString = result['translation'];
-        if(result.status!='OK'){
-            localsdict[str] ='<span class="unlocalized">'+localizedString+'</span>';
-        }
-        localsdict[str] = localizedString;
-        genro.setInStorage('local',localekey,localsdict);
-        return localizedString;
-    }else{
+    if(!isNullOrBlank(localsdict[str])){
         return localsdict[str];
     }
-    return str;
+    var pagelocals = genro._pageLocals[localekey] = genro._pageLocals[localekey] || {};
+    if(!isNullOrBlank(pagelocals[str])){
+        return pagelocals[str];
+    }
+    var toTranslate = noLocMarker?'!!'+str:str;
+    var result = genro.serverCall('getRemoteTranslation',{txt:toTranslate,language:language}) || {};
+    var localizedString = result['translation'];
+    if(result.status=='OK'){
+        localsdict[str] = localizedString;
+        genro.setInStorage('local',localekey,localsdict);
+    }else{
+        pagelocals[str] = localizedString;
+    }
+    return localizedString;
 }
 
 function _F(val,format,dtype){
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -1186,7 +1186,11 @@ class GnrWebPage(GnrBaseWebPage):
 
     @public_method
     def getRemoteTranslation(self, txt=None,language=None,**kwargs):
-        return self.localizer.getTranslation(txt,language=language or self.locale)
+        language = language or self.locale
+        result = self.localizer.getTranslation(txt,language=language)
+        if result['status'] != 'OK':
+            logger.debug("Missing translation (%s) for %s in %s", result['status'], txt, language)
+        return result
 
     def localize(self, txt, language=None,**kwargs):
         return self.localizer.translate(txt,language=language or self.locale)
--- a/gnrpy/tests/app/gnrlocalization_test.py
+++ b/gnrpy/tests/app/gnrlocalization_test.py
@@ -71,6 +71,27 @@ class TestGnrLocalization(BaseGnrAppTest):
         assert r["status"] == "NOKEY"
         assert r["translation"] == "bkasjklasjsd"
 
+    def test_missing_language_is_reported_as_nolang(self):
+        """
+        A fallback answer must stay distinguishable from a real translation:
+        the client caches a translation only when the status is OK.
+        """
+        al = gl.AppLocalizer(self.app)
+        al.localizationDict['en_records_to_delete'] = {'base': 'Records to delete'}
+        al.localizationDict['en_condition_op'] = {'base': 'Condition op', 'it': 'Operatore condizione'}
+
+        r = al.getTranslation('!!Records to delete', 'it')
+        assert r['status'] == 'NOLANG'
+        assert r['translation'] == 'Records to delete'
+
+        r = al.getTranslation('!!Condition op', 'it')
+        assert r['status'] == 'OK'
+        assert r['translation'] == 'Operatore condizione'
+
+        r = al.getTranslation(gl.GnrLocString('en_records_to_delete'), 'it')
+        assert r['status'] == 'NOLANG'
+        assert r['translation'] == 'Records to delete'
+
     def test_symbol_only_captions_not_translated(self):
         """
         Symbol-only captions ('=', '>=', '%', '#', whitespace, ...) all flatten
--- a/localization.xml
+++ b/localization.xml
@@ -246,7 +246,7 @@
 <en_you_are_about_to_set_archiviation_date_in_the_selected_count_records base="You are about to set archiviation date in the selected $count records" it="Stai per archiviare, impostando la data di archiviazione a $count records"></en_you_are_about_to_set_archiviation_date_in_the_selected_count_records>
 <en_date fr="Date" de="Datum" en="Date" it="Data" base="Date"></en_date>
 <en_n_records base="N.Records"></en_n_records>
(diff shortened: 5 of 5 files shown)


genro (approved):
Good diagnosis, and the right call on the marker. The second `localsdict[str] = localizedString` overwrote the marker branch unconditionally, so a `NOKEY`/`NOLANG` English fallback was persisted exactly like a real translation into a dictionary with no version and no expiry — a translation added later never reached that browser again. Removing the dead `unlocalized` line instead of resurrecting it is correct: no CSS defines the class and it would have landed as literal markup inside labels and captions. Moving the report to `getRemoteTranslation`, where the status is actually known, is the better place for it. `_pageLocals` is declared next to `_debugPaths` and `_aliasDatapaths`, matching the existing idiom, and the `localization.xml:249` entry is wired: `genro_components.js:6835` calls `lbl:_T('Records to delete')` with no `!!`, so the key resolves to `en_records_to_delete`.

One note on cost, non-blocking. `_T` calls `genro.serverCall` with no callback, so `funcCreate(undefined)` returns undefined, and `remoteCall` takes the no-callback branch and forces `sync = true` (`genro_rpc.js:478-482`). That is a blocking XHR on the UI thread. What the change moves is when it is paid:

- before: one call per untranslated string per browser, because the non-OK answer was persisted;
- after: one call per untranslated string per page load.

For scale, the root `localization.xml` holds 937 `en_` keys of which 117 carry no `it=`, there are eleven more localization files, and the d11 JS has 134 distinct `_T('...')` literals. On an Italian instance every one of those strings a page touches costs a blocking round trip on each load.

The body says "one server call per string per page, exactly as before". That holds inside a single page thanks to the memo, but not across loads, where it used to be once per browser. Worth correcting, because that is the sentence a reader uses to judge the cost.

The trade itself is defensible and I am not asking you to change it: the cost falls only on instances with missing translations and it disappears as soon as the translation exists. The proper remedy is the one the body already sets out of scope, a version token on the client dictionary keyed to `gnr.vendoredMtime`, which would let non-OK answers be cached too and expire them on deploy. Worth an issue so it does not get lost.

PR #1212, opened 2026-08-31
fix(web): make TableScriptToHtml output file names record and resource specific

### LINKED ISSUE #1205: TableScriptToHtml output filename is not record-specific: prints collide and serve stale files
## Defect

`TableScriptToHtml.outputDocName()` (`gnrpy/gnr/web/gnrbaseclasses.py:760`) builds the output filename as `<table>_<slugified recordCaption>`:

```python
caption = slugify(self.tblobj.recordCaption(self.record))
doc_name = '%s_%s%s' % (self.tblobj.name, caption, ext)
```

The record is loaded with `recordAs(record, virtual_columns=self.virtual_columns)`, and `virtual_columns` is `None` by default. So when a table's `caption_field` is a virtual/formula column, that column is simply not in the record — and `recordCaption()` substitutes the table name for the missing column instead of failing or returning blank (`gnrpy/gnr/sql/gnrsqltable/record.py:677`):

```python
cols = [(c, gnrstring.toText(record.get(c, tblname), locale=self.db.locale)) for c in fields]
```

Result: **every record of that table prints to the same filename**. The print resource identity is not part of the name either, so two different print resources on the same record collide as well.

Because `rpc_callTableScript` writes the file and then reads it back to serve it (`gnrpy/gnr/web/gnrwebpage.py:2669`), two prints in flight for the same page serve each other's file: the client receives a document belonging to another record, or to another print.

## Evidence

Production instance with `erpy_coge.cliente` (`caption_field='caption_cliente'`, a virtual column). Every connection folder holds exactly one file, for all clients and for both print resources:

```
sites/<site>/data/_connections/<conn>/<page>/pdf/cliente_cliente.pdf
```

Observed consequence: in that application two print resources on the same record overwrite each other's output, and the PDF attached to a dunning letter can be built from another print's file, since everything goes through the same path. (A stale-view symptom on the same page turned out to have a separate, client-side cause: the print was not re-run at all. Not this defect.)

## Affected code

Same pattern in all three print engines:

- `gnrpy/gnr/web/gnrbaseclasses.py:760` — `TableScriptToHtml.outputDocName`
- `gnrpy/gnr/web/gnrtablescript.py:85` — `TableScriptOnRecord.outputDocName`
- `gnrpy/gnr/web/gnrtablescript_new.py:94` — `TableScriptOnRecord.outputDocName`

## Proposed fix, in order of blast radius

1. `outputDocName` includes the record pkey and the print class name, keeping the caption for readability.
2. `TableScriptToHtml.__call__` loads the caption columns (from `rowcaptionDecode()`) into `virtual_columns`, so the caption is real data rather than a fallback.
3. `recordCaption` stops substituting the table name for a column missing from the record: a hole should read as blank, not as plausible-looking data. This is the actual defect generator, but it also feeds grids, selections and `db_caption`, so it needs a maintainer decision.

## Note

The output filename is user-visible (downloaded file name when `downloadAs` is not set, files under `print_debug`), so points 1-2 are a behaviour change to agree on before implementing.

Workaround in application code, and the current house idiom: override `outputDocName` in the print resource (as `packages/ciak/.../cost_report/html_res/*.py` in the pansotti project and `common/tables/_default/html_res/print_gridres.py` here already do).

fporcari:
## Problem

`TableScriptToHtml` built its output document name as `<table>_<slugified recordCaption>`. That name is neither record-specific nor resource-specific: on a table whose `caption_field` is a virtual column, every record of the table and every print resource on it wrote to the very same file (`cliente_cliente.pdf` in the reported production instance). Since `rpc_callTableScript` writes the file and then reads it back to serve it, two prints in flight on the same page serve each other's document.

## Root cause

Two independent holes, both in `gnrpy/gnr/web/gnrbaseclasses.py`:

- the record was loaded with `recordAs(record, virtual_columns=self.virtual_columns)`, and `virtual_columns` is `None` by default. When the rowcaption is built on a virtual column, that column is simply absent from the loaded record, and `recordCaption()` substitutes the table name for the missing column — so the caption degenerated to the table name for every record;
- neither the record pkey nor the identity of the print resource took part in the name, so even a correct caption left two resources on the same record colliding.

## Change

- `TableScriptToHtml.captionVirtualColumns()` adds to the declared `virtual_columns` the virtual columns the rowcaption is built on, and `__call__` loads the record with them. Copy-adapt of the same idiom in `gnrpy/gnr/web/gnrwebpage_proxy/apphandler/get_record.py:181`, which already merges caption columns into `virtual_columns` before reading a record.
- `TableScriptToHtml.outputDocIdentifier()` returns the print resource identity (from `_gnrPublicName`) plus the record pkey, a uuid when there is no single record (selection prints, `record='*'`) — following `getExportIdentifier()` a few lines above in the same file. `outputDocName()` appends it, keeping table name, caption and `record_idx` for readability.

Names go from `batch_log_batch-log.html` for everything to, e.g., `batch_log_invcinvoicealpha_print_first_main_<pkey>.html`.

The output name is user-visible (download name when `downloadAs` is not set, files under `print_debug`), so this is a deliberate behaviour change on the file name.

## Not in scope

- `recordCaption()` substituting the table name for a column missing from the record (`gnrpy/gnr/sql/gnrsqltable/record.py:677`) is the actual defect generator, but it has 18 call sites and also feeds grids, selections and `db_caption`: it needs a maintainer decision and an issue of its own.
- `gnrpy/gnr/web/gnrtablescript.py` and `gnrpy/gnr/web/gnrtablescript_new.py` carry the same pattern but are imported nowhere outside their own import-smoke tests, so they are left untouched.

## Verification

- New real tests in `gnrpy/tests/web/gnrbaseclasses_test.py`: a real sqlite instance, real records inserted in `sys.batch_log` (whose `caption_field` is the formula column `log_caption`, exactly the reported shape), rendered through a real `GnrDummySite` page. They check that the caption virtual columns are loaded, that declared `virtual_columns` are kept, that the caption is the record's and not the table name, and that names differ per record, per resource and per pkeyless print. 6 of the 7 fail on the unpatched module.
- `pytest gnrpy/tests/web/gnrbaseclasses_test.py -q`: 7 passed.
- `pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql`: 1025 passed, 59 skipped.
- Full suite: 2182 passed, 69 skipped.
- `flake8` on both touched files: zero errors.

Fixes #1205

### DIFF
--- a/gnrpy/gnr/web/gnrbaseclasses.py
+++ b/gnrpy/gnr/web/gnrbaseclasses.py
@@ -32,7 +32,7 @@ from gnr.core.gnrdict import dictExtract
 from gnr.core.gnrstring import splitAndStrip, slugify, templateReplace
 from gnr.core.gnrlang import GnrObject
 from gnr.core.gnrbag import Bag
-from gnr.core.gnrlang import getUuid
+from gnr.core.gnrlang import getUuid, uniquify
 from gnr.web import logger
 from gnr.web.gnrwebpage_proxy.gnrbaseproxy import GnrBaseProxy
 
@@ -347,7 +347,7 @@ class TableScriptToHtml(BagToHtmlWeb):
         if record=='*':
             record = None
         else:
-            record = self.tblobj.recordAs(record, virtual_columns=self.virtual_columns)
+            record = self.tblobj.recordAs(record, virtual_columns=self.captionVirtualColumns())
         html_folder = self.getHtmlPath(autocreate=True)
         self.locale = locale or self.page.locale
         self.language = language or self.page.language
@@ -370,6 +370,15 @@ class TableScriptToHtml(BagToHtmlWeb):
             return self.getPdfUrl(os.path.basename(self.pdfpath)) if resultAs=='url' else self.pdfpath
             #with open(temp.name,'rb') as f:
             #    result=f.read()
+
+    def captionVirtualColumns(self):
+        """:attr:`virtual_columns` plus the virtual columns the rowcaption is built on,
+        which would otherwise be missing from the loaded record"""
+        virtual_columns = self.virtual_columns.split(',') if self.virtual_columns else []
+        model_virtual_columns = self.tblobj.model.virtual_columns
+        caption_columns = [c.replace('$', '') for c in self.tblobj.rowcaptionDecode()[0]]
+        virtual_columns.extend([c for c in caption_columns if c in model_virtual_columns])
+        return ','.join(uniquify(virtual_columns)) or None
 
     def getDocName(self):
         return os.path.splitext(os.path.basename(self.filepath))[0]
@@ -757,19 +766,32 @@ class TableScriptToHtml(BagToHtmlWeb):
         """
         return self.builder.calcRowsNumber(text, width_mm=width_mm, font_name=font_name, font_size=font_size)
 
+    def outputDocIdentifier(self):
+        """Identity of the print resource and of the printed record, so that concurrent
+        prints never write to the same file"""
+        public_name = getattr(self, '_gnrPublicName', None) or self.__class__.__name__
+        resource_id = slugify(' '.join(public_name.rsplit('.', 2)[-2:]).replace('/', ' '), sep='_')
+        record_id = None
+        if self.record is not None and not self.record.get('selectionPkeys'):
+            record_id = self.record.get(self.tblobj.pkey)
+        record_id = re.sub(r'\W', '_', str(record_id)) if record_id else getUuid()
+        return '%s_%s' % (resource_id, record_id)
+
     def outputDocName(self, ext=''):
-        """TODO
-        :param ext: TODO"""
+        """Return the output file name for the current record
+
+        :param ext: the filename extension"""
         if ext and not ext[0] == '.':
             ext = '.%s' % ext
-        caption = ''
+        chunks = [self.tblobj.name]
         if self.record is not None:
             caption = slugify(self.tblobj.recordCaption(self.record))
-            idx = self.record_idx
-            if idx is not None:
-                caption = '%s_%i' %(caption,idx)
-        doc_name = '%s_%s%s' % (self.tblobj.name, caption, ext)
-        return doc_name
+            if caption:
+                chunks.append(caption)
+            if self.record_idx is not None:
+                chunks.append(str(self.record_idx))
+        chunks.append(self.outputDocIdentifier())
+        return '%s%s' % ('_'.join(chunks), ext)
 
 
 
--- a/gnrpy/gnr/web/gnrbaseclasses.py
+++ b/gnrpy/gnr/web/gnrbaseclasses.py
@@ -347,7 +347,9 @@ class TableScriptToHtml(BagToHtmlWeb):
         if record=='*':
             record = None
         else:
-            record = self.tblobj.recordAs(record, virtual_columns=self.captionVirtualColumns())
+            record = self.tblobj.recordAs(record,
+                                          virtual_columns=self.virtual_columns if isinstance(record, Bag)
+                                          else self.captionVirtualColumns())
         html_folder = self.getHtmlPath(autocreate=True)
         self.locale = locale or self.page.locale
         self.language = language or self.page.language
--- a/gnrpy/tests/web/gnrbaseclasses_test.py
+++ b/gnrpy/tests/web/gnrbaseclasses_test.py
@@ -90,3 +90,21 @@ class TestTableScriptOutputName(BaseGnrTest):
         first = self.buildScript(FirstPrint, '*').filepath
         second = self.buildScript(FirstPrint, '*').filepath
         assert first != second
+
+    def test_a_bag_record_keeps_what_the_caller_merged_in(self):
+        # btcprint and print_template pass a Bag already loaded with the batch's
+        # own virtual_columns and extra keys merged in; asking recordAs for the
+        # caption columns would reload it from the database and drop both
+        record = self.tblobj.record(pkey=self.pkeys[0], mode='bag')
+        record['batch_parameter'] = 'kept'
+        script = self.buildScript(FirstPrint, record)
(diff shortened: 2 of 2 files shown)


genro (changes requested):
The diagnosis is right and `outputDocIdentifier()` is the correct shape: resource identity from `_gnrPublicName` plus the pkey, a uuid when there is no single record, following `getExportIdentifier()` a few lines above. `captionVirtualColumns()` addresses the real generator of the collapsed name, and the tests build real records in `sys.batch_log`, whose formula `caption_field` is exactly the reported shape. Requesting changes for one side effect outside the declared scope.

## Loading the record with caption virtual columns reloads Bag records from the database

`captionVirtualColumns()` returns a non-empty string for every table whose rowcaption is built on a virtual column, and `recordAs` (`gnrpy/gnr/sql/gnrsqltable/record.py:369`) reads:

```python
if mode == 'bag' and (virtual_columns or not isinstance(record, Bag)):
    pkey = record.get('pkey', None) or record.get(self.pkey)
    if pkey:
        record = self.record(pkey=pkey, mode=mode, virtual_columns=virtual_columns)
```

With `virtual_columns` truthy, a Bag record that carries its pkey is reloaded. Before this PR `virtual_columns` was the class default `None` (`gnrbaseclasses.py:322`), so a Bag arrived at `__call__` untouched.

Two in-repo callers pass a Bag:

- `btcprint.do_print_record` (`gnrpy/gnr/web/batch/btcprint.py:97`) passes the records from `get_records()`, loaded with the **batch's** `virtual_columns` (`btcbase.py:318`);
- `resources/common/tables/_default/print/print_template.py:38-46` does `record.update(self.batch_parameters)` and passes that Bag.

After this change `self.record` inside the maker keeps only the caption's virtual columns: the batch's other virtual columns and any key the caller merged into the Bag are gone, plus one extra SELECT per printed record. In this repo the consequence is invisible, because `print_template` precomputes `htmlContent` and passes `filename` explicitly. A project print resource whose `html_res` reads those columns off `self.record` loses them from the PDF with no error and no log line.

The narrow fix is to keep the merge for the pkey case only, where the record has to be loaded anyway:

```python
record = self.tblobj.recordAs(record, virtual_columns=self.captionVirtualColumns()
                              if not isinstance(record, Bag) else self.virtual_columns)
```

Merging the caption columns with those already present in the received Bag would work too, if you prefer the caption to be complete in both cases.

## Minor, your call

`resources/common/tables/_default/html_res/print_gridres.py:162` and `print_gridtemplate.py:63` override `outputDocName` without calling `super()`, so two grid prints sharing a `print_title` still collide. Outside #1205 as scoped, but the PR title promises resource-specific names, so either cover them or say in the body that grid prints are excluded.

Merge order note: this PR conflicts with #1045 on `gnrpy/tests/web/gnrbaseclasses_test.py` only (`gnrbaseclasses.py` merges automatically). #1074 → #1045 → #1212 is the cheapest order.

fporcari:
Fixed in `23c048f1fe`, taking your narrow variant
([review](https://github.com/genropy/genropy/pull/1212#pullrequestreview-5098062755)):

> The narrow fix is to keep the merge for the pkey case only, where the record has to be loaded anyway

`__call__` now asks for the caption's virtual columns only when the record is
not already a Bag:

```python
record = self.tblobj.recordAs(record,
                              virtual_columns=self.virtual_columns if isinstance(record, Bag)
                              else self.captionVirtualColumns())
```

so a Bag reaches `recordAs` with exactly what it had before this PR — the class
default, `None` unless a subclass declares one — and `record.py:369` leaves it
alone. `btcprint.do_print_record` and `print_template` keep the batch's virtual
columns and whatever the caller merged in, and the extra SELECT per printed
record is gone.

I did not take the "merge the caption columns into the received Bag" option. It
would need the caption columns fetched separately and merged, which is the extra
SELECT back again on a path that has no use for it: the only consumer of those
columns is `recordCaption`, and a Bag caller that wants a full caption can
declare `virtual_columns` on its own class.

## What this costs, stated plainly

A Bag caller no longer gets the caption's virtual columns, so its `outputDocName`
caption falls back to what it produced before #1212. The **defect** #1205
describes is unaffected: `outputDocIdentifier()` reads the pkey off the Bag, so
the names stay distinct.

Two tests, both on real `sys.batch_log` records:

- `test_a_bag_record_keeps_what_the_caller_merged_in` — pins the fix. I checked
  the ratchet: with the previous line restored it fails (the merged
  `batch_parameter` is gone after the reload), with the fix it passes.
- `test_a_bag_record_still_prints_to_its_own_file` — pins the part that must
  *not* regress. It passes with and without the fix, deliberately: it is there
  so a future change to this line cannot quietly reintroduce #1205 on the Bag
  path.

`tests/web/gnrbaseclasses_test.py` 9 passed, flake8 clean on both files.

## Grid prints

Not covered here, as you suggested I could either do or declare: #1250, with
both call sites. `print_gridtemplate.py:64` is the worse of the two — it has no
fallback at all, so a null `titolo` writes to the literal name `None.<ext>`.

Your approvals fell with this push and nothing they covered has moved: the diff
is the one line above plus the two tests. Re-requesting.

genro (approved):
Re-approving on `23c048f1fe`. The narrow variant is in: the Bag branch passes the same expression the code had before this PR, so `recordAs` leaves a caller's Bag alone (`record.py:369` reloads only when `virtual_columns` is truthy or the record is not a Bag). `Bag` is imported at `gnrbaseclasses.py:34`. The pkey and dict paths still get `captionVirtualColumns()`, so the #1205 fix stands where the record has to be loaded anyway.

Verified rather than read: `tests/web/gnrbaseclasses_test.py` 9 passed, with `test_module_under_test` proving the module exercised is the one in the tree and not the editable install. Good call adding that assertion — without it the file answering is the main checkout.

The cost you state is the right one to accept: a Bag caller loses the caption's virtual columns, which reaches only the caption, while the output name stays record specific because the identifier reads the pkey off the Bag. Grid prints tracked in #1250.

PR #1216, opened 2026-09-01
Update esbuild dependency version

cgabriel:
CVSS:3.1/AV:N/AC:H/PR:N/UI:R/S:U/C:H/I:N/A:N

CWE-346
GHSA-67mh-4wv8-2f99

### DIFF
--- a/tools/codemirror6-build/package-lock.json
+++ b/tools/codemirror6-build/package-lock.json
@@ -26,7 +26,7 @@
         "thememirror": "^2.0.1"
       },
       "devDependencies": {
-        "esbuild": "^0.23.0"
+        "esbuild": "^0.28.2"
       }
     },
     "node_modules/@codemirror/autocomplete": {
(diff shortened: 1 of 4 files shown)


cgabriel:
Note to merger: this does not afflict the already generated build currently in the repo, so there is not immediate need to update them.

fporcari (approved):
Approved.

Verified on the diff: the two `package.json` files only move `esbuild` from `^0.23.0` to `^0.28.2` in devDependencies, the `prosemirror-build` dependency block is an alphabetical reorder with all 18 versions unchanged, and both `package-lock.json` diffs touch nothing but `esbuild` and its `@esbuild/*` platform binaries (plus the two new `netbsd-arm64` / `openharmony-arm64` entries). The build scripts keep the same CLI flags, all stable across this range.

Agreeing with your note: GHSA-67mh-4wv8-2f99 only affects `esbuild serve`, which neither tool uses, so the vendored bundles under `resources/js_libs/` are unaffected and there is no reason to regenerate them here.

One thing left open, not a blocker: no CI workflow invokes these two tools, so the green checks say nothing about the bump, and the bundle has not actually been rebuilt with 0.28.2 in this PR. Whoever next regenerates either bundle should confirm 0.28.2 produces an equivalent output. Your call whether to do it here or leave it to the next build.

PR #1218, opened 2026-09-01
fix(daemon): make page drop atomic with subscriber unindexing

### LINKED ISSUE #1217: [BUG] siteregister: dangling table subscribers survive the page item and break notifyDbEvents for the whole site
## Problem Description

A page_id can remain in `PageRegister.tableSubscribers` after its register item is gone. From that moment every `notifyDbEvents` touching one of those tables raises `KeyError: '<page_id>'` inside the daemon, the exception travels back through Pyro into `db.commit()` of **every** page of the site (the data is already committed: `onDbCommitted` runs after `connection.commit()`), and no subscriber receives db events anymore. Grids keep showing deleted records until the daemon is restarted.

Expected behavior: removing a page must atomically remove its subscriptions from the index, whatever happened to the item before, and a residual dangling key must never abort the notification for the live subscribers.

The index was introduced by #968 (PR #969): `subscribed_table_pages` reads `registerItems[k]` with no guard. The drop path does unindex (`PageRegister.drop` → `dropSubscriptions`), but it unindexes **what the item's own list says**, not what the index says. Two paths break that assumption:

1. **`PageRegister.create` on an existing page_id** — `addRegisterItem` overwrites the item wholesale with no guard (unlike `new_connection`, which asserts). The previous life's subscriptions stay in the index; the eventual drop cleans only the new list → dangling key forever.
2. **`BaseRegister.load`** — replaces `registerItems` from the pickle and neither rebuilds nor clears `tableSubscribers` (restored subscriptions invisible to `notifyDbEvents`; whatever the index held dangles).

## Reproduction Steps

```python
reg = PageRegister(siteregister)
reg.create('P')
reg.subscribeTable('P', table='foo.bar', subscribe=True)
reg.create('P')          # re-registration replaces the item, index keeps 'P'
reg.drop('P')            # cleans only the new (empty) list
reg.subscribed_table_pages('foo.bar')   # KeyError: 'P'
```

In production the re-registration path is `_check_page_id` → `_register_new_page(page_id=page_id)` (guest pages, `gnrwebpage.py:310`).

## Environment

Observed in production on a multidb/multidomain instance (genropy develop, deployed 2026-08-28): one dangling page_id in one workspace register kept `saveRecordCluster` / `deleteDbRows` of all users of that workspace failing on `wkfl.approval_flow`, `ciak.ordine_acquisto`, `ciak.rendiconto` for the whole daemon lifetime, verified live by comparing `tableSubscribers` against `registerItems` via Pyro.

## Additional Material

Fix direction (PR follows):
- `dropSubscriptions` driven by the index (scan `tableSubscribers`, discard the page_id) so drop is atomic with unindexing;
- `create` unindexes an existing page_id before overwriting;
- `PageRegister.load` rebuilds the index from the restored items;
- `subscribed_table_page_items/pages` prune a residual dangling key instead of raising — the index is a cache, not the truth.

fporcari:
Fixes #1217

A page_id could survive in `PageRegister.tableSubscribers` after its register item was gone. From then on every `notifyDbEvents` touching one of those tables raised `KeyError` inside the daemon, the exception travelled back through Pyro into `db.commit()` of every page of the site (with the data already committed), and no subscriber received db events until a daemon restart. Observed in production on a multidomain instance: one dangling page_id kept all users of a workspace failing on three tables for the whole daemon lifetime.

The drop path did unindex, but it unindexed what the item's own list said, not what the index said — so any item replacement that bypassed drop (a re-registered page_id overwriting the item via `create`, a pickle restored by `load`) left the two out of step, and the eventual drop cleaned the wrong list.

**Changes** (all in `PageRegister`):

- `dropSubscriptions` is now driven by the index: it scans `tableSubscribers` and discards the page_id, making unindexing atomic with the drop whatever happened to the item before. The item's own list is still emptied when passed.
- `create` unindexes an existing page_id before overwriting it (the counterpart of the `new_connection` assert; here re-registration is legal — guest pages do it — so clean instead of refuse).
- `load` rebuilds `tableSubscribers` from the restored items, instead of keeping whatever the previous lifetime held while `registerItems` is replaced wholesale.
- `subscribed_table_page_items` / `subscribed_table_pages` prune a residual dangling key instead of raising: the index is a cache of the per-page lists, and one stale key must not abort the notification for every live subscriber of the table.

**Tests**: six new cases in `subscribed_tables_index_test.py` — the re-registration regression (create → subscribe → create → drop leaves no dangling key), index/lists alignment across re-registration, drop with a diverged item list, `load` rebuild (including clearing stale pre-load entries), and the pruning lookup. Full gnrpy suite green (2261 passed).

Related observation, out of scope here: `SiteRegister.load` opens the pickle in text mode (`open(self.storage_path)`), which cannot work with `pickle.load`; autorestore likely never restores. Worth its own issue.

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -421,11 +421,19 @@ class PageRegister(BaseRegister):
             self.tableSubscribers[table].add(page_id)
         return bool(subscribed_tables)
 
-    def dropSubscriptions(self, page_id, subscribed_tables):
-        """Forget every subscription of a page that is going away."""
-        for table in list(subscribed_tables):
+    def dropSubscriptions(self, page_id, subscribed_tables=None):
+        """Forget every subscription of a page that is going away.
+
+        Driven by the index, not by the page's own list: the list says what the
+        current item believes, the index says what notifyDbEvents will actually
+        read. Scanning the index makes the removal atomic with the drop whatever
+        happened to the item before (an overwritten registration, a restored
+        pickle), so a dropped page can never leave a dangling subscriber behind.
+        """
+        for table in list(self.tableSubscribers):
             self._dropSubscriber(table, page_id)
-        del subscribed_tables[:]
+        if subscribed_tables:
+            del subscribed_tables[:]
 
     def _dropSubscriber(self, table, page_id):
         subscribers = self.tableSubscribers.get(table)
@@ -439,9 +447,24 @@ class PageRegister(BaseRegister):
         """Tables with at least one subscribed page."""
         return list(self.tableSubscribers)
 
+    def load(self, storagefile):
+        # A restored pickle replaces registerItems wholesale: rebuild the index from
+        # the restored per-page lists, or every restored subscription is invisible to
+        # notifyDbEvents and whatever the index held before dangles.
+        super(PageRegister, self).load(storagefile)
+        self.tableSubscribers = defaultdict(set)
+        for page_id, register_item in self.registerItems.items():
+            for table in register_item.get('subscribed_tables') or []:
+                self.tableSubscribers[table].add(page_id)
+
     def create(self, page_id, pagename=None, connection_id=None, subscribed_tables=None, user=None, user_ip=None, user_agent=None, relative_url=None, data=None):
         register_item_id = page_id
         start_ts = datetime.now()
+        if register_item_id in self.registerItems:
+            # Re-registering an existing page_id replaces the item wholesale: unindex
+            # the previous life first, or its subscriptions dangle in the index and
+            # outlive the item, breaking every notifyDbEvents on those tables.
+            self.dropSubscriptions(register_item_id)
         if subscribed_tables:
             subscribed_tables = subscribed_tables.split(',')
         subscribed_tables = subscribed_tables or []
@@ -467,7 +490,7 @@ class PageRegister(BaseRegister):
         # it must never point at one that is already gone.
         self.dropSubscriptions(register_item_id,
                                (self.registerItems.get(register_item_id) or {}
-                                ).get('subscribed_tables') or [])
+                                ).get('subscribed_tables'))
         register_item = self.drop_item(register_item_id)
         self.pageProfilers.pop(register_item_id, None)
         if cascade:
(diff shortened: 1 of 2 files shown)


genro (approved):
Approving. This is a bug I introduced in #969 and #1106, and the diagnosis is exactly right about why.

I chose the direct `registerItems[k]` deliberately and wrote the reasoning down in the review of #1106: the invariant is maintained by construction, `drop` unindexes before removing the item, so a filter would be defensive code against an impossible state. What I actually verified was the `drop` path — and only that one. `create` overwriting an existing item and `load` replacing `registerItems` from a pickle both break the invariant without going anywhere near `drop`, and I never looked at either. "Maintained by construction" was a claim about one of three writers.

The consequence is worse than the crash, and the description gets the mechanism right: `onDbCommitted` runs after `connection.commit()`, so the data is already written when the `KeyError` travels back through Pyro into `db.commit()`. Every page of the site loses db events, and grids keep showing deleted records until the daemon restarts. A single dangling key in one workspace register did that for a whole daemon lifetime.

**The fix is the right shape, and stronger than the symptom needed.** Driving `dropSubscriptions` from the index rather than from the item's own list is what makes the removal atomic whatever happened to the item before — the list says what the current item believes, the index says what `notifyDbEvents` will read, and only the second one matters at that moment. Rebuilding the index in `PageRegister.load` closes the pickle path, and the `create` guard closes re-registration. Pruning in `subscribed_table_page_items` instead of raising is the principle I should have applied from the start: the index is a cache of the per-page lists, so residual drift is a stale entry to discard, never a reason to abort the notification for every live subscriber of that table.

60 tests pass on this head across `subscribed_tables_index_test.py` and `siteregister_links_test.py`, checked on the branch with the module path printed first.

**One thing to carry forward, because the same shape survives on the other link I added.**

`connection_page_items` still does `self.registerItems[k]` with no guard, over `connection_item['pages']` — the connection→pages set from #1106, the twin of the index this PR just hardened. Same failure mode: a page_id in the set with no item raises `KeyError`.

I checked the paths and it is narrower than the table index was. `page_register.drop` is only reached through `SiteRegister.drop_page`, which unlinks first, and `expire_pages` goes through `drop_page` too — so there is no bypass. But `drop_page` unlinks only `if page_item:`, which means it skips the unlink in precisely the state where drift already exists, and nothing rebuilds the connection→pages sets in `load` the way this PR now rebuilds `tableSubscribers`.

So the exposure is smaller but it is the identical shape, and both of the writers this PR discovered — `create` overwriting and `load` restoring — are the ones I did not consider the first time. Worth its own issue rather than widening this PR: the same three moves (index-driven unlink, rebuild on load, prune instead of raise) applied to `connection_page_items`.

cgabriel (approved):

PR #1220, opened 2026-09-01
fix(daemon): make the register link sets converge on a vanished child

### LINKED ISSUE #1219: [BUG] siteregister: the connection→pages and user→connections links have the dangling-key shape #1217 just fixed for table subscribers
## Problem

#1217 fixed a dangling-key class of bug in `PageRegister.tableSubscribers`. The same shape survives on the **two other link sets** introduced by #1106, and neither was touched by that fix:

| link set | read with | rebuilt on `load` |
|---|---|---|
| `connection_item['pages']` | `registerItems[k]`, no guard (`siteregister.py:509`) | no |
| `user_item['connections']` | `registerItems[k]`, no guard (`siteregister.py:377`, `:380`) | no |

`PageRegister` now has its own `load` that rebuilds `tableSubscribers` from the restored items. `ConnectionRegister` and `UserRegister` have no `load` override: they inherit `BaseRegister.load`, which replaces `registerItems` from the pickle and leaves both link sets exactly as they were.

## Three ways the drift happens

**1. The unlink is skipped precisely when drift already exists.**

```python
def drop_page(self, page_id, cascade=None):
    page_item = self.page_register.registerItems.get(page_id)
    if page_item:                      # <-- skipped when the item is already gone
        self.updateRegisterLink(self.connection_register, page_item['connection_id'],
                                'pages', page_id)
    return self.page_register.drop(page_id, cascade=cascade)
```

`drop_connection` has the identical `if connection_item:`. The unlink is driven by the child item, so if the item has vanished the parent's set keeps the dead id forever — the same "unindex what the item's own list says, not what the index says" that #1217 identified.

**2. A restored pickle.** Same as the `PageRegister.load` half of #1217: neither set is rebuilt or validated against the restored items.

**3. `drop_connections` cannot converge.**

```python
def drop_connections(self, user):
    for connection_id in self.user_connection_keys(user):   # reads user_item['connections']
        self.drop_connection(connection_id)
```

A dead id in `connections` makes `drop_connection` find `connection_item is None`, skip the unlink, and leave the id in place. Every subsequent `drop_connections` for that user walks it again with the same outcome. The table index had no equivalent: nothing iterated it to drive deletions.

## Consequence

`connection_page_items`, `connection_pages`, `user_connection_items` and `user_connections` raise `KeyError` on a dangling id. The blast radius is smaller than #1217's — these are not on the `notifyDbEvents` path — but `connection_page_keys` feeds `drop_pages`, and `validate_page_id` (`gnrwebpage_proxy/connection.py:72`) reads `connection_item['pages']` on every request. A dead id there does not grant access — `_check_page_id` fails one step later on the missing register item — but it does mean the membership test answers about a page that no longer exists.

## Fix direction

The same three moves #1217 applied to `tableSubscribers`, on both sets:

1. **unlink driven by the parent sets, not by the child item** — a `dropRegisterLinks(register, link_name, child_id)` that scans the parents and discards the id, called unconditionally by `drop_page` / `drop_connection`;
2. **`ConnectionRegister.load` and `UserRegister.load`** that rebuild their set from the restored children, the way `PageRegister.load` now rebuilds `tableSubscribers`;
3. **prune instead of raise** in the four readers: a key with no item is residual drift to discard, not a reason to abort the call.

## Provenance

Mine. I introduced both link sets in #1106 and chose the unguarded `registerItems[k]` deliberately, arguing in that review that the invariant held by construction because `drop` unindexes before removing the item. That was true of the `drop` path and of nothing else — #1217 found the two writers I had not considered (`create` overwriting, `load` restoring), and they apply here unchanged.

genro:
## Problem

#1217 fixed a dangling-key bug in `PageRegister.tableSubscribers`. The same shape survived on the two link sets introduced by #1106, and neither was touched by that fix:

| link set | read with | rebuilt on `load` |
|---|---|---|
| `connection_item['pages']` | `registerItems[k]`, unguarded | no |
| `user_item['connections']` | `registerItems[k]`, unguarded | no |

**The unlink was driven by the child item.** `drop_page` read the connection id off the page, `drop_connection` read the user off the connection, and both were guarded by `if <child>_item:` — so the unlink was skipped in exactly the state where it mattered, and the dead id stayed in the parent's set.

**`drop_connections` made that permanent rather than merely wrong.** It iterates the user's set, so a dead id made `drop_connection` find no item, skip its own removal, and leave the id in place. Every later call walked the same id again. The table index had no equivalent, because nothing iterated it to drive deletions.

**A restored pickle left both sets blind.** The three registers are dumped and loaded independently, so a parent's set knows nothing of the children that actually came back — the `PageRegister.load` half of #1217, unaddressed here.

## Change

Three moves, mirroring #1217.

**`dropRegisterLinks(register, link_name, child_id)`** scans the parents' sets instead of reading the parent off the child, and `drop_page`/`drop_connection` call it **unconditionally**. The child item is precisely what may be missing, so it cannot be the thing that decides whether the unlink happens. The parent-known variant, `dropRegisterLink`, is what a pruning reader calls once it has already found the key with no item.

**`rebuildRegisterLinks(parent_register, child_register, parent_field, link_name)`** rebuilds a parent's set from the children that name it, called from `SiteRegister.load` after all three registers are in.

That placement is deliberate and it was not my first shape. I initially put the rebuild inside `PageRegister.load` and `ConnectionRegister.load`, and #1217's own test caught it by failing: a child register had to reach up into the site facade to repair its parent. The site-level call has no such coupling, and it is also the first point at which every item exists.

**The four readers prune instead of raising.** `connection_page_items`, `connection_pages`, `user_connection_items` and `user_connections` go through a shared `_live_children`. A link set is a cache of what the children say about their parent, so a key with no item is drift to discard — never a reason to raise on every caller of that parent.

## Tests

Eight cases in `gnrpy/tests/web/siteregister_links_test.py`, **all verified failing on the unfixed module**:

- dropping a page, and a connection, whose item had already vanished;
- `drop_connections` converging on a dangling id, and staying idempotent;
- each of the four readers pruning and still returning the live children;
- `dropRegisterLinks` reaching two parents that both claim the same child;
- the rebuild after an independent restore of the three registers.

## Verification

- `pytest gnrpy/tests/web/siteregister_links_test.py gnrpy/tests/web/subscribed_tables_index_test.py`: 109 passed
- `pytest gnrpy/tests`: **2239 passed**, with 2 pre-existing failures in `flatfiles_test.py` and 8 pre-existing errors in `tests/sql` from missing pg/mysql drivers, all present on a clean `develop`
- Not verified: no live run against a real daemon. As with #1217, the drift is reproduced in the tests by removing the item directly, which is what the three real writers do by other means.

## Provenance

Mine. I introduced both link sets in #1106 and chose the unguarded `registerItems[k]` deliberately, arguing in that review that the invariant held by construction because `drop` unindexes before removing the item. That was true of the `drop` path and of nothing else. #1217 found the two writers I had not considered — `create` overwriting an item, `load` restoring one — and they apply here unchanged.

Fixes #1219

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -746,21 +746,6 @@ class SiteRegister(BaseRemoteObject):
         children.discard(child_id)
         return True
 
-    def dropRegisterLink(self, register, parent_id, link_name, child_id):
-        """Discard *child_id* from one parent's link set, parent known.
-
-        Unlike updateRegisterLink this does not require the child item to exist:
-        it is what a pruning reader calls once it has found a key with no item.
-        """
-        parent_item = register.registerItems.get(parent_id)
-        if parent_item is None:
-            return False
-        children = parent_item.get(link_name)
-        if not children or child_id not in children:
-            return False
-        children.discard(child_id)
-        return True
-
     def dropRegisterLinks(self, register, link_name, child_id):
         """Discard *child_id* from every parent that holds it, parent unknown.
 
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -374,10 +374,30 @@ class ConnectionRegister(BaseRegister):
         return list(user_item['connections'])
 
     def user_connection_items(self, user):
-        return [(k, self.registerItems[k]) for k in self.user_connection_keys(user)]
+        return [(k, register_item)
+                for k, register_item in self._live_children(
+                    self.user_connection_keys(user),
+                    self.siteregister.user_register, user, 'connections')]
 
     def user_connections(self, user):
-        return [self.registerItems[k] for k in self.user_connection_keys(user)]
+        return [register_item for k, register_item in self.user_connection_items(user)]
+
+    def _live_children(self, keys, parent_register, parent_id, link_name):
+        """Pair each key with its item, pruning the keys that have none.
+
+        A link set is a cache of what the children say about their parent: a key
+        with no item is residual drift, not a reason to raise on every caller of
+        the parent. Pruning here keeps the set converging instead of carrying a
+        dead id for the life of the daemon.
+        """
+        for k in keys:
+            register_item = self.registerItems.get(k)
+            if register_item is None:
+                self.siteregister.dropRegisterLink(parent_register, parent_id,
+                                                   link_name, k)
+                continue
+            yield k, register_item
+
 
     def connections(self, user=None, include_data=None):
         if not user:
@@ -531,10 +551,30 @@ class PageRegister(BaseRegister):
         return list(connection_item['pages'])
 
     def connection_page_items(self, connection_id):
-        return [(k, self.registerItems[k]) for k in self.connection_page_keys(connection_id)]
+        return [(k, register_item)
+                for k, register_item in self._live_children(
+                    self.connection_page_keys(connection_id),
+                    self.siteregister.connection_register, connection_id, 'pages')]
+
+    def _live_children(self, keys, parent_register, parent_id, link_name):
+        """Pair each key with its item, pruning the keys that have none.
+
+        A link set is a cache of what the children say about their parent: a key
+        with no item is residual drift, not a reason to raise on every caller of
+        the parent. Pruning here keeps the set converging instead of carrying a
+        dead id for the life of the daemon.
+        """
+        for k in keys:
+            register_item = self.registerItems.get(k)
+            if register_item is None:
+                self.siteregister.dropRegisterLink(parent_register, parent_id,
+                                                   link_name, k)
+                continue
+            yield k, register_item
 
     def connection_pages(self, connection_id):
-        return [self.registerItems[k] for k in self.connection_page_keys(connection_id)]
+        return [register_item
+                for k, register_item in self.connection_page_items(connection_id)]
 
     def pages(self, connection_id=None, user=None, include_data=None, filters=None):
         # walk the link sets rather than the whole registry: a connection knows its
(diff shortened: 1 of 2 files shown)


fporcari (changes requested):
The three moves are right and the placement of the rebuild in `SiteRegister.load` is the correct call — a child register repairing its parent would have been the wrong coupling. Two things before merge, both inside the scope this PR sets for itself.

**1. `pages()` is the fifth reader on the same keys, and it is the exposed one.** `PageRegister.pages()` still walks `connection_page_keys` / `user_connection_keys` and then indexes raw: `siteregister.py:594`, `pages = [self.registerItems[page_id] for page_id in page_ids]` — the only unguarded `registerItems[...]` read left in the file. A dangling id raises `KeyError` there exactly as it did in the four readers you fixed. The `include_data` branch one line above is worse in a quieter way: `get_item` returns `None` for a missing key, so a dead id yields a `None` in the list that then fails in `Bag(page)` or in `v['user'] == user`.

This is the reader on the request path — `validate_page_id` (`gnrwebpage_proxy/connection.py:72`) falls back to `register.pages(connection_id=...)`, and `siteregister_client.py:514` calls it with `include_data=True`. If drift is worth pruning in `connection_page_items`, it is worth pruning here first. Routing the non-`include_data` branch through `_live_children` covers it; the `include_data` branch can build on the pruned keys the same way. `ConnectionRegister.connections(user, include_data=True)` has the identical shape.

**2. `_live_children` is duplicated verbatim.** Same body, same docstring, in `ConnectionRegister` (:385) and `PageRegister` (:559). It touches only `self.registerItems` and `self.siteregister`, both set in `BaseRegister.__init__` — so it belongs in `BaseRegister`, once. As it stands the next fix to the pruning rule has to be applied twice, which is the shape of drift this PR is about.

A test for `pages()` with a dangling id, in both the plain and the `include_data` form, would close the set of readers.

genro:
> **1. `pages()` is the fifth reader on the same keys, and it is the exposed one.**
> **2. `_live_children` is duplicated verbatim.**
> — [review of 1 Sep](https://github.com/genropy/genropy/pull/1220#pullrequestreview-5081100527)

Both in on `fbb28fa3a3`, and you were right about the second one in a way I should have caught myself: duplicating the helper is the shape this branch exists to remove.

**`pages()`.** Fixed, and the `include_data` branch is the half worth naming. You are right that it is the quieter failure: `get_item` returns `None` for a missing key, so a dead id did not raise — it travelled on as a `None` into `Bag(page)` or `v['user'] == user` and failed somewhere with no connection to the cause. Both branches now go through the pruned keys, and `ConnectionRegister.connections(user, include_data=True)` had the same shape and gets the same treatment.

There is no unguarded `registerItems[...]` read left in the module — the only remaining subscript is the write in `addRegisterItem`.

**`_live_children` moved to `BaseRegister`**, once. It touches only `registerItems` and `siteregister`, both set in `BaseRegister.__init__`, so there was never a reason for it to live in the subclasses.

**One change beyond what you asked, because `pages()` forced it.** The helper no longer takes a parent id. The user walk spans several connections — `page_ids` comes from `user_connection_keys` then `connection_page_keys` per connection — so no single parent covers that call, and passing one would have meant a special case. `dropRegisterLinks` already scans the parents, so the helper now uses it: one call cleans the key out of every parent that wrongly holds it, and the signature loses an argument instead of gaining a branch. Pruning is the rare corrective path, so the scan costs nothing on the ordinary one.

**Tests**, as you asked, in both forms: `pages()` pruning by connection, by user across two connections, and with `include_data=True` where the assertion is `None not in got` rather than a raise — plus the same for `connections()`. All four fail on `faea4acb7f`.

`pytest gnrpy/tests/web gnrpy/tests/app gnrpy/tests/core`: 973 passed, with the two pre-existing `flatfiles_test.py` failures. flake8 clean on both files.

fporcari (dismissed):
Both points are addressed, and the shape you landed on is better than what I asked for: dropping `parent_id` from `_live_children` and delegating to `dropRegisterLinks` means the `pages(user=...)` walk, which spans several connections, needs no special case. `pages()` now prunes on both branches, which was the one on the request path.

Approving. One leftover, your call whether it belongs here or in a follow-up: `dropRegisterLink` at `siteregister.py:749` has no caller left — `_live_children` was its only one and now uses the plural. It is dead code this PR introduces, and `SiteRegister` is the remote facade, so it is public surface with nothing behind it. I would delete it before merging.

genro:
> One leftover, your call whether it belongs here or in a follow-up: `dropRegisterLink` at `siteregister.py:749` has no caller left [...] I would delete it before merging.
> — [review of 1 Sep](https://github.com/genropy/genropy/pull/1220#pullrequestreview-)

Deleted in `88eec8cfdc`. It belongs here rather than in a follow-up: this branch introduced it two commits ago and this branch is what left it unreachable, so carrying it to `develop` would mean shipping dead code and then opening an issue to remove what we already know is dead.

Confirmed no caller anywhere in `gnrpy/`, `projects/` or `resources/` before removing — only the definition.

Your reason for wanting it gone is the better one and I have put it in the commit message: `SiteRegister` is the remote facade, so an unused method there is public surface with nothing behind it. And a later caller reaching for the singular would quietly reintroduce the assumption this branch removed — that the pruning reader knows which parent holds the dangling key. That assumption is exactly what `pages(user=...)` cannot satisfy.

72 tests pass on the head, flake8 clean. Your approval fell with the push; re-requesting.

fporcari (commented):
LGTM

cgabriel (approved):
Can't find any issue here, so I'm approving. Will check if the continuous scan for live connections under heavy load will affect performances.

genro:
Re-requesting because the approval fell with the push that removed `dropRegisterLink`, and the LGTM that followed landed as a comment rather than an approval, so the code owner gate still blocks. cgabriel has approved. Nothing changed since your LGTM — the diff is the same one you read.

fporcari (commented):
LGTM again

fporcari (approved):

Issue #1223, opened 2026-09-02
Register link maintenance is not atomic with the item, which is why a scan is needed


genro:
# Register link maintenance is not atomic with the item, which is why a scan is needed

## What #1220 did, and what it did not settle

#1219 was about readers carrying a dead key for the life of the daemon. #1220
fixed that by making the unlink unconditional and driven by the link sets rather
than by the child item: `dropRegisterLinks` discards a key from every parent that
holds it, without being told which parent that is.

That converges, and the drift is gone. It also pays for the rare case on the
ordinary path: `drop_page` walks the whole connection register on every page
close, which is O(live connections) where it used to be O(1). The concern raised
in review on #1220 is legitimate.

Looking closer, the scan is a symptom. A page item carries `connection_id`, a
connection item carries `user`, and those fields are the only statement of where
a child belongs. The `pages` and `connections` sets say the same thing in the
other direction, so they are an index: derived, reconstructible from the children
at any moment, never a source of truth. When the index is maintained by whatever
changes the item, the child field is always authoritative and no reader ever has
to search for the parent of an orphan key.

That is not where the maintenance lives today.

## Where it lives

`create` and `drop` on the child registers do not touch the parent's set. The
link is added in `SiteRegister.new_page` / `new_connection` and removed in
`SiteRegister.drop_page` / `drop_connection`, one level above the code that
changes the item. Anything reaching a register directly bypasses it, and
`get_register()` hands the registers out, remotely included.

Three consequences, each with a failing test:

1. `page_register.create(...)` — the page exists, its connection does not list it.
2. `page_register.drop(...)` — the page is gone, its connection still lists it.
3. `page_register.drop(pid, cascade=True)` — reads `connection_page_keys`, finds
   the key it should have removed, concludes the connection still has pages and
   does not close it. Emptied connections and users stop being collected. Same
   for `connection_register.drop(cid, cascade=True)` and its user.

The third is not index hygiene. It is cleanup that silently does not happen.

## Evidence

`gnrpy/tests/web/siteregister_authority_test.py` — 51 tests, one invariant
asserted after every access that changes a register: rebuilding the sets from the
child fields must reproduce exactly the sets that are stored, and no child may
point at a parent that is gone. Three tests check the invariant can fail.

38 pass. 9 fail, and they are the list above:

- `test_page_register_create_maintains_the_connection_set`
- `test_connection_register_create_maintains_the_user_set`
- `test_page_register_drop_maintains_the_connection_set`
[...]

PR #1225, opened 2026-09-02
fix(daemon): open the freeze file in binary mode when restoring

### LINKED ISSUE #1224: SiteRegister.load opens the freeze file in text mode, so the restore always fails
# `SiteRegister.load` opens the freeze file in text mode, so the restore always fails

## The defect

`dump` writes the pickle in binary:

```python
with open(self.storage_path, 'wb') as storagefile:   # siteregister.py:1077
```

`load` reads it back without a mode, which is text:

```python
with open(self.storage_path) as storagefile:         # siteregister.py:1084
```

`pickle.load` on a text stream raises `UnicodeDecodeError` on the first byte of
the protocol header. The `try` around it catches `EOFError` only, so the
exception propagates out of `load`, out of `GnrSiteRegisterServer.run`, and the
site register never starts.

Introduced in 5b9c00a7b7 (26 March 2026).

## Why it went unnoticed

`start` guards the call with `autorestore = autorestore and os.path.exists(self.storage_path)`
(siteregister.py:1198). `siteregister_client.py:244` always asks for
`autorestore=True`, but the file only exists when the previous shutdown ran
`stop(saveStatus=True)` — that is, when the daemon was stopped with
`--savestatus`. On an ordinary stop nothing is written, the guard turns
autorestore off, and the broken branch is never taken.

So the feature is not degraded, it is unreachable: whenever a saved state is
actually there to restore, the restore raises instead.

## Reproduction

`gnrpy/tests/web/siteregister_authority_test.py` (from #1223) round trips a
populated register through `dump` and `load`. Four tests fail on the
`UnicodeDecodeError`, none of them about the content:

- `test_a_round_trip_preserves_the_hierarchy`
- `test_a_round_trip_rebuilds_a_set_that_had_drifted`
- `test_a_round_trip_of_an_empty_register`
- `test_dropping_after_a_round_trip`

## Fix

Open with `'rb'`. Whether the bare `except EOFError` should also cover a
truncated or unreadable freeze file is a separate call: today a corrupt file
takes the daemon down rather than starting it empty.

genro:
## Problem

`dump` writes the pickle in binary, `load` read it back without a mode:

```python
with open(self.storage_path, 'wb') as storagefile:   # dump, siteregister.py:1077
with open(self.storage_path) as storagefile:         # load, siteregister.py:1084
```

A text stream makes `pickle.load` raise `UnicodeDecodeError` on the first byte of
the protocol header. The `try` around it catches `EOFError` only, so the
exception propagates out of `load`, out of `GnrSiteRegisterServer.run`, and the
site register never starts.

Introduced in 5b9c00a7b7 (26 March 2026).

## Why it went unnoticed

`start` guards the call with `autorestore = autorestore and os.path.exists(self.storage_path)`.
`siteregister_client.py:244` always asks for `autorestore=True`, but the freeze
file only exists when the previous shutdown ran `stop(saveStatus=True)` — that
is, a stop with `savestatus`. On an ordinary stop nothing is written, the guard
turns autorestore off, and the broken branch is never taken.

So the restore was not degraded, it failed whenever there was a state to restore.

## Change

`open(self.storage_path, 'rb')`, with a comment saying why the mode matters.

Nothing else. Whether the bare `except EOFError` should also cover a truncated or
unreadable freeze file is a separate call, left alone here: today a corrupt file
takes the daemon down rather than starting it empty.

## Tests

`gnrpy/tests/web/siteregister_freeze_test.py`, six cases that take the restore
branch every time. **Five verified failing on the unfixed module**, all on the
`UnicodeDecodeError` rather than on any content:

- a populated register dumps and loads back with the same keys in all three registers;
- an empty register round trips;
- the restored items keep their parent — a page its `connection_id`, a connection its `user`;
- the restored register still drops a page correctly;
- the freeze file is consumed, renamed to `_loaded.pik`, so a restart cannot restore the same state twice.

The sixth asserts that loading a file that was never written raises
`FileNotFoundError`; it passes either way and pins the behaviour of the guard
that hid this bug.

## Verification

- `gnrpy/tests/` — 2249 passed, 86 skipped, 2 failed, 8 errors
- the 2 failures and 8 errors are identical on `develop` without this change:
  `core/flatfiles_test.py` (CSV dialect) and `sql/test_gnrbaseadapter.py`
  (double colon), neither of which reaches the site register
- `gnrpy/tests/web/` alone — 377 passed, 76 skipped
- flake8 clean on both files

Closes #1224

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -1081,7 +1081,9 @@ class SiteRegister(BaseRemoteObject):
 
     def load(self):
         try:
-            with open(self.storage_path) as storagefile:
+            # 'rb': dump writes the pickle in binary, and a text stream makes
+            # pickle.load raise UnicodeDecodeError on the protocol header
+            with open(self.storage_path, 'rb') as storagefile:
                 self.user_register.load(storagefile)
                 self.connection_register.load(storagefile)
                 self.page_register.load(storagefile)
--- a/gnrpy/tests/web/siteregister_freeze_test.py
+++ b/gnrpy/tests/web/siteregister_freeze_test.py
@@ -1,0 +1,98 @@
+"""The freeze round trip: dump the three registers, load them back.
+
+``dump`` writes the pickle with ``'wb'``. ``load`` opened the same file without a
+mode, which is text, so ``pickle.load`` hit ``UnicodeDecodeError`` on the first
+byte of the protocol header -- and the ``try`` around it catches ``EOFError``
+only, so the exception left the site register unable to start.
+
+The branch was reachable but rarely taken: ``start`` turns autorestore off unless
+the freeze file exists, and it is only written when the daemon is stopped with
+``savestatus``. These tests take it every time.
+"""
+
+import pytest
+
+from gnr.web.daemon.siteregister import SiteRegister
+
+
+class _FakeDaemon:
+    def register(self, obj, name):
+        pass
+
+
+class _FakeServer:
+    daemon = _FakeDaemon()
+    gnr_daemon_uri = None
+    hmac_key = None
+
+
+def _register(tmp_path, name='register.pik'):
+    reg = SiteRegister(_FakeServer(), sitename='testsite')
+    reg.setConfiguration()
+    reg.storage_path = str(tmp_path / name)
+    return reg
+
+
+def _populate(reg, users=2, connections=2, pages=2):
+    for u in range(users):
+        user = 'user-%d' % u
+        for c in range(connections):
+            connection_id = '%s/conn-%d' % (user, c)
+            reg.new_connection(connection_id, user=user)
+            for p in range(pages):
+                reg.new_page('%s/page-%d' % (connection_id, p), pagename='p',
+                             connection_id=connection_id, user=user)
+    return reg
+
+
+def _round_trip(reg, tmp_path):
+    reg.dump()
+    restored = _register(tmp_path)
+    assert restored.load() is True
+    return restored
+
+
+def test_a_dumped_register_loads_back(tmp_path):
+    reg = _populate(_register(tmp_path))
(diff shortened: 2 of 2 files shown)


fporcari (approved):
Verified locally: the six new tests run against the base fail 5/6 on `UnicodeDecodeError`, and pass together with `siteregister_links_test.py` once the `'rb'` change is applied. flake8 clean with the repo config.

Two non-blocking nits, your call whether to fold them in here:
- `import os` in `test_the_freeze_file_is_consumed_by_the_load` could move to module level with the other imports.
- #1224 has no assignee; assigning it to yourself keeps the issue from reading as still-to-do for everybody else while this PR is open.

cgabriel (approved):
Approved. Just for the record, it was not introduced in 5b9c00a in March 2026, but rather in 41ab7be4ce on 25 October 2013 , please verify your history checker. So this has been sitting there for a looong time.

genro:
You are right that 5b9c00a7b7 is wrong, and the mistake is one of method: I read the line with `git log -L`, which reports the last commit that touched it, and wrote that up as the commit that introduced it. Two different questions, and I answered the second one.

Checking properly, the history is more specific than either date:

- **41ab7be4ce** (25 Oct 2013, fporcari) introduces both calls — `open(path, 'w')` for dump and `open(path)` for load. Consistent with each other, and this was Python 2.
- **ad4840b521** (23 Apr 2024, mbertoldi, "fix dump gndaemon") moves dump to `'wb'` and leaves load untouched. This is where the pair stops agreeing, so this is the actual break.
- **5b9c00a7b7** (26 Mar 2026) carries the line into the new module during the refactor.

So the defect is two years and four months old, not thirteen years: until April 2024 the two ends matched. I will correct the attribution in #1224.

@fporcari on your two nits: `import os` moves to the module header — I will fold that into the #1223 branch, which is already touching these tests. The #1224 assignee is moot now that the issue is closed.

Issue #1226, opened 2026-09-02
An emptied user survives or is dropped depending on what emptied it


genro:
# An emptied user survives or is dropped depending on what emptied it

## The two paths

A user item is dropped when its last connection goes, on one path but not the
other.

**Last page closed.** `PageRegister.drop(page_id, cascade=True)` finds the
connection has no pages left and calls `siteregister.drop_connection(connection_id)`
**without propagating `cascade`**. `ConnectionRegister.drop` therefore skips its
own cascade block, and the user survives with no connections.

**Connection expired.** `expire_connection` calls
`drop_connection(connection_id, cascade=True)`. `ConnectionRegister.drop` then
runs:

```python
if cascade:
    user = register_item['user']
    keys = self.user_connection_keys(user)
    if not keys:
        self.siteregister.drop_user(user)
```

and the user goes, taking its store with it.

Same end state — a user with no connections — and two opposite outcomes,
decided by which call emptied it.

## Why it matters

Keeping an emptied user is deliberate: its store can hold state that is global
to that user and outlives any single connection. The page path honours that. The
expiry path does not, and expiry is the one that runs unattended, with nobody
closing anything.

## What is the intended behaviour

Two shapes, both consistent:

- **No user cascade at all.** `ConnectionRegister.drop` stops closing the
  emptied user; a user is only removed by an explicit `drop_user`. Matches the
  intent above and makes the rule the same on every path. It is a removal, not
  an addition.
- **A user expiry of its own.** The emptied user stays, and a `user_max_age`
  collects it after a period of its own. More machinery, and only worth it if
  emptied users actually accumulate.

## Status

No behaviour changed for now. `siteregister_authority_test.py` (from #1223) pins
the current outcome in `test_cascade_from_a_page_stops_at_the_connection`, which
asserts the page path as it behaves rather than as the name suggests. Whichever
shape is chosen, that test changes with it.

PR #1227, opened 2026-09-02
fix(daemon): maintain the register link sets where the items change

### LINKED ISSUE #1223: Register link maintenance is not atomic with the item, which is why a scan is needed

genro:
## Problem

The link sets were maintained in the `SiteRegister` wrappers — `new_page` added
the key, `drop_page` removed it — one level above the code that actually adds and
removes an item. `create` and `drop` on the child registers never touched the
parent's set, and `get_register()` hands the registers out, remotely included, so
anything reaching one directly bypassed the maintenance entirely.

Three consequences:

1. `page_register.create(...)` — the page exists, its connection does not list it.
2. `page_register.drop(...)` — the page is gone, its connection still lists it.
3. `page_register.drop(pid, cascade=True)` — reads `connection_page_keys`, finds
   the key it should have removed, concludes the connection still has pages and
   does not close it. Emptied connections and users stop being collected.

The third is not index hygiene, it is cleanup that silently does not happen.

#1220 made the readers converge on a key with no item, which stopped the drift
from being permanent. It did not stop it from forming, and it paid for the rare
case on the ordinary path: `drop_page` scanned every live connection on each page
close, where it used to be O(1). That cost was raised in review on #1220.

## Change

A page item carries `connection_id`, a connection item carries `user`. Those
fields are the only statement of where a child belongs; the sets say the same
thing in the other direction, so they are an index — derived, reconstructible,
never a source of truth. This puts the maintenance where the items change.

`addRegisterItem` and `drop_item` are the only two calls that put an item in a
register and take it out. Each register declares where its items hang:

| register | `parent_field` | `parent_link_name` | `parent_register_name` | `child_link_name` |
|---|---|---|---|---|
| `PageRegister` | `connection_id` | `pages` | `connection` | — |
| `ConnectionRegister` | `user` | `connections` | `user` | `pages` |
| `UserRegister` | — | — | — | `connections` |

Those two calls maintain the parent's set, so no route can change the hierarchy
and leave the index behind. `drop_item` unlinks **before** the pop — the set is
read to reach the items, so it must never point at one already gone (the #1217
ordering). `child_link_name` names the set of an item's own children, so
replacing a parent item does not discard it.

`update_item` writes whatever it is given straight into the item, so a
`connection_id` or `user` in its `upddict` used to change the field and leave both
sets naming the parent the item no longer has. A parent field now goes to
`reparent_item` and the rest of the dict is applied as before (raised in review).

`reparent_item` moves an item under another parent — the field and both sets in
one call. It is the only place a parent field changes, and
`change_connection_user` now uses it instead of three separate statements, which
is what let the field and the sets disagree there.

**Removed.** With the child field authoritative on every path, `dropRegisterLinks`
has nothing left to find: `drop_page` and `drop_connection` become one line each
and the unlink is O(1) off the child's own field. The pruning in `_live_children`
goes with it — a key with no item can no longer exist, so the five readers index
the set directly rather than skipping silently over an impossible case.

Both were added by #1220, one day ago. They were the right fix while the drift
could form; this removes the reason they existed.

## Tests

**`gnrpy/tests/web/siteregister_authority_test.py`** (new) — one invariant,
asserted after every access that changes a register: rebuilding the sets from the
child fields reproduces exactly the sets that are stored, and no child points at a
parent that is gone. Three tests check the invariant itself can fail. It covers
the facade, the registers reached directly, `get_register`, the parent move,
expiry, cascade and the freeze round trip.

**`siteregister_links_test.py`** — eleven tests rewritten. Each manufactured its
orphan with `registerItems.pop(...)`, reaching past the API into the raw dict —
the one operation no real path can perform now that `drop_item` unlinks. They keep
their scenarios and use `drop_item`, asserting the orphan never forms rather than
that it is pruned afterwards.

**`subscribed_tables_index_test.py`** — built `PageRegister` on a stub site with a
single method. A register now maintains its parent's set as items come and go, so
it needs its siblings; the stub would have had to reimplement that. It builds a
real `SiteRegister` instead, like the other two files.

## Self-review found and fixed

Written up because each one is a defect this branch introduced, caught before
review rather than by it:

- **re-registering a page under another connection** left the key in the old one.
  `create` replaces the item wholesale and the replacement may name another
  parent — two connections claiming one page, exactly the drift a parent-unknown
  scan existed to clean up.
- **replacing a parent item wiped its children set.** The new dict arrives with
  `pages=set()` while the pages are still there and still name it.
- **`reparent_item` on a register whose items have no parent** wrote a `None` key
  into the item and then failed on the missing parent register, leaving it
  corrupted. Now an explicit error before any mutation.
- **the unlink was after the pop** in an earlier revision of this branch, against
  the #1217 ordering. Restored to before.
- **one new test compared a list built from a set without sorting it**: it passed
  alone and failed in the suite. The suite was then re-run under several
  `PYTHONHASHSEED` values.

## Open

`drop_page(cascade=True)` descends one level: the emptied connection is closed,
the emptied user is not, while `expire_connection` does drop it. Behaviour
unchanged here and pinned by
`test_cascade_from_a_page_stops_at_the_connection`; the intended shape is #1226.

The sets are still pickled inside the items and overwritten at load by
`rebuildRegisterLinks`. Harmless — the rebuild wins — but it persists a second
statement of the same fact. Left alone deliberately: stripping them at dump costs
a copy of every item on each freeze, for no correctness gain.

## Verification

- `gnrpy/tests/` — 2312 passed, 86 skipped
- the 2 failures and 8 errors are identical on `develop` without this change:
  `core/flatfiles_test.py` and `sql/test_gnrbaseadapter.py`, neither reaching the
  site register
- `gnrpy/tests/web/` — 440 passed, 76 skipped, re-run under `PYTHONHASHSEED` 0, 42, 12345
- flake8 clean

Closes #1223

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -298,6 +298,11 @@ class BaseRegister(BaseRemoteObject):
         register_item = self.get_item(register_item_id)
         if not register_item:
             return
+        if self.parent_field and upddict and self.parent_field in upddict:
+            # the parent field is one half of a link: writing it straight into the
+            # item would leave both sets naming the parent it no longer has
+            upddict = dict(upddict)
+            self.reparent_item(register_item_id, upddict.pop(self.parent_field))
         register_item.update(upddict)
         return register_item
 
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -103,6 +103,21 @@ class RemoteStoreBagHandler(BaseRemoteObject):
 
 class BaseRegister(BaseRemoteObject):
     """docstring for BaseRegister"""
+
+    # Where this register's items hang. The item field is the only statement of
+    # where a child belongs; the set on the parent says the same thing in the
+    # other direction, so it is an index, maintained here in the two calls that
+    # add and remove an item and derived from the children again on load.
+    # A register whose items have no parent leaves all three at None.
+    parent_field = None
+    parent_link_name = None
+    parent_register_name = None
+
+    # The set on this register's own items holding their children, named by the
+    # child register as parent_link_name. A register with no children leaves it
+    # at None; test_the_two_ends_of_a_link_agree_on_its_name checks the pair.
+    child_link_name = None
+
     def __init__(self, siteregister):
         self.siteregister = siteregister
         self.registerItems = dict()
@@ -110,6 +125,12 @@ class BaseRegister(BaseRemoteObject):
         self.itemsTS = dict()
         self.locked_items = dict()
         self.cached_tables = defaultdict(dict)
+
+    @property
+    def parent_register(self):
+        if not self.parent_register_name:
+            return None
+        return self.siteregister.get_register(self.parent_register_name)
 
     def lock_item(self, register_item_id, reason=None):
         locker = self.locked_items.get(register_item_id)
@@ -135,7 +156,27 @@ class BaseRegister(BaseRemoteObject):
 
     def addRegisterItem(self, register_item, data=None):
         register_item_id = register_item['register_item_id']
+        previous = self.registerItems.get(register_item_id)
+        if previous is not None:
+            # re-registering an id replaces the item wholesale, so the two ends of
+            # its links have to survive the replacement
+            if self.parent_field:
+                # the replacement may name another parent: unlink the previous one,
+                # or both end up claiming the same child
+                self.siteregister.updateRegisterLink(
+                    self.parent_register, previous.get(self.parent_field),
+                    self.parent_link_name, register_item_id)
+            if self.child_link_name:
+                # the replacement arrives with an empty set, but the children are
+                # still there and still name this item, so the set it had is true
+                register_item[self.child_link_name] = previous[self.child_link_name]
         self.registerItems[register_item_id] = register_item
+        if self.parent_field:
+            # the only insertion point, so no create can leave the parent's set
+            # behind whatever route reached it
+            self.siteregister.updateRegisterLink(
+                self.parent_register, register_item.get(self.parent_field),
+                self.parent_link_name, register_item_id, add=True)
         register_item['datachanges'] = list()
         register_item['datachanges_idx'] = 0
         register_item['subscribed_paths'] = set()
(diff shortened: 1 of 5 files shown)


fporcari (dismissed):
Approved.

Verified on b7394cd6:
- Whole diff read. The unlink now lives in `drop_item`, before the pop and off the child's own parent field, and the link in `addRegisterItem`: O(1) on every page close, which answers the review concern on #1220. `dropRegisterLinks` and `_live_children` are gone and the five readers index the set directly; `drop_page` / `drop_connection` are one line each.
- `addRegisterItem` covers the two replacement cases (the previous parent is unlinked, the children set is carried over) and `reparent_item` refuses a parentless register before mutating anything. `change_connection_user` is the only caller of a parent-field change in the tree (`gnrwebpage_proxy/connection.py`).
- Ratchet: the authority suite run against develop's `siteregister.py` fails 15 tests, the 9 listed in #1223 plus the 6 from the self-review section; all 58 pass on the branch. The four register test files pass under `PYTHONHASHSEED` 0, 1, 42 and 12345 (136 tests); `tests/web` 452 passed, 59 skipped; flake8 clean on the five files.
- Old freeze files: `rebuildRegisterLinks` resets the set on every parent at load, so a restored item never lacks the key that `addRegisterItem` copies on replacement.
- No caller of the removed methods in the repository or in the Sourcerer index; `get_register` is reached only through the SiteRegister remote dispatch.

Two notes, neither blocking:
1. `BaseRegister.update_item(register_item_id, upddict)` is the one route left that can rewrite `connection_id` or `user` on an item without touching the sets. Nothing in the tree calls it, but it is reachable through the remote dispatch like every other method. Either route parent-field keys through `reparent_item` there, or state in a comment that it must not be used for them.
2. `PageRegister.drop(unknown, cascade=True)` and `ConnectionRegister.drop(unknown, cascade=True)` raise `TypeError` on `register_item[...]` when the item does not exist. Pre-existing, not introduced here; the unknown-id tests only cover the no-cascade case.

cgabriel (dismissed):
Approving, verified the removal of what has bene added this morning via #1220, nice cleanup

genro:
Folded in your note 1, in 9fadfb1ee4.

`update_item` now routes a parent field through `reparent_item` and applies the rest of the dict as before. I took the fix rather than the comment because the PR claims no route can change the hierarchy and leave the index behind, and `update_item` was a route — documenting it would have left the claim false. Five tests on it, three verified failing without the change; `tests/web` 440 passed under PYTHONHASHSEED 0, 42 and 12345, full suite 2312 passed, flake8 clean.

Your note 2 I left alone: `drop(unknown, cascade=True)` raising TypeError is pre-existing and not what this branch is about. Opening it separately rather than widening this PR.

The push dismissed both approvals. @fporcari @cgabriel sorry for the second pass — the diff on top of what you read is the six lines in `update_item` plus its tests.

fporcari (approved):
Approved again on 9fadfb1.

The delta on top of b7394cd is exactly what the comment says: five lines in `update_item` that route the parent field through `reparent_item` and apply the rest of the dict as before, plus five tests. Verified: the three tests that exercise the move fail against b7394cd's `siteregister.py` and pass here; the four register test files pass under `PYTHONHASHSEED` 0 and 42 (141 tests); `tests/web` 457 passed, 59 skipped; flake8 clean; merge with develop clean. Taking the fix over the comment was the right call, the claim in the description is now true on every route.

PR #1229, opened 2026-09-02
guarantee temp-dir cleanup regardless of invocation directory

cgabriel:
pytest does not call setup_module/teardown_module defined in a conftest.py for other test modules, so the stubs in app/conftest.py and sql/conftest.py were dead code: app test modules that needed a Genropy environment never got one, and no cleanup was performed when the full suite ran from gnrpy/tests/.

Three changes:

* app/conftest.py: replace the dead setup_module/teardown_module pair with a gnr_test_config autouse fixture (scope=module), mirroring the fixture that sql/conftest.py already has.  The fixture checks GENRO_GNRFOLDER before acting so that test modules with their own setup_module remain undisturbed.

* sql/conftest.py: remove the same dead stubs; update the fixture docstring to reflect the corrected understanding.

* tests/conftest.py (new): add a session-scoped autouse fixture that scans tests/core/ for leftover tmp_* directories at session end and removes them. This is a belt-and-suspenders guarantee that temp dirs never accumulate in the source tree regardless of how the suite is invoked or whether an individual teardown or atexit handler fires.

The change now cleanup correctly the temporary environments created during test via tempfile, avoiding disk usage and dirt left around.

### DIFF
--- a/gnrpy/tests/app/conftest.py
+++ b/gnrpy/tests/app/conftest.py
@@ -29,12 +29,23 @@ def sqlite_temp_dir():
         shutil.rmtree(tmpdir, ignore_errors=True)
 
 
-def setup_module(module):
+@pytest.fixture(scope='module', autouse=True)
+def gnr_test_config():
+    """Ensure every module in this package runs with a genro configuration.
+
+    pytest does not call setup_module/teardown_module defined in conftest.py
+    for test modules, so this autouse fixture fills that role instead.
+    Modules that already provide their own GENRO_GNRFOLDER (via a
+    setup_module of their own) are detected and left alone.
+    """
+    if os.environ.get('GENRO_GNRFOLDER'):
+        yield
+        return
     BaseGnrTest.setup_class()
-
-
-def teardown_module(module):
-    BaseGnrTest.teardown_class()
+    try:
+        yield
+    finally:
+        BaseGnrTest.teardown_class()
 
 
 def _csv_dir():
--- a/gnrpy/tests/conftest.py
+++ b/gnrpy/tests/conftest.py
@@ -1,0 +1,25 @@
+"""Top-level test configuration.
+
+Provides a session-scoped cleanup fixture that removes any leftover
+tmp_* directories from tests/core/ at the end of the test session.
+These directories are created by BaseGnrTest.setup_class() and should
+be removed by each test's teardown_class() or its atexit backstop, but
+this fixture is a final guarantee that they never accumulate in the
+source tree regardless of how the test suite is invoked.
+"""
+
+import os
+import shutil
+
+import pytest
+
+
+@pytest.fixture(scope='session', autouse=True)
+def _cleanup_core_tmp_dirs():
+    yield
+    core_dir = os.path.join(os.path.dirname(__file__), 'core')
+    if not os.path.isdir(core_dir):
+        return
+    for entry in os.listdir(core_dir):
+        if entry.startswith('tmp_'):
+            shutil.rmtree(os.path.join(core_dir, entry), ignore_errors=True)
--- a/gnrpy/tests/sql/conftest.py
+++ b/gnrpy/tests/sql/conftest.py
@@ -25,19 +25,13 @@ def sqlite_temp_dir():
     finally:
         shutil.rmtree(tmpdir, ignore_errors=True)
 
-def setup_module(module):
-    BaseGnrTest.setup_class()
-def teardown_module(module):
-    BaseGnrTest.teardown_class()
-
-
 @pytest.fixture(scope='module', autouse=True)
 def gnr_test_config():
     """Make sure every module in this package runs with a genro configuration.
 
-    The setup_module/teardown_module above are never called: pytest honours
-    xunit module hooks on test modules, not on a conftest.  Every module that
-    needs a configuration therefore repeats the call itself -- and
+    pytest does not call setup_module/teardown_module defined in conftest.py
+    for test modules.  Every module that needs a configuration therefore
+    repeats the call itself -- and
     test_db_notify and test_gnrlistener do not, so on a machine without a
     ~/.gnr of its own, CI included, their fixtures died with
     'Missing genro configuration'.  Modules that already set one up keep it.


fporcari (commented):
Two findings from running this on b9b2eef, one of them worth a small change before merge.

1. The guard in `app/conftest.py::gnr_test_config` cannot see a module's own `setup_module`: autouse conftest fixtures are set up before the module's xunit fixture, so `GENRO_GNRFOLDER` is still unset when the check runs. Traced on `tests/app/test_api_engine.py` with `--setup-show` and a mkdtemp/rmtree tracer: `gnr_test_config` builds a config (dir A), then `setup_module` builds a second one (dir B); `teardown_class` reaps B only, and A is reaped by the new session fixture in `tests/conftest.py`. Same for `test_gnrdbo_df.py`. So the docstring's "detected and left alone" does not hold, and the session cleanup is not belt-and-suspenders here: it is what removes the leak the fixture itself creates. Two ways out, either is fine: drop the now redundant `setup_module`/`teardown_module` from those two modules (the fixture provides the environment), or detect the module hook instead of the env var (`getattr(request.module, 'setup_module', None)`). `sql/conftest.py` has the same latent guard, pre-existing.

2. The claim that app modules "never got an environment": `tests/app` passes on develop with `GENRO_GNRFOLDER` unset and an empty `HOME` (161 passed), so nothing in `tests/app` fails today for lack of one. The fixture is a harmless mirror of the sql one, not a fix of a failure; the description could say so. The invocation directory does not change the cleanup either: `BaseGnrTest` paths are `__file__`-relative and `tests/app` from `gnrpy/` and from `gnrpy/tests/` both leave nothing behind on develop.

What does hold: a full develop run under that clean environment left 9 `tmp_*` directories in `tests/core`, and the same run on this branch leaves 0. That is the value of `tests/conftest.py`, and it is worth having. flake8 clean on the three files; merge with develop clean.

genro (commented):
Verified on develop at 919a3a57 and on the branch.

The session fixture does what the description says: `tests/sql` leaves 9 directories in `tests/core` per run on develop — 62 before, 71 after — and 0 on this branch. That is worth having, and the two dead `setup_module`/`teardown_module` pairs were indeed dead.

Confirming @fporcari's first finding independently, and tracing it to its cause.

The guard cannot abstain. A conftest autouse fixture is set up **before** the module's xunit `setup_module`, so `GENRO_GNRFOLDER` is still unset when the check runs. Minimal case: the fixture prints the variable empty, `setup_module` sets it a moment later. The docstring's claim — "Modules that already provide their own GENRO_GNRFOLDER are detected and left alone" — does not hold.

What it costs: `tmp_conf_dir` is class-level state, so a second `setup_class()` before the matching teardown overwrites the path and orphans the first directory. The `atexit` backstop reads the current value, so it reaps the last one and never the orphans:

```
setup_class()   -> tmp_h8oyr0er
setup_class()   -> tmp_0jpfmn4b
teardown_class()
tmp_0jpfmn4b removed, tmp_h8oyr0er still on disk
```

That is the whole leak: one orphan per module that builds its own environment while the fixture has already built one. Nine such modules in `tests/sql`, nine directories. `tests/app` has two, so this PR adds two new leaks and then sweeps them up — the session fixture is not belt-and-suspenders here, it is the counterpart of the leak the module fixture creates.

**Requested before merge:** the change @fporcari asked for — detect the module hook (`getattr(request.module, 'setup_module', None)`) or drop the redundant `setup_module` from the two app modules — and a docstring that matches. `sql/conftest.py` carries the same guard and the same nine leaks; either fix it here or in a follow-up.

**Suggestion, not blocking:** `setup_class` writes into the repository — `mkdtemp(prefix=f"{cls.local_dir}/tmp_")`. `tmp_conf_dir` is only ever used as an absolute path (the `gnr/` config dir, the instances path, a `<custom path=...>` in the generated `environment.xml`), so nothing requires it under `tests/core`. A plain `mkdtemp()` puts it in the system temp dir, where a leak is the operating system's problem and never reaches `git status` — and `tests/conftest.py` stops being needed. Also note that a pattern sweep only sees `tmp_*`: a working tree here holds 158 leftovers, 93 of them without the prefix, from before #1025 added it on 13 Aug.

flake8 clean on the three files; CI green.

cgabriel:
Confirmed the guard can never fire since autouse fixtures run before a module's own setup_module.

Went with dropping the redundant setup_module/teardown_module pairs rather than detecting the hook: since the fixture already builds the environment for every module, keeping a code path that special-cases modules with their own hook is redundant once none exist. Removed the dead pairs from test_api_engine.py and test_gnrdbo_df.py, and simplified gnr_test_config in app/conftest.py to build/teardown unconditionally, with the docstring updated to match.

Applied the identical fix to sql/conftest.py in this same PR rather than a follow-up: same broken guard, same orphaning mechanism, and dropped the matching dead setup_module/teardown_module pairs from the 10 affected modules there too.

Verified: tests/app full suite green (198 passed), tests/sql green everywhere, zero leftover tmp_* in tests/core after both runs, flake8 clean on all touched files.

Left the mkdtemp()-into-tests/core-vs-system-temp-dir suggestion out of scope for this PR since it was flagged as non-blocking — happy to pick it up separately if you'd rather have it now.

fporcari (approved):
Approving, with one thing to record rather than to change.

The diagnosis is the useful part: `setup_module`/`teardown_module` in a `conftest.py` are never called by pytest, and every module in the package was repeating the call to work around it. Replacing that with one module-scoped autouse fixture per package removes the duplication from fifteen files and puts the configuration where it cannot be forgotten by the next test module — which is how `test_db_notify` and `test_gnrlistener` ended up without one. The session-scoped sweep in the new top-level `conftest.py` is a backstop rather than a fix, and it is written as one.

**The one change worth stating in the description.** In `gnrpy/tests/sql/conftest.py` this also drops the early return:

```python
if os.environ.get('GENRO_GNRFOLDER'):
    yield
    return
```

From here on a run that sets its own `GENRO_GNRFOLDER` no longer uses it: `BaseGnrTest._build_test_env` overwrites the variable with its own temp configuration and restores the previous value at teardown. That is arguably the better behaviour — the suite becomes hermetic instead of depending on whatever configuration the caller happens to have — but it is a behaviour change for anyone running the suite from a worktree with a scratch configuration of their own, and the description does not mention it. Worth a line there, so the next person who finds their `GENRO_GNRFOLDER` ignored does not go looking for a bug.

Not verified: I did not run the suite, and I did not check whether any module outside the two packages touched here still relies on the old module hooks.

fporcari:
Adding what my approval should have said: I wrote it without first reading this PR's own history, and the history contains a blocking ask.

@genro's review of 2 September ends with a **"Requested before merge"**: drop the redundant `setup_module`/`teardown_module` from the two app modules, or detect the module hook instead of the environment variable, and make the docstring match.

That ask is satisfied by the current branch, which is why the approval stands: `test_api_engine.py` and `test_gnrdbo_df.py` both lose their `setup_module`/`teardown_module`, the same happens across `tests/sql`, and the new docstrings no longer claim that a module providing its own configuration is detected and left alone — they say the fixture fills the role the conftest hooks never did. The leak mechanism the two of you traced — a second `setup_class()` overwriting class-level `tmp_conf_dir` and orphaning the first directory — has no occurrences left once nothing builds an environment twice.

One consequence of that, which my review framed as the only change and which is really the smaller half: with the guard gone, an **ambient** `GENRO_GNRFOLDER` — one exported before pytest runs, rather than set by a module hook — is no longer honoured either. The guard could never see a module hook, for the ordering reason @genro established, but it could see that one. It now gets overwritten for the module's duration and restored at teardown. Still worth a line in the description.

@genro's non-blocking suggestion — `mkdtemp()` in the system temp dir instead of `mkdtemp(prefix=f"{cls.local_dir}/tmp_")`, which would make `tests/conftest.py` unnecessary — is untouched here and is the better long-term shape. It belongs in its own issue rather than in this PR.

PR #1234, opened 2026-09-02
fix(storage): skip the <services><storage> container in the flat service scan

### LINKED ISSUE #1233: [BUG] The <services><storage> container node is registered as a phantom storage service
### Problem Description

`BaseStorageHandler._loadStorageParametersFromSiteConfig` registers the `<storage>` container node of the `<services>` section as if it were a storage service.

The method reads the nested form from `self.site.config['services.storage']`, and then also scans every child of `<services>` looking for flat services:

```python
for service_name, attrs in services.digest('#k,#a'):
    service_type = attrs.pop('service_type', None) or service_name
    if service_type == 'storage':
        ...
```

For the container node itself `service_name` is literally `'storage'`, so `or service_name` makes `service_type` equal to `'storage'` and the container is registered as a service.

Observed: `storage_params` contains a bogus `storage` entry with no implementation, besides the real services.
Expected: only the services declared inside the section are registered.

Harmless today (nothing resolves that mount), but it shows up in any listing of the configured storages, and `site.storage('storage')` would silently fall back to a local storage under `site_static_dir` instead of failing.

### Reproduction Steps

1. Declare a storage service in `siteconfig.xml` using the nested form documented by the method itself:

```xml
<services>
  <storage>
    <minio implementation="aws_s3" bucket="sandbox"/>
  </storage>
</services>
```

2. Inspect `site.storage_handler.storage_params`.

Observed result:

```python
{'minio': {'bucket': 'sandbox', 'implementation': 'aws_s3'}, 'storage': {}, ...}
```

The `storage` entry is the container node, not a configured service.

Module involved: `gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py`, `BaseStorageHandler._loadStorageParametersFromSiteConfig` (the flat-structure loop that follows the nested one).

The same either/or handling is done correctly in `ServiceHandler.serviceConfigurationsFromSiteConfig` (`gnrpy/gnr/lib/services/__init__.py`), which scans the flat children only when the type section is absent.

### Environment

genropy `develop` (26.09.01), Python 3.13, macOS — framework level, environment independent.

### Additional Material

The docstring of the method contributes to the confusion: its nested example shows the container node carrying the service attributes (`<storage service_name="my_storage" implementation="local">`), while the code reads the children of `<storage>` as the services.

fporcari:
Fixes #1233

`BaseStorageHandler._loadStorageParametersFromSiteConfig` reads the nested form from `services.storage` and then also scans every child of `<services>` for flat services, defaulting `service_type` to the node key. For the type section itself that key is literally `storage`, so the container node was registered as a phantom storage mount with no implementation: it showed up in `storage_params`, and `site.storage('storage')` would silently fall back to a local storage under `site_static_dir`.

The flat loop now skips that key. Both loops keep running: making it either/or as `ServiceHandler.serviceConfigurationsFromSiteConfig` does would drop flat `service_type="storage"` siblings whenever a nested section is present, which the storage handler supports today.

The docstring's nested example is corrected as well — it showed the container node carrying the service attributes, which is the ambiguity the code reads as a service.

### Tests

- `test_load_storage_from_siteconfig_nested_and_flat_structure`: a nested `minio` plus a flat `my_flat_store`, both registered, no `storage` key.
- The negative assertion added to the existing `test_load_storage_from_siteconfig_nested_structure`.

### Verification

- Repro confirmed against the unpatched code: `{'minio': {...}, 'storage': {}, 'my_flat_store': {...}}`; with the fix the `storage` entry is gone.
- `flake8` clean on both files; full `gnrpy` suite 2283 passed, 70 skipped.
- Note: `TestStorageHandler` inherits `BaseGnrDaemonTest`, so the whole class skips where the `gnrdevelop` site cannot be built (it does here). The new test bodies were driven directly against the patched module to check them; they are not verified through a normal pytest run in that condition.

### DIFF
--- a/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py
+++ b/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py
@@ -171,6 +171,15 @@ class BaseStorageHandler:
             for service_name, attrs in services.digest('#k,#a'):
                 if service_name == 'storage':
                     # <services><storage> is the type section handled above, not a service
+                    if attrs and attrs.get('service_name'):
+                        # the shape this docstring used to document: an instance that
+                        # followed it had the mount registered under the label 'storage',
+                        # and would now lose it without a word
+                        logger.warning(
+                            "<services><storage service_name=%r> is the storage type section, "
+                            "not a service: the mount is not loaded. Declare it as a child, "
+                            "<storage><%s implementation=... /></storage>",
+                            attrs['service_name'], attrs['service_name'])
                     continue
                 attrs = dict(attrs) if attrs else {}
                 service_type = attrs.pop('service_type', None) or service_name
--- a/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py
+++ b/gnrpy/gnr/web/gnrwsgisite_proxy/gnrstoragehandler.py
@@ -144,8 +144,8 @@ class BaseStorageHandler:
 
         Reads from siteconfig.xml services section:
         <services>
-            <storage service_name="my_storage" implementation="local">
-                <base_path>/path/to/storage</base_path>
+            <storage>
+                <my_storage implementation="local" base_path="/path/to/storage"/>
             </storage>
             <my_s3 service_type="storage" implementation="aws_s3" bucket="my-bucket" />
         </services>
@@ -169,6 +169,9 @@ class BaseStorageHandler:
                     self._setStorageParams(service_name, parameters=attrs, implementation=implementation)
             # Also check flat structure: <services><my_storage service_type="storage" .../></services>
             for service_name, attrs in services.digest('#k,#a'):
+                if service_name == 'storage':
+                    # <services><storage> is the type section handled above, not a service
+                    continue
                 attrs = dict(attrs) if attrs else {}
                 service_type = attrs.pop('service_type', None) or service_name
                 if service_type == 'storage':
--- a/gnrpy/tests/web/gnrstoragehandler_test.py
+++ b/gnrpy/tests/web/gnrstoragehandler_test.py
@@ -551,6 +551,47 @@ class TestStorageHandler(BaseGnrDaemonTest):
         assert params.get('implementation') == 'local'
         assert params.get('base_path') == '/tmp/nested'
 
+        # The <storage> container node is not a service itself
+        assert 'storage' not in handler.storage_params
+
+    def test_load_storage_from_siteconfig_nested_and_flat_structure(self):
+        """Test that nested and flat storage services coexist without a phantom 'storage' mount."""
+        from gnr.web.gnrwsgisite_proxy.gnrstoragehandler import BaseStorageHandler
+
+        # Create a mock site with both a nested section and a flat storage service
+        class MockSite:
+            def __init__(self):
+                self.config = Bag()
+                self.config['services.storage.minio'] = Bag()
+                self.config.setAttr('services.storage.minio',
+                    implementation='aws_s3', bucket='sandbox')
+                self.config['services.my_flat_store'] = Bag()
+                self.config.setAttr('services.my_flat_store',
+                    service_type='storage', implementation='local', base_path='/tmp/flat')
+                self.site_static_dir = '/tmp/test_site'
+
+            @property
+            def gnrapp(self):
+                class MockApp:
+                    packages = type('obj', (object,), {'keys': lambda self: []})()
+                return MockApp()
+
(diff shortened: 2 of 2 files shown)


genro (dismissed):
Correct fix in the right place: the flat scan in `gnrstoragehandler.py:171-179` treated the `<storage>` container as a service keyed by its own label, and the skip at :172-174 removes exactly that node without making the handler either/or, which would have dropped flat entries with `service_type="storage"`. Docstring matches the code.

cgabriel (changes requested):
I'm worried about existing deployment assuming this behaviour, I suggest if the 'storage' key is being hit with a service_name attribute to drop a logger.warning about the skip

fporcari:
Done in `fa3efd8598`
([review](https://github.com/genropy/genropy/pull/1234#pullrequestreview-5099402430)):

> I'm worried about existing deployment assuming this behaviour, I suggest if the 'storage' key is being hit with a service_name attribute to drop a logger.warning about the skip

The worry is founded, and more specifically than "some deployment might": the
docstring **this very PR rewrites** documented that shape.

```xml
<!-- what the docstring said before this PR -->
<storage service_name="my_storage" implementation="local">
    <base_path>/path/to/storage</base_path>
</storage>
```

An instance that followed it got the mount registered by the flat scan under
the label `storage`, since that is the node's key. After the skip it gets
nothing, so the warning has to name what was dropped:

```python
if service_name == 'storage':
    if attrs and attrs.get('service_name'):
        logger.warning(
            "<services><storage service_name=%r> is the storage type section, "
            "not a service: the mount is not loaded. Declare it as a child, "
            "<storage><%s implementation=... /></storage>",
            attrs['service_name'], attrs['service_name'])
    continue
```

Guarded on `service_name` as you asked, so a plain `<storage>` container — the
shape the docstring now documents — stays quiet. The message names the mount and
gives the new spelling, so whoever reads the log has the migration in the line.

## Verification, and something worth knowing

`tests/web/gnrstoragehandler_test.py` **49 passed**. New test
`test_load_storage_from_siteconfig_old_container_shape_is_warned`, with the
ratchet checked: reverting the warning makes it fail, the other six siteconfig
tests stay green. flake8 clean.

Getting those numbers needed the base merged in, which is in this push as
`a4ae483117`: this branch predated #1036, so the whole suite answered
`SKIPPED [1] Daemon not available: 'NoneType' object has no attribute
'services_handler'` — the exact message #1036 was opened about. With #1036 on
develop the 49 tests actually run. Worth knowing for anyone else rebasing a
branch that touches this file: before today the storage handler suite was not
being verified by anyone, here or on CI.

@genro your approval fell with this push. What is added since it is the warning
above plus its test; the skip you approved at `:172-174` is untouched, and the
merge brings only develop.

cgabriel (approved):
Looking good now, a little helper to handle misconfiguration. Approving.

genro (approved):
Re-approving on `662dae241`. My earlier approval fell to the push that added the warning, and the warning is the only thing that moved: the skip itself, the docstring correction and the two tests are unchanged from what I read on `9dfb889ad` — I diffed the two files this PR owns across the two heads, and the warning plus one new test is the whole delta.

The delta is right. `attrs and attrs.get('service_name')` fires only on the shape the old docstring documented — a `<storage>` container carrying the service attributes — so a plain nested section stays silent, and the instance that would otherwise lose its mount without a word now gets it named. `logger` is `gnr.web` (`gnrstoragehandler.py:22`), the same one the `volumes` deprecation already writes to, and the format arguments match the placeholders.

Verified at this head, merged with `develop` at `d882af9349`:

- `tests/web/gnrstoragehandler_test.py` — 67 passed. Worth correcting one line of the PR body: `TestStorageHandler` does **not** skip here. All three siteconfig tests, the two new ones included, run and pass. Whatever made the class skip on your machine is local, so the coverage you were unsure about is real coverage.
- flake8 clean on both files.
- No conflict with the current `develop`.

The reason not to make the two loops either/or still holds, and it is the part of this PR that would be easiest to get wrong: `ServiceHandler.serviceConfigurationsFromSiteConfig` can afford the exclusive form because it has no flat siblings to lose, while dropping the flat scan here would silently discard every `service_type="storage"` entry whenever a nested section is present.

fporcari:
Correction accepted: the body's line about `TestStorageHandler` skipping here is wrong. All three siteconfig tests run, the two new ones included, 67 passing in the file — so the coverage I was unsure about is real coverage, and there is no gap to note. Whatever made the class skip was local to my machine.

Squash-merging this now: the branch carries two merge commits from develop, so the history is worth flattening.

Issue #1237, opened 2026-09-02
[BUG] StorageNode.open() runs autocreate even for reads: 3 S3 requests where 1 would do


fporcari:
## Problem Description

`StorageNode.open()` calls the autocreate helper unconditionally, so opening a
file **for reading** creates its parent directory — or, on a remote service,
pays the round trips to find out whether it has to.

`gnrpy/gnr/lib/services/storage.py:357`:

```python
def open(self, mode='rb'):
    """Is a context manager that returns the open file pointed"""
    self.service.autocreate(self.path, autocreate=-1)
    ...
```

`StorageService.autocreate()` builds a `StorageNode` for the parent directory,
reads its `.exists`, and calls `makedirs()` when it is missing. For a read that
work is never needed: either the file is there, and so is its parent, or the
open is going to fail anyway.

`local_path()` (`storage.py:400`) does the same thing, and it is a read path far
more often than a write one.

Expected behaviour: autocreate only when the mode actually writes
(`w`, `a`, `x`, or `+`).

### What it costs

On a local service the cost is an `os.path.exists`, so it is invisible. On
`aws_s3` it is two HTTP requests per open, because `autocreate` → `.exists`
→ `isfile()` + `isdir()` map onto `HeadObject` + `ListObjectsV2`:

| operation | S3 API calls | of which the actual read |
|---|---|---|
| `node.open('rb')` + read | 3 | 1 (`GetObject`) |

Measured against a local MinIO, counting at
`botocore.client.BaseClient._make_api_call`:

```
read 4KB    3 calls  GetObject x1, HeadObject x1, ListObjectsV2 x1
```

Two thirds of the requests of every read are the autocreate. On a local
endpoint that is ~1.3 ms of the ~2.2 ms a small read takes; against a real
region, at ~30 ms per round trip, a read costs ~90 ms instead of ~30, and the
`ListObjectsV2` is a billed request.

There is a second, more subtle effect on a read-only mount: the legacy `aws_s3`
service returns early from its write paths when `readonly` is set, but the
autocreate on a read still issues its two requests to decide it has nothing to
do.

## Reproduction Steps

Any S3-compatible endpoint (a local MinIO is enough). With a storage service of
implementation `aws_s3` configured as `st`:

```python
import collections
import botocore.client

CALLS = collections.Counter()
_make = botocore.client.BaseClient._make_api_call


def counted(self, operation_name, api_params):
    CALLS[operation_name] += 1
    return _make(self, operation_name, api_params)


[...]

Issue #252, opened 2025-10-29
🐛 genro-storage adapter: Missing symbolic path resolution for rsrc:, pkg:, page:, etc.


fporcari:
## Description

The genro-storage adapter currently does not handle symbolic storage paths correctly. Services like `rsrc:`, `pkg:`, `page:`, `conn:`, `dojo:`, `user:`, `temp:`, etc. are configured with `implementation='symbolic'` but the adapter incorrectly maps them to local storage without implementing the dynamic path resolution logic.

## Problem

In the native storage system, symbolic paths are resolved dynamically by dedicated methods in `resources/common/services/storage/symbolic.py`:

- `rsrc:common/icons/save.png` → resolves to the actual resource folder path
- `pkg:mypackage/data/file.txt` → resolves to the package folder  
- `page:mydata.txt` → resolves to `site_path/data/_connections/{connection_id}/{page_id}/mydata.txt`
- `conn:uploads/file.pdf` → resolves to `site_path/data/_connections/{connection_id}/uploads/file.pdf`
- `user:documents/doc.pdf` → resolves to `site_path/data/_users/{user}/documents/doc.pdf`
- etc.

The current adapter in `gnr/lib/services/storage_genro_adapter.py` incorrectly treats these as simple local paths:

```python
IMPLEMENTATION_MAP = {
    'symbolic': 'local',  # ❌ Wrong: treats symbolic as local
    ...
}
```

## Impact

When `storage_backend=genro-storage` is enabled, all symbolic paths fail to resolve correctly:
- Resources (`rsrc:`) cannot be found
- Package files (`pkg:`) are inaccessible  
- Page/connection storage (`page:`, `conn:`) don't work
- User directories (`user:`) fail
- Temporary storage (`temp:`) doesn't resolve

This breaks core Genropy functionality.

## Root Cause

The symbolic service has special path resolution methods that are not implemented in the adapter:
- `path_rsrc(resource_id, *args)` - looks up in `site.resources`
- `path_pkg(pkg, *args)` - resolves to package folder
- `path_page(connection_id, page_id, *args)` - resolves to connection/page data
- `path_conn(connection_id, *args)` - resolves to connection data
- `path_dojo(version, *args)` - resolves to dojo library path
- `path_user(user, *args)` - resolves to user data directory
- `path_temp(*args)` - resolves to system temp directory
- `path_vol(volume, *args)` - resolves to configured volumes

[...]

Issue #251, opened 2025-10-28
🧪 Testing Request: genro-storage Integration (feature/genro-storage-integration)


genro:
## 🎯 Testing Request

The **`feature/genro-storage-integration`** branch is ready for testing\! This branch integrates the `genro-storage` library into Genropy while maintaining full backward compatibility with existing code.

### 1. Test Suite for Native StorageNode API
- **46 tests** documenting and verifying the current API
- File: `gnrpy/tests/core/gnrstorage_test.py`
- Status: ✅ **46/46 tests passing**

### 2. genro-storage Adapter
- **Complete wrapper** with Genropy API
- **Config converter** (Genropy config → genro-storage format)
- **Runtime switch** to choose native vs genro-storage
- File: `gnrpy/gnr/lib/services/storage_genro_adapter.py`

### 3. Wrapper Test Suite
- **32 tests** verifying 100% compatibility
- File: `gnrpy/tests/core/gnrstorage_wrapper_test.py`
- Status: ✅ **32/32 tests passing**

### Setup

```bash
# 1. Checkout the branch
git checkout feature/genro-storage-integration

# 2. Install genro-storage
pip install git+https://github.com/genropy/genro-storage.git

# 3. Run tests
cd gnrpy
export PYTHONPATH=$PWD:$PYTHONPATH

# Native tests
pytest tests/core/gnrstorage_test.py -v

# Wrapper tests
pytest tests/core/gnrstorage_wrapper_test.py -v

# All storage tests
pytest tests/core/gnrstorage*.py -v
```

### Expected Results
- Native tests: **46/46 ✅**
- Wrapper tests: **32/32 ✅**
- **TOTAL: 78/78 tests passing**

## 🧩 Runtime Switch Configuration

The adapter allows runtime selection of backend:

### Option 1: Global (in site config)
```python
STORAGE_BACKEND = 'genro-storage'  # or 'native'
```

### Option 2: Per-service override
```python
storage.configure([
    {'name': 'uploads', 'type': 's3', 'backend': 'genro-storage'},  # new
    {'name': 'legacy', 'type': 'local', 'backend': 'native'}  # old
])
```

### Option 3: Automatic fallback
If genro-storage is not installed, automatically falls back to native.

### ✅ Implemented and Tested
- Properties: `fullpath`, `basename`, `ext`, `size`, `mtime`, `isfile`, `isdir`, `exists`
- File operations: `open()`, `read`, `write`, `delete()`
[...]

PR #1238, opened 2026-09-02
feat(storage): optional genro-storage backend behind a default-off switch

fporcari:
Draft on purpose: the code is complete and green, but two round-trip defects
found by benchmarking this integration are still open upstream. See
**Before this leaves draft** at the bottom.

Refs #251. Related: #252 (symbolic mapped to local — this branch does not do
that, see below), #1079 (boto3 with no upper bound, which is why the dependency
is optional).

## What this does

Serves the replaceable part of the legacy storage layer through the
`genro-storage` package, behind a siteconfig switch that is **off by default**.
The legacy code is untouched and stays the default; nothing legacy is removed.

```xml
<storage use_genro_storage="True"/>
```

The design turned out smaller than planned: instead of adapting the node, this
adds a **`StorageService` whose backends are genro-storage**
(`gnr/lib/services/storage_genro.py`), with the legacy `StorageNode` unchanged
on top of it. The legacy node therefore keeps its `isinstance` identity, its
`.service`, and the `serve()`, `mkdir()`, `base64`, `internal_url`, `listdir`,
`autocreate`, `copy`/`move` behaviour it already has — including the bridging of
a copy between the two worlds, which `StorageService` does by content on its
own. `GenroStorageHandler` overrides exactly one method, `storage()`: which
service resolves a mount is the single point of variation.

| implementation | with the switch on |
|---|---|
| `local`, `raw`, `aws_s3` | served by genro-storage |
| `symbolic` (`rsrc`, `pkg`, `page`, `conn`, `user`, `temp`, `dojo`, `gnr`, `pages`), `http`, `compressed`, `vol:`, any unknown name | **stay legacy, by design** |
| `relative`, `sftp` | mappable, deferred: stay legacy |

`symbolic` stays legacy permanently and deliberately: it is nine dynamic
resolvers keyed on the mount name, three of them request-scoped, with no
`base_path` at all — things a storage library cannot know. This is the point
#252 got wrong by mapping `symbolic` to `local`; there is an explicit test that
`rsrc:`, `pkg:`, `temp:` and `gnr:` resolve to the very same path as on the
legacy handler with the switch on.

A mount is skipped with a warning, and stays legacy, when it is `readonly`
(genro-storage has no counterpart, and silently dropping the restriction would
be worse), when it has no bucket, or when it is local on a `base_path` that does
not exist yet (genro-storage's local backend requires the directory; the legacy
service creates it on first write — upstream `genropy/genro-storage#75`).

## Dependency

Optional extra `genro_storage` in `gnrpy/pyproject.toml`, plus genro-storage in
`developer` so the tests can import it. `s3fs` is deliberately **not** a base
dependency: it pulls `aiobotocore`, which caps `botocore<1.40.19`, while genropy
declares `boto3` with no upper bound — that is #1079, and making it mandatory
would put it on the critical path of every install.

`pip check` is unchanged by this branch: same five pre-existing lines before and
after (two of them are the `s3fs`/`gcsfs` exact pin on `fsspec`, an environment
matter that predates this work).

## Tests

- `gnrpy/tests/core/gnrstorage_compare_test.py` — one test body per operation,
  run **twice** through a mode fixture, over the whole legacy `StorageNode`
  surface. Real services on a real filesystem, and on a real S3 endpoint when
  `GNR_TEST_S3_ENDPOINT` is set; only the site is stood in for, never a storage
  operation. Where the two worlds legitimately differ, the difference is
  asserted and named rather than smoothed over.
- `gnrpy/tests/web/gnrgenrostoragehandler_test.py` — the switch (off by default,
  and each test sets the value it needs instead of assuming the local
  siteconfig), mount routing, registry sync, and the legacy-stays-legacy
  guarantee.
- `gnrpy/tests/core/gnrstorage_benchmark.py` — the same operations timed in both
  modes, `pytest -s`.
- `projects/gnrcore/packages/test15/webpages/tools/storage_genro.py` — a
  testhandler page that drives the switch through real requests: storage trees
  on both kinds of mount, an inspector over the node surface, a side-by-side of
  the two handlers on the same path, an uploader, a cross-world copy/move, a
  tree on an S3 mount.

Measured on this branch:

| | |
|---|---|
| switch off (the default) | `2469 passed, 169 skipped` — no test went passed→failed, no new skip beyond the S3 ones that skip without an endpoint |
| comparison tests, local | 94 passed, 0 skipped (both modes) |
| comparison tests, MinIO | 188 passed with `GNR_TEST_S3_ENDPOINT` set; 94 passed + 94 skipped, reason naming the variable, without it |
| handler routing | 29 passed, including the nine symbolic mounts and `_http_` |

CI has no S3 endpoint, so that half skips there with an explicit reason.

## Benchmark

Full tables, with the machine and the versions, in
`docs/development/genro-storage-migration.md` §11. The short version: **the
overhead is per call, not per byte**. On a local mount a metadata call goes from
0.002 ms to 0.047 ms (the 22x ratio), while on a 4 MB file the ratio collapses
to ~1. On S3 the writes are faster (a single `PutObject` against smart_open's
multipart), metadata is a wash, and small reads are slower. `md5hash` on S3 is
not a like-for-like comparison: the legacy service gives up on a multipart ETag
and returns `None`.

## Driving it in a real site found two defects

Both fixed here, both with a test:

- `children()` was not sorted, while the legacy local service sorts by basename
  — the order shows up in every tree built on the service.
- `size` and `md5hash` raised on a directory instead of answering `None`, which
  is what the legacy `aws_s3` service already does.

## Before this leaves draft

- `genropy/genro-storage#78` — on S3, `copy_to()` issues 6 API calls where one
  `CopyObject` would do (two of them `ListObjectVersions`, which a copy does not
  need), and `open('rb')` adds a `HeadObject`. The copy does stay server-side,
  so it is round trips and not a transfer, but on a real endpoint it is ~180 ms
  instead of ~30. Our `duplicateNode` could sidestep it by going through the
  backend's `copy()`; not done here, since the upstream fix would make it
  unnecessary.
- #1237 — `StorageNode.open()` runs autocreate even for reads, costing a
  `HeadObject` plus a `ListObjectsV2` per open on S3. Pre-existing and on the
  legacy path too, but it inflates every read on both sides.

Also worth a decision before merging, and written up in the plan document:
whether the hybrid of the table above is the intended end state (this branch
assumes it is), and whether `relative` and `sftp` should come in here or later.

Not yet exercised in a live site: the upload path itself (it needs a file picked
by hand) and multidomain.

### DIFF
--- a/docs/development/envvars.py
+++ b/docs/development/envvars.py
@@ -8,6 +8,11 @@ variables = {
             "GNR_TEST_PG_PORT": "Database server port for tests",
             "GNR_TEST_PG_USER": "Database server user for tests",
             "GNR_TESTING_INSTANCE_NAME": "Name of the instance when running tests",
+            "GNR_TEST_S3_ENDPOINT": "S3-compatible endpoint (e.g. a local MinIO at http://127.0.0.1:9000) for the storage comparison tests and benchmark; when unset those tests skip",
+            "GNR_TEST_S3_ACCESS_KEY": "Access key for GNR_TEST_S3_ENDPOINT",
+            "GNR_TEST_S3_SECRET_KEY": "Secret key for GNR_TEST_S3_ENDPOINT",
+            "GNR_TEST_S3_BUCKET": "Bucket the storage tests write to (default: sandbox)",
+            "GNR_TEST_S3_PREFIX": "Key prefix the storage tests create and delete under (default: gnrtest)",
     },
     "Deployment": {
         "GNR_RMS_CODE":"TBD",
--- a/docs/development/genro-storage-migration.md
+++ b/docs/development/genro-storage-migration.md
@@ -8,6 +8,20 @@ Goal: serve the replaceable part of the legacy storage layer through genro-stora
 behind a switch that is **off by default**, with the legacy code left intact and still
 the default. Nothing legacy is removed on this branch.
 
+> **Implementation note (§4.1 point 10, resolved further than planned).** The plan
+> called for a node adapter subclassing `StorageNode` over a service-shaped shim.
+> Implementing it showed the stronger form of the same idea: a **`StorageService` whose
+> backends are genro-storage**, with the legacy `StorageNode` unchanged on top. The
+> legacy node then keeps its own `isinstance` identity, `.service`, `serve()`, `mkdir()`,
+> `base64` shape, `internal_url`, `listdir`, `autocreate`, and the `StorageService`
+> copy/move machinery bridges the two worlds by content on its own. So there is no node
+> adapter and no bridging code: `gnr/lib/services/storage_genro.py` is the whole
+> variation, and `GenroStorageHandler` overrides exactly one method, `storage()`.
+> Consequences: planned Phase 3 collapsed into Phase 2, planned Phase 5 (cross-world
+> copy/move) needed no code — only tests — and planned Phase 8 (`ep_table` version keys)
+> became unnecessary, because the service reports `versions()` with the legacy boto3 key
+> names instead.
+
 Everything marked *(measured)* below was executed in this session on this machine
 (macOS 25.0.0, pyenv 3.13.2, MinIO at `http://127.0.0.1:9000`, bucket `sandbox`).
 Everything else is read from source.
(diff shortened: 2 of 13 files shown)


fporcari:
## Update since this left draft

Three things changed after the description was written. The **Benchmark** section of the
description predates them: the numbers below supersede it, and
`docs/development/genro-storage-migration.md` §11 carries the full tables.

### The benchmark was not measuring the operation that matters

It timed `exists`, `size`, `mtime` and `md5hash` separately but never `ext_attributes` —
the one call `StorageResolver` (`gnr/lib/services/storage.py:914`) makes per child of a
listing. So the operation that decides how fast a storage tree renders was the one
operation not measured. Two rows added: the single call, and the resolver's own shape
over a 100-file directory.

They change the reading of the table. **On an S3 mount a tree is ~128x faster with the
switch on** — 656 ms against 5.1 ms per 100-file listing. The legacy `aws_s3` service
issues a `HeadObject` per child, while s3fs answers each child from the listing it has
just made. Since `StorageResolver` is exactly this shape, an S3 folder in the UI is
where a user would feel the switch first, and in their favour.

Measured on genro-storage 0.8.1, MinIO on localhost, three runs:

| S3 mount | reps | legacy | genro | ratio |
|---|---|---|---|---|
| tree: children + ext_attributes | 5 | 3281 ms | 25.5 ms | **0.01** |
| ext_attributes | 50 | 80.1 ms | 31.8 ms | 0.40 |
| children (100 files) | 10 | 64.7 ms | 25.3 ms | 0.39 |
| write small (4KB) | 10 | 53.6 ms | 27.9 ms | 0.52 |
| md5hash | 50 | 30.3 ms | 94.1 ms | 3.11 |
| copy same mount | 10 | 32.1 ms | 63.0 ms | 1.97 |

On a **local** mount the overhead stays per call, not per byte: a metadata call goes from
0.002 ms to ~0.05 ms, while on a 4 MB file the ratio collapses to ~1. The one local row
that can reach a user is the same tree row, 0.27 ms → 21 ms per 100-file listing.
`md5hash` on S3 is not like-for-like: the legacy service gives up on a multipart ETag and
returns `None`.

**What 0.8.1 buys** — it reads mtime, size and directory status in one backend `info()`
instead of four calls. Same benchmark, only the package swapped:

| mount | operation | 0.8.0 | 0.8.1 | gain |
|---|---|---|---|---|
| S3 | ext_attributes (50) | 124.8 ms | 31.1-31.8 ms | ~4x |
| S3 | tree, 5x100 children | 93.3 ms | 25.4-25.9 ms | ~3.6x |
| local | ext_attributes (50) | 9.81 ms | 9.2-9.7 ms | none |
| local | tree, 5x100 children | 110.7 ms | 104-112 ms | none |

0.8.1 only helps the fsspec backends: `LocalStorage` does not override `ext_attributes`
and still makes the four separate calls. **The dependency pin deliberately stays
`>=0.8,<0.9`** — nothing here needs 0.8.1 to work, it is only faster with it, so there is
no reason to force an upgrade on anyone already holding 0.8.0.

### The url layer of the genro-storage service was losing three behaviours

genro-storage 0.8's `url()` carries only `expires_in`, so the service silently dropped the
content disposition it had just computed, the mount's `url_expiration` and the
non-expiring `public_url` (`public_base_url` included, which was never even mapped). On an
S3 mount with the switch on, `serve(download=True)` and a `?download=True` on the
`/_storage/` route opened the file inline instead of downloading it under the intended
name.

`url()` and `public_url()` now delegate to the legacy service of the same mount, which
already signs with `ResponseContentDisposition`, honours `url_expiration` and builds the
public address from `public_base_url` or the regional endpoint. The handler injects that
sibling; a remote mount without one raises, because a download that silently opens inline
is worse than an error.

### Two defects on the download path, neither depending on the switch

Both are on the legacy path today, with the flag off.

- `aws_s3.url()` read `_content_disposition` into a local **before** the `_download`
  branch wrote it back into `kwargs`, so the branch never reached the signed url and
  `url(_download=True)` always produced `inline`. The disposition is now derived once,
  from `_download` or `download`, which also makes `btcprint.py:202`
  (`url(nocache=True, download=True)`) behave on S3 as it already did on a local mount.
- `_download` had three writers — `aws_s3.internal_url()`, `sftp.internal_url()` and the
  genro-storage service — and no reader on the serving path, where every implementation
  reads `download`: the `?_download=True` that `internal_url()` puts in its own url was
  discarded, and the file opened inline. `StorageNode.serve()` now maps it onto `download`
  in one shared place, covering the legacy local, `aws_s3` and `sftp` services and the
  genro-storage one, with the switch on or off.

`sftp` is fixed by that shared mapping but **not exercised**: no SFTP host was available.

Every case above is pinned in both modes, on an S3 mount and on a local one. Each new
test was run against the pre-fix code first and fails there.

```
2582 passed,  69 skipped   (with MinIO configured)
2470 passed, 181 skipped   (no S3 endpoint, the CI shape)
flake8 clean
```

### Still open, and worth a decision in this review

- **genropy/genro-storage#78** — on S3, `copy_to()` issues 6 API calls where one
  `CopyObject` would do, and `open('rb')` adds a `HeadObject`. Still open on 0.8.1: that
  is the ~2x on the copy row. `duplicateNode` could sidestep it through the backend's
  `copy()`; not done here, since the upstream fix would make it unnecessary.
- **#1237** — `StorageNode.open()` runs autocreate even for reads, costing a `HeadObject`
  plus a `ListObjectsV2` per open on S3. Pre-existing, on both paths.
- Whether the hybrid of the description's table is the intended end state (this branch
  assumes it is), and whether `relative` and `sftp` should come in here or later.
- Not yet exercised in a live site: the upload path itself, and multidomain.

genro (commented):
The switch really is default-off and I checked it rather than assuming: one read in shipped code, `gnrwsgisite.py:198`, through `experimentalFlag`, which is `boolean(...)` over a missing tag and so `False`. Nothing else constructs the handler or the service. The optional import is properly guarded at `gnrstoragehandler.py:27-32` and turns a missing package into a `RuntimeError` naming the extra rather than an import error at module load. Routing symbolic mounts to the legacy handler answers #252 structurally, and the test that pins it is the right one.

I am not reviewing this to a verdict yet, because you say it is not ready and I agree with your reasons. Two things to settle first.

The draft flag is not set. The body opens "Draft on purpose", GitHub does not have it marked draft, so it is mergeable and it sits in three people's queues as ready. Set it, or drop the sentence.

Two changes ship outside the switch, and the body does not say so. `storage.py:392-395` rewrites `_download` into `download=True` in `StorageNode.serve()` for every service, legacy included, and `aws_s3.py:325-328` fixes the attachment `Content-Disposition` on the legacy S3 `url()`, where `_content_disposition` was read before `_download` could set it. Both are fixes in the right direction and I would take them — but they reach every install the moment this merges, which is not what "behind a default-off switch" tells a reader. Either split them out, where they can land now on their own merit, or say plainly in the body that they are unconditional.

Two notes for when it does leave draft.

`genro-storage>=0.8,<0.9` is in the `developer` extra, not only in the new `genro_storage` one, so every editable dev install pulls it whether the flag is ever used. Deliberate, since the comparison tests need it, but it is a new mandatory dev dependency and the body is silent.

`storage_genro.py:169-173` raises `NotImplementedError` for `local_path(keep=True)` on a non-local mount, which the legacy `aws_s3` service supports through `S3TemporaryFilename`. An instance flipping the flag loses that at runtime with no static signal. It fails loudly, which is the right choice, but it belongs in the migration document as a known gap.

Also: `closingIssuesReferences` is empty, and neither #251 nor #1237 has an assignee.

fporcari:
Answering the [review](https://github.com/genropy/genropy/pull/1238#pullrequestreview-5235135470) on the point that has an action attached.

> Two changes ship outside the switch, and the body does not say so. [...] Either split them out, where they can land now on their own merit, or say plainly in the body that they are unconditional.

Split out: **#1347**, against `develop`, closing **#1346**, which describes both defects. It carries only the two production hunks — `StorageNode.serve()` mapping `_download` onto `download`, and `aws_s3.url()` deriving the disposition once — with new tests written against what `develop` already has, since the comparison file they were pinned in here does not exist there and depends on the optional package. The post-images are byte-identical to `71b27676f`. One correction to that commit message while moving it: on `develop` `_download` has two writers, not three; the third was the genro-storage service itself.

This branch still contains them until it is realigned on a `develop` that has #1347, so the overlap is real but temporary, and the order is the one you suggested: they land on their own merit first.

On the rest, for the record rather than as changes:

- **`genro-storage>=0.8,<0.9` in the `developer` extra** is deliberate, and you are right that it is a new mandatory dev dependency: the comparison tests import the package, so an editable dev install has to have it whether or not the flag is ever turned on. It stays out of the base dependencies, which is what keeps it off every install's critical path.
- **`local_path(keep=True)` on a non-local mount** raising `NotImplementedError` where the legacy `aws_s3` service answers through `S3TemporaryFilename` is a real gap with no static signal, and it belongs in the migration document as one. It is not in there yet.
- **The draft flag** is still not set, and I am leaving that decision, and the description, alone rather than editing either from here.

`closingIssuesReferences` stays empty on this one: the feature has #251 behind it and the two defects now have #1346.

PR #1247, opened 2026-09-03
fix(siteregister): report serverstore changes for an unknown page

### LINKED ISSUE #1231: SiteRegister: set_serverstore_changes on an unknown page writes into a detached Bag, silently
## What happens

`SiteRegister.set_serverstore_changes(page_id, datachanges)` (`gnrpy/gnr/web/daemon/siteregister.py:1083`) calls `page_register.get_item_data(page_id)`; for a page the register does not hold, `get_item_data` (`:219`) returns a fresh `Bag()` that is attached to nothing. The client's `_serverstore_changes` are written into it and the Bag is dropped. No error, no log.

## When

A race, not a use case: `_check_page_id` runs in the page `__init__` before `_rpcDispatcher`, so a request for an unknown page dies there. `set_serverstore_changes` meets a missing page only when the page is removed between `__init__` and the dispatch (cleanup racing a request).

## Proposal

Log a `WARNING` naming the `page_id` and the paths lost, and return. Behaviour otherwise unchanged.

## Test (red today, same stand-in fixture as `gnrpy/tests/web/subscribed_tables_index_test.py`)

```python
import logging

class _FakeDaemon:
        pass

class _FakeServer:
    daemon = _FakeDaemon()
    gnr_daemon_uri = None
    hmac_key = None

def test_serverstore_changes_for_an_unknown_page_are_reported(caplog):
    sr = SiteRegister(_FakeServer(), sitename='testsite')
    with caplog.at_level(logging.WARNING):
        sr.set_serverstore_changes('ghost', {'rootenv.language': 'en'})
    assert any('ghost' in rec.getMessage() for rec in caplog.records)
```

Today: no record is emitted.

Context: the ASGI bridge (genropy-asgi) will log a WARNING in the same case; aligning the daemon keeps the two behaviours identical.

fporcari:
## Problem

`SiteRegister.set_serverstore_changes(page_id, datachanges)` wrote the client's
`_serverstore_changes` into a Bag attached to nothing when the register no longer held the
page: no error, no log, the changes lost. The page can only be missing because cleanup
raced the request (`_check_page_id` rejects an unknown page in `__init__`, before the
dispatch), which is exactly the case that has to be visible.

## Root cause

`PageRegister.get_item_data` returns a fresh `Bag()` on a cache miss without inserting it
into `self.itemsData`, so the object the caller writes into is detached and dies with the
method. Fixing `get_item_data` itself would regress `handle_ping`, which chains
`get_item_data(user).getItem(...)` with no guard — so the guard belongs in the caller.

## Change

`set_serverstore_changes` checks `page_register.exists(page_id)` first and, when the page
is gone, logs a `WARNING` naming the page and the discarded paths, then returns. Behaviour
for a live page is unchanged. Same level and same lazy `%s` style as
`gnrasync.registerPage`, which already reports this condition; the ASGI bridge logs a
WARNING in the same case, so the two stay aligned.

## Verification

- `gnrpy/tests/web/siteregister_serverstore_changes_test.py` (new): the unknown page is
  reported, and a known page still stores its changes with no warning. Red before the
  change (the first test emits no record), green after.
- `pytest gnrpy/tests/ -q`: 2291 passed, 69 skipped.
- flake8 clean on both touched files.

## Related issue

Fixes #1231

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -1037,8 +1037,10 @@ class SiteRegister(BaseRemoteObject):
         if not page_item:
             return False
         catalog = self.catalog
-        if _serverstore_changes:
-            self.set_serverstore_changes(page_id, _serverstore_changes)
+        if _serverstore_changes and not self.set_serverstore_changes(page_id, _serverstore_changes):
+            # refresh answered above, so the page went while this ping travelled
+            logger.warning('page %s vanished from the register: serverstore changes discarded (%s)',
+                           page_id, ','.join(sorted(_serverstore_changes)))
         if _children_pages_info:
             for k, v in list(_children_pages_info.items()):
                 child_lastUserEventTs = v.pop('_lastUserEventTs', None)
@@ -1081,13 +1083,27 @@ class SiteRegister(BaseRemoteObject):
         return result
 
     def set_serverstore_changes(self, page_id=None, datachanges=None):
+        """File the client's serverstore changes on the page's register item.
+
+        Answers whether they were filed. `get_item_data` builds a detached Bag
+        for an id it does not know, so without the guard the changes went into
+        an object that died with the call - and the client has already dropped
+        its own buffer by then, so there is nothing to recover: the only
+        available answer is to report. Whether absence is exceptional depends on
+        the caller, so this one does not log.
+
+        `exists` rather than `get_item`: the latter stamps `itemsTS` before
+        testing, which would leak a timestamp for a page that never existed.
+
+        :param page_id: id of the target page
+        :param datachanges: the changes to file, as sent by the client
+        :returns: True when filed, False when the page is not in the register"""
         if not self.page_register.exists(page_id):
-            logger.warning('page %s not existing in register: serverstore changes discarded (%s)',
-                           page_id, ','.join(sorted(datachanges or {})))
-            return
+            return False
         page_item_data = self.page_register.get_item_data(page_id)
         for k, v in list(datachanges.items()):
             page_item_data.setItem(k, self._parse_change_value(v))
+        return True
 
     def _parse_change_value(self, change_value):
         if isinstance(change_value, (bytes, str)):
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -1081,6 +1081,10 @@ class SiteRegister(BaseRemoteObject):
         return result
 
     def set_serverstore_changes(self, page_id=None, datachanges=None):
+        if not self.page_register.exists(page_id):
+            logger.warning('page %s not existing in register: serverstore changes discarded (%s)',
+                           page_id, ','.join(sorted(datachanges or {})))
+            return
         page_item_data = self.page_register.get_item_data(page_id)
         for k, v in list(datachanges.items()):
             page_item_data.setItem(k, self._parse_change_value(v))
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -641,8 +641,11 @@ class GnrWebPage(GnrBaseWebPage):
         self._lastUserEventTs = kwargs.pop('_lastUserEventTs', None)
         self._lastRpc = kwargs.pop('_lastRpc', None)
         self._pageProfilers = kwargs.pop('_pageProfilers', None)
-        if _serverstore_changes:
-            self.site.register.set_serverstore_changes(self.page_id, _serverstore_changes)
+        if _serverstore_changes and not self.site.register.set_serverstore_changes(
+                self.page_id, _serverstore_changes):
+            # the page passed _check_page_id in __init__, so this is the cleanup race
+            logger.warning('page %s vanished from the register: serverstore changes discarded (%s)',
+                           self.page_id, ','.join(sorted(_serverstore_changes)))
         auth = AUTH_OK
         if method not in ('doLogin', 'onClosePage'):
             auth = self._checkAuth(method=method, **kwargs)
--- a/gnrpy/tests/web/siteregister_serverstore_changes_test.py
+++ b/gnrpy/tests/web/siteregister_serverstore_changes_test.py
@@ -1,0 +1,42 @@
+"""Tests for ``SiteRegister.set_serverstore_changes`` on a page the register lost (#1231).
+
+``get_item_data`` returns a detached ``Bag()`` on a cache miss, so writing the client's
+``_serverstore_changes`` into it used to drop them with no error and no log. The page can
+only be missing because cleanup raced the request, which is exactly the case that has to
+be visible.
+
+The register is a real ``SiteRegister`` built with a stand-in server -- its constructor
+only needs ``server.daemon.register`` -- as in ``subscribed_tables_index_test.py``.
+"""
+
+import logging
+
+from gnr.web.daemon.siteregister import SiteRegister
+
+
+class _FakeDaemon:
+    def register(self, obj, name):
+        pass
+
+
+class _FakeServer:
+    daemon = _FakeDaemon()
+    gnr_daemon_uri = None
+    hmac_key = None
+
+
+def test_serverstore_changes_for_an_unknown_page_are_reported(caplog):
+    sr = SiteRegister(_FakeServer(), sitename='testsite')
+    with caplog.at_level(logging.WARNING):
+        sr.set_serverstore_changes('ghost', {'rootenv.language': 'en'})
+    messages = [rec.getMessage() for rec in caplog.records]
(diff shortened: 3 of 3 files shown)


genro (changes requested):
Changes request on `be3e23f062`, and the request is one thing: the answer shape, so this lands as the same convention @genro wrote in #1253 rather than a second rule for a sibling call.

## What holds

The diagnosis is right and the severity is higher than the body says. `get_item_data` returns a detached `Bag()` on a miss (`siteregister.py:219-223`), so the write went into an object that died with the method. And the client has already cleared its buffer by then: `genro_rpc.js:306-308` attaches `_serverstore_changes` to the call and nulls it in the same breath, so nobody holds those changes any more. They are unrecoverable at the moment the server notices, which is why the only available answer is to report, not to repair.

What that costs is worth stating in the body, because it is not just "the two copies diverge". `get_dbenv` folds everything under `rootenv` into the page's database environment (`siteregister.py:349-350`), and the declared serverpaths are `rootenv.language` and `rootenv.context_subtables` (`gnrwebpage.py:2424` and `:2509`). A lost serverstore change means the server keeps querying with the previous language, or the previous subtable context.

I also checked the guard, since it asks one dictionary and the write uses another: `exists` reads `registerItems`, `get_item_data` reads `itemsData`. They are coupled at every point — both written in `add_item`, both popped in `drop_item`, both pickled and reloaded in `dump`/`load` — so the guard is a sound proxy. Two tests pass on a real `SiteRegister`.

## The request: answer, do not log

#1253 sets the convention for exactly this situation, on the sibling call: return `True` when the change is filed, `False` when the target does not exist, no exception, and callers that do not care ignore it. `set_serverstore_changes` should answer the same way, and drop the `logger.warning`.

The reason is that the callee cannot know whether absence is exceptional, and its three callers disagree:

- `GnrWebPage._rpcDispatcher` (`gnrwebpage.py:645`) — the page passed `_check_page_id` in `__init__`, so a missing page here is the cleanup race your body describes. Exceptional, worth a log.
- `SiteRegister.handle_ping` (`siteregister.py:1041`) — same, `refresh` returned an item at `:1036` or the method already returned.
- `SiteRegister.handle_ping` (`siteregister.py:1048`) — the children loop. Here `k` comes from the browser's own walk over `window.frames` (`genro.getChildrenInfo`, `genro.js:1007`) and passes no check at all: the page is missing because a child's registration is gone while its iframe is alive. Ordinary, and it repeats on every RPC that carries changes for that child. Two lines below, `refresh(k)` returns `None` for the same dead child in silence, and the datachange collection for it proceeds.

So the warning as written makes one of the three routes for the same dead child loud and leaves the other two quiet. With a boolean, each caller decides, and the body's claim about the cleanup race becomes true where it is made instead of true for two callers out of three.

## Keep `exists()`, not `get_item()`

Worth saying because the obvious refactor is a trap. `get_item` calls `updateTS` at `:227` **before** testing whether the item is there, so asking it about a ghost id writes a timestamp into `itemsTS` for a page that does not exist. `itemsTS` is only popped in `drop_item`, which never runs for an id that was never registered, and it is pickled in the freeze. The children loop is precisely where ghost ids come from, so `get_item` would feed a slow leak that survives a restore. Your `exists()` avoids it.

Same trap waits for whoever implements #1253: `set_datachange` already calls `get_item`, so it stamps the ghost before returning.

## A question rather than a request

Would you rather do one change that closes #1253 as well? The two issues are the same rule applied to two write paths on the register, and landing them together means one convention arrives at once instead of two PRs each establishing it for its own method. #1253 adds `set_datachange` returning the boolean and `setInClientData` propagating it, which is more surface than this PR but the same decision.

Your call. If you would rather keep this one narrow, the boolean here is enough and #1253 stays where it is.

One thing to know if you take the wider road: #1254 asks for `set_datachange` to stop being reachable from application code. It does not conflict — the public surface in #1253 is `setInClientData` — but the two want to be read together.

## Perimeter

I did not run the full suite, only your new test file, and I did not exercise the children path in a browser: the third caller's behaviour is read from `getChildrenInfo` and the ping handler, not observed.

fporcari:
Done in `0a1d04bf56`, base realigned in `8071817874`
([review](https://github.com/genropy/genropy/pull/1247#pullrequestreview-5100662980)).

## The answer shape

> `set_serverstore_changes` should answer the same way, and drop the `logger.warning`.

Taken, same convention as #1253: `True` when filed, `False` when the page is not
in the register, no exception, and the callee logs nothing.

I went one step further than dropping the warning, because dropping it alone
would lose the information at all three callers instead of at one — and your
argument is that two of them *should* be loud. So the two that reach the call
after establishing the page existed now log it themselves:

- `GnrWebPage._rpcDispatcher` — the page passed `_check_page_id` in `__init__`,
  so absence here is the cleanup race the body describes;
- `SiteRegister.handle_ping` at `:1041` — `refresh` answered at `:1036`, same
  reasoning.

The children loop at `:1048` ignores the return, as you say it should: its ids
come from the browser's own walk over `window.frames`, nothing checked them, and
`refresh(k)` two lines below already returns `None` for the same dead child in
silence. The body's claim about the cleanup race is now true where it is made.

## `exists()`, and a test so the trap does not get walked into

Kept, and thank you for naming the mechanism — I had it as "cheaper" rather than
as a leak. It is now pinned rather than commented:

```python
def test_asking_about_a_ghost_leaves_no_timestamp_behind():
    sr = SiteRegister(_FakeServer(), sitename='testsite')
    sr.set_serverstore_changes('ghost', {'rootenv.language': 'en'})
    assert 'ghost' not in sr.page_register.itemsTS
    assert 'ghost' not in sr.page_register.itemsData
```

Ratchet checked: swap `exists` for `get_item` and it fails on the first
assertion. Worth having because the refactor is attractive and the damage is
invisible — `itemsTS` is only popped in `drop_item`, which never runs for an id
that was never registered, and it goes into the freeze. Whoever implements #1253
inherits the same trap, since `set_datachange` already calls `get_item`; I have
left a note to that effect on it.

## Staying narrow

> Would you rather do one change that closes #1253 as well?

Keeping this one narrow. #1253 is not the same size: it adds `set_datachange`'s
boolean plus propagation through `SiteRegister.setInClientData`,
`WebPage.setInClientData` and `setInClientData_legacy` — public surface, in the
same methods #1254 wants to make unreachable from application code. Those two
want to be decided together, and I would rather that decision arrive in its own
PR than ride in behind a two-line guard on a sibling call.

The convention arrives here first and #1253 follows it, which is the same
outcome as landing them together minus the coupling.

## Verification

`tests/web/siteregister_serverstore_changes_test.py` 3 passed on a real
`SiteRegister`, flake8 clean on all three files. Same perimeter you named: I did
not exercise the children path in a browser either — the third caller's
behaviour is read from `getChildrenInfo` and the ping handler, and what my tests
cover is the callee's contract, not the three callers' use of it.

Re-requesting.

genro (approved):
This is what I asked for. The callee answers with a boolean and says nothing, the two callers that care about it log, and the children loop of `handle_ping` ignores the return — so the ghost page no longer produces noise on the path where it is expected.

Keeping `exists()` rather than `get_item()` matters more than it looks, and the test that pins it is the right one: `get_item` stamps `itemsTS` at `siteregister.py:227` before it tests the item, so asking about a ghost through it would leave a timestamp behind for a page that does not exist. `exists` reads `registerItems`, which is written and popped in lockstep with `itemsData`, so it is a sound proxy.

One thing to do before you squash. The description still describes the previous design — it says the callee logs a WARNING naming the page and the discarded paths, and the Verification section says the unknown page is reported. At head the callee logs nothing. That text becomes the commit message, so it would be the only surviving account of the change and it would be wrong.

For the record: the warning at `siteregister.py:1040` is nearly unreachable, since `refresh(page_id)` four lines above already found the item in the same dict. Only a concurrent drop between the two statements reaches it. Your inline comment says exactly that, so it is honest rather than decorative — I just want it written down that no test covers it and none easily could.

cgabriel (approved):

fporcari:
Merged, with the squash message written by hand rather than the description.

> The description still describes the previous design — it says the callee logs a WARNING naming the page and the discarded paths […] That text becomes the commit message, so it would be the only surviving account of the change and it would be wrong.

`develop` now carries the design that is actually at head:

> The callee now guards on `page_register.exists` and answers whether it filed them. It logs nothing itself, because whether the absence is exceptional depends on the caller: the two that care, `_rpcDispatcher` and `handle_ping`, log a warning naming the page and the discarded paths, while the children loop of `handle_ping` ignores the return.

The `exists` / `get_item` reasoning and the `itemsTS` stamp are in there too — it was the part of your review that would have been lost first, and it is the one a reader needs to not "simplify" it back.

Your last point is on the record as well, in the commit:

> The warning in `handle_ping` is nearly unreachable — `refresh(page_id)` found the item four lines above, so only a concurrent drop between the two statements reaches it — and no test covers it.

The description and the title stay as they were: they are what you and @cgabriel reviewed against, and rewriting them after the fact leaves a record matching neither the review nor the merge.

PR #1248, opened 2026-09-03
fix(siteregister): do not announce autocreated parents to the client

### LINKED ISSUE #1230: SiteRegister: autocreated parents are announced to the client as serverChange
## What happens

`PageRegister._on_data_trigger` (`gnrpy/gnr/web/daemon/siteregister.py:186`) does not look at the Bag event's `reason`. A `setItem('a.b.c', 1)` on a page store whose `a` and `a.b` do not exist yet makes the Bag insert them on the fly (`reason='autocreate'`); if `a` sits under a subscribed path, the trigger queues THREE `ClientDataChange`s with `reason='serverChange'`: `a` (an empty Bag), `a.b` (an empty Bag), then `a.b.c`.

## Why it matters

On the client the two parents add nothing: writing the leaf recreates them. And when the client already holds its own children under `a` (client-side data under the same client path, never sent to the server), the `a = Bag()` change overwrites that subtree with an empty Bag an instant before the leaf arrives (`genro_rpc.js:setDatachangesInData` → `updater` writes whenever the value differs).

## Proposal

In `_on_data_trigger`, skip the `ins` events whose `reason == 'autocreate'`. Only explicitly written nodes travel.

## Test (red today, on a real SiteRegister with a stand-in server, no Pyro — same fixture as `gnrpy/tests/web/subscribed_tables_index_test.py`)

```python

class _FakeDaemon:
        pass

class _FakeServer:
    daemon = _FakeDaemon()
    gnr_daemon_uri = None
    hmac_key = None

def test_autocreated_parents_are_not_announced_as_server_changes():
    reg.create('p1')
    reg.subscribe_path('p1', 'a')
    reg.get_item_data('p1').setItem('a.b.c', 1)
    changes = reg.get_datachanges('p1')
    assert [c.path for c in changes] == ['a.b.c']
    assert changes[0].reason == 'serverChange'
```

Today: `assert ['a', 'a.b', 'a.b.c'] == ['a.b.c']` fails, the two parents carrying an empty Bag each.

Context: the ASGI bridge (genropy-asgi) filters them already; aligning the daemon keeps the two behaviours identical.

fporcari:
## Problem

A `setItem('a.b.c', 1)` on a page store whose `a` and `a.b` do not exist yet announced
THREE changes to the client: `a` (an empty Bag), `a.b` (an empty Bag), then `a.b.c`, all
with `reason='serverChange'`. The two parents add nothing — writing the leaf recreates
them — and when the client holds its own children under `a`, the `a = Bag()` change wipes
that subtree an instant before the leaf arrives: `setDatachangesInData` → `updater` writes
whenever the value differs, with no guard.

## Root cause

`_on_data_trigger` did not look at the Bag event's `reason`, while the on-the-fly parents
are emitted as `reason='autocreate'` by `Bag._onNodeInserted`.

## Change

`_on_data_trigger` takes `reason` and returns immediately on `autocreate`, so only
explicitly written nodes travel. This is the filter `gnrasync._on_data_trigger` already
applies to the identical event, and the whole client side already ignores
`reason == 'autocreate'` — the daemon was the odd one out. The ASGI bridge filters them
too, so the two behaviours stay aligned.

The method lives on `BaseRegister`, so the guard covers the page, user, connection and
global stores at once. Nothing legitimate is dropped: a parent can only match a
`subscribed_paths` prefix if the leaf written under it matches too, so the leaf that
recreates the parents is always announced.

## Verification

- `gnrpy/tests/web/siteregister_datachanges_test.py` (new): the issue's case, plus the two
  counter-cases that pin down what must still travel — an empty Bag written explicitly,
  and a leaf's rewrites and attributes. All three red before the change, green after.
- `pytest gnrpy/tests/ -q`: 2351 passed, 10 skipped.
- flake8 clean on both touched files.

## Related issue

Fixes #1230

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -184,7 +184,10 @@ class BaseRegister(BaseRemoteObject):
         data.subscribe('datachanges', any=lambda **kwargs: self._on_data_trigger(register_item=register_item, **kwargs))
         self.itemsData[register_item_id] = data
 
-    def _on_data_trigger(self, node=None, ind=None, evt=None, pathlist=None, register_item=None, **kwargs):
+    def _on_data_trigger(self, node=None, ind=None, evt=None, pathlist=None, register_item=None,
+                         reason=None, **kwargs):
+        if reason == 'autocreate':
+            return
         if evt == 'ins':
             pathlist.append(node.label)
         path = '.'.join(pathlist)
--- a/gnrpy/tests/web/siteregister_datachanges_test.py
+++ b/gnrpy/tests/web/siteregister_datachanges_test.py
@@ -1,0 +1,63 @@
+"""Tests for the datachanges a page store announces to the client (#1230).
+
+A ``setItem('a.b.c', 1)`` on a Bag whose ``a`` and ``a.b`` do not exist yet makes the Bag
+insert the parents on the fly (``reason='autocreate'``). Those parents carry an empty Bag,
+and the client rewrites whatever it holds under that path whenever the value differs
+(``genro_rpc.js:setDatachangesInData``), so announcing them wipes the client-side subtree
+an instant before the leaf arrives. Writing the leaf recreates the parents anyway.
+
+``_on_data_trigger`` lives on ``BaseRegister``, so the filter covers the page, user,
+connection and global stores at once. ``gnrasync._on_data_trigger`` already filters the
+identical event.
+
+The register is a real ``SiteRegister`` built with a stand-in server -- its constructor
+only needs ``server.daemon.register`` -- as in ``subscribed_tables_index_test.py``.
+"""
+
+from gnr.core.gnrbag import Bag
+from gnr.web.daemon.siteregister import SiteRegister
+
+
+class _FakeDaemon:
+    def register(self, obj, name):
+        pass
+
+
+class _FakeServer:
+    daemon = _FakeDaemon()
+    gnr_daemon_uri = None
+    hmac_key = None
+
+
+def _subscribed_page():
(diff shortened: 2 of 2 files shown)


genro (approved):
Approving `429ff398d9`.

The claim that decides whether the guard drops anything legitimate is the one about `subscribed_paths`, and I checked it rather than took it: the match is `path.startswith(subscribed)`, so if a parent's path starts with the subscription, every leaf beneath it does too, since the leaf carries the parent's path as a prefix. There is no combination where a parent passes the filter and the leaf that recreates it does not.

Returning before the `_caching_table` branch loses nothing either: `_htraverse` builds the on-the-fly node as `BagNode(curr, label=label, value=curr.__class__())`, with no attributes, so `node.attr.get('_caching_table')` was always `None` for an autocreate.

The two counter-cases are the useful part of this PR over the issue: an explicitly written empty Bag and a leaf's rewrites with attributes both still travel, which is what pins down that the filter is on the reason and not on the value.

Ratchet run independently, with the same three tests against the unpatched module rather than the working tree: 3 of 3 red, green at this head. Real `SiteRegister` and real Bags throughout, no mocks.

One observation, not a request. The guard sits at the top of the method, so it covers any event carrying that reason, while the issue asked for the `ins` events. The two are equivalent today, because `reason='autocreate'` is emitted only from `_onNodeInserted` (`gnrbag.py:798`).

Perimeter: I did not run the full suite, only the new test file, and I did not check a browser. The client-side subtree wipe this prevents is deduced from `updater`'s `getItem(path) != value` comparison against an empty Bag, not observed.

Merge order: this and #1252 change the same two lines and conflict in git, as you note over there. Whichever lands second needs the rebase; the two changes stack.

cgabriel (approved):

PR #1252, opened 2026-09-03
fix(siteregister): announce a popped node on its own path, as a delete

### LINKED ISSUE #1249: [BUG] SiteRegister: a pop on a page store is announced to the client as a write of the removed value on the parent path
## Problem Description

`BaseRegister._on_data_trigger` (`gnrpy/gnr/web/daemon/siteregister.py:187`) appends
`node.label` to the path only on the `ins` branch. For an `evt == 'del'` the label is
missing, so a `pop('a.b.c')` is announced as a `ClientDataChange` on path **`a.b`**
carrying the **removed value**, with `delete=False`.

Two consequences on the client (`gnrjs/gnr_d11/js/genro_rpc.js:526` →
`setDatachangesInData`): `isDelete` is false, so it takes the `updater` branch and writes
`a.b = 1` — the parent Bag, siblings included, is replaced by a scalar — and the deletion
itself never reaches the client at all.

Expected: the change is announced on `a.b.c` with the delete flag set, so
`genro._data.delItem` removes exactly the popped node.
`gnrasync._on_data_trigger` (`gnrpy/gnr/web/gnrasync.py:294-299`) already appends the
label for both `ins` and `del` on the identical trigger signature.

The method lives on `BaseRegister`, so the page, user, connection and global stores are
all affected.

## Reproduction Steps

Real `SiteRegister` with a stand-in server (no daemon, no Pyro), the fixture of
`gnrpy/tests/web/subscribed_tables_index_test.py`:

```python

class _FakeDaemon:
        pass

class _FakeServer:
    daemon = _FakeDaemon()
    gnr_daemon_uri = None
    hmac_key = None

reg.create('p1')
reg.subscribe_path('p1', 'a')
data = reg.get_item_data('p1')
data.setItem('a.b.c', 1)
data.setItem('a.b.d', 2)
reg.get_datachanges('p1')
data.pop('a.b.c')
print([(c.path, c.value, c.delete) for c in reg.get_datachanges('p1')][-1:])
```

Today on `develop`: `[('a.b', 1, False)]`.
Expected: the change on `a.b.c`, with `delete=True` and the sibling `a.b.d` untouched.

Reproducible in-process, no site and no browser needed — the snippet above is enough.
The visible effect in a running application needs a page holding a subscribed subtree and
a server-side `pop` on it: the sibling nodes under the parent disappear from the client
store while the popped node stays.

## Environment

`develop` at 919a3a572a, Python 3.13.2, macOS. Independent of the adapter and of the OS:
the code path is pure Python plus the JS updater.

## Additional Material

Found while fixing #1230 and deliberately left out of #1248, which only filters the `ins`
events whose `reason == 'autocreate'`. Verified that the two are independent: with and
without the #1248 guard, the `pop` line is identical (`('a.b', 1, False)`) — `autocreate`
is only emitted on insertion, so the `del` branch never crosses that filter.

The fix looks like one line plus the delete flag, but it changes what the client does with
an existing message, so it deserves its own PR and its own test.

/cc @genro

fporcari:
## Problem

`BaseRegister._on_data_trigger` appended `node.label` to the path only for `evt == 'ins'`.
A `pop('a.b.c')` was therefore announced on the parent path `a.b`, carrying the **removed
value**, with `delete=False`. Two effects on the client: the parent node was overwritten
with that value — the surviving siblings under `a.b` wiped with it — and the deletion
itself never arrived.

## Root cause

Two halves, one behaviour:

- `gnrpy/gnr/web/daemon/siteregister.py` — the label was appended for `ins` only, and the
  `ClientDataChange` was built with the default `delete=False` whatever the event was.
- `gnrjs/gnr_d11/js/genro_rpc.js` — `setDatachangesInData` tests `reason == 'serverChange'`
  **before** `isDelete`, so for a store change the flag was ignored and the `updater`
  branch ran regardless. Fixing the daemon alone would have stopped the sibling wipe but
  still not deleted anything on the client.

`gnrasync._on_data_trigger` already appends the label for both `ins` and `del` on the
identical trigger signature.

## Change

- The label is appended for `del` too, and the change is marked `delete=evt == 'del'`.
- The serverChange branch of `setDatachangesInData` calls `genro._data.delItem` on the
  mapped client path when the change is a delete, instead of writing the value.
- The label is appended to a **copy** of `pathlist`. The Bag hands the same list object to
  every subscriber of the event and then to its parent (`Bag._onNodeInserted`,
  `_onNodeDeleted`), so mutating it in place was a latent corruption of the paths seen by
  anything else subscribed to that store. No behaviour change for a single subscriber at
  the root — which is why nothing showed today — and one test pins it down.

The trigger lives on `BaseRegister`, so the page, user, connection and global stores are
all fixed by the same lines.

## Verification

`gnrpy/tests/web/siteregister_datachanges_del_test.py` (new, 11 tests). The daemon side of
this defect had **zero** coverage before: no test in the suite named `get_datachanges`,
`set_datachange`, `_on_data_trigger` or `subscribe_path`. The file covers the whole
observable surface of the trigger:

- a written leaf is announced on its own path, an overwrite keeps that path, a write
  outside the subscribed paths is ignored;
- a popped leaf is announced on its own path with `delete=True`; a pop does **not** announce
  the parent, and the sibling survives; popping a subtree announces the subtree path; a pop
  outside the subscribed paths is ignored;
- the flag reaches the client: `handle_ping_get_datachanges` carries `change_delete` true
  for a pop and false for a write;
- the user store behaves identically, as the trigger is `BaseRegister`'s;
- the trigger leaves the event's `pathlist` intact for the other subscribers of the Bag.

6 of the 11 are red on `develop` and green with the change; the other 5 are the regression
guards on the write path. Full suite: 2359 passed, 10 skipped. flake8 clean on both Python
files.

The JS half is **not** covered by an automated test: there is no test harness for `gnrjs`
in this repo (only the vendored dojo suites). It needs a manual check — a server-side `pop`
on a subscribed subtree, with a sibling under the same parent, must remove exactly the
popped node in the client store.

## Note for whoever merges

This touches `_on_data_trigger` two lines away from #1248, so the two conflict in git
(verified with `git merge-tree`). The resolution is mechanical — the changes stack, the
`autocreate` guard first and the label append after — but whichever merges second needs a
rebase.

## Related issue

Fixes #1249

### DIFF
--- a/gnrjs/gnr_d11/js/genro_rpc.js
+++ b/gnrjs/gnr_d11/js/genro_rpc.js
@@ -548,7 +548,10 @@ dojo.declare("gnr.GnrRpcHandler", null, {
                     if (stringStartsWith(changepath, serverpath_prefix)) {
                         let clientpath = clientpath_prefix + changepath.slice(serverpath_prefix.length);
                         if (isDelete) {
-                            genro._data.delItem(clientpath);
+                            // mark the removal as the server's, like the write path
+                            // below and the two other server-side pops: an unmarked
+                            // event is echoed back as a client write (genro.js:1621)
+                            genro._data.delItem(clientpath, reason);
                         } else {
                             updater(clientpath, value, attr, reason);
                         }
@@ -556,7 +559,7 @@ dojo.declare("gnr.GnrRpcHandler", null, {
                 }
             }
             else if (isDelete) {
-                genro._data.delItem(changepath);
+                genro._data.delItem(changepath, reason);
             }
             else {
                 updater(changepath, value, attr, reason);
--- a/gnrjs/gnr_d11/js/genro_rpc.js
+++ b/gnrjs/gnr_d11/js/genro_rpc.js
@@ -547,7 +547,11 @@ dojo.declare("gnr.GnrRpcHandler", null, {
                     var serverpath_prefix = genro._serverstore_paths[clientpath_prefix];
                     if (stringStartsWith(changepath, serverpath_prefix)) {
                         let clientpath = clientpath_prefix + changepath.slice(serverpath_prefix.length);
-                        updater(clientpath, value, attr, reason);
+                        if (isDelete) {
+                            genro._data.delItem(clientpath);
+                        } else {
+                            updater(clientpath, value, attr, reason);
+                        }
                     }
                 }
             }
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -185,8 +185,8 @@ class BaseRegister(BaseRemoteObject):
         self.itemsData[register_item_id] = data
 
     def _on_data_trigger(self, node=None, ind=None, evt=None, pathlist=None, register_item=None, **kwargs):
-        if evt == 'ins':
-            pathlist.append(node.label)
+        if evt in ('ins', 'del'):
+            pathlist = pathlist + [node.label]
         path = '.'.join(pathlist)
         if evt != 'del' and node.attr.get('_caching_table'):
             caching_subscribers = self.cached_tables[node.attr['_caching_table']]
@@ -198,7 +198,8 @@ class BaseRegister(BaseRemoteObject):
         for subscribed in register_item['subscribed_paths']:
             if path.startswith(subscribed):
                 register_item['datachanges'].append(
-                    ClientDataChange(path=path, value=node.value, reason='serverChange', attributes=node.attr))
+                    ClientDataChange(path=path, value=node.value, reason='serverChange', attributes=node.attr,
+                                     delete=evt == 'del'))
                 break
 
     def invalidateTableCache(self, table):
--- a/gnrpy/tests/web/siteregister_datachanges_del_test.py
+++ b/gnrpy/tests/web/siteregister_datachanges_del_test.py
@@ -1,0 +1,179 @@
+"""Tests for the datachanges a register announces to the client (#1249).
+
+``BaseRegister._on_data_trigger`` appended ``node.label`` only for ``evt == 'ins'``, so a
+``pop('a.b.c')`` was announced on the PARENT path ``a.b`` carrying the removed value and
+``delete=False``: the client wrote that value over the parent node -- siblings included --
+and the deletion never reached it. ``gnrasync._on_data_trigger`` appends the label for
+both events; the daemon now does the same and marks the change ``delete``.
+
+The trigger lives on ``BaseRegister``, so every store built on it (page, user, connection,
+global) is covered by the same code path -- the user register is exercised here for that
+reason.
+
+The register is a real ``SiteRegister`` built with a stand-in server -- its constructor
+only needs ``server.daemon.register`` -- as in ``subscribed_tables_index_test.py``.
+
+The fixture creates the subscribed root explicitly and clears the changes it produced, so
+what each test asserts is only the event it fires: the parents a ``setItem`` autocreates
+are announced too, which is a separate defect (#1230) and not what these tests measure.
+"""
+
+from gnr.core.gnrbag import Bag
+from gnr.web.daemon.siteregister import SiteRegister
+
+
+class _FakeDaemon:
+    def register(self, obj, name):
+        pass
+
+
+class _FakeServer:
+    daemon = _FakeDaemon()
+    gnr_daemon_uri = None
+    hmac_key = None
+
+
+def _site_register():
+    return SiteRegister(_FakeServer(), sitename='testsite')
+
+
+def _subscribed_page(sr=None):
+    reg = (sr or _site_register()).page_register
+    reg.create('p1')
+    reg.subscribe_path('p1', 'a')
+    reg.get_item_data('p1').setItem('a', Bag())
+    reg.get_datachanges('p1', reset=True)
+    return reg
+
+
+def _paths(changes):
+    return [(c.path, c.value, c.delete) for c in changes]
+
+
+# ---------------------------------------------------------------------------
+# what a write announces
+# ---------------------------------------------------------------------------
+
+def test_a_written_leaf_is_announced_on_its_own_path():
+    reg = _subscribed_page()
+    reg.get_item_data('p1').setItem('a.b', 1)
+    changes = reg.get_datachanges('p1')
+    assert _paths(changes) == [('a.b', 1, False)]
+    assert changes[0].reason == 'serverChange'
+
+
+def test_overwriting_a_leaf_keeps_the_leaf_path():
+    reg = _subscribed_page()
+    data = reg.get_item_data('p1')
+    data.setItem('a.b', 1)
+    data.setItem('a.b', 2)
+    assert _paths(reg.get_datachanges('p1')) == [('a.b', 1, False), ('a.b', 2, False)]
+
+
+def test_a_write_outside_the_subscribed_paths_is_ignored():
+    reg = _subscribed_page()
+    reg.get_item_data('p1').setItem('z.b', 1)
+    assert reg.get_datachanges('p1') == []
+
+
+# ---------------------------------------------------------------------------
+# what a pop announces -- the defect
+# ---------------------------------------------------------------------------
+
+def test_a_popped_leaf_is_announced_on_its_own_path_as_a_delete():
+    reg = _subscribed_page()
+    data = reg.get_item_data('p1')
+    data.setItem('a.b', 1)
+    reg.get_datachanges('p1', reset=True)
+    data.pop('a.b')
+    assert _paths(reg.get_datachanges('p1')) == [('a.b', 1, True)]
+
+
+def test_a_pop_does_not_announce_the_parent_path():
+    reg = _subscribed_page()
+    data = reg.get_item_data('p1')
+    data.setItem('a.b.c', 1)
+    data.setItem('a.b.d', 2)
+    reg.get_datachanges('p1', reset=True)
+    data.pop('a.b.c')
+    changes = reg.get_datachanges('p1')
+    assert [c.path for c in changes] == ['a.b.c']
+    assert data['a.b.d'] == 2
+
+
+def test_popping_a_subtree_announces_the_subtree_path():
+    reg = _subscribed_page()
+    data = reg.get_item_data('p1')
+    data.setItem('a.b.c', 1)
+    reg.get_datachanges('p1', reset=True)
+    data.pop('a.b')
+    changes = reg.get_datachanges('p1')
+    assert [(c.path, c.delete) for c in changes] == [('a.b', True)]
+    assert isinstance(changes[0].value, Bag)
+
+
+def test_a_pop_outside_the_subscribed_paths_is_ignored():
+    reg = _subscribed_page()
+    data = reg.get_item_data('p1')
+    data.setItem('z.b', 1)
+    reg.get_datachanges('p1', reset=True)
+    data.pop('z.b')
+    assert reg.get_datachanges('p1') == []
+
+
+# ---------------------------------------------------------------------------
+# the flag has to reach the client
+# ---------------------------------------------------------------------------
+
+def test_the_delete_flag_travels_in_the_client_envelope():
+    sr = _site_register()
+    reg = _subscribed_page(sr)
+    data = reg.get_item_data('p1')
+    data.setItem('a.b', 1)
+    reg.get_datachanges('p1', reset=True)
+    data.pop('a.b')
+    envelope = sr.handle_ping_get_datachanges('p1')
+    nodes = envelope.getNodes()
+    assert [n.attr['change_path'] for n in nodes] == ['a.b']
+    assert nodes[0].attr['change_delete'] is True
+
+
+def test_a_write_travels_in_the_client_envelope_without_the_delete_flag():
+    sr = _site_register()
+    reg = _subscribed_page(sr)
+    reg.get_item_data('p1').setItem('a.b', 1)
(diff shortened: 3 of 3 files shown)


genro (changes requested):
Changes request on `964f526a87`, for one line. The rest holds, including the part that is easy to get wrong, and the split is a question for you rather than a request from me.

## What holds

Measured before and after on a page subscribed to `a`, popping `a.b.c` with a sibling `a.b.sibling` present:

```
develop      announces  path=a.b      value=1   delete=False
this head    announces  path=a.b.c    value=1   delete=True
```

The server store is right either way. The defect was entirely in what got announced, and the client consequence follows from `updater`: comparing the Bag it holds under `a.b` against the scalar `1` always differs, so the container was replaced by the removed value and every sibling under it went with it.

The `pathlist` copy is the part worth calling out, because it is invisible and you fixed it anyway. `_onNodeInserted` hands the same list object to every subscriber of the event and only then builds `[parentlabel] + pathlist` for its parent, so an in-place append is a mutation the next subscriber would read. It never showed because the register subscribes once, at the root, where there is no parent above to receive the corrupted list.

Ratchet run independently, with your test file against the unpatched module rather than the working tree: 6 of the 11 red, the other 5 green as the write-path guards they are meant to be. `node --check` clean.

I also checked the two invariants this defect turns on, since a delete that goes wrong halfway is worse than one that never travels.

- **Every removal carries the flag.** True structurally rather than by audit: there are four producers of a `ClientDataChange` in the repo and only one door a real removal can come through, the Bag trigger, which is the one you fix. `addLocalDatachange` has a single caller and it writes, `setInClientData_legacy` does not pass the parameter, and nothing passes `delete=True` to `set_datachange`. There is no server API for removing a path from the client store at all: the habit is writing `None`, which leaves the node in place with a null value.
- **Every removal is handled.** `setDatachangesInData` has exactly three branches, and after this PR each one that can receive a removal performs one.

## The one request: say the removal comes from the server

`genro._data.delItem(clientpath)` reaches `_pop` with `doTrigger` unset, which becomes `true`, so the event fires with `reason === true`. The other two places where the server pushes a removal into the client store mark it: `genro._data.popNode(fullpath,'serverChange')` in `gnrsharedobjects.js:53` and in `gnrwebsocket.js:181`. The write path in this very function marks it too, through `updater`'s `{'doTrigger':reason}`.

It matters because `genro.js:1621` branches on `kw.reason != 'serverChange'` with no guard on the event, and inside that branch a change under a registered shared object path is sent back to the server as a client-originated write. So an unmarked removal can be echoed. The condition is a path registered both as a serverstore path and under a shared object, and I could not find one in repo, so I am not claiming it happens today: the inconsistency with the two siblings is enough on its own.

`delItem(clientpath, 'serverChange')`.

## Splitting the browser half: your call

@genro raised it and I think the decision is yours, so here is both sides rather than a request.

The argument for splitting is not tidiness, it is that the two halves have different verification regimes: eleven tests and a ratchet on the daemon side, none at all on the client side, in a single merge. Splitting makes the untested half revertible on its own.

The argument against is the cost, and it is not zero. This already conflicts with #1248, which is approved and will land first, so a rebase is due anyway; splitting means a rebase plus a second branch for five lines of JavaScript. And it buys less safety than it looks: reverting this PR whole costs the same, since it restores today's behaviour and nothing else.

Neither merge order is dangerous, which is why I think it is a preference and not a correctness question. I checked: with the daemon half alone the announcement carries the right path and the flag, the client ignores the flag and takes the write path, and `updater` finds the value it already holds and writes nothing. So removals stop arriving, exactly as today, and the sibling wipe is gone. With the client half alone nothing changes at all, since no producer sets the flag.

## Before the merge

Please run the browser check and record the result in the PR. This is the point that has no automated cover and, more than that, this branch is not code being changed: it is code being switched on. Nothing in the repo has ever set `delete=True` on a store change, so the client's removal path for store changes has never run. A server-side pop on a subscribed subtree, with a sibling under the same parent, must remove exactly the popped node.

One follow-up out of this review, not for this PR: #1254, on `set_datachange` being reachable from application code.

fporcari:
Done in `3afeb9c10d`
([review](https://github.com/genropy/genropy/pull/1252#pullrequestreview-5100382419)).

## The marking

> `delItem(clientpath, 'serverChange')`

Taken, written as `delItem(clientpath, reason)` — inside that branch `reason`
**is** `'serverChange'`, and passing the change's own reason is what the write
path two lines below already does (`updater`'s
`reason != null ? {'doTrigger':reason,...} : null`).

I applied the same to the second delete branch, `:558-560`. It is the one reached
when `reason != 'serverChange'`, so marking it `'serverChange'` would be a lie —
but dropping the reason there is the same inconsistency in its milder form, and
it costs nothing to fix: `_pop` turns `null` into `true` (`gnrbag.js:1420-1422`),
so a change carrying no reason behaves exactly as today, and one that carries a
different reason now propagates it instead of losing it. One rule for both
branches, matching the write path.

## The split: not splitting

Your own comparison is the reason. You write that reverting this PR whole costs
the same as reverting the client half alone, since it restores today's behaviour
and nothing else — so the safety the split buys is the safety we already have.
Against that: this conflicts with #1248, which is approved and lands first, so a
rebase was due anyway (done, `develop` merged in), and splitting would add a
second branch for five lines of JavaScript that cannot be verified on its own
either.

The asymmetry of verification regimes you name is real, and I would rather answer
it with the browser check than with a branch boundary.

## The browser check: not done, and I am not claiming it

This is the one thing still open on this PR, and I am leaving it open rather than
recording something I did not run. You are right that it is not code being
changed but code being switched on: nothing in the repo has ever set
`delete=True` on a store change, so the client's removal path has never executed.

What it needs is a page subscribed to a subtree, a server-side pop of a node with
a sibling under the same parent, and the removal observed to take exactly the
popped node. That is not one of the `gnrcomponents/testhandler` pages, so it is
somebody sitting in front of a real instance — worth saying plainly rather than
approximating it with a unit test that cannot see the client store at all.

## What did run

`tests/web/siteregister_datachanges_del_test.py` 11 passed after the realign,
flake8 clean on the daemon side, `node --check` on `genro_rpc.js`. Thank you for
re-running the ratchet against the unpatched module rather than the working tree —
6 red of 11 is the number I should have quoted in the body.

Re-requesting.

fporcari:
Realigned on `develop` (`fad3206602`), which had gone `DIRTY` after #1248 merged. Both changes land in `_on_data_trigger` and are complementary, so the resolution keeps both sides:

```python
def _on_data_trigger(self, node=None, ind=None, evt=None, pathlist=None, register_item=None,
                     reason=None, **kwargs):
    if reason == 'autocreate':
        return
    if evt in ('ins', 'del'):
        pathlist = pathlist + [node.label]
```

`#1248`'s guard on `reason` sits above and is untouched; this PR's `('ins', 'del')` branch and its non-mutating `pathlist` replace the `ins`-only line. The two do not interact: `reason='autocreate'` is emitted only from `_onNodeInserted` (`gnrbag.py:798`), so no `del` event carries it.

Verified: `siteregister_datachanges_del_test.py` + `siteregister_datachanges_test.py` 14 passed together — including the 3 tests that arrived from #1248 — flake8 clean, and the full suite through the commit hook 2373 passed / 10 skipped.

No code review point is answered here: the changes requested on `3afeb9c10d` are still open.

genro (changes requested):
The code point I raised is answered, and answered better than I asked. Nothing below is a request to change a line: the only thing still open is the browser check, and that is what this state is holding on — I would rather the block be visible than approve and rely on a promise in a comment thread.

## What is closed

> `delItem(clientpath, 'serverChange')`

Written as `delItem(clientpath, reason)`, and inside that branch `reason` **is** `'serverChange'`. I followed it through: `delItem` → `pop` → `popNode` → `_pop`, which fires `onNodeTrigger({... 'reason': doTrigger})` (`gnrbag.js:1434`). So the removal now arrives marked, like the two server-side pops in `gnrsharedobjects.js:53` and `gnrwebsocket.js:181`, and cannot be echoed back through the shared-object branch at `genro.js:1592` (the `:1621` in my earlier review was off by the lines that have moved since; the branch is the same one).

Applying the same rule to the second branch (`:558-560`) was the right generalisation, and I checked the case that could have made it a regression rather than a tidy-up. `_pop` turns `null` into `true`, so a change carrying no reason behaves as before — but only if the reason really is absent rather than empty, because `''` is not `null` and would have silently switched the trigger off. It is absent: `Bag.toXml()` drops an attribute whose value is `None`, so `change_reason=change.reason` with `reason=None` (`siteregister.py:1078`) emits no attribute at all and the client reads `undefined`. Verified, not assumed.

The realign on `fad320660` keeps both sides of the #1248 overlap: the `reason == 'autocreate'` guard sits above, untouched, and the `('ins', 'del')` branch with the non-mutating `pathlist` replaces the `ins`-only line. Correct — `autocreate` is emitted only from `_onNodeInserted`, so no `del` event carries it.

Re-verified at this head merged with `develop` at `d882af9349`: `siteregister_datachanges_del_test.py` + `siteregister_datachanges_test.py` 14 passed together, flake8 clean, no conflict with the current `develop`. Ratchet run again with your test file against the unpatched module rather than the working tree: 6 red of 11, the other 5 green as the write-path guards.

The decision not to split was yours to make and the reasoning holds.

## What is open

The browser check. You said it plainly and did not claim it, which is the right way to leave it — but it cannot be left open through the merge, and here is why it is not a formality.

Nothing in this repository has ever set `delete=True` on a store change. There are four producers of a `ClientDataChange` and only one door a real removal can come through, the Bag trigger this PR fixes; no server API removes a path from the client store, and the habit is writing `None`. So `setDatachangesInData`'s removal path for store changes has never executed, in any deployment, ever. This branch does not modify behaviour that works — it switches on a path that has never run, in a file with no test harness in this repo.

What it needs is the one thing a unit test cannot see: a page subscribed to a subtree, a server-side `pop` of a node that has a sibling under the same parent, and the client store observed to lose exactly the popped node and keep the sibling. Record the result in the PR and I will flip this immediately.

## One thing for whoever merges

#1258 rewrites the same block of `genro_rpc.js` — its segment guard and this PR's `isDelete` branch both replace the `updater(clientpath, ...)` line, and `git merge-tree` confirms the conflict (`siteregister.py` merges by itself, the JS does not). Whichever lands second combines them like this, with the delete inside the guard rather than beside it:

```javascript
if (stringStartsWith(changepath, serverpath_prefix)) {
    let inner = changepath.slice(serverpath_prefix.length);
    if (!inner || inner[0] == '.') {
        let clientpath = clientpath_prefix + inner;
        if (isDelete) { genro._data.delItem(clientpath, reason); }
        else { updater(clientpath, value, attr, reason); }
    }
}
```

fporcari:
Recording what this PR now owes, since #1258 landed first (merged as of today).

#1258 changed the same block of `genro_rpc.js`: it wraps the client-leg update in a segment-boundary guard, so an update is applied only when the remainder after the subscribed prefix is empty or starts with `.`. This branch rewrites the same `updater(clientpath, ...)` line to add the `isDelete` path. `git merge-tree` confirmed the two do not merge on their own — `siteregister.py` does, `genro_rpc.js` does not — so this is a hand combination, and the delete has to happen *inside* the segment guard rather than beside it:

```js
if (stringStartsWith(changepath, serverpath_prefix)) {
    let inner = changepath.slice(serverpath_prefix.length);
    if (!inner || inner[0] == '.') {
        let clientpath = clientpath_prefix + inner;
        if (isDelete) { genro._data.delItem(clientpath, reason); }
        else { updater(clientpath, value, attr, reason); }
    }
}
```

The reason the delete belongs inside and not outside: a delete announced on a path that only shares a character prefix with a subscribed path is the same defect #1258 fixes for updates, and putting `delItem` beside the guard would reintroduce it on the delete leg alone — where it is worse, because a spurious delete removes data rather than writing the wrong node.

This is the resolution to apply when realigning this branch on develop. The CHANGES_REQUESTED on it is separate and still open.

fporcari:
Realigned on develop (3b57e285e1). The conflict is the one predicted on #1258: `genro_rpc.js` only — `siteregister.py` auto-merged.

Resolved as written in that review, with the delete **inside** the segment guard:

```js
if (stringStartsWith(changepath, serverpath_prefix)) {
    let inner = changepath.slice(serverpath_prefix.length);
    if (!inner || inner[0] == '.') {
        let clientpath = clientpath_prefix + inner;
        if (isDelete) {
            // mark the removal as the server's, like the write path
            // below and the two other server-side pops: an unmarked
            // event is echoed back as a client write (genro.js:1621)
            genro._data.delItem(clientpath, reason);
        } else {
            updater(clientpath, value, attr, reason);
        }
    }
}
```

Beside rather than inside would have reintroduced #1255's defect on the delete leg alone, where it is worse: a spurious delete removes data rather than writing the wrong node.

One detail from your review I did not carry over. You noted the `genro.js:1621` in the comment was off "by the lines that have moved since"; in the merged tree it is right again — `:1607` is the `evt == 'upd'` branch, and `:1621` is the unguarded `kw.reason != 'serverChange'` one, which is the one the comment is about. Left as it stands.

Verified on this head: the four siteregister test files together — `siteregister_datachanges_del_test.py`, `siteregister_datachanges_test.py`, plus `siteregister_subscribed_path_test.py` and `siteregister_setinclientdata_test.py`, which arrived from develop with #1258 and #1260 — 28 passed. `node --check` clean on `genro_rpc.js`. Full suite through the commit hook: 2448 passed, 10 skipped.

**No review point is answered here.** The browser check is still not done, and your CHANGES_REQUESTED stands on exactly the right thing: this removal branch has never executed in any deployment, so it is not a formality and I am not going to claim it in a thread. This push only removes the second blocker, the `DIRTY`.

PR #1258, opened 2026-09-03
fix(siteregister): match subscribed_paths on segment boundaries

### LINKED ISSUE #1255: SiteRegister: subscribed_paths match by character prefix, so a change on 'formats.x' is announced to a page subscribed to 'form' and lands on a wrong client path
## What happens

`BaseRegister._on_data_trigger` (`gnrpy/gnr/web/daemon/siteregister.py:198`) filters a page store write with `path.startswith(subscribed)` — a prefix of characters, not of path segments. A page subscribed to `form` (via `setPendingContext` / `subscribe_path`) is announced every write under `formats.*` too.

On the client, `setDatachangesInData` (`gnrjs/gnr_d11/js/genro_rpc.js:545-548`) matches the same way and rebuilds the client path as `clientpath_prefix + changepath.slice(len(serverpath_prefix))`: with `form → cliente.dati`, the change on `formats.date` is written to **`cliente.datiats.date`** — a path nothing reads, silently.

The upload side of the same mechanism is already segment-aware: `genro.js:1579-1582` accepts a registered path only when the remainder is empty or starts with `'.'`. Download and upload disagree.

## When

Two server paths of one page where one is a character prefix of the other: `form`/`formats`, `order`/`orders`, `srv.a`/`srv.ab`. Ordinary names.

## Proposal

Match on segment boundaries on both legs: `path == subscribed or path.startswith(subscribed + '.')` in `_on_data_trigger`, and the same rule in `setDatachangesInData` before slicing.

## Test (red today, stand-in fixture of `gnrpy/tests/web/subscribed_tables_index_test.py`)

```python

class _FakeDaemon:
        pass

class _FakeServer:
    daemon = _FakeDaemon()
    gnr_daemon_uri = None
    hmac_key = None

def test_a_subscribed_prefix_matches_whole_segments_only():
    reg.create('p1')
    reg.subscribe_path('p1', 'form')
    data = reg.get_item_data('p1')
    data.setItem('form.name', 'Ada')
    data.setItem('formats.date', 'dmy')
    assert [c.path for c in reg.get_datachanges('p1')] == ['form.name']
```

Today: `['form.name', 'formats.date']`.

Context: sibling of #1230 and #1249. The ASGI core (genro-asgi, `RegisterRegistry.subscribe_page_store`) already matches on segment boundaries; aligning the daemon keeps the two behaviours identical.

fporcari:
## Problem

A page subscribed to the server path `form` is announced every write under `formats.*` as
well. On the client the same match rebuilds the client path by slicing off the server
prefix, so a change on `formats.date` with `form -> cliente.dati` is written to
`cliente.datiats.date` — a path nothing reads, with no error anywhere.

It takes only two server paths on one page where one is a character prefix of the other:
`form`/`formats`, `order`/`orders`, `srv.a`/`srv.ab`.

## Root cause

Both legs of the datachange mechanism match a subscription by characters instead of by
path segments:

- `BaseRegister._on_data_trigger` (`gnrpy/gnr/web/daemon/siteregister.py`) filtered the
  write with `path.startswith(subscribed)`.
- `setDatachangesInData` (`gnrjs/gnr_d11/js/genro_rpc.js`) matched with
  `stringStartsWith(changepath, serverpath_prefix)` and then sliced the prefix off
  unconditionally.

The upload leg of the very same mechanism is already segment-aware: `genro.js` accepts a
registered path only when the remainder is empty or starts with `'.'`. Download and upload
disagreed.

## Change

Both legs now apply that same rule:

- server: `if path == subscribed or path.startswith(subscribed + '.')`
- client: compute the remainder after the prefix and update only when it is empty or
  starts with `'.'`, mirroring the `genro.js` precedent.

The client loop over `genro._serverstore_paths` keeps its shape — no `break` added: two
distinct client prefixes may legitimately mirror overlapping server subtrees, and stopping
at the first match would drop a real update. That is orthogonal to this fix.

`drop_datachanges` carries the same character-prefix defect and is deliberately left out
of this PR; a twin issue is opened for it (linked below).

## Verification

- New `gnrpy/tests/web/siteregister_subscribed_path_test.py`, built on the stand-in
  server fixture of `subscribed_tables_index_test.py` (real `SiteRegister`, no daemon, no
  Pyro): 6 tests, red before the change (`['formats', 'formats.date']` announced to a page
  subscribed to `form`), green after.
- `pytest gnrpy/tests/ -q` — 2356 passed, 10 skipped.
- `flake8` on the touched Python files — zero errors.
- The client leg has **no JS test harness in this repository** (no jest/karma/mocha, no
  `package.json`), so it was verified by reading against the `genro.js` precedent plus a
  `node --check` parse. No JS test was run, and none is claimed.

Note on the assertion the issue proposes: `== ['form.name']` stays red even with this fix,
because the autocreated parent `form` is still announced — that is #1230, not this bug.
The test here asserts the segment rule without depending on #1230.

## Related issues

Fixes #1255

Twin defect in `drop_datachanges`, out of scope here: see the linked follow-up issue.

Two PRs are already open on the same file, and the three will need a merge order —
textual conflict with #1248 is likely, since it touches the same `_on_data_trigger`:

- #1248 (`fix/1230-autocreate-serverchange`) — same function
- #1252 (`fix/1249-del-datachange-path`) — same file

### DIFF
--- a/gnrjs/gnr_d11/js/genro_rpc.js
+++ b/gnrjs/gnr_d11/js/genro_rpc.js
@@ -546,8 +546,10 @@ dojo.declare("gnr.GnrRpcHandler", null, {
                 for (var clientpath_prefix in genro._serverstore_paths) {
                     var serverpath_prefix = genro._serverstore_paths[clientpath_prefix];
                     if (stringStartsWith(changepath, serverpath_prefix)) {
-                        let clientpath = clientpath_prefix + changepath.slice(serverpath_prefix.length);
-                        updater(clientpath, value, attr, reason);
+                        let inner = changepath.slice(serverpath_prefix.length);
+                        if (!inner || inner[0] == '.') {
+                            updater(clientpath_prefix + inner, value, attr, reason);
+                        }
                     }
                 }
             }
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -196,7 +196,7 @@ class BaseRegister(BaseRemoteObject):
             else:
                 caching_subscribers[register_item_id].add(path)
         for subscribed in register_item['subscribed_paths']:
-            if path.startswith(subscribed):
+            if path == subscribed or path.startswith(subscribed + '.'):
                 register_item['datachanges'].append(
                     ClientDataChange(path=path, value=node.value, reason='serverChange', attributes=node.attr))
                 break
--- a/gnrpy/tests/web/siteregister_subscribed_path_test.py
+++ b/gnrpy/tests/web/siteregister_subscribed_path_test.py
@@ -1,0 +1,81 @@
+"""Tests for the segment-aware matching of ``subscribed_paths`` (#1255).
+
+``BaseRegister._on_data_trigger`` filters a page store write against the paths the page
+subscribed. Matching them by character prefix announces every write under ``formats.*``
+to a page subscribed to ``form``; the client then rebuilds the client path by slicing off
+the server prefix, so the change lands on a path nothing reads. A subscription must match
+whole path segments only.
+
+The page register is taken from a real ``SiteRegister`` built with the stand-in server of
+``subscribed_tables_index_test``: its constructor only needs ``server.daemon.register``,
+so no daemon and no Pyro are involved.
+"""
+
+from gnr.web.daemon.siteregister import SiteRegister
+
+
+class _FakeDaemon:
+    def register(self, obj, name):
+        pass
+
+
+class _FakeServer:
+    daemon = _FakeDaemon()
+    gnr_daemon_uri = None
+    hmac_key = None
+
+
+def _register():
+    return SiteRegister(_FakeServer(), sitename='testsite').page_register
+
+
+def _announced(reg, page_id='page-1'):
+    return [dc.path for dc in reg.get_datachanges(page_id)]
+
+
+def _subscribed_page(subscribed='form'):
+    reg = _register()
+    reg.create('page-1')
+    reg.subscribe_path('page-1', subscribed)
+    return reg, reg.get_item_data('page-1')
+
+
+def test_a_sibling_sharing_a_character_prefix_is_not_announced():
+    """The regression: 'formats' is not under 'form', it only starts with its letters."""
+    reg, data = _subscribed_page()
+    data.setItem('form.name', 'Ada')
+    data.setItem('formats.date', 'dmy')
+    announced = _announced(reg)
+    assert not [p for p in announced if p == 'formats' or p.startswith('formats.')]
+
+
+def test_the_subscribed_subtree_is_still_announced():
+    reg, data = _subscribed_page()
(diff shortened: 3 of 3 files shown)


fporcari:
The twin defect in `drop_datachanges`, referenced above as out of scope for this PR, is now tracked in #1259.

genro (approved):
Correct on both legs, and the precedent you point at is the right one. Approving, with a merge-order note at the end that is worth reading before this lands.

**The rule.** `genro.js:1579-1587` already accepts a registered path only when the remainder after the prefix is empty or starts with `'.'`, and that is exactly what both legs now do. So this is not a new convention: it is the upload leg's rule finally applied to the download leg, which is why the change reads as small as it does.

**The server leg.** I checked the one way `path == subscribed or path.startswith(subscribed + '.')` could be narrower than `startswith(subscribed)` in a way that matters — an empty `subscribed`, which used to match everything and now matches nothing. It cannot happen: the only producer is `setPendingContext` (`siteregister.py:717`), fed by `addToContext`, and `gnrwebstruct/base.py:822` calls it under `if serverpath:`. An empty string never reaches `subscribe_path`.

**The client leg.** `inner[0] == '.'` on a remainder computed from the same slice that builds the client path, so the path that gets written and the path that was matched can no longer disagree. Keeping the loop without a `break` is right for the reason you give, and it is a real case rather than a hypothetical one.

**Verified**, at `96bcb43c6` merged with `develop` at `d882af9349`:

- ratchet run with your test file against the unpatched module rather than the working tree: 2 red of 6 on `develop`, 6 green with the change. The body says "6 tests, red before the change", which reads as all six; two is the honest number, and it is the right two.

**Merge order — a real conflict, not a textual one.** #1252 touches the same block of `genro_rpc.js`. `git merge-tree` confirms `siteregister.py` merges by itself but `genro_rpc.js` does not: your `inner` guard and its `isDelete` branch both rewrite the `updater(clientpath, ...)` line. Whichever lands second has to combine them, and the combination is this — the delete has to happen inside the segment guard, not beside it:

```javascript
if (stringStartsWith(changepath, serverpath_prefix)) {
    let inner = changepath.slice(serverpath_prefix.length);
    if (!inner || inner[0] == '.') {
        let clientpath = clientpath_prefix + inner;
        if (isDelete) { genro._data.delItem(clientpath, reason); }
        else { updater(clientpath, value, attr, reason); }
    }
}
```

#1260 touches the same file but disjoint hunks; it merges clean against both.

One loose end for the record: the twin defect in `drop_datachanges` is #1259, which the body mentions without the number. It is open and unassigned.

fporcari:
Correction accepted, and it is mine to record: the body says "6 tests, red before the change", which reads as all six. Two of the six are red on the unpatched module — the server-leg boundary case and the client-leg one — and the other four pass either way, since they pin behaviour the change preserves. Your number is the honest one.

Merging this now. Noting the `genro_rpc.js` combination on #1252 so it is written down where the work will happen rather than in a merged thread.

PR #1260, opened 2026-09-03
feat(siteregister): report whether a data change was actually filed

### LINKED ISSUE #1253: SiteRegister: setInClientData answers True when filed, False when the target page does not exist
## What happens

`WebPage.setInClientData(page_id=…)` / `setInClientData(filters=…)` (`gnrpy/gnr/web/gnrwebpage.py:2248-2280`) returns nothing. Underneath, `SiteRegister.set_datachange` (`gnrpy/gnr/web/daemon/siteregister.py:322-325`) returns silently when `get_item` finds no such page. The caller cannot tell whether the change was filed or dropped — a `publish` toward a page that has just been closed, or that does not exist yet, is indistinguishable from a delivered one.

## Proposal

- `set_datachange` returns `True` when the change is appended, `False` when the item does not exist.
- `SiteRegister.setInClientData` returns `True` when every resolved target was filed, `False` when at least one was not (with `filters`, the pages resolved at call time).
- `WebPage.setInClientData` and `setInClientData_legacy` propagate that value.

Callers that do not care ignore the return, as today. A caller writing toward a page that may not exist checks for `False`. No exception is introduced.

## Test (red today, stand-in fixture of `gnrpy/tests/web/subscribed_tables_index_test.py`)

```python

class _FakeDaemon:
        pass

class _FakeServer:
    daemon = _FakeDaemon()
    gnr_daemon_uri = None
    hmac_key = None

def test_setInClientData_reports_whether_the_target_exists():
    sr = SiteRegister(_FakeServer(), sitename='testsite')
    sr.page_register.create('p1')
    assert sr.setInClientData('a.b', value=1, page_id='p1') is True
    assert sr.setInClientData('a.b', value=1, page_id='ghost') is False
```

Today both calls return `None`.

Context: sibling of #1231 (`set_serverstore_changes` on an unknown page). The ASGI bridge (genropy-asgi) will answer the same booleans, so the two stacks stay identical on this call.

fporcari:
Fixes #1253

## Motivation

Filing a client data change against a page that is no longer registered is a silent
no-op. `BaseRegister.set_datachange` did a bare `return` when `get_item` found nothing
and returned nothing when it appended either, and `PageRegister.setInClientData` looped
over its targets ignoring every outcome. A caller had no way to distinguish a delivered
change from one dropped on the floor.

This is a missing contract, not a defect: no caller reads the return value today (all
call sites checked), so nothing changes behaviourally — the register simply starts
answering the question.

## Change

Scope stops at the register.

- `BaseRegister.set_datachange` returns `True` when the change is queued and `False`
  when the register item is not (or no longer) registered. Same shape as the existing
  `lock_item` in the same class.
- `PageRegister.setInClientData` collects the outcomes of the `set_datachange` calls it
  makes — one per target page, or one per node for a `Bag` path — and returns `all(...)`.
- Docstrings on both state what the boolean means.

The remote plumbing already propagates a return value (`siteregister_client.py`
`ServerStore.set_datachange`), so nothing there needed touching.

**`WebPage.setInClientData` is deliberately not touched** (third bullet of the issue).
Its websocket branch (`gnrwebsockethandler.py`) posts and forgets and cannot answer the
same question, so returning a boolean there would mean a three-valued contract in which
`if not page.setInClientData(...)` is a false alarm on any `wsk_enabled` site. Better to
leave the page-level API alone than to ship a boolean that lies on half the deployments.

### The one point to confirm

A `filters` that matches **zero** pages returns `True` — the vacuous `all([])`: nothing
was addressed, so nothing was dropped. It is documented in the docstring and covered by
a test. The alternative reading is that "nothing matched" is itself a non-delivery and
should be `False`. Happy to flip it if you prefer that reading; it is a contract call,
not an implementation one.

## Verification

New `gnrpy/tests/web/siteregister_setinclientdata_test.py`, on the `_FakeDaemon` /
`_FakeServer` stand-in already used by `subscribed_tables_index_test.py`: a real
`SiteRegister`, no mocks of the register internals. Covers live page, unknown page,
dropped page, `filters` matching pages, `filters` matching nothing, `Bag` path on a live
and on an unknown page, and the `set_datachange` primitive directly.

All 8 tests are red before the change (both primary assertions return `None`) and green
after.

```
pytest gnrpy/tests/web/siteregister_setinclientdata_test.py -q   ->  8 passed
pytest gnrpy/tests/ -q --ignore=gnrpy/tests/sql                  ->  1194 passed
pytest gnrpy/tests/ -q  (pre-commit hook, full)                  ->  2358 passed, 10 skipped
flake8 on both touched files                                     ->  clean
```

## Related issues

- Merge order: `gnrpy/gnr/web/daemon/siteregister.py` is also touched by #1258
  (issue #1255, `_on_data_trigger`) and by the open PRs #1248 and #1252. The hunks here
  (`set_datachange`, `PageRegister.setInClientData`) are disjoint from all three, so the
  order does not matter, but it is worth knowing they overlap in the file.
- Coherence: the closest sibling, #1231 / PR #1247, answered the same class of silent
  drop with a **WARNING** log instead of a return value. The two answers should end up
  coherent — a boolean here plus a warning there is defensible (different layers), but
  that is a reviewer call worth making explicitly.

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -320,9 +320,14 @@ class BaseRegister(BaseRemoteObject):
         return self.update_item(register_item_id, dict(datachanges=list(), datachanges_idx=0))
 
     def set_datachange(self, register_item_id, path, value=None, attributes=None, fired=False, reason=None, replace=False, delete=False):
+        """Queue a client data change on a register item.
+
+        Return True if the change was queued, False if the item is not registered
+        (any more) and the change was therefore dropped.
+        """
         register_item = self.get_item(register_item_id)
         if not register_item:
-            return
+            return False
         datachanges = register_item['datachanges']
         register_item['datachanges_idx'] = register_item.get('datachanges_idx', 0)
         register_item['datachanges_idx'] += 1
@@ -332,6 +337,7 @@ class BaseRegister(BaseRemoteObject):
         if replace and datachange in datachanges:
             datachanges.pop(datachanges.index(datachange))
         datachanges.append(datachange)
+        return True
 
     def drop_datachanges(self, register_item_id, path):
         register_item = self.get_item(register_item_id)
@@ -718,20 +724,29 @@ class PageRegister(BaseRegister):
 
     def setInClientData(self, path, value=None, attributes=None, page_id=None, filters=None,
                         fired=False, reason=None, public=False, replace=False):
+        """Queue one or more client data changes on the target pages.
+
+        Return True if every change was queued, False as soon as one target page is
+        not registered (any more) and its change was therefore dropped. A ``filters``
+        that matches no page returns True: nothing was dropped because nothing was
+        addressed.
+        """
         if filters:
             pages = [p['register_item_id'] for p in self.pages(filters=filters)]
         else:
             pages = [page_id]
+        done = []
         for page_id in pages:
             if isinstance(path, Bag):
                 changeBag = path
                 for changeNode in changeBag:
                     attr = changeNode.attr
-                    self.set_datachange(page_id, path=attr.pop('_client_path'), value=changeNode.value,
-                                        attributes=attr, fired=attr.pop('fired', None))
+                    done.append(self.set_datachange(page_id, path=attr.pop('_client_path'), value=changeNode.value,
+                                                    attributes=attr, fired=attr.pop('fired', None)))
             else:
-                self.set_datachange(page_id, path=path, value=value, reason=reason,
-                                    attributes=attributes, fired=fired)
+                done.append(self.set_datachange(page_id, path=path, value=value, reason=reason,
+                                                attributes=attributes, fired=fired))
+        return all(done)
 
 
 class SiteRegister(BaseRemoteObject):
--- a/gnrpy/tests/web/siteregister_setinclientdata_test.py
+++ b/gnrpy/tests/web/siteregister_setinclientdata_test.py
@@ -1,0 +1,115 @@
+"""Tests for the boolean contract of ``setInClientData``/``set_datachange`` (#1253).
+
+Filing a data change against a page that is no longer registered is a silent no-op: the
+change is dropped and the caller has no way to tell. The register now answers with a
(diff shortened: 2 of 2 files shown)


genro (approved):
Approving, and answering the contract question you raised rather than leaving it to the merge.

**The vacuous `True` is the right reading, and it is the only one that composes.** `False` has to mean "a change I was told to deliver was dropped", so that `if not setInClientData(...)` says something a caller can act on. A `filters` matching nothing is not a drop: no target existed, and the caller who wrote the filter is the one who decides whether an empty match is interesting — they can ask `pages(filters=...)`. Fold that case into `False` and the boolean stops being about delivery and starts being about two unrelated things at once, which is the shape that makes callers write `if result is False and pages:`. Keep it as it is; the docstring already says so.

**Verified**, at `ca9c92a0c` merged with `develop` at `d882af9349`:

- **no caller reads the return value today.** I checked all 187 `setInClientData` / `set_datachange` occurrences in `gnrpy/gnr`, `projects` and `resources`: not one assigns, tests or returns the result. So the claim of zero behavioural change holds, and there is no site where the new `False` becomes a false alarm.
- **the value survives the wire.** This is the part that could have made the whole PR dead on the path applications actually use, so it is worth stating: `ServerStore.set_datachange` returns it (`siteregister_client.py:418`), and `SiteRegister.__getattr__`'s dispatcher ends in `return h(*args, **kwargs)` (`siteregister.py:1169`), so a call carrying `register_name` propagates the boolean rather than swallowing it.
- ratchet with your test file against the unpatched module rather than the working tree: 8 red of 8 on `develop`, 8 green with the change.

Leaving `WebPage.setInClientData` alone is the right call for the reason you give: a boolean that is honest on the legacy path and a guess on the websocket path is worse than no boolean, because the caller cannot tell which one they got.

On the coherence point with #1247: a boolean here and a WARNING there is defensible and I would keep both. They answer different callers — this one answers code that can react, the log answers an operator reading a running site — and neither replaces the other.

Merge order: `siteregister.py` is also touched by #1252 and #1258, on disjoint hunks. `git merge-tree` confirms this one merges clean against both, so it can land in any position.

PR #1262, opened 2026-09-04
fix(core): keep BagResolver instances out of tracebackBag locals

### LINKED ISSUE #1261: [BUG] Errors raised inside a BagResolver are reported with an Exception Id but never written to sys.error
### Problem Description

When a server exception is raised inside (or in a frame that holds) a `BagResolver` — for example the `StorageResolver` used by the storage tree of the `sys.service` form — the user gets the "Server Error" dialog with an `Exception Id`, but no row is written to `sys.error`. Opening the linked `/sys/ep_error?error_code=...` page shows "Error not found".

Expected: every error that is reported to the user with an error id is persisted in `sys.error`.

Observed on an S3 storage service configured with a custom endpoint and wrong credentials:

```
Server Error: An error occurred (SignatureDoesNotMatch) when calling the ListObjectsV2 operation: ...
Exception Id: 260904-HF65W
```

Root cause: `tracebackBag()` (`gnrpy/gnr/core/gnrerror.py`) copies every frame's `f_locals` into a `Bag`. It neutralizes `Bag` and `GnrStructData` values, but not `BagResolver` instances: `Bag.setItem` stores a resolver as a lazy node. When `sys.error.errorHandler` inserts the record, the adapter serializes the `error_data` column with `toXml()`, the lazy node is resolved, `load()` runs again (here: the S3 `list_objects_v2` call), the same exception is raised inside the insert, and `sys` `Package.errorHandler` swallows it:

```
ERROR gnr.pkg: sys.errorHandler: failed to write to sys.error
```

`GnrApp.errorHandler` has already generated the error id and returns it regardless, so the user sees an id that leads nowhere.

The same happens for any error raised inside `TableResolver`, `BagCbResolver`, `DirectoryResolver`, `XmlStorageResolver`, etc., or whenever a resolver object is a local variable of a frame in the traceback.

### Reproduction Steps

1. Load a `GnrApp` that includes `gnrcore:sys` (e.g. `test_invoice` on sqlite, `db.model.check(applyChanges=True)`).
2. Define a `BagResolver` subclass whose `load()` raises (e.g. a `botocore.exceptions.ClientError`), put it in a `Bag` and access the node so that `load()` runs and raises.
3. In the `except` block call `app.errorHandler(exception=e, error_type='rpc_exception', traceback=True, rpc_method='m', rpc_kwargs=Bag(a=1))`.
4. Call `db.closeConnection()` (what `site.cleanup()` does at the end of a request) and query `sys.error` by `error_code`.

Observed result: an `error_id` is returned, `sys.error` has no row for it, and the log contains `sys.errorHandler: failed to write to sys.error` with the resolver's exception raised from `Bag.toXml` inside `prepareRecordData`.

The same steps with a plain exception (no resolver involved) write the row correctly. Also verified: boto3 client objects, `ClientError`, bytes with NUL, a previous commit on the `system` connection in the same thread, and an uncommitted write on the main connection do not lose the row.

### Environment

genropy `develop` (bd0a43f624), Python 3.13, macOS, sqlite and PostgreSQL

### Additional Material

Proposed fix: in `tracebackBag()` treat `BagResolver` instances like `Bag` values and store a placeholder (`*RESOLVER* <class name>`) instead of the object, so the traceback snapshot is inert. Verified on the reproduction above: the row is written in every scenario.

fporcari:
## Summary

Errors raised inside (or in a frame holding) a `BagResolver` were shown to the user with an `Exception Id` but never persisted in `sys.error`. Reported on the storage tree of an S3 service with a custom endpoint (`SignatureDoesNotMatch` from `ListObjectsV2`).

Fixes #1261

## Root cause

`tracebackBag()` copies each frame's `f_locals` into a `Bag`. `Bag` and `GnrStructData` values are replaced by placeholders, `BagResolver` instances are not: `Bag.setItem` stores them as lazy nodes. When `sys.error.errorHandler` inserts the record, the adapter serializes `error_data` with `toXml()`, the lazy node is resolved, `load()` runs again and raises the same exception inside the insert. `sys` `Package.errorHandler` logs `sys.errorHandler: failed to write to sys.error` and swallows it, while `GnrApp.errorHandler` has already returned the error id.

## Change

- `gnrpy/gnr/core/gnrerror.py`: `BagResolver` locals are stored as `*RESOLVER* <class name>`, like `Bag` locals.

## Tests

- `gnrpy/tests/core/gnrlang_test.py`: `tracebackBag()` on an exception raised inside `BagResolver.load()` must not run `load()` again when serialized and must carry the placeholder.
- `gnrpy/tests/app/test_error_handler.py`: real-DB tests on the `test_invoice` sqlite fixture (`gnrcore:sys` included): `GnrApp.errorHandler` writes the `sys.error` row for a plain exception and for an exception raised inside a resolver, checked after `closeConnection()` as `site.cleanup()` does.

Both new tests fail on `develop` without the fix and pass with it. flake8 clean on touched files; `pytest core web app sql xtnd`: 2353 passed, 10 skipped.

### DIFF
--- a/gnrpy/gnr/core/gnrerror.py
+++ b/gnrpy/gnr/core/gnrerror.py
@@ -5,7 +5,7 @@ import linecache
 import json
 
 from gnr.core.gnrstructures import GnrStructData
-from gnr.core.gnrbag import Bag
+from gnr.core.gnrbag import Bag, BagResolver
 
 
 def _is_library_frame(filename):
@@ -61,6 +61,8 @@ def tracebackBag(limit=None, full_stack=False):
                     v = '*STRUCTURE*'
                 elif isinstance(v,Bag):
                     v = '*BAG*'
+                elif isinstance(v,BagResolver):
+                    v = '*RESOLVER* %s' % v.__class__.__name__
                 elif isinstance(v,(dict,list,tuple)):
 
                     json.dumps(v)
--- a/gnrpy/tests/app/test_error_handler.py
+++ b/gnrpy/tests/app/test_error_handler.py
@@ -1,0 +1,66 @@
+"""Tests for GnrApp.errorHandler persistence into sys.error.
+
+Uses the shared test_invoice fixture (``db_sqlite``), whose instance
+includes ``gnrcore:sys``.
+"""
+
+from core.common import BaseGnrTest
+
+from gnr.core.gnrbag import Bag, BagResolver
+
+
+def setup_module(module):
+    BaseGnrTest.setup_class()
+
+
+def teardown_module(module):
+    BaseGnrTest.teardown_class()
+
+
+class FailingResolver(BagResolver):
+    classKwargs = {'cacheTime': 0, 'readOnly': True}
+    classArgs = []
+
+    def load(self):
+        raise RuntimeError('resolver failure')
+
+
+def _handle(app, exc):
+    return app.errorHandler(exception=exc, error_type='rpc_exception',
+                            traceback=True, notify_user=True,
+                            rpc_method='m', rpc_kwargs=Bag(a=1))
+
+
+def _error_rows(db, error_id):
+    db.closeConnection()
+    return db.table('sys.error').query(
+        columns='$description,$error_type,$rpc_method,$error_data',
+        where='$error_code=:c', c=error_id).fetch()
+
+
+class TestErrorHandlerPersistence:
+
+    def test_plain_exception_is_written(self, db_sqlite):
+        app = db_sqlite.application
+        try:
+            raise ValueError('plain failure')
+        except ValueError as e:
+            error_id = _handle(app, e)
+        rows = _error_rows(db_sqlite, error_id)
+        assert len(rows) == 1
+        assert rows[0]['description'] == 'plain failure'
+        assert rows[0]['error_type'] == 'rpc_exception'
+        assert rows[0]['rpc_method'] == 'm'
+
+    def test_exception_inside_resolver_is_written(self, db_sqlite):
+        app = db_sqlite.application
+        bag = Bag()
+        bag.setItem('root', FailingResolver())
+        try:
+            bag['root']
+        except RuntimeError as e:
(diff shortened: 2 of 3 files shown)


genro (approved):
Right diagnosis, minimal fix, and the tests go through the real database. Approving.

**The fix uses the same criterion the `Bag` itself uses**, which is what makes it complete rather than a patch for the case you hit. `Bag` decides a value is a resolver with `isinstance(value, BagResolver)` — `gnrbag.py:236`, `:1623`, `:1645` — so anything the guard in `gnrerror.py:64` now catches is exactly the set of values that would otherwise become lazy nodes and re-run `load()` at serialization. I checked the one class that could have slipped past: `BagResolverNew` (`gnrbag.py:3163`) does not subclass `BagResolver`, but by that same criterion the `Bag` does not treat it as a resolver either, so it is stored as a plain value and never resolved. Nothing is left uncovered.

I also confirmed the branch order matters and is correct: `BagResolver` is a plain `object` subclass, so before this change it fell past `GnrStructData`, past `Bag`, past `dict/list/tuple`, and landed on the bare `loc[k] = v` — which is precisely how the frame that raised ended up holding a live resolver.

**Verified**, at `c7ab2dd2f` merged with `develop` at `d882af9349`:

- ratchet with your test files against the unpatched module rather than the working tree: `test_exception_inside_resolver_is_written` red on `develop`, green with the change; the plain-exception test green on both, as the regression guard it is meant to be.
- 46 passed on `tests/core/gnrlang_test.py` plus `tests/app/test_error_handler.py` at the merged head.

`tests/app/test_error_handler.py` earns a word of its own: it uses the `db_sqlite` fixture and the real `test_invoice` instance, which does include `gnrcore:sys` (`instanceconfig.xml:6`), so the row goes through the actual insert and the actual `error_data` serialization — the step where the defect lived. A mocked version of this test would have passed on `develop`.

cgabriel (approved):

PR #1263, opened 2026-09-04
fix(web): point the error detail link at the root domain

fporcari:
## Problem

On a multidomain instance the developer link shown on an rpc exception is a dead
link: `https://<host>/sys/ep_error?error_code=260904-HN6HT` returns 404.

Two facts add up:

- `sys.error` rows are always written in the root store: `sys.error.errorHandler`
  opens `tempEnv(connectionName='system', storename=db.rootstore)` explicitly,
  because in multidomain the redirect of `_syspackage` tables to the root store
  is skipped (`gnrsql/execute.py`, `if not self.multidomain`).
- The link was a literal `/sys/ep_error?error_code=...`, with no domain segment.
  In multidomain the first path segment must be a domain (`_main_` or a
  workspace), otherwise `DbStoreRouter.route()` fails and the dispatcher returns
  `NotFound`. Even with a workspace prefix it would read that store's own
  `sys.error`, where the row does not exist.

So the link has to point at the root domain: `/_main_/sys/ep_error?error_code=...`,
which is what the `detail_url` formula column of the model already emits.

## Change

- New `site.rootDomainHomeUri`, next to `current_home_uri`: the home URI of the
  root domain, where the records of the `_syspackage` tables live. It also
  honours the configured `home_uri`, which the old literal ignored.
- The developer link in `_rpcDispatcher` is built through it.
- Same prefix in the `error_test` pages, whose `window.open` calls had the same
  literal.

Single-domain instances are unaffected: the property returns `default_uri`.

## Not in this PR

- `request_host` is stored as `request.host_url`, which ends with a slash, so
  the `detail_url` of the grid comes out with a double slash
  (`https://host//_main_/sys/...`). It works, the path is normalised on parsing,
  but it is one `rstrip('/')` away from being clean.
- From a workspace domain `/<workspace>/sys/th_error` shows that store's own
  `sys.error`, which is always empty. Fixing that means making `_syspackage`
  tables read from the root store in multidomain too: a design change, worth its
  own issue.

## Test plan

- New `test_root_domain_home_uri` in `gnrpy/tests/web/gnrwsgisite_test.py`,
  covering both modes on the real site object.
- Full suite: 2351 passed, 10 skipped.
- flake8 clean on the touched files.

### DIFF
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -674,7 +674,7 @@ class GnrWebPage(GnrBaseWebPage):
             result = '<div>%s</div>' %str(e)
             if error_id:
                 if self.isDeveloper():
-                    detail_url = '/sys/ep_error?error_code=%s' % error_id
+                    detail_url = '%ssys/ep_error?error_code=%s' % (self.site.rootDomainHomeUri, error_id)
                     result = '%s <br/> Exception Id: <a href="%s" target="_blank">%s</a>' % (result, detail_url, error_id)
                 else:
                     result = '%s <br/> Check Exception Id: %s' % (result, error_id)
--- a/gnrpy/gnr/web/gnrwsgisite.py
+++ b/gnrpy/gnr/web/gnrwsgisite.py
@@ -675,6 +675,17 @@ class GnrWsgiSite(object):
             return f'{self.default_uri}{self.currentDomain}/'
         return self.default_uri
 
+    @property
+    def rootDomainHomeUri(self):
+        """Returns the home URI of the rootDomain.
+
+        In multidomain mode _syspackage records live in the root store,
+        so they are reachable only through the rootDomain.
+        """
+        if self.multidomain:
+            return f'{self.default_uri}{self.rootDomain}/'
+        return self.default_uri
+
     def getSubscribedTables(self,tables):
         domain_proxy = self.domains[self.currentDomain]
         if domain_proxy and domain_proxy._register is not None:
--- a/gnrpy/tests/web/gnrwsgisite_test.py
+++ b/gnrpy/tests/web/gnrwsgisite_test.py
@@ -16,6 +16,15 @@ class TestGnrWsgiSite(BaseGnrDaemonTest):
         assert self.site.remote_edit is None
         assert self.site.locale
         
+    def test_root_domain_home_uri(self, monkeypatch):
+        assert not self.site.multidomain
+        assert self.site.rootDomainHomeUri == self.site.default_uri
+        monkeypatch.setattr(type(self.site.db), 'multidb_config',
+                            property(lambda db: {'multidomain': 'true'}))
+        assert self.site.multidomain
+        assert self.site.rootDomainHomeUri == f'{self.site.default_uri}{self.site.rootDomain}/'
+        assert self.site.rootDomainHomeUri == '/_main_/'
+
     def test_apikeys(self):
         r = self.site.getApiKeys("booger")
         assert r is None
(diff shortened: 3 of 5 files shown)


genro (approved):
Correct, and the property is the right shape — a sibling of `current_home_uri` rather than a literal repeated in three places. Approving, with one follow-up that this PR should not carry.

**The premise checks out.** `sys.error.errorHandler` does open `tempEnv(connectionName='system', storename=self.db.rootstore)` (`sys/model/error.py:58-59`), so the rows really are only ever in the root store, and a link without a domain segment really cannot reach them on a multidomain instance.

**The two things that could have made the concatenation wrong, and do not:**

- `default_uri` is normalised to end with `'/'` (`gnrwsgisite.py:462-463`), so `'%ssys/ep_error?...'` cannot produce `//sys` or `/appsys`.
- `/_main_/` is not multidomain-only. `DbStoreRouter.route()` handles `first_segment == self.site.rootDomain` with no guard on `multidomain` (`gnrwsgisite.py:325`), so the prefix the model already emits resolves on a single-domain instance too. I checked this because the natural worry is that the two now disagree — the property returns a bare `default_uri` in single-domain while the model always writes `/_main_/` — and both work.

**Verified**, at `9ca0bdeb3` merged with `develop` at `d882af9349`: `tests/web/gnrwsgisite_test.py` 13 passed; ratchet with your test file against the unpatched module rather than the working tree, `test_root_domain_home_uri` red on `develop`; no conflict with the current `develop`. flake8 reports only `gnrwsgisite.py:69 F824`, which is on `develop` already and not yours.

**The follow-up, which belongs in the same issue as the `rstrip('/')` you already listed.** `sys/model/error.py:28,31` hardcodes `'/_main_/sys/ep_error?error_id=' || $id`. That literal ignores `home_uri` in exactly the way the one you just removed did, so on an instance with a configured `home_uri` the grid's Detail link stays broken while the rpc toast link is now correct. The formula cannot call the property, so it needs its own answer — worth naming in the issue rather than discovering it the next time someone reports a dead link.

Two smaller things while you are in there, neither blocking: the model formula uses `error_id=$id` and the dispatcher uses `error_code=%s`, so the two links reach `ep_error` through different parameters; and the same issue could cover the workspace-domain case your "Not in this PR" already describes.

gnrpy/gnr/web/gnrwebpage.py
```diff
@@ -674,7 +674,7 @@ def _rpcDispatcher(self, *args, **kwargs):
             result = '<div>%s</div>' %str(e)
             if error_id:
                 if self.isDeveloper():
-                    detail_url = '/sys/ep_error?error_code=%s' % error_id
+                    detail_url = '%ssys/ep_error?error_code=%s' % (self.site.rootDomainHomeUri, error_id)
```
cgabriel: can't a more moden and efficient f-string be used here?

cgabriel (approved):
LGTM

gnrpy/gnr/web/gnrwebpage.py
fporcari: Fair, and yes — the file is mixed anyway (22 f-strings against 57 `%` interpolations in `gnrwebpage.py`), so there is no idiom to defend here.

Not folding it into this PR, for one reason that has nothing to do with the change itself: `dismiss_stale_reviews` is on for `develop`, so the push would drop your approval and @genro's on an otherwise-ready PR, and both of you would have to come back for one cosmetic line.

It is item 5 of #1272 instead, together with @genro's follow-up on the same link — `sys/model/error.py:24-34` still hardcodes `/_main_/sys/ep_error?error_id=` and ignores `home_uri`, which is the same literal this PR just removed from the Python side. That PR touches this exact line anyway, so the f-string lands with it rather than on its own.

Resolving this and merging.

PR #1267, opened 2026-09-07
fix: create the archive folders only when archive_and_delete writes an archive

### LINKED ISSUE #836: [BUG] archive_and_delete batch creates empty folders even in delete-only mode
## Problem Description

The `archive_and_delete` resource action ([resources/common/tables/_default/action/_common/archive_and_delete.py](resources/common/tables/_default/action/_common/archive_and_delete.py)) supports three modes via the `mode` batch parameter:

- `A` — Archive only
- `D` — Delete only
- `AD` — Archive and delete

However, in `step_get_dependencies` the batch unconditionally creates the archive folders on the static export path:

```python
self.archive_path   = self.page.site.getStaticPath('site:export_archive','source',name,'records', autocreate=True)
self.files_to_copy  = self.page.site.getStaticPath('site:export_archive','source',name,'files',   autocreate=True)
self.result_path    = self.page.site.getStaticPath('site:export_archive','results','%s.zip' %name, autocreate=-1)
```

When `mode == 'D'` (Delete only), `step_archive` returns immediately:

```python
def step_archive(self):
    if self.mode == 'D':
        return
    ...
```

So those folders (`records/`, `files/`, plus the `results/` parent) are created on disk and never written to. Over time, repeated delete-only runs leave behind empty `archive_for_<table>/...` directory trees in the `export_archive` site folder.

### Expected behavior

The archive paths should only be created when the batch is actually going to write an archive (i.e. `mode in ('A', 'AD')`). For `mode == 'D'` no folders should be touched.

## Reproduction Steps

1. Open any table that exposes the `Archive and delete` action (the action is registered as `_common`, available on every table for `_DEV_,superadmin`).
2. Select one or more records and launch the action.
3. In the dialog, choose **Mode = Delete only** (`D`). The Filename field is intentionally disabled in this mode, so `name` falls back to the default `archive_for_<package>_<table>`.
4. Run the batch.

**Observed result:** under `site:export_archive/source/archive_for_<package>_<table>/` the folders `records/` and `files/` are created and remain empty. No zip is produced (correctly), but the empty directory tree stays on disk.

**Expected result:** in delete-only mode no directory should be created under `export_archive`.

- Module/feature: batch resource action `archive_and_delete` (BaseResourceAction subclass)
- The issue is purely filesystem-side; verify by listing `getStaticPath('site:export_archive','source')` after a delete-only run.

## Additional notes — worth a deeper review of this batch

While looking at it, a few other things would benefit from a closer look:

1. **Folder creation should be lazy / mode-aware.** The cleanest fix is to move the `autocreate=True` calls into `step_archive` (guarded by `mode != 'D'`), and keep `step_get_dependencies` to compute paths only (`autocreate=False`).
2. **Default `name` in delete-only mode.** Since the Filename field is disabled when `mode=='D'`, the batch always falls back to `archive_for_<table>`. That default is meaningless for a pure delete and is what causes a generic, shared folder name to be touched on every delete-only run.
3. **`reltblobj` scoping in `step_archive`.** Inside the `for t in self.index_tables` loop, `onArchiveExport` is called on `reltblobj` (line ~59), but `reltblobj` is only assigned inside the `else` branch a few lines earlier. When `tablename == self.tblobj.fullname` the variable carries over from a previous iteration (or is undefined on the first one). This looks like a latent bug and should be checked alongside the folder fix.
4. **`subtable='*'` + `excludeLogicalDelete` spelling.** Line 26 uses `excludeLogicalDelete=False` while line 54 uses `excludeLogicalDeleted=False`. Worth verifying which one the query layer actually honors.

Points 3 and 4 are out of scope for the title of this issue but they suggest the whole batch deserves a review pass, not just a one-line patch.

## Environment

- Repo: `genropy/genropy`, branch `develop`
- File: `resources/common/tables/_default/action/_common/archive_and_delete.py`
- Reproducible on any Genropy instance exposing the action.

dgpaci:
Fixes #836

## Problem

`archive_and_delete` computed the `export_archive` paths in `step_get_dependencies` with `autocreate`, regardless of the selected mode. In delete-only mode (`D`) `step_archive` returns immediately, so every launch left an empty directory tree behind:

```
export_archive/source/archive_for_<pkg>_<table>/records/
export_archive/source/archive_for_<pkg>_<table>/files/
export_archive/results/
```

Reproduced on a `gnrtest` instance before the fix — all four folders created, all empty, no zip.

## Fix

Path computation moved to `prepareArchiveFolders`, called from `step_archive` **after** the mode guard, so a delete-only run touches nothing under `export_archive`. This also settles point 2 of the issue: the generic `archive_for_<table>` fallback name is still computed (it is the archive filename for `A`/`AD`) but no longer names a folder on a pure delete.

Two further corrections in the same area, both of which were making folders that stayed empty even while archiving:

- `records` is the basename of the pickled bag (`records.pik`), not a folder, so `autocreate=True` was creating a `records/` directory that was never written to. It now autocreates its parent only (`autocreate=-1`), which is what the `records.pik` write needs.
- the `files/<table>/<pkey>/` folder was created before checking `sn.exists`, so a record whose attachment file is missing left an empty folder in the archive. It is now created while copying, only when there is a file to copy.

## Also in scope — points 3 and 4 of the issue

Both were verified against a real instance and both are real defects:

**Point 3 — `reltblobj` scoping.** `step_archive` called `onArchiveExport` on `reltblobj`, which is only bound inside the dependencies branch. Archiving an attachment table itself (`tablename == self.tblobj.fullname`) raised:

```
UnboundLocalError: cannot access local variable 'reltblobj' where it is not associated with a value
```

The table object is now resolved once per iteration as `archivingTable` and used for both the dependencies query and the export hook — `archivingTable` was already being computed for exactly this purpose and then not used.

**Point 4 — query kwarg spelling.** The query layer honours `excludeLogicalDeleted`; the misspelling `excludeLogicalDelete` on line 26 was the only occurrence in the whole repository. `SqlQuery.__init__` collects unknown kwargs into `self.sqlparams`, so it was silently swallowed and the default `excludeLogicalDeleted=True` applied: a logically deleted record produced `curr_records == 0`, and the batch archived and deleted nothing for it. Same class of typo on `addPkeyColumns` (correct name: `addPkeyColumn`), fixed alongside — the effect is that the extra aliased `pkey` column is no longer added to the archived records, which matches both the original intent and the related-tables query a few lines below.

## Test case

`gnrpy/tests/web/archive_and_delete_test.py` — 10 tests, no mocks of the framework. The batch is loaded through the real site resource loader and its steps are run against the `gnrtest` instance, so the assertions look at the real `export_archive` tree on disk and at the real records in the database.

`adm.htmltemplate_atc` is the fixture table: it is an `AttachmentTable`, so it exercises the `onArchiveExport` branch on the archived table itself, and it carries `__del_ts`, so it also covers the logically deleted case.

Coverage:

- delete-only creates no folder at all, and still deletes the records
- archive-only writes `records.pik` and the zip, keeps the records, and creates no `records/` folder
- the pickled archive holds the selected records
- the attachment of the archived table is copied under `files/<table>/<pkey>/` with its content
- a missing attachment file leaves no empty `files/` folder
- a logically deleted record is archived and deleted
- `result_handler` reports the zip url only when archiving

Against the pre-fix resource 8 of these 10 fail (the other two are non-regression guards). Full `gnrpy` suite: 2473 passed, 10 skipped. flake8 clean on both files.

## Follow-up worth its own issue

Not changed here, but found while reviewing: the selected-records query does not pass `bagFields=True`, while the related-tables query a few lines below does. Bag (`X`) columns of the very records being archived are therefore missing from the archive — for example `adm.htmltemplate.data`. On an archive-then-delete that is silent data loss. Happy to open a separate issue for it.

### DIFF
--- a/gnrpy/tests/web/archive_and_delete_test.py
+++ b/gnrpy/tests/web/archive_and_delete_test.py
@@ -1,0 +1,144 @@
+"""Real checks on the ``archive_and_delete`` resource action (#836).
+
+The batch is loaded through the site resource loader and its steps are run
+against the ``gnrtest`` instance, so the assertions look at the real
+``export_archive`` tree on disk and at the real records in the database.
+
+``adm.htmltemplate_atc`` is the fixture table: it is an ``AttachmentTable``,
+so it exercises the ``onArchiveExport`` branch of ``step_archive`` on the
+archived table itself, and it carries ``__del_ts``, so it also covers the
+logically deleted records the batch is expected to archive and delete.
+"""
+
+import os
+import shutil
+import sys
+from datetime import datetime
+
+from core.common import BaseGnrTest
+
+from gnr.app.gnrapp import GnrApp
+from gnr.core.gnrbag import Bag
+from gnr.core.gnrlang import getUuid
+from gnr.web.gnrdummysite import GnrDummySite
+
+TABLE = 'adm.htmltemplate_atc'
+RESPATH = 'action/_common/archive_and_delete'
+ARCHIVE_NAME = 'test_archive'
+
+
+class TestArchiveAndDelete(BaseGnrTest):
+
+    @classmethod
+    def setup_class(cls):
+        super().setup_class()
+        app = GnrApp(cls.test_instance_name)
+        app.db.model.check(applyChanges=True)
+        app.db.commit()
+        cls.site = GnrDummySite(cls.test_instance_name, site_name=cls.test_instance_name)
+        cls.db = cls.site.db
+        cls.tblobj = cls.db.table(TABLE)
+        cls.export_archive = cls.site.getStaticPath('vol:export_archive')
+
+    def setup_method(self):
+        """Start every test from an empty export_archive and one stored attachment"""
+        shutil.rmtree(self.export_archive, ignore_errors=True)
+        letterhead_tbl = self.db.table('adm.htmltemplate')
+        self.letterhead = letterhead_tbl.newrecord(name='letterhead %s' % getUuid())
+        letterhead_tbl.insert(self.letterhead)
+        self.filepath = 'archive_and_delete_test/%s.txt' % getUuid()
+        self.attachment_node = self.site.storageNode('home:%s' % self.filepath)
+        with self.attachment_node.open(mode='w') as attachment:
+            attachment.write('attached content')
+        self.record = self.tblobj.newrecord(filepath=self.filepath, description='attachment',
+                                            maintable_id=self.letterhead['id'])
+        self.tblobj.insert(self.record)
+        self.db.commit()
+
+    def buildBatch(self, **batch_parameters):
+        batch = self.site.loadTableScript(page=self.site.dummyPage, table=TABLE, respath=RESPATH)
+        batch.batch_parameters = batch_parameters
+        batch.selectedPkeys = [self.record['id']]
+        return batch
+
+    def runBatch(self, **batch_parameters):
+        batch = self.buildBatch(**batch_parameters)
+        for step in batch.batch_steps.split(','):
+            getattr(batch, 'step_%s' % step)()
+        return batch
+
+    def sourceFolder(self, name=ARCHIVE_NAME):
+        return os.path.join(self.export_archive, 'source', name)
+
+    def recordExists(self):
+        return self.tblobj.query(where='$%s=:pkey' % self.tblobj.pkey, pkey=self.record['id'],
+                                 excludeLogicalDeleted=False, excludeDraft=False).count()
+
+    def logicalDelete(self):
+        record = self.tblobj.record(pkey=self.record['id'], for_update=True).output('dict')
+        old_record = dict(record)
+        record[self.tblobj.logicalDeletionField] = datetime.now()
+        self.tblobj.update(record, old_record)
+        self.db.commit()
+
+    def test_resource_under_test(self):
+        # the editable install resolves gnr.* to the main checkout, so make sure the
+        # resource being exercised is the one in this working tree
+        checkout = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', '..', '..'))
+        batch = self.buildBatch(mode='D')
+        resource_file = sys.modules[type(batch).__module__].__file__
+        assert resource_file.startswith(checkout + os.sep)
+
+    def test_delete_only_creates_no_folder(self):
+        self.runBatch(mode='D')
+        assert not os.path.exists(self.export_archive)
+
+    def test_delete_only_deletes_the_records(self):
+        self.runBatch(mode='D')
+        assert self.recordExists() == 0
+
+    def test_archive_only_keeps_the_records(self):
+        self.runBatch(mode='A', name=ARCHIVE_NAME)
+        assert self.recordExists() == 1
+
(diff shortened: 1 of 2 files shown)


fporcari (approved):
The fix is in the right place. Moving the path computation out of `step_get_dependencies` into `prepareArchiveFolders`, called from `step_archive` only, means a delete-only run never learns where the archive would have gone — that is a stronger guarantee than creating the folders and cleaning them up, and it is what makes `test_delete_only_creates_no_folder` an assertion about the tree rather than about a cleanup path. `autocreate=-1` on `records` is the right call too: `records` is the basename of `records.pik`, and `autocreate=True` was creating it as a sibling *directory* of the pickle it names. Same for moving `os.makedirs(destfolder)` inside the file loop, past the `sn.exists` guard — a record whose attachments are all gone now leaves nothing behind.

**What the title does not say, and should be findable later.** Three latent defects are fixed in this diff, and none of them is about empty folders:

- `excludeLogicalDelete=False` → `excludeLogicalDeleted=False`. `query()` (gnrpy/gnr/sql/gnrsqltable/query.py:129) names the option `excludeLogicalDeleted` and defaults it to `True`; the misspelled name fell through `**kwargs` into `SqlQuery`, where it became an unused bind parameter rather than an error. So on develop the batch never fetched logically deleted records — it silently skipped exactly the rows that the "archive and delete" mode exists to clear.
- `reltblobj` → `archivingTable` in the `onArchiveExport` branch. In the `tablename == self.tblobj.fullname` branch `reltblobj` is never assigned, so archiving a table that is itself an `AttachmentTable` raised. Confirmed: with only this resource restored to its develop version, the new tests fail with `UnboundLocalError: cannot access local variable 'reltblobj'`. That branch was dead code, not a subtle mis-selection.
- `addPkeyColumns=False` → `addPkeyColumn=False`. Cosmetic in effect — with `columns=None` the pkey is selected anyway, which is why `step_delete_archived`'s `r[self.tblobj.pkey]` kept working — but worth having the name right.

Since the PR title and body describe only the folder fix, whoever reads the merge log later will not find the logical-delete fix, which is the one that changes what the batch actually archives. Worth a line in the body, or a sentence in the merge commit.

**The tests.** Ten cases through the real resource loader against the real `gnrtest` instance, real records, real files on disk, no mocks — and `test_resource_under_test` guards against the editable install resolving `gnr.*` to the main checkout, which is the trap that makes a resource test pass while testing the wrong file. `adm.htmltemplate_atc` is a good fixture choice: being an `AttachmentTable` it reaches the `onArchiveExport` branch, and carrying `__del_ts` it reaches the logical-delete path.

Verified rather than taken from the description: `pytest gnrpy/tests/web/archive_and_delete_test.py` gives 10 passed on 40b6a00b57. Restoring only `resources/common/tables/_default/action/_common/archive_and_delete.py` from `develop` and re-running, 8 of the 10 fail — including `test_delete_only_creates_no_folder` on an `export_archive` that exists, and the `UnboundLocalError` above. The two that still pass are `test_resource_under_test` and `test_delete_only_deletes_the_records`, which is correct: neither depends on the change.

Not verified: no run against PostgreSQL (the fixture is the sqlite `gnrtest` instance), no run of the batch from the UI, and I did not check whether any project overrides `archive_and_delete` or `onArchiveExport` outside this repository.

PR #935, opened 2026-06-05
Fix CSV Sniffer failure with semicolon delimiter (rebased on flatfiles)

EdoFede:
## Fix CSV Sniffer failure with semicolon delimiter

  **Fixes:** #427
  **Supersedes:** #437 (rebased on top of `develop` after the `flatfiles` refactor in #832)

  ### Why a new PR

  PR #437 was opened from a branch that lived on the upstream repo and I no
  longer have write access there, so I cannot push the rebase+port commits to
  its head. This PR carries the same fix from my fork, rebased on the current
  `develop` and ported on top of #832. The previous review history lives in
  #437 — all the requested changes from that review have been addressed here.

  ### Summary

  This PR resolves the CSV sniffer failure when parsing files with semicolon
  delimiters and improves overall CSV dialect detection reliability.

  ### Problem

  The standard Python `csv.Sniffer` had multiple issues:
  - **Inconsistent delimiter detection**: failed to recognize semicolons and
    other non-comma delimiters, even when processing the entire file.
  - **Quote character detection failures**: almost never correctly identified
    the quoting character, causing parsing errors with quoted fields.
  - **Unreliable results**: behavior varied unpredictably based on file content
    and structure.

  ### Solution

  - **Replaced `csv` module with `clevercsv`** for robust CSV dialect detection
    (delimiters: comma, semicolon, tab, pipe, colon; reliable quote detection).
  - **Graceful fallback to the stdlib `csv` module** when `clevercsv` is not
    installed, so the framework keeps working during partial upgrades
    (addresses @mbertoldi's review on #437).
  - **Added centralized file encoding detection** via the `getEncoding()`
    function in `gnrlang.py` (#436).
  - **Introduced `getCsvDialect()`** for flexible CSV file analysis with:
    - automatic delimiter and quote character detection,
    - configurable `start_at_line` to skip initial lines,
    - configurable `detector_max_lines` to bound analysis scope.
  - **Added `start_at_line` support across the CSV processing chain**
    (`getReader()`, `CsvReader`, `getCsvDialect()`) to enable future processing
    of non-standard CSV files generated by banking systems, third-party ERPs,
    and other legacy software that include headers, comments, or metadata
    before the actual data.

  ### Rebase on top of the `flatfiles` refactor (#832)

  After PR #832 moved `CsvReader`, `XmlReader`, `XlsReader`, `XlsxReader`,
  `getReader`, `readCSV`, etc. from `gnrlist.py` to a new `gnr.core.flatfiles`
[...]

PR #1271, opened 2026-09-07
fix(flatfiles): CsvReader with an explicit delimiter, and rows lost to doubled line terminators

### LINKED ISSUE #1270: [BUG] CsvReader breaks on real-world CSVs: ValueError with an explicit delimiter (regression from #935), IndexError on doubled line terminators
Two defects in `gnr.core.flatfiles.CsvReader`, found together while importing a real payroll CSV. The first is a `develop`-only regression, the second a long-standing robustness gap; both are one-liners and both are fixed in the same PR.

### Problem Description

**1. `ValueError: Unknown dialect type: None` — regression**

Any `CsvReader(..., delimiter=X)` / `getReader(..., delimiter=X)` call that does **not** also pass a `dialect` crashes at construction, before a single row is read:

```
ValueError: Unknown dialect type: None
  File "gnr/core/flatfiles.py", line 384, in CsvReader.__init__
    self.rows = csv.reader(self.filecsv, dialect=dialect, delimiter=delimiter)
  File "clevercsv/read.py", line 54, in _make_simple_dialect
    raise ValueError("Unknown dialect type: %r" % dialect)
```

`gnr/core/flatfiles.py:382-388`:

```python
# Delimiter argument has priority over dialect in clevercsv.reader
if delimiter:
    self.rows = csv.reader(self.filecsv, dialect=dialect, delimiter=delimiter)   # dialect is None here
elif dialect:
    self.rows = csv.reader(self.filecsv, dialect=dialect)
else:
    self.rows = csv.reader(self.filecsv, delimiter=',')
```

In this module `csv` is clevercsv (`try: import clevercsv as csv`, a declared dependency in `gnrpy/pyproject.toml`). clevercsv **rejects** `dialect=None`; the stdlib `csv` accepts it and falls back to its own defaults. The `elif` and `else` branches are safe: only the `if delimiter:` branch forwards the `None`.

- Expected: `CsvReader(path, delimiter=';')` reads the file with that delimiter.
- Observed: `ValueError` at construction.

**Scope: `develop` only.** `master` (26.09.04) still does `import csv` (stdlib), and the stdlib accepts `dialect=None` — no released version is affected, hence no `hotfix` label. It does hit every installation running `develop`.

**Regression introduced by** ef436e8b84 (PR #935, *Fix CSV Sniffer failure with semicolon delimiter*, merged into `develop` on 2026-06-10), which swapped the module's `csv` for clevercsv while leaving `dialect=dialect` in the `delimiter` branch. The same line was first written in 95c953026d (2026-01-28), rebased away and reshipped through #935.

**Why the tests did not catch it:** every `CsvReader` test in `gnrpy/tests/core/flatfiles_test.py` either passes an explicit dialect (`test_CsvReader_auto_dialect`, via `getCsvDialect`) or passes neither dialect nor delimiter. The `if delimiter:` branch is never exercised.

**2. `IndexError: list index out of range` on doubled line terminators — not a regression**

CSV files whose records end with a doubled CR (`\r\r\n`, emitted by some Windows exporters) are split by **every** csv reader into an extra **zero-field** row (`[]`) per record — verified identically with clevercsv and the stdlib, in text mode and with `newline=''`, so no reader option avoids it. `CsvReader.__call__` yields those as `GnrNamedList(index, [])`, and any consumer that touches a column dies:

```
File ".../th_alfa_importazione.py", line 99, in rpc_importFileAlfa
  fixed_r = dict((k.lower()... ,v) for k,v in r.items())
File "gnr/core/gnrlist.py", line 256, in items
  result[v] = (k, self[v])
File "gnr/core/gnrlist.py", line 205, in __getitem__
  return list.__getitem__(self, x)
IndexError: list index out of range
```

A blank physical line carries no data in any CSV, so the reader can skip those rows without any caller losing information.

### Reproduction Steps

**Defect 1**

1. Any environment with clevercsv installed (it is a declared dependency, so: any environment).
2. Take a semicolon-separated CSV, e.g. `gnrpy/tests/core/data/test_CsvAuto_SemiColon.csv`.
3. Run:

```python
from gnr.core.flatfiles import CsvReader
CsvReader('gnrpy/tests/core/data/test_CsvAuto_SemiColon.csv', delimiter=';')
```

Observed result: `ValueError: Unknown dialect type: None`. Same through the web proxy: `self.utils.getReader(sn, delimiter=';', detect_encoding=True)`.

**Defect 2**

```python
open('/tmp/doubled.csv', 'w', newline='').write('id;name\r\r\n1;Alice\r\r\n2;Bob\r\r\n')
rows = list(CsvReader('/tmp/doubled.csv', delimiter=';')())
[r['name'] for r in rows]
```

Observed result: 4 rows instead of 2, and `IndexError` on the first column access.

Real-world hit for both: the ts_alfa payroll importer in the teamset project calls `getReader(sn, delimiter=';', detect_encoding=True)` over a 58-record semicolon CSV with doubled terminators — defect 1 killed it at construction, defect 2 on the first row. Across all the projects indexed in Sourcerer, that is the only call site passing `delimiter` without a dialect: small blast radius, total for whoever is in it.

Both reproducible in isolation, see above.

### Environment

genropy `develop` @ 8907f605f7 (VERSION 26.09.04), clevercsv 0.8.5, Python 3.13.2, macOS (Darwin 25.0.0)

### Additional Material

Fixed in #1271: `dialect and delimiter` guard on the first branch (with `delimiter or ','` in the `else`, so an explicit delimiter still wins over the dialect's own), and `if not r: continue` in `__call__`. One test each, both red before the change.

Related: #427 (closed — the `csv_auto`/Sniffer bug whose fix introduced defect 1), #1002 (closed — the other regression from the #832 reader refactor).

fporcari:
Fixes #1270

Two independent defects in `CsvReader`, one commit each.

**1. `ValueError: Unknown dialect type: None` (regression)**

`CsvReader` forwarded `dialect=None` to `clevercsv.reader` whenever an explicit
`delimiter` was passed, and clevercsv rejects it where the stdlib `csv` accepts
it — so every `CsvReader(path, delimiter=';')` / `getReader(sn, delimiter=';')`
died at construction, before reading a single row.

The guard is `dialect and delimiter` on the first branch; the `else` now carries
`delimiter or ','`, so an explicit delimiter still wins over the dialect's own
and the plain no-argument case is unchanged.

Regression from ef436e8b84 (#935), which swapped this module's `csv` for
clevercsv while leaving `dialect=dialect` in the delimiter branch. `develop`
only: `master` still imports the stdlib `csv` here, so no released version is
affected and no hotfix is needed there.

**2. `IndexError: list index out of range` on doubled line terminators**

Records ending with `\r\r\n` — some Windows exporters emit that — are split by
every csv reader into an extra zero-field row per record, verified identically
with clevercsv and the stdlib, in text mode and with `newline=''`. Those were
yielded as `GnrNamedList(index, [])`, so any consumer touching a column got an
IndexError. A blank physical line carries no data in any CSV, so `__call__` now
skips it and no caller loses information. Not a regression: long-standing, and
the same crash happens on `master`.

**Tests:** one per defect, each covering a branch nothing exercised — every
existing `CsvReader` test passes either an explicit dialect or neither argument,
and none feeds a malformed terminator. Both are red on `develop` (with the
ValueError and the IndexError above) and green with this change; the full `gnrpy`
suite passes (2419 passed, 10 skipped).

Verified on the case that surfaced this: a 58-record semicolon payroll CSV with
doubled terminators, which now reads back its 57 data rows through
`getReader(sn, delimiter=';', detect_encoding=True)`.

### DIFF
--- a/gnrpy/gnr/core/flatfiles.py
+++ b/gnrpy/gnr/core/flatfiles.py
@@ -396,6 +396,8 @@ class CsvReader(BaseReader):
 
     def __call__(self):
         for r in self.rows:
+            if not r:
+                continue
             yield GnrNamedList(self.index, r)
         if self._owns_filecsv:
             self.filecsv.close()
--- a/gnrpy/gnr/core/flatfiles.py
+++ b/gnrpy/gnr/core/flatfiles.py
@@ -379,13 +379,14 @@ class CsvReader(BaseReader):
         for _ in range(start_at_line):
             next(self.filecsv)
 
-        # Delimiter argument has priority over dialect in clevercsv.reader
-        if delimiter:
+        # Delimiter argument has priority over dialect in clevercsv.reader,
+        # which rejects dialect=None where the stdlib csv accepts it
+        if dialect and delimiter:
             self.rows = csv.reader(self.filecsv, dialect=dialect, delimiter=delimiter)
         elif dialect:
             self.rows = csv.reader(self.filecsv, dialect=dialect)
         else:
-            self.rows = csv.reader(self.filecsv, delimiter=',')
+            self.rows = csv.reader(self.filecsv, delimiter=delimiter or ',')
 
         self.headers = next(self.rows)
         name = docname if isinstance(docname, str) else (getattr(docname, 'name', None) or '<file-like>')
--- a/gnrpy/tests/core/flatfiles_test.py
+++ b/gnrpy/tests/core/flatfiles_test.py
@@ -572,6 +572,27 @@ def test_CsvReader_auto_dialect():
         assert last_row[9] == expected_description
 
 
+
+def test_CsvReader_delimiter_without_dialect():
+    """Test CsvReader and getReader with an explicit delimiter and no dialect.
+
+    clevercsv rejects dialect=None where the stdlib csv accepts it, so the
+    delimiter has to be passed on its own.
+    """
+    test_file = os.path.join(DATA_DIR, 'test_CsvAuto_SemiColon.csv')
+
+    reader = CsvReader(test_file, delimiter=';', encoding='utf-8')
+    assert reader.ncols == 11
+    assert reader.headers[0] == 'Data contabile'
+    rows = list(reader())
+    assert len(rows) == 6
+    assert rows[5][2] == '-50,00'
+
+    reader = getReader(test_file, delimiter=';', encoding='utf-8')
+    assert reader.ncols == 11
+    assert len(list(reader())) == 6
+
+
 ### ported from gnrlist_test
 def test_getReader():
 
--- a/gnrpy/tests/core/flatfiles_test.py
+++ b/gnrpy/tests/core/flatfiles_test.py
@@ -593,6 +593,28 @@ def test_CsvReader_delimiter_without_dialect():
     assert len(list(reader())) == 6
 
 
+
+def test_CsvReader_doubled_line_terminators():
+    """Test CsvReader on a file whose records end with CR CR LF.
+
+    Some Windows exporters emit a doubled CR, which every csv reader (clevercsv
+    and the stdlib alike, in text mode and with newline='') splits into an extra
+    zero-field row per record.
+    """
+    with tempfile.NamedTemporaryFile(mode='w', suffix='.csv', delete=False, newline='') as f:
+        csv_file = f.name
+        f.write('id;name\r\r\n1;Alice\r\r\n2;Bob\r\r\n')
+
+    try:
+        reader = CsvReader(csv_file, delimiter=';')
+        assert reader.headers == ['id', 'name']
+        rows = list(reader())
+        assert len(rows) == 2
+        assert [r['name'] for r in rows] == ['Alice', 'Bob']
+    finally:
+        os.unlink(csv_file)
+
+
 ### ported from gnrlist_test
 def test_getReader():
 


mbertoldi (dismissed):

genro (approved):
Correct fix, in the right branch, with the right scope. Approving.

The diagnosis holds at every step I could check independently.

**The premise.** With `clevercsv` 0.8.5 installed, `clevercsv.reader(f, dialect=None, delimiter=';')` raises `ValueError: Unknown dialect type: None`; the same call on the stdlib `csv` returns the reader. `flatfiles.py:38` binds `csv` to `clevercsv` when it imports, and `clevercsv` is a declared dependency (`gnrpy/pyproject.toml:31`), so the failing branch is the one that runs on a normal install.

**The origin.** `ef436e8b84` (#935) replaced the single line `csv.reader(self.filecsv, dialect=dialect, delimiter=delimiter or ',')` with the three branches and kept `dialect=dialect` in the first one. On the stdlib that line was fine — the `None` was absorbed — so the regression arrived with the import swap in the same commit, not with the branching.

**The scope.** `master` has no `clevercsv` import in `flatfiles.py`: it uses the stdlib, which accepts `dialect=None`. No released version is affected and no hotfix is due there. Confirmed on `origin/master`.

**The three branches after the change.** I checked each one against `clevercsv`, since the first is now the only place a dialect and a delimiter still meet:

- `dialect and delimiter` — works, and the explicit delimiter does win: with a dialect sniffed on `,` and `delimiter=';'` the reader splits on `;`. The comment above the line is accurate.
- `elif dialect` — untouched.
- `else` with `delimiter or ','` — covers both the broken case and the no-argument case, which stays byte-identical to before.

**The ratchet.** Run with your test file against the unpatched module rather than the working tree, with `clevercsv` present: `1 failed, 43 passed` on `develop`, `44 passed` at `46c4aaaf29`. Without `clevercsv` the new test passes on `develop` too, which is worth knowing for anyone reading it later: the test only exercises the regression where the dependency is actually installed.

**Blast radius.** Nothing in this repo passes an explicit delimiter — the entry points are `page.utils.getReader(file_path, filetype=..., **kwargs)` (`gnrwebpage_proxy/utils.py:412`) and one bare call in `lgcy/th_converter.py:59`. The delimiter comes from application code, so the whole cost of this defect fell outside the repo, on every importer running `develop` that names its separator. That also explains why no existing test caught it: `getReader(path, delimiter=';')` without `filetype='csv_auto'` leaves `dialect` at `None` (`flatfiles.py:537`), and no test in the suite took that path.

flake8 clean on both files.

mbertoldi (approved):

mbertoldi (approved):

cgabriel (approved):

genro:
My approval landed on `850fd2274`, but the review above was written against `46c4aaaf29`: you pushed `fix(flatfiles): skip zero-field rows in CsvReader` two minutes before I submitted, and it also dropped @mbertoldi's approval. I have read that commit since, and the approval stands for it too — this comment records what I checked, so the review is not read as covering three lines when it covers five.

`if not r: continue` in `CsvReader.__call__` is right, and the guard is narrower than it looks: the csv reader yields `[]` for a blank line, while a row of empty fields is `['', '']` and survives. Before, a blank line became `GnrNamedList(self.index, [])`, i.e. a phantom record with every column reading `None` — so this removes a row that was never wanted rather than one that was. Whole file green with `clevercsv` installed: 45 passed, flake8 clean on both files.

One behaviour change worth knowing, since nothing in the diff says it. In a **single-column** CSV an unquoted empty value and a blank line are the same thing to the reader — both `[]` — so an empty value there is now dropped instead of yielding an all-`None` record. Neither result is useful, and I am not asking you to change it; it just belongs in the record.

Please extend the title and the body: both still describe only the dialect fix, and the zero-field skip is a separate defect in a different method, with its own test. Whoever reads this PR later has no way to know it is in here.

genro (approved):

PR #1276, opened 2026-09-09
fix(siteregister): drop datachanges on segment boundaries

### LINKED ISSUE #1259: SiteRegister: drop_datachanges matches by character prefix, so dropping 'form' also discards pending changes on 'formats.*'
## What happens

`BaseRegister.drop_datachanges` (`gnrpy/gnr/web/daemon/siteregister.py`) drops the pending
datachanges of a page with a character-prefix test:

```python
datachanges[:] = [dc for dc in datachanges if not dc.path.startswith(path)]
```

`startswith` compares characters, not path segments, so dropping the changes queued under
`form` also throws away every pending change on `formats.*`, `formatter`, `formdata` — any
sibling whose name happens to begin with the same letters. The page never receives them:
they are removed from the queue before the next poll, silently.

## When

Two paths on one page where one is a character prefix of the other: `form`/`formats`,
`order`/`orders`, `srv.a`/`srv.ab`. Ordinary names.

## Proposal

The same segment-aware rule applied to `_on_data_trigger` in #1255:

```python
datachanges[:] = [dc for dc in datachanges
                  if not (dc.path == path or dc.path.startswith(path + '.'))]
```

## Context

Same defect as #1255, in the other direction (dropping instead of announcing). It was
deliberately left out of the scope of the fix for #1255 so that the two can be reviewed
and tested independently.

fporcari:
## Problem / Root cause

`BaseRegister.drop_datachanges` (`gnrpy/gnr/web/daemon/siteregister.py`) discarded the
datachanges a page has not polled yet with a character-prefix test:

```python
datachanges[:] = [dc for dc in datachanges if not dc.path.startswith(path)]
```

`str.startswith` compares characters, not Bag segments, so dropping the queue under
`gnr.batch.sync` also removed every pending change on `gnr.batch.sync_log.*` — the page
never received them and nothing reported it. Every caller passes a subtree root
(`serverbatch.py:115,147,163`, `gnrwebpage.py:2343`, `chat_component.py:168`), so any
sibling sharing a character prefix (`form`/`formats`, `order`/`orders`, `srv.a`/`srv.ab`)
loses its queued changes silently.

Same defect as #1255 in the dropping direction; #1258 (96bcb43c6b) fixed the announcing
leg at `siteregister.py:202`.

## Change

One-line filter change, mirroring the predicate the announcing leg already uses: a change
is dropped only when its path is the dropped path itself or continues after a dot.

## Verification

From `gnrpy`, with `PYTHONPATH` pointing at this worktree (`gnr.__file__` asserted inside it):

- narrow gate: `python3 -m pytest tests/web/siteregister_datachanges_test.py tests/web/siteregister_subscribed_path_test.py tests/web/siteregister_client_test.py tests/web/siteregister_drop_datachanges_test.py -q` → **20 passed**
- the two regression tests (sibling `sync_log.x`, sibling segment `srv.ab`) were confirmed to **fail on the pre-fix line** and pass after it
- `python3 -m pytest tests/ -q --ignore=tests/sql` → **1292 passed**
- full suite via the pre-commit hook → **2456 passed, 10 skipped**
- `flake8` on both touched files → zero errors

New tests in `gnrpy/tests/web/siteregister_drop_datachanges_test.py` run against a real
`SiteRegister` with the `_FakeDaemon`/`_FakeServer` stand-in of
`siteregister_datachanges_test.py` (no mocks of the register itself): exact path, subtree,
sibling character prefix, sibling segment, unrelated path, missing register item.

## Follow-up (out of scope)

`subscription_storechanges` (`siteregister.py` ~1037) matches the same way and is left for
a separate change.

## Related issue

Fixes #1259

### DIFF
--- a/gnrpy/gnr/web/daemon/siteregister.py
+++ b/gnrpy/gnr/web/daemon/siteregister.py
@@ -347,7 +347,8 @@ class BaseRegister(BaseRemoteObject):
         if not register_item:
             return
         datachanges = register_item['datachanges']
-        datachanges[:] = [dc for dc in datachanges if not dc.path.startswith(path)]
+        datachanges[:] = [dc for dc in datachanges
+                          if not (dc.path == path or dc.path.startswith(path + '.'))]
 
     def subscribe_path(self, register_item_id, path):
         register_item = self.get_item(register_item_id)
--- a/gnrpy/tests/web/siteregister_drop_datachanges_test.py
+++ b/gnrpy/tests/web/siteregister_drop_datachanges_test.py
@@ -1,0 +1,74 @@
+"""Tests for the segment-aware drop of queued datachanges (#1259).
+
+``BaseRegister.drop_datachanges`` discards the changes a page has not polled yet under a
+subtree root. Matching them by character prefix also discards every pending change on a
+sibling whose name merely begins with the same letters -- dropping ``gnr.batch.sync``
+takes ``gnr.batch.sync_log.x`` with it, and the page never receives it. Same defect as
+#1255/#1258 in the dropping direction, so the same predicate applies.
+
+The register is a real ``SiteRegister`` built with the stand-in server of
+``siteregister_datachanges_test``: its constructor only needs ``server.daemon.register``,
+so no daemon and no Pyro are involved.
+"""
+
+from gnr.web.daemon.siteregister import SiteRegister
+
+
+class _FakeDaemon:
+    def register(self, obj, name):
+        pass
+
+
+class _FakeServer:
+    daemon = _FakeDaemon()
+    gnr_daemon_uri = None
+    hmac_key = None
+
+
+def _page_with_queue(*paths):
+    reg = SiteRegister(_FakeServer(), sitename='testsite').page_register
+    reg.create('p1')
+    for path in paths:
+        reg.set_datachange('p1', path, value=1)
+    return reg
+
+
+def _queued(reg):
+    return [dc.path for dc in reg.get_datachanges('p1')]
+
+
+def test_a_sibling_sharing_a_character_prefix_survives_the_drop():
+    """The regression: 'sync_log' is not under 'sync', it only starts with its letters."""
+    reg = _page_with_queue('gnr.batch.sync', 'gnr.batch.sync.thermo', 'gnr.batch.sync_log.x')
+    reg.drop_datachanges('p1', 'gnr.batch.sync')
+    assert _queued(reg) == ['gnr.batch.sync_log.x']
+
+
+def test_the_dropped_path_itself_is_removed():
+    reg = _page_with_queue('gnr.batch.sync')
+    reg.drop_datachanges('p1', 'gnr.batch.sync')
+    assert _queued(reg) == []
+
+
+def test_the_whole_subtree_under_the_dropped_path_is_removed():
+    reg = _page_with_queue('gnr.batch.sync.thermo', 'gnr.batch.sync.log.line')
+    reg.drop_datachanges('p1', 'gnr.batch.sync')
+    assert _queued(reg) == []
+
+
+def test_an_unrelated_path_is_untouched():
+    reg = _page_with_queue('other.node')
+    reg.drop_datachanges('p1', 'gnr.batch.sync')
+    assert _queued(reg) == ['other.node']
+
+
+def test_a_longer_sibling_segment_survives_the_drop():
+    """'srv.ab' shares every character of 'srv.a' but is a different node."""
+    reg = _page_with_queue('srv.a', 'srv.ab.x')
+    reg.drop_datachanges('p1', 'srv.a')
+    assert _queued(reg) == ['srv.ab.x']
+
+
+def test_dropping_on_a_missing_register_item_does_not_raise():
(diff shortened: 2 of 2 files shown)


fporcari (commented):
Independent verification of the approved checklist, run against tested SHA `37287fd6b188df9b95878f161b6fe39baa874f99` in a clean detached worktree (`PYTHONPATH` and `GENRO_GNRFOLDER` pointed at that worktree; `gnr.__file__` asserted to resolve inside it before any test ran).

1. `python3 -m pytest tests/web/siteregister_drop_datachanges_test.py -v` — PASS. 6 passed, 0 failed, 1 warning (pre-existing `getReader` deprecation), 0.58s. All six expected cases present, including the two sibling cases (`gnr.batch.sync_log.x` and `srv.ab.x`) and the missing-register-item case. The tests build a real `SiteRegister` with a stand-in server object, no mocking of the queue itself.

2. `python3 -m pytest tests/web -k siteregister -v` — PASS. 150 passed, 0 failed, 433 deselected, 1.26s. The announcing-direction coverage from #1255/#1258 is included and green: `tests/web/siteregister_subscribed_path_test.py` contributes all six of its tests, among them `test_a_sibling_sharing_a_character_prefix_is_not_announced` and `test_a_segment_subscription_does_not_match_a_longer_sibling_segment`.

3. Caller census for `drop_datachanges` (grep over the whole repo, excluding `.git`) — PASS, no caller depended on sibling clearing.
   - `gnrpy/gnr/web/daemon/siteregister_client.py:422-423` — the store-side pass-through, forwards `path` unchanged; no path shaping of its own.
   - `gnrpy/gnr/web/gnrwebpage_proxy/serverbatch.py:115` (`batch_create`), `:147` (`batch_aborted`), `:163` (`remove_batch`) — all three pass `self.batch_path`, i.e. `'gnr.batch.%s' % batch_id` (line 59), a subtree root. Siblings here are other batches keyed by uuid or caller-supplied `batch_id`; clearing one because another's id happens to share leading characters was never intended, and `remove_batch(all_batches=True)` already iterates batch ids one by one (`:156-164`) rather than relying on a shared prefix. Under the fix `gnr.batch.<id>` still drops the batch node and everything under it (`.end`, `.error`, `.thermo.<line>`), which is all these call sites need.
   - `gnrpy/gnr/web/gnrwebpage.py:2343` (`chatMessageToUser`, disconnect branch) — passes `'gnr.chat.msg.%s' % roomId`, a subtree root; siblings are other rooms and must not be cleared.
   - `resources/common/gnrcomponents/chat_component/chat_component.py:168` (`ct_send_message`, disconnect branch) — same path shape, same conclusion.
   No call site passes a path with a trailing separator or a non-dot separator, so no caller changes behaviour other than by no longer losing a sibling's pending changes.

4. `flake8 gnr/web/daemon/siteregister.py tests/web/siteregister_drop_datachanges_test.py` — PASS, exit 0, zero errors.

Observation, outside the scope of #1259 and not a blocker for this PR: `gnrpy/gnr/web/daemon/siteregister.py:1038`, in the user-store subscription leg, still matches with `changepath.startswith(subpath)` — the same character-prefix comparison this PR fixes at line 351 and #1255/#1258 fixed at line 202. Worth a separate ticket rather than a change here.

Verification result: PASS

genro (approved):
The predicate is the right one, and the argument for it is the strongest kind: the announcing leg of the same register already matches this way. `siteregister.py:202` is `path == subscribed or path.startswith(subscribed + '.')`, and the dropping leg now says the same thing. Two halves of one mechanism that disagreed on what a subtree is.

I checked every caller rather than taking the subtree claim on trust. `serverbatch.py:115`, `:147` and `:163` all pass `self.batch_path`, built at `:58` as `'gnr.batch.%s' % batch_id`; `chat_component.py:168` and the webpage caller pass `'gnr.chat.msg.%s' % roomId`. All five are subtree roots, so nothing that used to be dropped legitimately is missed now.

The tests build a real `SiteRegister` and drive the real `create` / `set_datachange` / `get_datachanges`, and `get_datachanges` defaults to `reset=False`, so reading the queue does not perturb it. Two of the six cases fail on the old line, which is what makes them worth having.

One thing for the record rather than for this PR: the example in the description, `gnr.batch.sync` against `gnr.batch.sync_log`, does not exist in the repo — no path named `sync_log` anywhere. The mechanism is real, since `batch_id` is caller-supplied at `serverbatch.py:95`, so the illustration is doing no harm; it just is not a case anyone has hit.

`subscription_storechanges` at `siteregister.py:1037` still matches by bare prefix. You call it out as a follow-up, so this is only a reminder that the other half is still open.

fporcari:
Merged. The other half you pointed at is now #1321:

> `subscription_storechanges` at `siteregister.py:1037` still matches by bare prefix.

Confirmed on the merged `develop` — `siteregister.py:1038` — and the issue carries the predicate the two fixed legs use, plus the things I checked that turned out not to be involved.

On the example in the description: you are right that no path named `sync_log` exists in the repo. `batch_id` is caller-supplied at `serverbatch.py:95`, so the shape is reachable, but the illustration is invented and I have left it as it stands rather than rewriting the body of a merged PR.

PR #1277, opened 2026-09-09
fix(slotbar): updateSlotsAttr rebuilds only the slots it addresses

### LINKED ISSUE #1269: [BUG] slotbar updateSlotsAttr rebuilds every slot and duplicates registered nodeIds
## Problem Description

`slotbar_updateslotsattr` (`gnrpy/gnr/web/gnrwebstruct/dojo11.py`) pops and re-adds **every** slot of the bar so the new attributes reach the slot handlers:

```python
for slot in slots:
    if slot!='*' and slot!='|' and not slot.isdigit():
        self.pop(slot)
        self._addSlot(slot, ...)
```

`pop` removes the node from the struct but does not unregister the nodeIds its content registered (`checkNodeId` / `page._register_nodeId`), so any slot whose handler assigns a nodeId fails on the second build. On the main bar of a table handler view the query menu does (`th_view.py`: `nodeId='%s_searchMenu_a' % th_root`), so from a `th_top_custom`:

```python
top.bar.updateSlotsAttr(addrow_enable="^.sections.tipo.current?=#v!='tutte'")
```

raises `Server Error: V_<pkg>_<table>_searchMenu_a is duplicated`.

Expected: `updateSlotsAttr` either rebuilds only the slots whose parameters changed, or unregisters the nodeIds of the popped subtree before re-adding it (the same applies to `pop` in general, but the rebuild-all loop is where it bites).

Workaround in use: pop the one slot and `replaceSlots('<slot>', '<slot>', <slot>_<param>=...)`, which re-adds only the missing slot.

## Reproduction Steps

1. Any TH View resource on a table whose view bar has the standard slots (`searchMenu` included).
2. In `th_top_custom(self, top)`: `top.bar.updateSlotsAttr(addrow_enable='^.somepath')`.
3. Open the page: `... searchMenu_a is duplicated`.

## Environment

genropy `develop` as of 2026-09-07; observed on a Demetra instance.

## Additional Material

Found while implementing softwellsrl/demetra#42. The only application-level precedent of `updateSlotsAttr` (a sections-only bottom bar) works because none of its slots register a nodeId.

fporcari:
## Problem / Root cause

`slotbar_updateslotsattr` popped and re-added **every** slot of the bar so that new attributes could reach the slot handlers. Two independent faults followed.

1. **nodeId leak.** `Bag.pop` only detaches the node: nothing removes the nodeIds its subtree registered in `page._register_nodeId` (written only by `GnrDomSrc.checkNodeId`). The rebuild therefore hit the `checkNodeId` assertion. On a table handler view bar the query menu declares `nodeId='<th_root>_searchMenu_a'` (`resources/common/th/th_view.py:1544`), so a `th_top_custom` calling `top.bar.updateSlotsAttr(...)` broke the page build with `Server Error: V_<pkg>_<table>_searchMenu_a is duplicated`.
2. **Parameter loss.** `_addSlot` consumes the slot parameters out of the bar attributes (`toolbarArgs.pop` + `dictExtract(..., pop=True)`), so the rebuild-all loop re-added every *other* slot with an empty parameter dict, silently dropping its original arguments.

## Change

- `gnrpy/gnr/web/gnrwebstruct/base.py`: new `GnrDomSrc.unregisterNodeIds(childname)` next to `checkNodeId`, which walks the child subtree and removes its nodeIds from the register.
- `gnrpy/gnr/web/gnrwebstruct/dojo11.py`: `slotBar` snapshots the slot parameters (`tb._slotArgs`) before the consuming `_addSlot` loop; `slotbar_updateslotsattr` now rebuilds **only** the slots addressed by the incoming kwargs, re-injecting that slot's remembered arguments and unregistering its nodeIds before `pop` + `_addSlot`. Incoming kwargs are merged into the snapshot, so repeated calls accumulate. Bar-level attributes keep being updated as before.

Scope kept minimal on purpose: `Bag.pop` is untouched, no generic `pop` override on `GnrDomSrc`, `slotbar_replaceslots` untouched. Every other `pop` of a struct subtree keeps its current behaviour; widening the unregistration to all removals would need its own regression pass (`th.py:369`, `th_picker.py:21`).

Slot order is unaffected: the client builds the bar from the `slots` attribute and looks children up by name (`gnrjs/gnr_d11/js/genro_components.js`, `createContent_horizontal`), not from child order.

No JS changes.

## Verification

Run in this session, from `gnrpy` with `PYTHONPATH` on this worktree (`gnr.__file__` asserted inside it):

- `python3 -m pytest tests/web/gnrwebstruct_test.py -q` → **66 passed** (57 before, 9 added).
- Regression proof: with the two source files reverted and the new tests in place, **7 of the 9 new tests fail** (`fixed_id is duplicated` among them); the two that pass without the fix cover pre-existing `slotBar` / `replaceSlots` behaviour.
- `python3 -m pytest tests/ -q --ignore=tests/sql` → **1295 passed**; the pre-commit hook ran the whole suite: **2459 passed, 10 skipped**.
- `flake8` on the three touched files → zero errors.

New tests (real structs via `GnrDomSrc_dojo_11.makeRoot(_PageStub())`, two `@struct_method` slot handlers, one of which assigns a fixed nodeId like the query menu): nodeId register cleared for a nested subtree and a no-op on a missing child; `updateSlotsAttr` on a sibling slot no longer duplicates the nodeId; untouched slots keep their parameters and are not rebuilt at all; the addressed slot is rebuilt with the register pointing at the new subtree; repeated calls accumulate; bar attributes still updated; `replaceSlots` still adds the missing slot with its parameters.

## Related issue

Fixes #1269

### DIFF
--- a/gnrpy/gnr/web/gnrwebstruct/base.py
+++ b/gnrpy/gnr/web/gnrwebstruct/base.py
@@ -151,7 +151,25 @@ class GnrDomSrc(GnrStructData):
         :param nodeId: the :ref:`nodeid`"""
         assert nodeId not in self.register_nodeId,'%s is duplicated' %nodeId
         self.page._register_nodeId[nodeId] = self
-        
+
+    def unregisterNodeIds(self, childname):
+        """Remove from the :ref:`nodeid` register every :ref:`nodeid` declared
+        inside a child subtree, so that the child can be rebuilt
+
+        :param childname: the :ref:`childname` of the subtree"""
+        childnode = self.getNode(childname)
+        if childnode is None:
+            return
+        register = self.register_nodeId
+        nodes = [childnode]
+        childvalue = childnode.getStaticValue()
+        if isinstance(childvalue, Bag):
+            nodes += list(childvalue.traverse())
+        for node in nodes:
+            nodeId = node.attr.get('nodeId')
+            if nodeId:
+                register.pop(nodeId, None)
+
     @property
     def register_nodeId(self):
         """TODO"""
--- a/gnrpy/gnr/web/gnrwebstruct/dojo11.py
+++ b/gnrpy/gnr/web/gnrwebstruct/dojo11.py
@@ -723,6 +723,7 @@ class GnrDomSrc_dojo_11(GnrDomSrc):
         namespace = namespace or self.parent.attributes.get('namespace')
         tb = self.child('slotBar',slotbarCode=slotbarCode,slots=slots,childname=childname,**kwargs)
         toolbarArgs = tb.attributes
+        tb._slotArgs = dict(toolbarArgs) #_addSlot consumes the slot parameters out of the attributes
         slots = gnrstring.splitAndStrip(str(slots))
         frame = self.parent
         frameCode = self.getInheritedAttributes().get('frameCode')
@@ -734,7 +735,17 @@ class GnrDomSrc_dojo_11(GnrDomSrc):
         
         #se ritorni la toolbar hai una toolbar vuota 
     
+    def _extractSlotArgs(self,slot,args):
+        slotName = slot.split('@')[0]
+        slotPrefix = '%s_' %slot.replace('@','_')
+        return dict([(k,v) for k,v in list(args.items())
+                        if k==slotName or k.startswith(slotPrefix)])
+
     def slotbar_updateslotsattr(self,**kwargs):
+        slotArgs = getattr(self,'_slotArgs',None)
+        if slotArgs is None:
+            slotArgs = self._slotArgs = dict()
+        slotArgs.update(kwargs)
         self.attributes.update(kwargs)
         toolbarArgs = self.attributes
         slotstr = toolbarArgs['slots']
@@ -746,9 +757,14 @@ class GnrDomSrc_dojo_11(GnrDomSrc):
         frame = self.parent.parent
         prefix = slotbarCode or frameCode
         for slot in slots:
-            if slot!='*' and slot!='|' and not slot.isdigit():
-                self.pop(slot)
-                self._addSlot(slot,prefix=prefix,frame=frame,frameCode=frameCode,namespace=namespace,toolbarArgs=toolbarArgs)
+            if slot=='*' or slot=='|' or slot.isdigit():
+                continue
+            if not self._extractSlotArgs(slot,kwargs):
+                continue
+            toolbarArgs.update(self._extractSlotArgs(slot,slotArgs))
+            self.unregisterNodeIds(slot)
+            self.pop(slot)
+            self._addSlot(slot,prefix=prefix,frame=frame,frameCode=frameCode,namespace=namespace,toolbarArgs=toolbarArgs)
 
     def slotbar_replaceslots(self, toReplace, replaceStr,**kwargs):
         """Allow to redefine the preset bars of the :ref:`slotBars <slotbar>` and the
--- a/gnrpy/gnr/web/gnrwebstruct/dojo11.py
+++ b/gnrpy/gnr/web/gnrwebstruct/dojo11.py
@@ -775,6 +775,7 @@ class GnrDomSrc_dojo_11(GnrDomSrc):
         :param replaceStr: MANDATORY. A string with the list of the slots to add
         """
         self.attributes.update(kwargs)
+        self._slotArgs = dict(getattr(self,'_slotArgs',None) or dict(),**kwargs) #_addSlot consumes the slot parameters out of the attributes
         toolbarArgs = self.attributes
         slotstr = toolbarArgs['slots']
         slotbarCode= toolbarArgs.get('slotbarCode')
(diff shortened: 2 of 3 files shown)


fporcari (commented):
## Independent verification

Tested SHA: `82d433f63cef23dbdae043c403cf155e75a14994` (the head after the one commit pushed here, see step 2). Own worktree on `pr/1277`, own scratch `GENRO_GNRFOLDER`, `PYTHONPATH` on the worktree with `gnr.__file__` asserted inside it.

1. **Targeted tests** (`pytest gnrpy/tests/web/gnrwebstruct_test.py -k 'unregisterNodeIds or slotbar or updateslotsattr or replaceslots' -v`) - 9 passed. `unregisterNodeIds` clears a nested subtree and is a no-op on a missing child; `updateSlotsAttr` rebuilds only the addressed slot, keeps the parameters of the untouched ones, accumulates over repeated calls and no longer duplicates the registered nodeId.

2. **Loose end found and fixed.** `slotbar_replaceslots` fed its slot parameters to `self.attributes` only, and `_addSlot` consumes them out of there: the `_slotArgs` snapshot never saw them, so a later `updateSlotsAttr` on that same slot rebuilt it from a snapshot missing those values - the very defect this PR closes, surviving through `replaceSlots`. That path is the workaround the issue reports as in use, so it matters. Reproduced with a new test (`replaceSlots('*','plain,*',plain_x='zz')` then `updateSlotsAttr(plain_y='ww')`): `KeyError: 'x'`. Fixed by feeding the snapshot in `slotbar_replaceslots`, same idiom as `slotBar`, one line; test green afterwards. Pushed as `82d433f6` on this branch (no force, no rebase). The push resets any standing approval - there were none.

3. **Non-regression, whole file** (`pytest gnrpy/tests/web/gnrwebstruct_test.py -v`) - 67 passed. The pre-commit hook of the pushed commit also ran the whole suite: 2460 passed, 10 skipped.

4. **Real case on a served page.** `sandboxpg` served from the worktree (`gnr web serve sandboxpg --noreload --nodebug -p 8097`, scratch `GENRO_GNRFOLDER`, private gnrdaemon on its own port), plus the same instance and the same scratch folder served on port 8098 with `PYTHONPATH` on a checkout without this change, as a control: only the library differs. Probe: a `test.booking` view resource with `th_top_custom` calling `top.bar.updateSlotsAttr(addrow_enable='^.probe_enable')`, i.e. the issue's reproduction, on the standard view bar carrying the query menu with its fixed nodeId. On 8098 (without the change) the page is blank and the error dialog reads `V_test_booking_searchMenu_a is duplicated` - the issue verbatim, which is also the proof that the served code is what it claims to be. On 8097 (this branch) the page builds: grid with its columns, query row, run, query menu, view configurator, pin, favourites, export, print, mail, chart, totals, menu, delete-row and lock all present and identical to the same view built without the `updateSlotsAttr` call; the only difference is the add-row button, which the incoming `addrow_enable` correctly turns off. Untouched slots intact, no duplicated nodeId. The probe resource and pages were local to the worktree and are not part of the branch.

5. **flake8** on the touched files (`base.py`, `dojo11.py`, `gnrwebstruct_test.py`) - zero errors.

Verification result: PASS

genro (approved):
The diagnosis is right and the evidence for it is checkable: `checkNodeId` is written from exactly one place, `base.py:252` inside `child()`, and nothing ever removed an entry. So `Bag.pop` detaching a node while its nodeId stayed registered was guaranteed to collide on the second build. `unregisterNodeIds` walking `traverse()` covers every descendant, which is what the query menu at `th_view.py:1544` needs.

I verified `_extractSlotArgs` mirrors what `_addSlot` consumes rather than approximating it: `dojo11.py:706-707` pops the bare key and the `'%s_' % slotCode` prefix, and the new helper matches exactly those two. That is the part that would have been easy to get subtly wrong.

Slot order is unaffected because the client looks children up by name, `genro_components.js:6269`.

One behaviour that disappears, worth a line in the description. `dojo11.py:760-762` skips any slot with no matching kwargs, so `updateSlotsAttr(slots='a,b,c')` no longer creates slots that did not exist yet — the names land in the `slots` attribute with nothing behind them. The old code rebuilt everything and so happened to support it. There are no callers in the repo, so this only reaches applications.

Minor: with slots `a` and `a_b`, a kwarg `a_b_x` matches both and rebuilds `a` needlessly. Same ambiguity `dictExtract` already has; no data is lost.

fporcari:
Recording the behaviour change you flagged, since it reaches applications rather than this repo:

> `dojo11.py:760-762` skips any slot with no matching kwargs, so `updateSlotsAttr(slots='a,b,c')` no longer creates slots that did not exist yet — the names land in the `slots` attribute with nothing behind them.

That is right, and it is the intended trade: the old code only supported it by rebuilding everything, which is the bug. An application that relied on it has to create the slot before addressing it. No caller in the repo does.

The `a` / `a_b` ambiguity stays as it is — same shape `dictExtract` already has, a needless rebuild and no data lost.

Merging as is.

PR #1278, opened 2026-09-09
fix(stores): fetchItemByIdentity keeps a stale reply out of the node's validation state

### LINKED ISSUE #1273: [BUG] fetchItemByIdentity's finalize overwrites the node's validation state for a reply it cannot know is current
### Problem Description

`GnrStoreQuery.fetchItemByIdentity`'s `finalize` raises a node-level validation error for a reply it cannot know is still current, and it does so by overwriting the whole validation state rather than adding to it.

`gnrjs/gnr_d11/js/gnrstores.js:644-647`:

```js
var finalize = dojo.hitch(this, function(r) {
    var scope = request.scope ? request.scope : dojo.global;
    if(r.attr.errors){
        this._parentSourceNode.widget._lastQueryError = r.attr.errors;
        this._parentSourceNode.setValidationError({error:r.attr.errors});
    }
```

and `setValidationError` is wholesale, `gnrjs/gnr_d11/js/gnrdomsource.js:1838-1843`:

```js
setValidationError: function(validation_result) {
    this._validations.error = validation_result.error;
    this._validations.warnings = validation_result.warnings;
    this._validations.required = validation_result.required;
    this.updateValidationClasses();
},
```

Called with `{error: r.attr.errors}` and nothing else, it sets `warnings` and `required` to `undefined`.

Two consequences, and the second is the one that matters:

1. `finalize` runs **before** `request.onItem` (`:663`). So by the time any consumer callback can compare the reply against the live value and decide it is stale, the damage is already done: the validation state belonging to the value the datastore has since moved to has been replaced by the stale reply's resolver error, with `warnings` and `required` wiped.
2. `resetValidationError()` in a consumer's discard branch — the shape #1057 added to `DynamicBaseCombo.mixin_setCondition` — is therefore a *normalisation of already-lost state*, not a repair. It clears the stale error, which is right, but it cannot restore the `warnings` and `required` that the overwrite dropped.

The fix belongs in `finalize`, not in its consumers: a reply whose currency the store cannot establish should not raise a node-level validation error at all, or should raise it in a form the consumer can discard without losing the node's own state.

### Reproduction Steps

The interleaving, as an extension of `test_8_condition_race` (`projects/gnrcore/packages/test/webpages/inputfields/dbselect.py`, added by #1057):

1. Open `/test/inputfields/dbselect/8`. Start from `glbl.regione = VAL`, `glbl.provincia = AO`.
2. In one tick, change the condition to `LOM` and write `MI` to the bound datapath (the fixture's `dataRpc`/`dataController` pair already does this).
3. Have the pending identity lookup for `AO` resolve **with resolver errors** — i.e. make the delayed reply also fail under the new condition, rather than simply arriving late as the existing fixture does.
4. The widget now holds `MI`. Its `_validations.error` is the resolver error for `AO`, and `_validations.warnings` / `_validations.required` are `undefined` — the state that belonged to `MI` is gone.

Observed: the node's validation state is that of a value the datastore no longer holds, with the other two fields cleared.
Expected: a reply the store cannot prove current does not touch the node's validation state; if it must, `warnings` and `required` survive.

The `_lastValue` skip (`gnrstores.js:612-617`) and the 60s cache (`:578`, `:603`) make the timing sensitive: `clearCache()` has to have run and `_lastValue` has to be cleared for the remote path to be taken at all, which is exactly the sequence `setCondition` produces. `#1057`'s `test_8_condition_race` is the fixture that reaches it — it uses `method=self.delayedDbSelect`, sleeping one second on `_id=='AO'`, to force the inverse ordering deterministically. The same fixture with a failing delayed reply is the reproduction.

### Environment

develop @ c3fb6cfb3f. The shared `GnrStoreQuery` path, so any `dbSelect` / `dbComboBox` with a bound `condition_*`.

### Additional Material

Raised independently by @genro and @dgpaci in the review of #1057 and left out of it deliberately: the guard there is in the consumer, and this is the store.

**Whoever fixes `finalize` must revisit `DynamicBaseCombo.mixin_setCondition` at the same time.** #1057's discard branch relies on the current ordering — it treats a truthy `widget._lastQueryError` inside `onItem` as proof that *this* reply's `finalize` installed it, which holds only because `fetchItemByIdentity` deletes the flag at `:619` before issuing the call and every other exit that can reach `onItem` is excluded from that call site. Change when or whether `finalize` sets the flag and that inference stops being sound.

**A related residual, deliberately not fixed in #1057** (@dgpaci's own note, recorded here so it is findable rather than tracked): with two *overlapping* condition changes, the inner `setValue(currvalue,false)` is synchronous only while the cache the outer fetch warmed is still there. A second `setCondition` calls `clearCache()`, so if the first reply lands after that, the inner `setValue` goes remote — and that reply travels dojo's own path, with no guard, and can write a stale caption. Much narrower than what #1057 closes. Hardening it means using the item `onItem` already receives instead of re-entering `setValue`, which would re-split the code by widget type (`_callbackSetLabel` is FilteringSelect-only) — the split #1057 removed. Judged not worth doing on its own; worth reconsidering together with the `finalize` fix, since both live on the same path.

fporcari:
## Problem / Root cause

`GnrStoreQuery.fetchItemByIdentity`'s `finalize` (`gnrjs/gnr_d11/js/gnrstores.js`) raised a node-level validation error for a reply whose currency it never checked:

- it set `widget._lastQueryError` and called `setValidationError({error:r.attr.errors})` unconditionally, with no comparison between `request.identity` and the value the datastore actually holds now;
- `finalize` runs **before** `request.onItem`, so a consumer callback could only normalise state that was already lost;
- `setValidationError` is wholesale (`gnrdomsource.js`), and the store was the only caller passing a partial literal, so `warnings` and `required` became `undefined`;
- it dereferenced `._parentSourceNode.widget` without the `.widget` guard used a few lines above on the same path.

The result: on any `dbSelect` / `dbComboBox` with a bound `condition_*`, a late identity reply installed the resolver error of a value the widget no longer holds, and blanked the two other validation fields. The consumer guard added by #1057 could clear the stale error but could not restore `warnings` and `required`.

## Change

`gnrstores.js` — `finalize` compares `request.identity` with the node's live bound value (`sn.getRelativeData(sn.attr.value)`, the same comparison `DynamicBaseCombo.mixin_setCondition` makes). When the reply is stale it writes neither `_lastQueryError` nor the validation state; `cached_values` and `request.onItem` are untouched, so caching and consumers behave exactly as before. When the reply is current it now passes a complete validation object — `{error, warnings: sn.getValidationWarnings(), required: sn.isValidationRequired()}` — so the node's own `warnings` and `required` survive the error write. The missing `.widget` guard is added.

`genro_widgets.js` — `mixin_setCondition` loses the now-dead `_lastQueryError` repair branch: a stale reply no longer sets the flag, so there is nothing left to undo. The plain stale `return` stays.

`projects/gnrcore/packages/test/webpages/inputfields/dbselect.py` — new `test_10_condition_race_validation`: `test_8`'s race fixture (`method=self.delayedDbSelect` sleeping on `_id=='AO'`, plus the `dataRpc`/`dataController` pair that writes `MI` in the same tick as the condition change) with `validate_notnull=True` on the raced node, so the node carries `required`/`warnings` state before the stale reply lands.

The request-token alternative discussed on the issue (a monotonic counter on `GnrStoreQuery`) was deliberately not taken: it adds state to a store shared by every `dbSelect` to cover the narrower overlapping-`setCondition` case, which the issue itself records as a residual rather than a tracked defect. `gnrdomsource.js` is untouched — the fix preserves `warnings`/`required` at the store call site rather than making `setValidationError` partial-safe.

## Verification

Run in this session, from the worktree (`gnr` imports asserted to resolve inside it):

- `flake8` on `projects/gnrcore/packages/test/webpages/inputfields/dbselect.py` — zero errors.
- `node --check` on both touched JS files (`gnrstores.js`, `genro_widgets.js`) — both parse. There is no JS test infrastructure in the repo.
- `python3 -m pytest tests/ -q --ignore=tests/sql` — **1286 passed**.
- The repo's own pre-commit gate (`pytest core sql web app xtnd` from `gnrpy/tests`) ran on the commit — **2450 passed, 10 skipped**.

**Still owed, and the reason this PR is a draft:** the browser pass on the testhandler page `/test/inputfields/dbselect` has NOT been run. It needs a served instance and was left to the operator. It must cover:

- `test_10` (the new case): press *Set LOM / MI*, then after settle assert the widget caption and the datapath are both `MI`, `_validations.error` is null, `_validations.warnings` is still an array, `_validations.required` is unchanged, and the `gnrrequired` class is still on the `focusNode`;
- `test_8` — a stale reply still neither restores nor invalidates the newer value;
- `test_9` — condition change plus `SET null` in the same tick: caption clears, no validation error;
- `test_7` — the single-option branch still updates both widgets;
- `test_1_condition` — a **current** reply that does not fit the condition still raises the error (the guard must not suppress that one).

## Related issue

Fixes #1273

### DIFF
--- a/gnrjs/gnr_d11/js/genro_widgets.js
+++ b/gnrjs/gnr_d11/js/genro_widgets.js
@@ -4999,15 +4999,6 @@ dojo.declare("gnr.widgets.DynamicBaseCombo", gnr.widgets.BaseCombo, {
                 this.setValue(null,false);
                 this.store.fetchItemByIdentity({identity:currvalue,onItem:function(){
                     if(self.sourceNode.getRelativeData(vpath) != currvalue){
-                        // the stale reply may have flagged a value that is no longer
-                        // there. fetchItemByIdentity clears _lastQueryError before
-                        // the call, so a value here belongs to this reply: without
-                        // one there is nothing of ours to undo, and the node may
-                        // meanwhile hold the current value's own error or required
-                        if(self._lastQueryError){
-                            delete self._lastQueryError;
-                            self.sourceNode.resetValidationError();
-                        }
                         return;
                     }
                     self.setValue(currvalue,false);
--- a/gnrjs/gnr_d11/js/gnrstores.js
+++ b/gnrjs/gnr_d11/js/gnrstores.js
@@ -641,10 +641,15 @@ dojo.declare("gnr.GnrStoreQuery", gnr.GnrStoreBag, {
             }
             var finalize = dojo.hitch(this, function(r) {
                 var scope = request.scope ? request.scope : dojo.global;
-                if(r.attr.errors){
-                    this._parentSourceNode.widget._lastQueryError = r.attr.errors;
-                    
-                    this._parentSourceNode.setValidationError({error:r.attr.errors});
+                var sn = this._parentSourceNode;
+                // the reply's error belongs to the identity it asked about: the datastore
+                // may have moved to another value, and setValidationError is wholesale
+                var stale = sn && sn.attr.value && request.identity != sn.getRelativeData(sn.attr.value);
+                if(r.attr.errors && sn && sn.widget && !stale){
+                    sn.widget._lastQueryError = r.attr.errors;
+                    sn.setValidationError({error:r.attr.errors,
+                                           warnings:sn.getValidationWarnings(),
+                                           required:sn.isValidationRequired()});
                 }
                 var result = r.getValue();
                 if (result instanceof gnr.GnrBag) {
--- a/projects/gnrcore/packages/test/webpages/inputfields/dbselect.py
+++ b/projects/gnrcore/packages/test/webpages/inputfields/dbselect.py
@@ -100,7 +100,24 @@ class GnrCustomWebPage(object):
                     lbl='DbSelect',width='25em',nodeId='dbselect_condition_null')
         fb.div('^.sigla',lbl='Datastore value')
         fb.button('Set LOM / null',action="SET .regione='LOM'; SET .sigla=null;")
-        
+
+    def test_10_condition_race_validation(self,pane):
+        "A stale condition reply must not overwrite the validation state of the newer value"
+        pane.data('.regione','VAL')
+        pane.data('.sigla','AO')
+        fb = pane.formbuilder(cols=2, border_spacing='4px')
+        fb.dbSelect(table='glbl.regione',value='^.regione',lbl='Regione',width='25em')
+        fb.dbSelect(table='glbl.provincia',value='^.sigla',method=self.delayedDbSelect,
+                    condition='$regione=:regione',condition_regione='^.regione',
+                    lbl='DbSelect',width='25em',validate_notnull=True,
+                    validate_notnull_error='Manca il valore',
+                    nodeId='dbselect_race_validation')
+        fb.div('^.sigla',lbl='Datastore value')
+        pane.dataRpc('.race_province',self.raceProvince,regione='^.regione',
+                     _if="regione=='LOM'",_else='null')
+        pane.dataController("SET .sigla=province;",province='^.race_province',_if='province')
+        fb.button('Set LOM / MI',action="SET .regione='LOM';")
+
     def test_2_clientmethod(self,pane):
         "Manually set what to display with callbackSelect"
         fb = pane.formbuilder(cols=1, border_spacing='4px')


fporcari:
Browser pass on the testhandler page, served from this branch at f297220b4e3f636be5bcf0712d063fbfb5ac88bd (instance `sandboxpg`, worktree checkout, port 8091; the served `gnrstores.js` was checked to contain the new guard before the first click). Each test loaded alone via `/test/inputfields/dbselect/<n>`, state read from the source nodes after a 3s settle.

| test | action | widget / datastore | `_validations` | `_lastQueryError` |
|---|---|---|---|---|
| test_10_condition_race_validation | Set LOM / MI | MI / MI | error null, warnings [], required null | null |
| test_8_condition_race | Set LOM / MI | MI / MI | error null, warnings [], required null | null |
| test_9_condition_null | Set LOM / null | "" / null | error null, warnings [], required null | null |
| test_7_condition_single_option | Set VAL / AO | AO / AO on both dbSelect and dbComboBox, both names updated | error null on both | null |
| test_1_condition | SET regione=LOM, sigla=AO (current reply, non-fitting value) | AO / AO | error "current value does not fit condition", warnings [] | "current value does not fit condition" |

Before the fix the raced node in test_10 ended with the stale error installed and warnings/required undefined; now the stale reply is discarded and the current-reply error path (test_1) is unchanged.

Unrelated, pre-existing on `develop`: loading the whole page (`/test/inputfields/dbselect` with no test number) fails with `'hdbSelect' is not defined in page 'dbselect.py'` in `test_6_hdbselect`, added by #1088; this branch does not touch that test.

fporcari (commented):
## Independent verification

Tested SHA `f297220b4e3f636be5bcf0712d063fbfb5ac88bd` (current head, single commit). Instance `sandboxpg`, served from a detached worktree of that SHA on port **8095**, own scratch `GENRO_GNRFOLDER` whose `environment.xml` repoints `gnrhome`, the `genropy` project path, `resources`, `webtools` and the `gnr_11`/`dojo_11` static roots at the worktree.

**Served-code proof, before the first click.** `GET /_gnr/11/js/gnrstores.js` (200, 29992 bytes) is byte-identical to the worktree's `gnrjs/gnr_d11/js/gnrstores.js` and carries the new predicate at line 647:

```
var stale = sn && sn.attr.value && request.identity != sn.getRelativeData(sn.attr.value);
```

The same file on `develop` contains no such line. In the live page `gnr.GnrStoreQuery.prototype.fetchItemByIdentity.toString()` matches the guard, and `gnr.widgets.DynamicBaseCombo.prototype.mixin_setCondition.toString()` no longer contains the `_lastQueryError` repair branch (the one `_lastQueryError` still in that method is the pre-existing single-option branch, untouched by this PR).

### Checklist

**1. Re-read the standing verification report and confirm its SHA is still the head — PASSED.**
The browser-pass comment of 2026-09-09 12:44Z cites `f297220b4e3f636be5bcf0712d063fbfb5ac88bd`; `pulls/1278` still reports that as `head.sha`, the PR carries one commit and there are no submitted reviews. Nothing was pushed after the report, so its table stands and did not need redoing. Its central claim was nevertheless re-run here (step 2b below).

**2. The NON-stale case with non-empty `warnings`/`required` — PASSED.**
Run on `/test/inputfields/dbselect/1` (`test_1_condition`, `validate_notnull=True`).

*2a — the error really is installed on a current reply.* With `regione='LOM'` and `sigla='AO'` (a value that does not fit the condition), the node's own `setValidationError` was called by the store with `{error:"current value does not fit condition", warnings:[]}`, and the node ended at `_validations = {error:"current value does not fit condition", warnings:[]}`, `widget._lastQueryError = "current value does not fit condition"`, datapath `AO`, caption `Valle d'Aosta/Vallee d'Aoste (AO)`. The guard does not suppress a current reply.

*2b — `warnings`/`required` survive that write.* On this page the node's own state is always reset to `{error:null,warnings:[],required:null}` by `resetValidationError` before the identity fetch, so a non-empty pre-state is not reachable by clicking alone. It was produced by seeding `_validations = {error:null, warnings:['probe warning'], required:true}` at the moment the store's `fetchItemByIdentity` is entered, and then letting the **unmodified** `finalize` write over it. The condition change to `LOM` (which clears the store cache and so forces a real query) produced:

```
setValidationError({error:"current value does not fit condition",
                    required:true,
                    warnings:["probe warning"]})
```

Both fields came through, and `gnrrequired` stayed on the `focusNode`. Control on the same seeded state, calling `setValidationError` with the pre-fix literal `{error:...}`: `warnings` and `required` both become `undefined` and the `gnrrequired` class is removed. That is exactly the loss this PR closes.

*2c — the new fixture, re-run here.* `/test/inputfields/dbselect/10` (`test_10_condition_race_validation`), *Set LOM / MI*, 6s settle: caption `Milano (MI)`, `widget.getValue()` `MI`, datapath `.sigla` `MI`, `.regione` `LOM`, `.race_province` `MI`, `_validations = {error:null, warnings:[], required:null}`, `widget._lastQueryError` absent. Instrumenting the node showed the stale `AO` reply produced **no** `setValidationError` call at all — the only validation call in the whole interaction was the `resetValidationError` that the value change itself triggers. Every page RPC returned 200.

**3. The branch where `sn.attr.value` is absent — PASSED (exercised, not out of scope).**
Not reachable by clicking, since every `dbSelect` on the page declares `value`, but reachable on a live node and worth pinning down because the short-circuit changes the outcome. With the node's live value at `null` and an identity fetch for `AO` under the `LOM` condition (a reply that does carry `errors`), on the same node:

- `attr.value` **present** — `'AO' != null`, predicate true, reply stale: no `setValidationError` call, `_validations` left at `{error:null,warnings:[],required:null}`, `_lastQueryError` never set.
- `attr.value` **absent** — predicate short-circuits falsy, `stale` false: `setValidationError({error:"current value does not fit condition", warnings:[]})` and `_lastQueryError` set, i.e. the pre-fix behaviour is preserved.

Correct by construction: a node with no bound value path has no live value to compare an identity against, so there is no staleness to detect and the error is installed as before.

**4. `flake8` on `projects/gnrcore/packages/test/webpages/inputfields/dbselect.py` — PASSED.** Zero errors (flake8 7.1.2, pyflakes 3.2.0, CPython 3.13.2).

**5. Report published — this comment.**

### Two notes, neither blocking

- `isValidationRequired()` returns nothing rather than `false` when the node is not required, so the non-stale write stores `required: undefined` where `resetValidationError` had left `null`. Both are falsy and `updateValidationClasses` behaves identically, so there is no observable difference; it is only a slight loss of tidiness in the node's own state.
- The PR's verification checklist expects `test_10` to end with `_validations.required` unchanged and `gnrrequired` still on the `focusNode`. That is not observable on this page: `validate_notnull` returns `required:true` only for an empty value, and the raced node ends holding `MI`, so `required` is `null` and the class is off both before and after. The load-bearing claim — the stale reply writes nothing into the validation state — does hold, as recorded above. The checklist wording, not the code, is what is inaccurate.

Environmental, not attributable to the branch: this sandbox ran without a sitedaemon, which produces the console's `Not registered service` and two 400s on the service-worker registration; they appear before any interaction and no page RPC failed.

Verification result: PASS

genro (commented):
The `setValidationError` half is clearly right and I am not questioning it. `gnrdomsource.js:1842-1847` assigns all three fields unconditionally, so passing `{error:...}` alone wiped `warnings` and `required`; the store was the only caller passing a partial literal, the other three at `genro_frm.js:536`, `:1989` and `:2094` all pass the full validate result. Passing the complete object is the fix.

My question is about the staleness guard itself, and it is the one your own description lists as unverified.

`fetchItemByIdentity` has a second caller: `patch_setValue` at `genro_widgets.js:5427`, which passes the value being set on the widget. If the datastore has not been written with that value yet, `sn.getRelativeData(sn.attr.value)` returns the old value or `undefined`, `stale` is true, and a genuine resolver error is swallowed — the widget keeps a value the condition rejects and shows nothing. Whether that can happen depends on whether `setValue(null,false)` at `genro_widgets.js:5217` touches the datastore before `finalize` runs, which I cannot settle by reading.

That is exactly `test_1_condition`, which you list under what has not been checked. Please run it and say what happens; if the datastore is always written first, the guard is sound and I will approve on that.

Two smaller things.

The description says "the reason this PR is a draft", but the draft flag is not set, so it is sitting in the queue as ready.

And `test_10` does not demonstrate the half of the fix it is named for. `validate_notnull` sets `required` only when the value is null or blank, `genro_frm.js:2380-2384`, and in that fixture `.sigla` goes from `'AO'` to `'MI'`, never blank. So at the moment the stale reply lands, `required` is falsy and `warnings` is empty: the fixture reproduces the race, but nothing in it would notice if `required` and `warnings` were still being destroyed.

fporcari:
Answering the [review](https://github.com/genropy/genropy/pull/1278#pullrequestreview-5235135295). You were right, and the case is worse than the one you described. **This PR should not go in as it stands.**

> Whether that can happen depends on whether `setValue(null,false)` at `genro_widgets.js:5217` touches the datastore before `finalize` runs, which I cannot settle by reading.

It does not, on the direction you were worried about. Run against the current head `f297220b4`, instance `sandboxpg` served from a worktree of that SHA on port 8096, the served `gnrstores.js` checked to carry the new predicate before the first click.

**The call path is the dijit method, not ours.** `gnr.widgets.FilteringSelect.patch_setValue` at `genro_widgets.js:5172` is not in a dbSelect's chain — `dbSelect` is `[dbBaseCombo, BaseSelect]` at `:5403` and `BaseSelect` only sets `_dojotag`. What runs is `dijit.form.FilteringSelect.setValue`, `dojo_libs/dojo_11/dojo_src/dijit/form/FilteringSelect.js:107-140`, whose body is the same and whose line 132 is the `fetchItemByIdentity`. Your reading holds through it.

**On that direction the datastore is written last.** The only write is dijit's `onChange`, which fires from `_handleOnChange` (`_FormWidget.js:224-235`) reached through `_callbackSetLabel`, i.e. **inside `onItem`, after `finalize`**. Every binding-driven write instead precedes the fetch: `SET` / `setRelativeData` / a record load go through `doUpdateAttrBuiltObj` (`gnrdomsource.js:1412-1569`), where the datastore change *is* what triggers `setValue`, and `mixin_setCondition` writes in the same tick at `genro_widgets.js:4994` and `:5000`.

**Reproduced.** With the datastore holding `MI` under `regione=LOM`, the cache cleared, and `w.setValue('AO', true)`: the reply carried the genuine `current value does not fit condition`, and at the moment `finalize` ran `request.identity` was `AO` while `sn.getRelativeData(sn.attr.value)` was still `MI`. `stale` true, `setValidationError` never called, `_lastQueryError` null, no `dijitError` — and the datastore then ended holding `AO`.

**That last part is a regression, not just a suppression.** Emulating the pre-PR function on the same page (the `!stale` term removed and the old partial literal restored) installs the error at the store, and the following validation pass carries it forward — `validate_select` reads `widget._lastQueryError` at `genro_frm.js:2350-2364`, and `BaseSelect.connectForUpdate` turns `validate_select` on for every dbSelect at `genro_widgets.js:5319-5325`. With no form handler, `_doChangeInData` then returns at `genro_widgets.js:1463-1471` *before* `setValueInData` at `:1484`. So today the rejected value does not reach the datastore; with the guard, it does.

Scope, to be fair to the change: a plain user gesture does not reach it. Typing `Valle` and tabbing out produces no identity fetch at all (`_setBlurValue` hits the `_lastValue` / null guards at `gnrstores.js:592` and `:613-617`) and the error arrives from `validate_select`; picking from the dropdown goes through `_doSelect` / `_setValueFromItem`, no fetch. It needs application code calling `setValue` directly. Also unmeasured: the same widget inside a form handler, where `_doChangeInData` publishes the error and falls through to the write anyway, so only the "shows nothing" half would apply.

**What I think the fix is, and I would like your opinion before I write it.** The guard infers staleness from "the datastore disagrees", which cannot distinguish *the datastore moved on* from *the datastore has not caught up yet*. Those are the two directions, and they need different answers. A pending-identity marker set on the widget by `fetchItemByIdentity` before the call, with `finalize` skipping only when a newer fetch superseded it, separates them — which is the request-token idea the issue records as a residual and this PR's description deliberately declined. The evidence above is what changes my mind about it. The alternative, writing the datastore before the fetch on the widget-driven path, touches dijit's own method and I like it less.

The regression table is otherwise unchanged on this head: test_10 and test_8 discard the stale `AO` reply and end on `MI` with no error, test_9 clears the caption with no error, test_7 takes the single-option branch with no identity fetch, and the data-first case of test_1 still installs the genuine error (`stale` false, `dijitError` on the node). Every page request returned 200; the two console errors are present at load on every page and unrelated.

Your two smaller points, both accepted:

- the "the reason this PR is a draft" sentence is stale — the browser pass it refers to was run and posted on 2026-09-09, and the flag was never set. The description stays as it was opened; treat that sentence as withdrawn by this comment.
- `test_10` does not demonstrate the half it is named for. My own verification comment of 09-10 reached the same conclusion from the other side: on that fixture `.sigla` never goes blank, so `required` is falsy and `warnings` empty when the stale reply lands. A fixture that pins the `warnings`/`required` half needs the raced node to be genuinely required at that moment, and that is what the next revision should carry.

Not verified: dbSelect as a grid cell editor, the `switch_*` store callback, the synchronous `finalize` branch (`rpc_sync`), the `cached_values` hit path, and the widget-driven case inside a form handler.

fporcari:
Fixed in `5603297b4`, and the fixture you objected to is rewritten.

**The predicate no longer looks at the datastore.** `fetchItemByIdentity` records the identity the node is asking about and `finalize` discards a reply only when a later ask superseded it:

```js
this._askedIdentity = request.identity;   // first statement of fetchItemByIdentity
...
var stale = this._askedIdentity != request.identity;   // in finalize
```

The store is one per sourceNode (`genro_widgets.js:4938`), so the marker is per node. Loose `!=` so a pkey arriving once as `5` and once as `'5'` is not a false stale.

Why the comparison goes rather than gets a second condition: `setValidationError` is wholesale and describes what the node is *currently asking about*, not what the datastore happens to hold. Every binding-driven write reaches `widget.setValue` at `gnrdomsource.js:1569`, so the marker always moves with the datastore — while the converse is false, and that asymmetry was the whole defect.

**The marker is written before every branch**, which is what makes it correct where the old test was not: each early return resolves an identity without issuing a query, so it supersedes anything in flight exactly as a query does.

| entry point | with the marker |
|---|---|
| `!request.identity` (`:592`) | supersedes an in-flight ask — this is `test_9` and the revised `test_10` |
| `cached_values` hit (`:603`) | resolved from cache, supersedes; `finalize` never runs on that path |
| `isNullOrBlank(id)` (`:611`) | same |
| `_lastValue == identity` (`:616`) | **the trap you would have hit next**: no query is issued, but the ask is recorded, so a slow older reply is discarded |
| `recordNode.caption` shortcut (`:635`) | resolved without a query, supersedes |
| query issued | current → `_lastQueryError` plus the complete `setValidationError`; stale → neither, with `cached_values` and `onItem` untouched |

Browser pass from a worktree of this code, instance `sandboxpg` on port 8097, the served `gnrstores.js` checked to contain `_askedIdentity` before the first click.

- **Widget-driven (`test_11`, new)** — datastore `MI` under `regione=LOM`, cache cleared, `w.setValue('AO',true)`: `setValidationError` fires with `current value does not fit condition` while `_askedIdentity` is `AO` and the datastore still reads `MI` — precisely the reply the old predicate swallowed. `dijitError` on, `_lastQueryError` set, and **the datastore still holds `MI`**: the rejected value does not land, as on `develop`.
- **`test_10` (rewritten)** — the race now ends on a blank, driven through an rpc and a controller calling `widget.setValue(null,true)` so the real validation pipeline runs, with notnull as a *warning* (`validate_notnull_iswarning`) so the blank actually reaches the datastore instead of being stopped by `_doChangeInData`'s error return. At the moment the stale `AO` reply lands the node genuinely carries `{"warnings":["Manca il valore"],"required":true}` — which is what your objection was about — and the reply installs nothing. End state: datastore `null`, `gnrrequired` still on, no `dijitError`.
- **`test_8`**, **`test_9`**: stale `AO` reply installs nothing, end states `MI` and blank respectively. **`test_7`**: single-option branch, no identity fetch at all, both widgets and both captions updated. **`test_1_condition`**: the current reply still installs the error (`_askedIdentity` `MI`, `dijitError` on). For that last one the condition has to move to `PIE`, not `VAL` — with `VAL` the single-option branch pre-empts the fetch and the test proves nothing.
- **The `_lastValue` branch, constructed explicitly** on test_9's widget: `setValue('AO',false)` with a slow reply, then `setValue('MI',false)` which hits the early return without issuing a query. The late `AO` reply installed nothing, *while the datastore still read `AO`* — so the old predicate would have installed a stale error there too. That case was never reachable by the old test in either direction.

Console shows only the service-worker pair present on every page load before any interaction; every rpc returned 200. flake8 clean on the fixture, `node --check` clean, and the commit hook ran the full suite at 2450 passed, 10 skipped.

Not exercised: the cached / `isNullOrBlank` / caption early returns as the *superseding* event in a real race — no fixture reaches them mid-flight, and their marker write is the same statement as the others'. Also untouched and pre-existing: `hdbSelect`, `RemoteBaseCombo`, and the widget inside a grid editor or a form with a `formHandler`, where `_doChangeInData` does not return before `setValueInData`, so only the "shows nothing" half of the old behaviour would have applied.

The description still describes the first commit's predicate. I am leaving it as it was opened; this comment and the commit message are the record of what the check is now.

genro (commented):
Reviewed `5603297b46`. The code is right; the description is not, and that is what I am holding this on.

The body says `finalize` compares `request.identity` with the node's live bound value, `sn.getRelativeData(sn.attr.value)`, "the same comparison `DynamicBaseCombo.mixin_setCondition` makes". The code does something else: `this._askedIdentity != request.identity`, a last-ask marker on the store, introduced by `5603297b46` — your own second commit, whose subject says so. The description was never rewritten after that push.

The implemented logic is the better of the two and I have no objection to it. `_parentSourceNode` is singular, so the store belongs to one node and the marker cannot be crossed by another widget's fetch; recording it before the early returns is the right call, since an identity resolved without a query would otherwise leave an older reply looking current.

Two smaller mismatches in the same direction. `test_11_widget_value_out_of_condition` is in the diff and not in the description. And the description says "the reason this PR is a draft" — it is not a draft, and the browser pass it lists as still owed is still owed.

Rewrite the body against the current head and I will approve it.

fporcari:
@genro re-requesting your review — the push that answered you never re-requested it, so this has been sitting on nobody since the 17th rather than on you. My apologies for the four days.

What changed since [your review](https://github.com/genropy/genropy/pull/1278#pullrequestreview-5235135295), both on the head `5603297b4`:

- **The predicate no longer looks at the datastore.** You were right that the comparison could not tell *the datastore moved on* from *the datastore has not caught up yet*, and the [answer of the 17th](https://github.com/genropy/genropy/pull/1278#issuecomment-5745022490) shows the second direction was a regression rather than a suppression: with the old guard the rejected value reached the datastore, where on `develop` it does not. It is now a per-sourceNode `_askedIdentity` marker, written as the first statement of `fetchItemByIdentity` and before every early return, so a branch that resolves without issuing a query still supersedes an ask in flight.
- **`test_10` is rewritten**, because you were right that it did not demonstrate the half it is named for. The raced node now genuinely carries `{"warnings":["Manca il valore"],"required":true}` at the moment the stale reply lands, driven through a real rpc and the real validation pipeline.

The `_lastValue` early return is the case worth your eye: no query is issued there, so the old predicate had no way to supersede an in-flight ask, and the constructed run showed the late reply installing a stale error while the datastore still read the old value. That branch was unreachable by the old test in either direction.

CI is green on the whole matrix and there are no open threads. CODEOWNERS on `gnrjs/` and `projects/` is you and me, so your approval is the one this needs.

Issue #1165, opened 2026-08-25
[BUG] page.loadTemplate() cannot run on sqlite: adm.userobject formula columns emit Postgres-only string_to_array


dgpaci:
### Problem Description

`page.loadTemplate()` cannot run on any sqlite instance.

Before falling back to the table's `tpl/` resource, `loadTemplate` always queries `adm.userobject` (`gnrpy/gnr/web/gnrwebpage.py:880`):

```python
data, metadata = self.db.table('adm.userobject').loadUserObject(objtype='template', code=pkey, tbl=table)
```

`adm.userobject` carries three formula columns — `is_mail`, `is_row`, `is_print` — that compile to the Postgres-only `string_to_array`:

```sql
( :env_is_mail_email  LIKE ANY(string_to_array("t0"."flags",',')) ) AS t0_is_mail,
( :env_is_row_row     LIKE ANY(string_to_array("t0"."flags",',')) ) AS t0_is_row,
( :env_is_print_print LIKE ANY(string_to_array("t0"."flags",',')) ) AS t0_is_print
```

On sqlite the statement dies with `sqlite3.OperationalError: no such function: string_to_array`, **before** the resource fallback is ever reached. The generator is `variantColumn_captions`-style SQL, but here the three columns come from the `userobject` model itself.

Observed behaviour: any template load on a sqlite instance raises.
Expected behaviour: `loadTemplate` works on every db implementation. A template stored as a `tpl/*.xml` resource must be loadable without any Postgres-only SQL — either by making the `userobject` lookup db-portable, or by not letting a failure there abort the resource fallback.

Two consequences worth stating explicitly:

1. no sqlite instance can print or render through a template at all;
2. no test in the suite can cover template loading end to end, because the whole `BaseGnrTest` infrastructure runs on sqlite. This is what forced the template-printing tests in #1045 to stub `page.loadTemplate`.

### Reproduction Steps

1. On `develop`, create `gnrpy/tests/web/probe_test.py`:

```python
from core.common import BaseGnrTest

from gnr.app.gnrapp import GnrApp
from gnr.web.gnrdummysite import GnrDummySite


class TestLoadTemplateOnSqlite(BaseGnrTest):

    @classmethod
    def setup_class(cls):
        super().setup_class()
        app = GnrApp(cls.test_instance_name)
        app.db.model.check(applyChanges=True)
        app.db.commit()
        cls.site = GnrDummySite(cls.test_instance_name, site_name=cls.test_instance_name)

[...]

PR #1279, opened 2026-09-09
fix(print): honour a string pdf path in TableTemplateToHtml

### LINKED ISSUE #1264: [BUG] TableTemplateToHtml.__call__ silently ignores a string pdf path
### Problem Description

`TableTemplateToHtml.__call__` accepts a `pdf` argument but only honours it when it is exactly `True`. The two guards disagree:

```python
if pdf :                                                  # gnrbaseclasses.py:311
    filepath = filepath or self.getHtmlPath(f'{self.getDocName()}.html')
result = super(TableTemplateToHtml, self).__call__(record=record, htmlContent=htmlContent, filepath=filepath, **kwargs)
if pdf is True:                                           # gnrbaseclasses.py:314
    return self.writePdf()
return result
```

A string path is truthy at :311, so the HTML file is written as if a PDF were coming, and then falsy at :314, so `writePdf()` never runs. `script(record=pkey, pdf='/some/path.pdf')` therefore returns the rendered HTML string, writes nothing at the requested path, and raises no error.

`TableScriptToHtml` handles the same argument correctly:

```python
callingPdfPath = pdf if isinstance(pdf,str) else None     # gnrbaseclasses.py:350
...
self.writePdf(docname=self.getDocName(), pdfpath=callingPdfPath)   # :368
self.writePdf(filepath=result, docname=self.getDocName(), pdfpath=callingPdfPath)  # :370
```

so the two table-script base classes disagree on the meaning of a string `pdf`.

Expected: either the string is threaded through as the output path the way `TableScriptToHtml` does it, or a non-boolean `pdf` is refused outright. Silently returning HTML and writing nothing is the one behaviour that should not survive.

This is pre-existing and #1045 does not touch either line, but #1045 is what makes it likely to be hit: `tables/_default/html_res/print_recordtemplate` becomes the standard entry point for printing one record, so it will be reached by callers that have a destination path in hand rather than a temporary folder.

### Reproduction Steps

1. Take any table with a template resource, e.g. `adm.user` and its `homepage` template.
2. `script = page.loadTableScript(table='adm.user', respath='html_res/print_recordtemplate', record_template='homepage')`
3. `out = script(record=<pkey>, pdf='/tmp/badge.pdf')`

Observed result: `out` is the rendered HTML string, `/tmp/badge.pdf` does not exist, no exception is raised.
Expected result: either the PDF is written to `/tmp/badge.pdf`, or the call fails telling the caller that `pdf` must be a boolean.

For the contrast, the same `pdf='/tmp/x.pdf'` on a `TableScriptToHtml` resource writes the PDF at the given path.

Reproducible in isolation: `gnrpy/tests/web/gnrbaseclasses_test.py::TestTemplatePrinting` already builds the whole fixture (real `GnrDummySite`, real `adm.user` record, real `homepage` template resource) and `test_default_respath_writes_a_pdf` exercises the `pdf=True` path a string `pdf` should join.

### Environment

develop @ 31f95562e5 — `gnrpy/gnr/web/gnrbaseclasses.py` lines 311 and 314

### Additional Material

Related: #1045 (introduces the generic `print_recordtemplate` resource that makes this path common), #1166 (the sibling trap: a template name that resolves to nothing produces a valid document instead of an error).

fporcari:
## Problem / Root cause

`TableTemplateToHtml.__call__` guarded the same `pdf` argument twice with
different predicates:

```python
if pdf :                 # truthy: a string passes
    filepath = filepath or self.getHtmlPath(f'{self.getDocName()}.html')
result = super(TableTemplateToHtml, self).__call__(...)
if pdf is True:          # identity: a string fails
    return self.writePdf()
return result
```

A string path therefore passed the first guard (the html was written as if a
PDF were coming) and failed the second, so `writePdf()` never ran:
`script(record=pkey, pdf='/tmp/badge.pdf')` returned the rendered HTML string,
wrote nothing at the requested path and raised nothing.

The sibling base class `TableScriptToHtml.__call__` already gives a string
`pdf` the meaning of an output path (`callingPdfPath = pdf if isinstance(pdf,str) else None`,
forwarded as `pdfpath` to `writePdf`), so the two table-script base classes
disagreed on the same argument.

## Change

One method, `gnrpy/gnr/web/gnrbaseclasses.py` `TableTemplateToHtml.__call__`:
copy-adapt the `TableScriptToHtml` idiom — keep the string as `callingPdfPath`
and make the second guard `if pdf: return self.writePdf(pdfpath=callingPdfPath)`.
`BagToHtmlWeb.writePdf` already honours an explicit `pdfpath`, so neither
`writePdf` nor `getPdfPath` needed a change.

Behaviour: a string `pdf` is written at the requested path and returned;
`pdf=True` still writes the PDF beside the html and returns its path;
`pdf=None` still returns the html.

Tests in `gnrpy/tests/web/gnrbaseclasses_test.py` (`TestTableTemplatePdfPath`),
following the `TestTableScriptOutputName` pattern: `BaseGnrTest` + `GnrApp` +
a real `GnrDummySite` over the `gnrtest` instance, a real `adm.user` record and
the real `homepage` template resource, `page.loadTemplate` stubbed to
`page.templateFromResource` (the #1165 sqlite workaround, `adm.userobject` is
empty on the test instance). The PDF is rendered by the real htmltopdf service
and read back with pymupdf to check the record's own content is in it; the two
PDF cases carry the `weasyprint`/`pymupdf` `importorskip` guard of
`gnrpy/tests/core/htmltopdf_test.py`.

## Verification

Run in this session, from `gnrpy/`, with `PYTHONPATH` pointing at this
worktree (asserted by `test_module_under_test`):

- Narrow gate: `python3 -m pytest tests/web/gnrbaseclasses_test.py tests/web/gnrtablescript_test.py tests/web/gnrtablescript_new_test.py -q` — **12 passed**. weasyprint and pymupdf are both installed on this machine, so **the two PDF cases actually ran, they were not skipped**.
- Regression proof: with the fix reverted and the new tests in place, `test_a_string_pdf_is_the_output_path` fails (**1 failed, 11 passed**) — it returns the HTML instead of the path.
- `python3 -m pytest core web app xtnd -q` from `gnrpy/tests` — **1233 passed**.
- `python3 -m pytest tests/ -q --ignore=tests/sql` — 1288 passed, 1 failed: `tests/web/gnrasync_test.py::test_wsproxy_push` (TimeoutError), which passes on its own (`7 passed`) and is unrelated to this change.
- `flake8 gnrpy/gnr/web/gnrbaseclasses.py gnrpy/tests/web/gnrbaseclasses_test.py` — zero errors.
- `tests/sql` was not run: `testing.postgresql` cannot boot on this machine right now (`shmget: No space left on device`). Nothing in this change touches SQL.

## Note for #1045

PR #1045 (`feat/123-generic-template-print-resource`) rewrites this same
`__call__`, so it will need a small adjacent-line rebase on top of this.

## Related issue

Fixes #1264

### DIFF
--- a/gnrpy/gnr/web/gnrbaseclasses.py
+++ b/gnrpy/gnr/web/gnrbaseclasses.py
@@ -320,11 +320,12 @@ class TableTemplateToHtml(BagToHtmlWeb):
         if not htmlContent:
             htmlContent = self.contentFromTemplate(record,template=template,locale=locale)
             record = self.record
+        callingPdfPath = pdf if isinstance(pdf,str) else None
         if pdf :
             filepath = filepath or self.getHtmlPath(f'{self.getDocName()}.html')
         result = super(TableTemplateToHtml, self).__call__(record=record,htmlContent=htmlContent,filepath=filepath,**kwargs)
-        if pdf is True:
-            return self.writePdf()
+        if pdf:
+            return self.writePdf(pdfpath=callingPdfPath)
         return result
     
     def getDocName(self):
--- a/gnrpy/tests/web/gnrbaseclasses_test.py
+++ b/gnrpy/tests/web/gnrbaseclasses_test.py
@@ -3,16 +3,24 @@
 ``sys.batch_log`` is the fixture table: its ``caption_field`` is the formula
 column ``log_caption``, the shape that made every record of a table print to
 the same file name.
+
+TestTableTemplatePdfPath covers issue #1264 on the sibling base class
+TableTemplateToHtml: a string ``pdf`` argument is the requested output path,
+the meaning TableScriptToHtml already gives it. The PDF cases render through
+the real htmltopdf service and are skipped when weasyprint or pymupdf are not
+importable (same guard as tests/core/htmltopdf_test.py).
 """
 
 import os
 import re
 
+import pytest
+
 from core.common import BaseGnrTest
 
 from gnr.app.gnrapp import GnrApp
 from gnr.web import gnrbaseclasses
-from gnr.web.gnrbaseclasses import TableScriptToHtml
+from gnr.web.gnrbaseclasses import TableScriptToHtml, TableTemplateToHtml
 from gnr.web.gnrdummysite import GnrDummySite
 
 
@@ -108,3 +116,59 @@ class TestTableScriptOutputName(BaseGnrTest):
         assert len(set(names)) == len(self.pkeys)
         for pkey, name in zip(self.pkeys, names):
             assert name.endswith('%s.html' % re.sub(r'\W', '_', pkey))
+
+
+class TestTableTemplatePdfPath(BaseGnrTest):
+    """``adm.user`` and its ``homepage`` template resource are the fixture: the
+    template prints the username, so the rendered document can be recognized."""
+
+    @classmethod
+    def setup_class(cls):
+        super().setup_class()
+        app = GnrApp(cls.test_instance_name)
+        app.db.model.check(applyChanges=True)
+        app.db.commit()
+        cls.site = GnrDummySite(cls.test_instance_name, site_name=cls.test_instance_name)
+        cls.tblobj = cls.site.db.table('adm.user')
+        record = cls.tblobj.newrecord(username='tpluser')
+        cls.tblobj.insert(record)
+        cls.site.db.commit()
+        cls.pkey = record[cls.tblobj.pkey]
+
+    def buildScript(self):
+        page = self.site.dummyPage
+
+        def loadTemplate(template_address, **kwargs):
+            # adm.userobject is empty on the sqlite test instance and querying it
+            # raises (#1165): go straight to the resource the address points at
+            table, tplname = template_address.split(':', 1)
+            data, _dataInfo = page.templateFromResource(table=table, tplname=tplname)
+            return data['compiled']
+        page.loadTemplate = loadTemplate
+        return TableTemplateToHtml(page=page, table=self.tblobj, record_template='homepage')
+
+    def pdfText(self, pdfpath):
+        fitz = pytest.importorskip('fitz')
+        with fitz.open(pdfpath) as doc:
+            return '\n'.join(page.get_text() for page in doc)
+
+    def test_a_string_pdf_is_the_output_path(self, tmp_path):
+        pytest.importorskip('weasyprint')
+        target = str(tmp_path / 'badge.pdf')
+        script = self.buildScript()
+        assert script(record=self.pkey, pdf=target) == target
+        assert os.path.getsize(target) > 0
+        assert 'tpluser' in self.pdfText(target)
+
+    def test_pdf_true_still_writes_beside_the_html(self):
+        pytest.importorskip('weasyprint')
+        script = self.buildScript()
+        result = script(record=self.pkey, pdf=True)
+        assert result == script.filepath.replace('.html', '.pdf')
+        assert os.path.getsize(result) > 0
+
+    def test_no_pdf_returns_the_html(self):
+        script = self.buildScript()
+        result = script(record=self.pkey)
+        assert result.startswith('<!DOCTYPE html')
+        assert 'tpluser' in result


fporcari (commented):
## Independent verification

Tested SHA: `45e0bf726efa019fdd507a3a99501ae34a359b39`, in a dedicated detached worktree off `refs/pull/1279/head`. Every run used `PYTHONPATH=<worktree>/gnrpy`, and the imported `gnr` module resolved to `<worktree>/gnrpy/gnr/__init__.py`, so the results come from this branch and not from an installed copy.

**Step 1 — `pytest gnrpy/tests/web/gnrbaseclasses_test.py -k TestTableTemplatePdfPath -v`**: PASSED. 3 passed, 0 failed, 0 skipped, 9 deselected.

**Step 2 — the two PDF cases must be PASSED and not SKIPPED**: PASSED. `test_a_string_pdf_is_the_output_path` PASSED and `test_pdf_true_still_writes_beside_the_html` PASSED; neither was skipped. Both PDF rendering dependencies were present in the environment (weasyprint 64.1, pymupdf importable), and the run shows the PDF stack actually loading during `test_a_string_pdf_is_the_output_path`, so the #1264 regression is genuinely exercised rather than skipped over.

**Step 3 — `pytest gnrpy/tests/web/gnrbaseclasses_test.py -v` (whole file)**: PASSED. 12 passed, 0 failed, 0 skipped. The 9 sister-class `TestTableScriptOutputName` tests are unaffected.

**Step 4 — flake8 on the touched files**: PASSED, zero findings. Files touched by the branch: `gnrpy/gnr/web/gnrbaseclasses.py` and `gnrpy/tests/web/gnrbaseclasses_test.py`.

The change in `TableTemplateToHtml.__call__` is minimal and matches what the tests assert: the string form of `pdf` is captured before the html path is derived and handed to `writePdf(pdfpath=...)`, while the `pdf is True` behaviour of writing beside the html is preserved.

Verification result: PASS

genro (approved):
Correct, and the argument that makes it obviously correct is the one you give: `TableScriptToHtml.__call__` at `gnrbaseclasses.py:362` already gives a string `pdf` the meaning of an output path, so the sibling class was the odd one out. Copying an idiom that exists three dozen lines away is the right kind of fix.

I checked the two claims the change rests on. `BagToHtmlWeb.writePdf` at `:310-316` does `pdfpath = pdfpath or self.getPdfPath(...)`, so nothing below needed touching. And with `callingPdfPath=None` the `pdf=True` path falls through to exactly the same default as before, so that case is byte-identical.

One gap in the behaviour table in the description. It covers `str`, `True` and `None`, but not a truthy value that is neither — a StorageNode, a Bag. Before, those chose the html branch and returned the html without writing anything; now they write a PDF to the default path. No caller in the repo does it, so this is a sentence in the description rather than a code change.

Also, unlike its sibling at `:406`, `BagToHtmlWeb.writePdf` never assigns `self.pdfpath`, so after a call with an explicit path the attribute is stale. Nothing reads it on this class today.

fporcari:
Both recorded, one of them with a correction.

> It covers `str`, `True` and `None`, but not a truthy value that is neither — a StorageNode, a Bag. Before, those chose the html branch and returned the html without writing anything; now they write a PDF to the default path.

Correct, and worth having on the record here since the description keeps the shape it was reviewed with. A truthy `pdf` that is neither a string nor `True` now writes the PDF to the default path instead of returning the html. `callingPdfPath` stays `None` for those, so the path is the same one `pdf=True` has always produced — the change is that something gets written at all. No caller in the repo passes one.

> unlike its sibling at `:406`, `BagToHtmlWeb.writePdf` never assigns `self.pdfpath`, so after a call with an explicit path the attribute is stale. Nothing reads it on this class today.

This one I am leaving alone, because the last sentence does not hold: `getPdfPath` at `:291` is `return self.pdfpath or self.filepath.replace('.html','.pdf')`, so the attribute is read as the default for the *next* call. Assigning it in `writePdf` would make it sticky — a later `writePdf()` with no argument would return the previous call's path instead of deriving one from the current `filepath`, which is the shape of #1263.

The asymmetry with `:406` is necessity rather than an oversight: `TableScriptToHtml` reads `self.pdfpath` back in its own `__call__` at `:384` and `:388`, so it has to store it. `TableTemplateToHtml` gets the path as `writePdf`'s return value and has nothing to read back.

Merging as is.

Issue #1275, opened 2026-09-09
Experiment: fetch-based asynchronous RPC and TYTX/NBag codecs behind instanceconfig switches


genro:
## Proposal

Experiment with replacing the legacy Dojo HTTP RPC transport with native `fetch`, while preserving the application-facing `genro.rpc` API. Coordinate a second, independently selectable experiment with TYTX/NBag serialization. Enable the experiments explicitly through `instanceconfig.xml`; keep current behavior as the default.

This is a separate workstream from WebSocket support. It should benefit existing HTTP applications without requiring them to migrate to WebSockets or change their page/component model.

## Measured motivation

A browser benchmark against a local GenroPy site hosted by the ASGI bridge compared the same lightweight RPC, using an ephemeral page and the same legacy Bag/XML result handler. After 12 warm-up triplets, three rounds of 50 rotating triplets measured 150 calls per path, one request at a time, with the WebSocket channel already open.

| Path | Median callback latency | p95 | Median server time (X-GnrTime) |
| --- | ---: | ---: | ---: |
| HTTP through legacy Dojo | 51.80 ms | 52.50 ms | 3.06 ms |
| HTTP through native fetch | 5.60 ms | 8.30 ms | 2.91 ms |
| WebSocket through WSX | 4.50 ms | 6.20 ms | 2.76 ms |

The legacy Dojo implementation checks async completion with `setInterval(_watchInFlight, 50)` in `dojo_libs/dojo_11/dojo_src/dojo/_base/xhr.js` (compact version: `dojo/dojo/_base/xhr.js`). The source inspection and fetch control indicate that most of the observed HTTP/WSK gap comes from client completion polling, not an inherent HTTP cost.

These are local development measurements on a lightweight method, not a claim that all GenroPy operations become nine times faster. SQL, page construction, application logic, payload size, concurrency and network latency need separate measurement. The benchmark fixture and detailed results currently live in the dedicated genropy-asgi working branch and should accompany the experiment as reproducible artifacts.

## Staged experiment

1. Inventory the existing RPC transport contract in `genro_rpc.js`: callbacks, `dojo.Deferred`, result parsing, datachanges, resource loading, HTTP/application errors, timeout, cancellation, credentials, headers and call bookkeeping.
2. Add a fetch-backed path for ordinary **asynchronous** HTTP RPC, retaining legacy parameter serialization and XML/Bag responses initially. Adapt Promise completion to the existing caller contract. Do not change widgets or require application rewrites.
[...]

Issue #1274, opened 2026-09-09
Feature: native TYTX JSON/MessagePack serialization for legacy Bag


genro:
## Summary

Add native TYTX serialization/deserialization to the legacy `gnr.core.gnrbag.Bag`, using a flattened row representation that can be encoded directly as JSON or MessagePack and reconstructed as a **legacy Bag**.

Related to #329 (TYTX integration with GnrClassCatalog). This proposal concerns the Bag representation, codec contract and measured migration path; it does not require replacing the catalog globally.

## Problem or Need

The current ASGI bridge registers the legacy Bag with TYTX under `BAG`, serializing it through `toXml(catalog=GnrClassCatalog.convert())` and reconstructing it with `Bag(text)`. MessagePack around this registered value still carries XML text, so it cannot remove XML generation/parsing costs.

Modern `genro_bag.Bag` already provides `bag.to_tytx(transport="json" | "msgpack")` and `Bag.from_tytx(...)`: it flattens the hierarchy into rows `(parent, label, tag, value, attributes)`, then passes the structure to the selected codec. This offers a reference for a focused backport/adapter.

Important distinction: calling the generic `genro_tytx.to_tytx(bag, "msgpack")` currently invokes the registered serializer without forwarding the transport. Merely adding methods or changing a format string at generic call sites does not guarantee native binary Bag encoding.

## Proposed Solution

1. Prototype an adapter against the real legacy Bag, then expose suitable `to_tytx` / `from_tytx` methods in core if compatibility and measurements support the backport. Use current public genro-tytx APIs and agree the dependency policy.
2. Serialize the flattened hierarchy directly through JSON or MessagePack; do not embed an XML or JSON string inside MessagePack for the native path.
3. Reconstruct the original legacy Bag class. Keep legacy and modern Bag type identities distinct, including when nested in dictionaries, lists, values and attributes.
4. Define an explicit, versioned format/codec contract and matching producer/consumer selection. The existing `BAG` registration means XML today: do not silently redefine that payload in a way that breaks existing readers. Coordinate transport-aware registration with genro-tytx as needed.
5. Preserve existing XML APIs and browser-facing protocols. Start with an opt-in internal communication path; a global default switch is outside the initial change.
6. Implement on a working branch based on the latest `origin/develop`.

### Compatibility requirements

[...]

Issue #329, opened 2025-12-03
Feature: Integrate genro-tytx library with GnrClassCatalog


fporcari:
## Summary

Integrate the `genro-tytx` library (TYTX - Typed Text protocol) into the Genropy framework, enabling the new serialization/deserialization system as an alternative to `GnrClassCatalog`.

## Background

The `genro-tytx` library is a new protocol developed by our team for exchanging typed data over text-based formats (JSON, XML, MessagePack). It provides functionality similar to `GnrClassCatalog` but with a more modern and extensible architecture.

- **Package name**: `genro-tytx` (PyPI)
- **Import name**: `genro_tytx` (Python)
- **Available functions**: `as_text`, `from_text`, `as_typed_text`, `from_typed_text`, `as_typed_json`, etc.

### 1. Create `gnrtytx` module with `GnrClassCatalogTyTX`

Create a new module `gnrpy/gnr/core/gnrtytx.py` that implements `GnrClassCatalogTyTX`:

- New class that wraps `genro_tytx` functionality
- API compatible with existing `GnrClassCatalog`
- Methods: `asText()`, `fromText()`, `asTypedText()`, `fromTypedText()`, `getType()`, etc.
- Leverage `genro_tytx.TypeRegistry` for custom types

### 2. Add instanceconfig option under experimental-features

Add configuration support in `instanceconfig.xml`:

```xml
<experimental-features tytx-integration="true"/>
```

### 3. Automatic switch mechanism

- When `tytx-integration` is enabled, the system uses `GnrClassCatalogTyTX` instead of `GnrClassCatalog`
- Automatic fallback to `GnrClassCatalog` if `genro_tytx` is not installed
- Factory method in `gnrclasses.py` for dynamic switching

## Files Involved

| File | Change |
|------|--------|
| `gnrpy/gnr/core/gnrtytx.py` | **NEW** - `GnrClassCatalogTyTX` implementation |
| `gnrpy/gnr/core/gnrclasses.py` | Factory method for dynamic switch |
| `gnrpy/gnr/app/gnrapp.py` | Read experimental-features configuration |

## Acceptance Criteria

- [ ] `GnrClassCatalogTyTX` class implements all methods from `GnrClassCatalog`
- [ ] Configuration option `tytx-integration` works in instanceconfig.xml
- [ ] Graceful fallback when `genro_tytx` is not installed
- [ ] Existing functionality remains unchanged when feature is disabled
- [ ] Unit tests for the new module

## Technical Notes

[...]

PR #1280, opened 2026-09-09
Add an opt-in fetch transport for asynchronous Dojo requests

fporcari:
## Summary

Supported asynchronous Dojo HTTP requests currently wait for the legacy 50 ms completion polling loop. Add an instance-controlled fetch transport that completes the existing `dojo.Deferred` directly, preserving callback chaining, parameter serialization, XML/Bag responses and RPC bookkeeping.

Enable explicitly in `instanceconfig.xml`:

```xml
<experimental-features dojo-xhr-patch="fetch"/>
```

Absent or other values preserve the original transport. The switch wraps the shared Dojo HTTP entry points for the enabled instance, including supported callers outside `genro.rpc`. Synchronous, cross-origin and unsupported requests retain XHR. Fetch failures are never retried through XHR, avoiding duplicate writes. Remove the switch, restart the instance and reload pages to roll back.

`ioArgs.xhr` is a compatibility adapter, not a native XMLHttpRequest; XHR event/progress listeners are outside its supported surface. Preserve native XML behavior for missing Content-Type and malformed documents across browser parser-error namespaces.

Related to #1275: this implements the initial HTTP transport experiment. TYTX/NBag integration (#1274, #329), broader network/concurrency measurements and any default change remain separate work. The parent issue remains open for those acceptance criteria.

## Validation

- Rebased onto current `develop` before final validation.
- Full Python suite: **2,509 passed, 10 skipped**, Python 3.13; temporary PostgreSQL test databases.
- Configured flake8 checks pass on all changed Python files; new Python files also pass default flake8. JavaScript syntax and `git diff --check` pass.
- Real HTTP contract: **20 checks per bundled Dojo build** (compact/source/release) in the integrated browser; compact also passed in Chrome. The new XML test reproduced both regressions before their fixes.
- Verified methods, encoding, raw bodies, headers/cookies, real Dojo Deferred type, callbacks/chaining, HTTP/parse errors, recovery, timeout through body consumption, cancellation, fallback and bypass of `_ioWatch`.
- Dedicated **sandboxpg_fetchmode** instance/database, fetch enabled and Python/static resources resolved from the feature worktree. Verified RPC, Bag results, callback chaining, datachanges, resource loading, HTTP/application error recovery and a TableHandler form save. Confirmed the changed province name directly in PostgreSQL, then restored it through the form and verified restoration.
- Compact contract and application RPC/error recovery repeated successfully after rebase.

Reproduce the transport contract with `python gnrjs/tests/dojo_xhr_server.py`, then open `/gnrjs/tests/dojo_xhr.html` on port 8765. The application testhandler is `/test/datastore/dojo_xhr`; use a disposable instance/database. Configuration and compatibility details are in `docs/development/dojo-xhr-patch.md`.

## Local benchmarks

Measured on 2026-09-09 using the real HTTP fixture, 12 warm-up requests followed by 50 sequential requests per transport. Original Dojo runs first, fetch second. Server time is read from `X-GnrTime`; all values are milliseconds.

| Browser / build | Transport | Client p50 | Client p95 | Server p50 |
| --- | --- | ---: | ---: | ---: |
| Chrome / compact | Original Dojo | 52.00 | 52.50 | 0.19 |
| Chrome / compact | Fetch | 3.90 | 5.70 | 0.13 |
| Integrated browser / compact, after rebase | Original Dojo | 51.70 | 52.60 | 0.13 |
| Integrated browser / compact, after rebase | Fetch | 0.50 | 0.70 | 0.03 |

These measurements demonstrate removal of the local completion-polling delay, not equivalent application-wide speedups. The fixture is a lightweight local HTTP server, not the TableHandler workload. SQL, application logic, serialization cost, payload size, latency, concurrency and alternating transport rounds need separate measurement. No throughput claim is made.

### DIFF
--- a/docs/development/dojo-xhr-patch.md
+++ b/docs/development/dojo-xhr-patch.md
@@ -75,7 +75,8 @@ Dojo 1.1 builds. Reloading starts again with original Dojo.
 
 Coverage includes methods and raw bodies, parameter encoding, request and
 response headers, XML attributes, callback order/context, chained Deferreds,
-HTTP and parse errors, recovery, cancellation, timeout, synchronous calls,
+HTTP and parse errors, XML without Content-Type, malformed XML, recovery,
+cancellation, timeout, synchronous calls,
 form serialization, cookies, fallback routing and fetch network failure.
 An explicit assertion verifies that fetch does not enter `dojo._ioWatch`.
 
@@ -114,3 +115,29 @@ The two CSV failures (`test_getCsvDialect_limited_lines` and
 `test_CsvReader_auto_dialect`) also reproduce in isolation without importing
 the modified webpage module. This run does not establish a green full suite.
 Real form/TableHandler validation and the broader benchmark remain pending.
+
+## XML compatibility and instance validation, 2026-09-09
+
+Responses without Content-Type now use XHR's XML default. Explicit text/plain
+responses still have no responseXML. Malformed XML is recognized with both
+the Mozilla and XHTML parser-error namespaces used by browser engines.
+The added contract check first reproduced both regressions, then passed
+after the corrections. The updated suite passed 20 checks on each bundled
+Dojo build in the integrated browser, and on the compact build in Chrome.
+
+The `/test/datastore/dojo_xhr` testhandler page exercises real application
+RPC, the returned Dojo Deferred, callback chaining, Bag results, datachanges,
+error recovery and a province TableHandler. Run it against a disposable
+instance with the transport option enabled. Set the site's `wsgi.debug` to
+`false` to exercise application error callbacks; debug mode can re-raise
+the deliberate exception as HTTP 500.
+
+Validation on the isolated `sandboxpg_fetchmode` PostgreSQL instance passed
+the RPC and datachange checks, HTTP and application error recovery, resource
+loading, and form load/save. A province name was changed through the form,
+verified directly in PostgreSQL, and restored through the form. The runtime
+used Python and JavaScript from the feature worktree. Four Python bootstrap
+and webpage tests passed. After rebasing onto current develop, the full Python
+suite passed with 2,509 passed and 10 skipped (Python 3.13). The compact browser
+contract and application RPC/error-recovery checks also passed after the rebase.
+Broader latency/concurrency measurements remain pending.
--- a/docs/development/dojo-xhr-patch.md
+++ b/docs/development/dojo-xhr-patch.md
@@ -1,0 +1,116 @@
+# Experimental Dojo HTTP transport
+
+The `dojo-xhr-patch` experiment replaces supported asynchronous Dojo HTTP
+requests with native fetch. It applies to `dojo.xhr`, `xhrGet`, `xhrPost`,
+`xhrPut`, `xhrDelete`, `rawXhrPost` and `rawXhrPut`, including callers outside
+`genro.rpc`. Fetch completes the existing Dojo Deferred directly, bypassing
+the legacy 50 ms completion polling interval.
+
+## Configuration and rollback
+
+Add this child to the instance's `instanceconfig.xml`:
+
+```xml
+<experimental-features dojo-xhr-patch="fetch"/>
+```
+
+Restart the instance to reload its configuration, then reload browser pages.
+The server supplies `dojoXhrPatch` at bootstrap; a page request parameter
+cannot override the instance setting. The patch is installed when the client
+is constructed, before application initialization. Requests made before that
+point continue to use the original transport.
+
+To roll back, remove the attribute (or the node if otherwise empty), restart
+the instance, and reload browser pages. Existing pages keep their installed
+transport until reloaded. Only the exact value `fetch` enables the patch;
+absent or other values leave Dojo unchanged.
+
+The patch lives in `genro_patch.js`, so no vendor bundle rebuild is required.
+Deploy the matching Python and JavaScript changes together and refresh any
+deployment-managed static caches.
+
+## Compatibility boundary
+
+Supported requests use same-origin HTTP(S) with existing parameter/form
+serialization, headers, cookies and the built-in Dojo text, JSON, XML and
+JavaScript response handlers. XML/Bag serialization and WebSocket transport
+are unchanged. The experiment targets the application's UTF-8 responses;
+fetch's text decoding does not reproduce XHR decoding for other charsets.
+
+Callbacks retain their context, ordering and transformations, including
+Deferred chaining and errback recovery. `ioArgs.xhr` provides response text,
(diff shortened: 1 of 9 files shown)


genro (approved):
Reviewed the transport adapter, bootstrap switch, XML compatibility fixes and regression coverage. No blocking findings in this focused review. The implementation preserves the Dojo Deferred contract, retains the original synchronous/unsupported paths and avoids retrying failed writes through XHR.

CI is green for Python 3.11–3.14 and CodeQL on the reviewed commit. The documented browser and real-instance validation support this opt-in experiment; I did not rerun those manual checks during this review. The explicit XHR adapter limitations and separate follow-up for broader performance measurements are appropriate for this scope.

cgabriel (approved):
I'm approving because it's an opt-in experimental feature, tests are ok but lighthouse analysis reports different results in the async js side to those offered in the PR description (performance score drops from 51 to 19 by activating the feature). So I'll leave more detailed performance analysis using different prod grade deployment. I'm suggesting to perform lighthouse tests as well so the whole stack (not just the xhr request timing based on the X-Gnr-Time header)

I'm also enjoying gnrjs/tests new folder!

genro:
@cgabriel Following up on the performance concern: I benchmarked RPC latency after the page had fully loaded, separately from Lighthouse navigation scores.

On PR head `44f0b6e9864aebbb627dfe066ccfde83d5df7e4e`, I used two isolated local Genropy instances with identical settings except the fetch switch. Each ran 20 warmup calls, then 300 measured calls in six alternating blocks of 50, using the same `genro.serverCall('transport_echo', {}, callback)` invocation. Timing ran from invocation to application callback entry, including Genropy response parsing. All responses were validated, and there were no browser errors.

| Metric | Legacy Dojo XHR | Fetch patch |
|---|---:|---:|
| Median RPC latency | 52.0 ms | 17.8 ms |
| p95 RPC latency | 52.7 ms | 26.6 ms |
| Median server duration (`X-GnrTime`) | 16.4 ms | 15.4 ms |

Median latency fell by approximately **66% (34.2 ms per call)**, while server duration was comparable. This is consistent with legacy Dojo's 50 ms polling interval for request completion; the fetch path avoids that polling delay.

This test used Chrome headless without CPU/network throttling and a small local RPC with real Bag parsing and a datachange. It supports a benefit for short, sequential RPCs on an already running page; it does not resolve the reported Lighthouse startup regression or establish the same relative gain for slower production RPCs, larger payloads, or concurrent requests.

fporcari:
@cgabriel I tried to reproduce the Lighthouse regression you reported (performance score 51 → 19 with the feature on) and could not, on either page I could reach without authentication. Here is what I measured and, more importantly, one structural fact that came out of it.

**Setup.** PR head `44f0b6e9864aebbb627dfe066ccfde83d5df7e4e`, instance `sandboxpg` served from a detached worktree of that SHA on port 8107, scratch gnrfolder of its own. Two things were proven before measuring rather than assumed: the served code is the worktree's (`/_gnr/11/js/genro_patch.js` carries `genropatches.dojoXhr`, absent from the main checkout's build) and the switch really takes effect per configuration — with the flag absent the live page has `genropatches.dojoXhr.installed === false` and a native `dojo.xhr`, with `<experimental-features dojo-xhr-patch="fetch"/>` it has `installed === true` and `dojo.xhr` is the adapter. Lighthouse CLI 12.8.2, default mobile preset with simulated throttling, 5 runs per side.

**Index page `/`, 5 runs per side**

| metric | OFF median (min–max) | ON median (min–max) | delta |
|---|---|---|---|
| performance score | **51** (46–52) | **52** (50–52) | +1 |
| FCP | 12 591 ms | 12 551 ms | −40 ms |
| LCP | 15 243 ms | 15 218 ms | −25 ms |
| TBT | 240 ms (211–375) | 210 ms (195–256) | −30 ms |
| Speed Index | 12 591 ms | 12 551 ms | −40 ms |
| CLS | 0.000 | 0.000 | 0 |

Per-run scores: OFF 46, 51, 51, 51, 52 · ON 51, 52, 52, 52, 50. The run-to-run spread inside a single side (6 score points OFF, ~165 ms of TBT) is larger than the between-side delta on every metric, so the honest reading is "no measurable difference", not "slightly better".

`/test/datastore/dojo_xhr`, 3 runs per side: score 48 vs 48, FCP +11 ms, LCP −133 ms, TBT 336 → 322 ms, CLS 0 on both. Long tasks during boot on the index page: 3 tasks both sides, 630 ms → 571 ms in total, attributed on both sides to the `_jslib` bundle parse (370 ms) and `dojo.js` (215 ms, 161 ms) — not to reply handling.

**The part that matters more than the tables: boot traffic does not go through the adapter.** Counted from the Lighthouse `network-requests` audit and independently from `PerformanceResourceTiming.initiatorType` in a live page, both stable across runs:

| target | side | total | XHR | fetch |
|---|---|---|---|---|
| `/` | OFF | 155 | 72 | 0 |
| `/` | ON | 155 | 71 | **1** |
| `/test/datastore/dojo_xhr` | OFF | 226 | 94 | 0 |
| `/test/datastore/dojo_xhr` | ON | 226 | 91 | **3** |

The mechanism is in the request list: of the 69 boot requests on the index page, 66 are Dojo's own module loader pulling individual `.js` files, and it loads them synchronously. The adapter's `supported(args)` declines `args.sync`, so every one of those stays on `original.xhr` by construction. The remaining few are genro RPCs, of which exactly one crossed the adapter, at or after onload. The adapter is installed and demonstrably working — a post-load `serverCall` does produce `fetch` entries — but 1.4 % of index boot traffic and 3.2 % of test-page boot traffic actually touch it. That share is a consequence of the `supported()` guard, not of my machine or page, so a startup regression caused by the transport swap itself would have nothing to act through.

**What this does not establish**, and where I need you. It does not establish that you did not see 51 → 19. I could not measure a post-login application page at all: Lighthouse starts from a fresh profile and lands on the login wall, and I did not attempt credentials against the instance. Your page, browser, Lighthouse version, throttling settings, jslib cache state and machine are all unknown to me, and a per-instance difference — a heavier menu with many boot RPCs, a rebuilt bundle, a plugin — stays untested. So: **do you have a test case where it reproduces that you can attach?** A page or an instance recipe, plus the two Lighthouse JSON reports if you still have them. I will run exactly that, and add to it the XHR-vs-fetch request split from the same trace, which is the number that tells us whether the transport is even in the path on your page.

On genro's benchmark: it measures post-load RPC latency, a different phase, and it says so itself. The two results are not in conflict — one is about a page already running, the other about the load window — and neither of them alone answers your point. Your Lighthouse suggestion on the whole stack stands as the right instrument; this is a first pass with it on the only two pages I could reach.

PR #1282, opened 2026-09-09
fix(web): tagged public methods accept the page session again

fporcari:
## Problem

`AuthorizationBaseTagsVerifier` asks for the `Authorization` header before
looking at the page session, so **every** `@public_method(tags=...)` called
from a logged in browser fails with `AUTH-901`. It surfaced on an ECR emulator
call from a sale page, but it hits any tagged handler reached through
`genro.serverCall` / `dataRpc`.

The check regressed in 4ecc2f8 (#700), which moved the inline logic into the
verifier module. Before that move `getPublicMethod` read:

```python
userTags = self.userTags or self.basicAuthenticationTags()
if not self.application.checkResourcePermission(handler.tags, userTags):
    raise self.exception(GnrUserNotAllowed, method=method)
```

Session tags first, Basic Authentication only for a caller with no session.
The verifier dropped that order, so the fallback became mandatory. It went
unnoticed because the tagged handlers in use are external endpoints that do
send Basic credentials.

## Change

- `verify()` checks `page.userTags` first and only falls back to
  `basicAuthenticate()` when there is no session, restoring the pre-refactor
  order. Tags that do not match give `user_not_allowed` (AUTH-001) again
  instead of an authentication error.
- The Basic Authorization decoding moves to `basicAuthenticate()`, unchanged in
  behaviour and error codes, so the same handler still serves both a browser
  page and an external client.
- `method` reaches the verifier, so the AUTH-001 caption names the method again.
- `GnrBasicAuthenticationError` gets a caption: the base one renders only code
  and description, so the `msg` kwarg never reached the user and the dialog
  showed just `AUTH-901` plus a `file:line`.

## Verification

`gnrpy/tests/web/verifiers_test.py` covers the four paths: session tags pass
with no header, non matching session tags give `user_not_allowed`, a caller
with no session still needs the header, and Basic credentials are still checked
against the required tags. Full suite green (2454 passed, 10 skipped).

### DIFF
--- a/gnrpy/gnr/web/gnrwebpage.py
+++ b/gnrpy/gnr/web/gnrwebpage.py
@@ -110,6 +110,7 @@ class GnrUserNotAllowed(GnrException):
 
 class GnrBasicAuthenticationError(GnrException):
     code = 'AUTH-901'
+    caption = "!!Error code %(code)s : %(msg)s."
 
 EXCEPTIONS = {
     'user_not_allowed': GnrUserNotAllowed,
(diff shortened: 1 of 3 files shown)


genro (approved):

cgabriel (approved):
LGTM.

PR #1284, opened 2026-09-10
Let a templateChunk opt into the Parameters tab (#1283)

### LINKED ISSUE #1283: [BUG] Template parameters are unreachable in the templateChunk palette editor
## Problem Description

Template **parameters** — the framework's registry for variables that are *not* columns of the template's table — can never round-trip in the `templateChunk` palette editor. Two independent defects combine, and each one alone would be enough to make the feature unreachable there.

**1. The Parameters tab is never instantiated in the chunk editor.**

`_te_frameChunkInfo` has the call commented out:

```python
# resources/common/gnrcomponents/tpleditor.py:719-726
def _te_frameChunkInfo(self,frame,table=None,datasourcepath=None):
    frame.top.slotToolbar('5,parentStackButtons,*',parentStackButtons_font_size='8pt')
    bc = frame.center.borderContainer()
    self._te_info_vars(bc,table=table,region='center', ...)
    #self._te_info_parameters(bc,region='center')
```

The full editor is fine — `_te_frameInfo` calls it normally at `:364`. Only the chunk variant is affected.

**2. The chunk editor's Save button reads a datapath nothing writes.**

```python
# tpleditor.py:755  (_te_saveButton)
vb='=.data.varsbag', pb='=.data.parametersbag', data='=.data')
```

`.data.parametersbag` is a dangling path. Every other site in the file uses `.data.parameters`:

- `_te_info_parameters` stores the grid at `#ANCHOR.data.parameters` (`:353`)
- `_te_mainstack`'s compile dataRpc reads `parametersbag='=.data.parameters'` (`:232`)
- the save/copy helpers read `data['parameters']` (`:635`, `:636`, `:642`, `:659`, `:660`)

So even with the tab uncommented, `_te_saveButton` would send an empty bag to `te_compileTemplate`, and the parameters would be dropped on every save.

`_te_saveButton` is used only by the chunk palette bars (`:694`, `:708`, `:715`); `te_templateEditor` saves through `_te_mainstack` and is unaffected.

## Reproduction Steps

1. Put a `templateChunk(..., table='some.table', editable=True)` on any form.
2. Double-click it to open the template palette, go to the Metadata tab.
3. Observe: there is no **Parameters** grid — only **Variables**. (The full template editor, e.g. the userobject one, does show it.)
4. To see the second defect independently, uncomment `tpleditor.py:726` and reload: the Parameters grid appears and accepts rows, but after Save the compiled template contains none of them — `te_compileTemplate` receives an empty `parametersbag`, so `formats`/`masks` for those codes are never applied.

## Environment

- Branch: `develop` (verified at commit currently on `origin/develop`)
- File: `resources/common/gnrcomponents/tpleditor.py`
- Affects every `templateChunk` with `editable=True`; does not affect `te_templateEditor`

## Suggested fix

Two lines:

- uncomment `self._te_info_parameters(bc,region='center')` at `:726` (region may need adjusting, since `_te_info_vars` already claims `center` in this variant — in `_te_frameInfo` vars take `bottom` and parameters take `center`);
- change `pb='=.data.parametersbag'` to `pb='=.data.parameters'` at `:755`.

## Why it matters

`parameters` is the only mechanism the template editor offers for **non-column** variables: `te_compileTemplate` consumes them for `formats`/`masks` (`:175-179`) without touching `columns`/`virtual_columns`, and `_te_pickers` publishes them as draggable `$code` tokens. `varsbag` is not a substitute — every row there lands in `main?columns` (`:166-167`) and the "Copy From" menu replaces the whole bag (`:679-686`).

Concretely: a table-rooted chunk template currently has no supported way to reference a value that does not come from the record. Consumers work around it by typing raw `$tokens` that survive compilation through `safe_substitute`, which works but is undiscoverable and unformattable.

Found while reviewing the `dbwatcher` package in `gnrxadm`, which needs exactly this to expose a result list, an occurrence count and a notification URL inside a template rooted on a business table.

dgpaci:
## Problem

Template **parameters** — the editor's registry for variables that are *not* columns of the template's table — could never be reached from the `templateChunk` palette editor (#1283). Two independent defects:

1. `_te_frameChunkInfo` had `self._te_info_parameters(bc,region='center')` commented out, so the grid was never instantiated in the chunk variant (the full `te_templateEditor` was fine).
2. The chunk Save button read `pb='=.data.parametersbag'`, a dangling datapath. Every other site uses `.data.parameters` (the grid store, the compile `dataRpc`, the save helpers), so `te_compileTemplate` received an empty bag and the `formats`/`masks` of those codes were dropped from the compiled chunk — even for a template that already carried parameters in its stored bag. As a side effect, reading that path materialised a junk `parametersbag` node in saved templates (visible in several `tpl/*.xml` under `projects/`).

## Solution

Instead of switching the tab on for every chunk, `templateChunk` now accepts a **`showParameters`** kwarg, defaulted to `False` and plumbed exactly like the existing `showLetterhead`:

- `createContent` pops it, resolving a `^`/`=` datapath to an absolute one;
- `openTemplatePalette` evaluates it and forwards it as `remote_showParameters`;
- `te_chunkEditorPane` passes it to `_te_frameChunkInfo`, which adds the Parameters grid at `region='center'` and moves the Variables grid to `region='bottom', height='60%'` — the same layout the full editor uses.

With the default, the Metadata tab is byte-for-byte what it is today; with `showParameters=True` the grid appears and round-trips through save.

The dangling `pb` path is fixed unconditionally: it is a bug on its own, and without it the opt-in would still compile to an empty parameters bag.

```python
pane.templateChunk(template='mytemplate', table='pkg.mytable',
                   editable=True, showParameters=True)
```

## Test case

New `gnrpy/tests/web/tpleditor_chunk_parameters_test.py` — 5 tests, no mocks: each pane is built on a real `GnrDummySite` through the same entry point the browser uses (`remoteBuilder` → `te_chunkEditorPane`), and assertions run on the real struct.

- `test_component_under_test` — the component exercised is the one in the working tree
- `test_parameters_hidden_by_default` — no `.parametersgrid`, Variables still owns the whole tab
- `test_parameters_shown_on_request` — grid present on `#ANCHOR.data.parameters`, Variables moved to `bottom`/`60%`
- `test_save_button_forwards_the_parameters` — every compile button carries `pb='=.data.parameters'`
- `test_parameter_formats_reach_the_compiled_template` — a `parametersbag` lands in the compiled `formats`/`masks` without polluting `columns`

The two behavioural tests fail on `develop` and pass on this branch. Full `gnrpy/tests` suite green: 2511 passed, 10 skipped.

Manual check: put `templateChunk(..., editable=True, showParameters=True)` on a form, double-click it, add a parameter row in Metadata, drag its `$code` from the Parameters folder of the Edit tab, save, reopen — the row is still there and the format applies in the preview.

Fixes #1283

### DIFF
--- a/gnrjs/gnr_d11/js/genro_components.js
+++ b/gnrjs/gnr_d11/js/genro_components.js
@@ -3801,7 +3801,7 @@ dojo.declare("gnr.widgets.TemplateChunk", gnr.widgets.gnrwdg, {
         genro.setData(paletteRoot+'.status','info');
     },
 
-    openTemplatePalette:function(chunkNode,editorConstrain,showLetterhead){
+    openTemplatePalette:function(chunkNode,editorConstrain,showLetterhead,showParameters){
         let componentNode = chunkNode.getParentNode();
         var paletteCode = componentNode.getAttributeFromDatasource('paletteCode');
         if(!paletteCode){
@@ -3821,6 +3821,7 @@ dojo.declare("gnr.widgets.TemplateChunk", gnr.widgets.gnrwdg, {
             var table = componentNode.getAttributeFromDatasource('table');
             var remote_datasourcepath = chunkNode.attr.datasource? chunkNode.absDatapath(chunkNode.attr.datasource):null;
             showLetterhead = typeof(showLetterhead) == 'string' ? chunkNode.getRelativeData(showLetterhead) : showLetterhead;
+            showParameters = typeof(showParameters) == 'string' ? chunkNode.getRelativeData(showParameters) : showParameters;
             var kw = {'paletteCode':paletteCode,'dockTo':'dommyDock:open',
                     title:'Template Edit '+table?table.split('.')[1]:'',width:'750px',
                     maxable:true,
@@ -3833,6 +3834,7 @@ dojo.declare("gnr.widgets.TemplateChunk", gnr.widgets.gnrwdg, {
                     remote_resource_mode:!table || (templateHandler.dataInfo && templateHandler.dataInfo.respath!=null),
                     remote_datasourcepath:remote_datasourcepath,
                     remote_showLetterhead:showLetterhead,
+                    remote_showParameters:showParameters,
                     remote_editorConstrain: editorConstrain
                     };  
            //kw.remote__onRemote = function(){
@@ -3946,6 +3948,7 @@ dojo.declare("gnr.widgets.TemplateChunk", gnr.widgets.gnrwdg, {
         var tplpars = objectExtract(kw,'template,editable');
         var editorConstrain = objectExtract(kw,'constrain_*',null,true);
         var showLetterhead = objectPop(kw, 'showLetterhead');
+        var showParameters = objectPop(kw, 'showParameters');
         if(paletteCode && (paletteCode[0]=='^' || paletteCode[0]=='=')){
             paletteCode = paletteCode[0]+sourceNode.absDatapath(paletteCode);
         }
@@ -3960,6 +3963,9 @@ dojo.declare("gnr.widgets.TemplateChunk", gnr.widgets.gnrwdg, {
         }
         if(typeof(showLetterhead)=='string'){
             showLetterhead = sourceNode.absDatapath(showLetterhead);
+        }
+        if(typeof(showParameters)=='string'){
+            showParameters = sourceNode.absDatapath(showParameters);
         }
         var record_id = objectPop(kw, 'record_id');
         if(record_id){
@@ -3992,11 +3998,11 @@ dojo.declare("gnr.widgets.TemplateChunk", gnr.widgets.gnrwdg, {
         var handler = this;
         if(tplpars.editable){
             kw.selfsubscribe_openTemplatePalette = function(){
-                handler.openTemplatePalette(this,editorConstrain,showLetterhead);
+                handler.openTemplatePalette(this,editorConstrain,showLetterhead,showParameters);
             }
             kw.connect_ondblclick = function(evt){
                 if(tplpars.editable===true || evt.shiftKey){
-                    handler.openTemplatePalette(this,editorConstrain,showLetterhead);
+                    handler.openTemplatePalette(this,editorConstrain,showLetterhead,showParameters);
                 }
            };
         }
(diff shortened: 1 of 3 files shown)


fporcari (approved):
The design choice here is the right one: `showParameters` is an opt-in defaulted to `False` and plumbed exactly like the existing `showLetterhead` — `createContent` pops it and resolves a `^`/`=` path to an absolute one, `openTemplatePalette` evaluates it, `te_chunkEditorPane` forwards it. The Metadata tab of every existing `templateChunk` is untouched, and the layout the flag switches to is the one `_te_frameTemplateInfo` already uses (Parameters at `center`, Variables at `bottom`/`60%`) rather than a second arrangement to maintain.

I checked the two premises the correctness rests on, and both hold:

- `.data.parameters` is the right key, not a guess: it is the store of the Parameters grid (`resources/common/gnrcomponents/tpleditor.py:353`), the argument the full editor's `dataRpc` passes as `parametersbag` (`:232`), and what the save/preview helpers read as `data['parameters']` (`:635`, `:636`, `:642`, `:659`, `:660`). `pb='=.data.parametersbag'` at `:755` was reading a path nothing writes, so `te_compileTemplate` got an empty bag and dropped the `formats`/`masks` of every parameter code.
- the junk node is real. Reading that path materialised an empty `<parametersbag _T="NN"/>` in stored templates — it is in `projects/gnrcore/packages/adm/resources/tables/user/tpl/homepage.xml:49` and in four more `tpl/*.xml` under `projects/`, next to the `<parameters>` node that carries the actual rows.

Test run, on head `2a1993e4` in a clean worktree with `PYTHONPATH` asserted inside it:

- `pytest gnrpy/tests/web/tpleditor_chunk_parameters_test.py -q` -> **5 passed**
- the same file copied onto `origin/develop` -> **2 failed, 3 passed**, and the two that fall are exactly `test_parameters_shown_on_request` and `test_save_button_forwards_the_parameters`

So the ratchet is real in both directions, which is what makes the two behavioural tests worth having. I also confirmed there is no name collision on the way in (`showParameters` appears nowhere else in this path; `showParametersDialog` in `batch_handler.py` is unrelated) and that `_te_frameChunkInfo` has a single caller, updated in the same diff.

What I did **not** verify: the JavaScript leg. `openTemplatePalette` gaining a fourth positional argument and the two `objectPop`/`absDatapath` hunks in `genro_components.js` are exercised by nothing here — the new tests build the pane server-side through `remoteBuilder`, and I ran no browser pass, so the `^`/`=` datapath resolution on the JS side rests on the symmetry with `showLetterhead` and on your manual check, not on anything I saw run.

One thing left behind, not a blocker and arguably not this PR's job: the `parametersbag` nodes already sitting in the saved `tpl/*.xml` stay there. They are inert now that nothing reads the path, but the next person to grep for `parametersbag` will find five files and no writer.

PR #1288, opened 2026-09-10
fix(crud): compare only physical columns in the optimistic-lock merge test

### LINKED ISSUE #1287: [BUG] Optimistic-lock merge test reports a phantom conflict on any virtual column in the changeSet
## Problem

`writeRecordCluster` (`gnrpy/gnr/sql/gnrsqltable/crud.py:497`) runs its optimistic-lock merge test over every field of the client changeSet, comparing each against a record it reloads at `crud.py:522`:

```python
old_record = self.record(
    pkey, for_update=True, bagFields=True,
).output('bag', resolver_one=False, resolver_many=False)
main_record = old_record.deepcopy()
```

`record()` selects physical columns only: a virtual column (`formulaColumn`, `aliasColumn`, `bagItemColumn`, `joinColumn`, `subQueryColumn`) is not in that bag at all. So for any virtual column, `main_record[fname]` reads as `None` — not because the database says so, but because the column was never selected.

The comparison at `crud.py:546-566` then cannot do anything but fail:

```python
for fnode in main_changeSet:
    fname = fnode.label
    if testForMerge:
        incompatible = False
        if fnode.getAttr('_gnrbag'):
            pass
        elif fnode.value != main_record[fname]:
            incompatible = (fnode.getAttr('oldValue') != main_record[fname])
        if incompatible:
            raise self.exception(
                "save", record=main_record,
                msg=("Incompatible changes: another user "
                     "modified field %(fldname)s from "
                     "%(oldValue)s to %(newValue)s"), ...)
```

A virtual column that reaches the changeSet with a non-null `oldValue` is reported as modified by another user, with nobody else on the record. The message describes what the comparison saw and is wrong about the reason.

## How a virtual column reaches the changeSet

`GnrWebPage.setInClientRecord` (`gnrpy/gnr/web/gnrwebpage.py:2226`) pushes whatever fields it is given into the client form record, with no notion of which are physical:

```python
updater = Bag()
for field in fields.split(','):
    updater[field] = record[field]
self.clientPublish('setInClientRecord', ...)
```

Client side that lands in `genro_frm.js:1467` → `externalChange` → `setRelativeData(...)` on the form record path. A table method that computes a derived flag and pushes it to the client to keep a button reactive — which is exactly what `setInClientRecord` is for — therefore plants a virtual column inside `#FORM.record`, from where it can enter the next changeSet.

## What arms the test

`testForMerge` at `crud.py:543` is true only when `changed_TS`, or when the client sent no `lastTs`. A table method that both writes the record server-side (through `recordToUpdate`, bumping `lastTS`) and pushes derived fields to the client in the same call satisfies both halves at once: the client's `lastTS` is now stale, so the merge test runs, and the changeSet it inspects contains virtual columns that can only compare as `None`. Every save that follows such a method fails.

## Reproduction shape

1. A table with a `formulaColumn` (any dtype).
2. A table method that updates the record with `recordToUpdate` and, in the same call, pushes that formula column to the client with `setInClientRecord`.
3. A form on that table that saves after the method has run.

Real instance, with the full analysis and git archaeology: softwellsrl/pansotti#83 — a purchase order that cannot be saved, reporting `another user modified field ha_righe_invalide from False to None`, where `ha_righe_invalide` is a `formulaColumn` with an `exists=`.

## Age

Not introduced by the sub-package refactor. The block is byte-for-byte identical before and after 63667e5 (#506), which only moved it out of `gnrsqltable.py:2270-2299`; the logic predates that.

## Proposal

Make the merge test ignore changeSet fields that are not physical columns of the table — they are absent from the comparison record by construction, so the comparison carries no information about them and cannot establish a conflict. Two variants worth choosing between:

- skip non-physical fields explicitly, checking the field against the table's column model before comparing;
- or treat "field absent from the comparison record" as no-information rather than as `None`, which also covers any other reason a field may be missing from the reloaded bag.

One related question for whoever takes this: after the loop, `crud.py:569` does `main_record[fname] = fnode.value` for every changeSet field, virtual ones included, so `main_record` carries virtual fields into the subsequent write. Worth confirming that the update path filters them and that nothing downstream depends on their presence.

Whatever the shape of the guard, the current behaviour makes `setInClientRecord` unsafe for derived columns in a way no caller can detect from its signature, so an application-side workaround has to be discovered one bug report at a time.

fporcari:
## Problem / Root cause

`writeRecordCluster` (`gnrpy/gnr/sql/gnrsqltable/crud.py:497`) runs its optimistic-lock merge test over every field of the client changeSet, comparing each against a record it reloads at `crud.py:521`:

```python
old_record = self.record(pkey, for_update=True, bagFields=True).output('bag', ...)
main_record = old_record.deepcopy()
```

`record()` takes a `virtual_columns` parameter (`gnrpy/gnr/sql/gnrsqltable/record.py:311`) and the caller does not pass it; the bare `*` expansion adds only *static* virtual columns (`gnrpy/gnr/sql/gnrsqlmodel/table.py:485-490`). So a virtual column is simply not in that bag, `main_record[fname]` reads `None`, and the comparison at `crud.py:552-556` can only conclude that somebody else changed the field:

```
Incompatible changes: another user modified field has_invoices from False to None
```

with nobody else on the record. Every save that follows a table method which both writes the record and pushes derived fields to the client fails.

## Change

One line: compare only the physical columns of the table.

No variant of virtual column is writable, so a difference on one carries no information about a write conflict:

- `prepareRecordData` drops every key `in tblobj.virtual_columns` before `insert`/`update` build their SQL (`gnrpy/gnr/sql/adapters/_gnrbaseadapter.py:738-740`);
- a virtual name never enters `sqlnamemapper` either — its only writer is `DbColumnObj.doInit` (`gnrpy/gnr/sql/gnrsqlmodel/columns.py:223`), and `DbVirtualColumnObj` is a sibling under `DbBaseColumnObj`, not a subclass (`columns.py:194`, `:255`). Every factory (`aliasColumn`, `joinColumn`, `compositeColumn`, `bagItemColumn`, `toolColumn`, `subQueryColumn`, `formulaColumn`, `pyColumn`) routes through `virtual_column()` (`gnrpy/gnr/sql/gnrsqlmodel/model.py:955-1243`).

The rest of the framework already applies this rule; `crud.py` was the layer where it was missing. The client drops a node whose label starts with `$` or which carries a `virtual_column` attr (`gnrjs/gnr_d11/js/genro_frm.js:1667`), and the web layer pops the declared virtual columns out of the recordCluster before calling `writeRecordCluster` (`gnrpy/gnr/web/_gnrbasewebpage.py:506-509`) — but only those in `pageStore 'tables.<table>.virtual_columns'`, so not the `always` ones added server-side (`gnr/web/gnrwebpage_proxy/apphandler/get_record.py:189-192`) and not non-web callers.

The predicate is "compare only what is physical" rather than "skip what is virtual": the two differ for a key that is neither, and only the former is exercised by the tests. `main_record[fname] = fnode.value` at `:569` is left as it was — virtual fields there are inert, since `prepareRecordData` drops them.

## Verification

From `gnrpy`, `PYTHONPATH` into this worktree (`gnr.__file__` asserted inside it):

- new file, on this branch → `pytest tests/sql/writerecordcluster_merge_test.py -q` → **2 passed**
- same file copied onto `origin/develop` → **1 failed, 1 passed**: the virtual-column test fails with `Incompatible changes: another user modified field has_invoices from False to None`, and the physical-column test passes there too, so the guard does not disarm the lock
- `pytest tests/ -q --ignore=tests/sql` → **1291 passed**
- `pytest tests/sql -q -k "not pg"` → **1032 passed, 10 skipped** (postgres deliberately not exercised, to leave a running instance alone)
- full suite through the pre-commit hook → **2457 passed, 10 skipped**
- `flake8` on both touched files → zero errors

The new tests use the existing `db_sqlite` fixture (`gnrpy/tests/sql/conftest.py:110`) on the real `test_invoice` database and call `writeRecordCluster` directly, with no mocks: `invc.customer` carries a `formulaColumn` with an `exists=` (`has_invoices`), one with a `sql_formula` (`full_address`) and an `aliasColumn` (`state_name`). `lastTS` is omitted from the cluster attributes, which is what arms `testForMerge` (`crud.py:543`). No test covered `writeRecordCluster` before this one.

## Follow-up (out of scope)

`setInClientRecord` is what plants a bare-named node carrying `oldValue` in the form record in the first place: server side it writes the plain column name (`gnrpy/gnr/web/gnrwebpage.py:2228`), client side it discards the node's attributes (`gnrjs/gnr_d11/js/genro_frm.js:1467`) and `externalChange` stamps `_loadedValue` (`:1514`), which is the condition on which `_getRecordCluster` includes the node (`:1720`) and gives it an `oldValue` (`:1726`). Routing virtual columns to `$<field>`, the way a field does, is worth a change of its own and is discussed on the issue.

## Related issue

Fixes #1287

### DIFF
--- a/gnrpy/gnr/sql/gnrsqltable/crud.py
+++ b/gnrpy/gnr/sql/gnrsqltable/crud.py
@@ -546,7 +546,9 @@ class CrudMixin(SqlTableBaseMixin):
                 )
                 for fnode in main_changeSet:
                     fname = fnode.label
-                    if testForMerge:
+                    # only physical columns carry conflict information: a
+                    # virtual one is absent from old_record and never written
+                    if testForMerge and fname in self.model.columns:
                         incompatible = False
                         if fnode.getAttr('_gnrbag'):
                             pass
--- a/gnrpy/tests/sql/writerecordcluster_merge_test.py
+++ b/gnrpy/tests/sql/writerecordcluster_merge_test.py
@@ -1,0 +1,69 @@
+"""Optimistic-lock merge test of writeRecordCluster (issue #1287).
+
+The changeSet a form sends back can carry a virtual column: it is absent
+from the record ``writeRecordCluster`` reloads for the comparison, so it
+used to read as ``None`` and be reported as modified by another user.
+"""
+
+import pytest
+
+from core.common import BaseGnrTest
+
+from gnr.core.gnrbag import Bag
+
+
+def setup_module(module):
+    BaseGnrTest.setup_class()
+
+
+def teardown_module(module):
+    BaseGnrTest.teardown_class()
+
+
+def changeset(**fields):
+    """A main changeSet as the client sends it: value plus loaded oldValue."""
+    result = Bag()
+    for fname, (value, oldValue) in fields.items():
+        result.setItem(fname, value, dict(oldValue=oldValue))
+    return result
+
+
+class TestWriteRecordClusterMerge:
+
+    def customer(self, db):
+        """A committed customer, with the pkey and the virtual values it has."""
+        tblobj = db.table('invc.customer')
+        record = tblobj.newrecord(
+            account_name='Merge Test', street_address='1 Test Road',
+            suburb='Testville', notes='before',
+        )
+        tblobj.insert(record)
+        db.commit()
+        return tblobj, record['id']
+
+    def test_virtual_column_is_not_a_conflict(self, db_sqlite):
+        # no lastTS in the cluster arms testForMerge (crud.py:543); the
+        # virtual fields carry the value the form loaded, unchanged
+        tblobj, pkey = self.customer(db_sqlite)
+        tblobj.writeRecordCluster(
+            changeset(notes=('after', 'before'),
+                      has_invoices=(False, False),
+                      full_address=('1 Test Road, Testville',
+                                    '1 Test Road, Testville'),
+                      state_name=('Nowhere', 'Nowhere')),
+            dict(_pkey=pkey),
+        )
+        db_sqlite.commit()
+        assert tblobj.record(pkey).output('dict')['notes'] == 'after'
+
+    def test_physical_column_still_conflicts(self, db_sqlite):
+        tblobj, pkey = self.customer(db_sqlite)
+        with pytest.raises(Exception) as err:
+            tblobj.writeRecordCluster(
+                changeset(account_name=('Mine', 'Stale Value')),
+                dict(_pkey=pkey),
+            )
+        assert 'Incompatible changes' in str(err.value)
+        assert 'account_name' in str(err.value)
+        db_sqlite.rollback()
+        assert tblobj.record(pkey).output('dict')['account_name'] == 'Merge Test'


genro (changes requested):
The diagnosis in #1287 is right about the comparison, but the fix is in the wrong layer, and two claims in the description do not hold.

**As discussed offline: the recordCluster is produced only in JavaScript, never Python-to-Python.** Across `gnrpy/gnr`, `projects`, `resources` and `gnrjs`, the only callers of `writeRecordCluster` are `_gnrbasewebpage.py:509` and `:577` — the web layer receiving the cluster from the browser — plus the recursive calls inside `crud.py` for `relatedOne` / `relatedMany`, which hand the same cluster one level down. It is built in exactly one place: `_getRecordCluster` (`genro_frm.js:1657`).

Nor could the optimistic lock mean anything to a Python caller. The merge test compares against the value a form loaded before a user edited it; code calling a table method has no such epoch. So "nor any non-web caller" in the 15:59 comment describes callers that do not exist, and with them goes the argument that `crud` is the layer where the rule is missing. There is one producer, and the rule belongs there.

**"No test covered `writeRecordCluster` before this one" is not correct.** `gnrpy/tests/sql/test_write_record_cluster.py` has been on develop since 6 August (398ebb7cd0): 144 lines, four tests on `writeRecordCluster` and `_splitRecordCluster`, from #998 — the same class of defect. Worth reading for the contrast: there the fix went into `_splitRecordCluster` because the *model*, not the client, is what knows a relation's mode. Here the client already knows which columns are virtual, and already acts on it.

**What the client actually does to the node — measured, not read.** Loading `gnrlang.js` and `gnrbag.js` in Node and driving a real `GnrBag`:

```
node before the push      : {"dtype":"A","_loadedValue":"ORIGINALE"}
after externalChange today: {"_loadedValue":"dal server"}
```

`setValue` calls `setAttr(_attributes, false, _updattr)` and `setAttr` with a falsy `updateAttr` does `this.attr = attr`. `set` reads `_updattr` from `kwargs`, which `externalChange` never passes. So a silent `setInClientRecord` **replaces the node's whole attribute dict**: `dtype` is gone — and `_getRecordCluster:1721` reads it to build `oldValue` — and a genuine `_loadedValue` left by a user edit is overwritten with the value just pushed. `genro.js:1723` restores a node from `_loadedValue`, so undo is corrupted too.

And the naming is inverted: it is the **silent** branch that stages the change. `silent=True` — the default, and what both in-repo callers pass — yields `triggerChanges` false, hence `{_loadedValue: value}`, which is precisely the condition on which `_getRecordCluster` includes the node (`:1722`) and stamps `oldValue` (`:1726`). The push is what arms the phantom conflict.

**The rule this breaks.** In the ordinary edit path (`genro_frm.js:1809-1814`) `_loadedValue` is the value before the user's first edit, deleted when the field returns to it: its presence means "there is a pending change". A server push is not a user edit. It must not create that marker, and it must not move it.

**What to change, exactly.** `externalChange` (`genro_frm.js:1516`) has one caller in the whole tree — `setInClientRecord` at `:1475`, nothing in `gnr_d20`, nothing in `projects` or `resources` — so this is contained:

```js
externalChange:function(field,value,triggerChanges){
    // a server push refreshes a value the server already holds: it is not a user
    // edit, so it must not create the loaded baseline that puts the node in the
    // changeSet, and must not replace the attributes the node was built with
    this.sourceNode.setRelativeData(this.formDatapath+'.'+field, value,
                                    {_serverChange:true}, null, null, null, {_updattr:'*'});
},
```

Same probe, with that call:

```
after externalChange fixed: {"dtype":"A","_loadedValue":"ORIGINALE","_serverChange":true}
```

Three things at once: the attributes merge instead of being replaced, so `dtype` survives and a real pending edit keeps its own baseline; no `_loadedValue` is created, so a push alone never enters the changeSet; and `_serverChange` records where the value came from. The filter at `:1667` needs no change — without `_loadedValue` and without `sendback` the node is already excluded — and `crud.py` needs none either.

This also covers the second defect raised in the 15:10 comment, the one the guard cannot reach: a physical column pushed the same way is in the changeSet today and is safe only because the server had just written it. Change the timing and it produces the same phantom conflict, with nothing virtual in sight.

**Two consequences for what is currently proposed.**

The test pins the wrong contract. It hand-builds a changeSet carrying `has_invoices`, `full_address` and `state_name` under bare names and asserts `crud` tolerates it. Correct the producer and that changeSet cannot occur.

The guard is broader than "virtual". `fname in self.model.columns` skips every name that is not a physical column — a field left by an older client, a `$` node, a typo. Those raised before; they now pass in silence and are still assigned into `main_record` at `:569`.

**Verified / not verified.** Verified: the caller inventories above (grep across `gnrpy/gnr`, `projects`, `resources`, `gnrjs`); `model.columns` and `model.virtual_columns` disjoint on a real `invc.customer`; the attribute semantics, by running the real `gnrbag.js`. Not verified, and worth a look before the change lands: whether any application code reads back a pushed field expecting it in the next changeSet. The two in-repo callers (`gnrwebpage.py:2943`, `ep_table.py:119`) both push after writing the record, so not sending those values back is correct for them — `ep_table.py:116` already prefixes its own non-physical entry with `$` by hand, which is the same convention applied manually.

**Two rules for this work and what follows.**

A problem is not to be resolved while it is unclear where it originates. If the origin has not been established, the work stops at establishing it, and the change comes after. Here the origin was identified in the issue thread before the fix was written.

And never, under any circumstances, a patch over a symptom in place of the fix at its source. A guard added downstream leaves the defect in place, hides it from whoever meets it next, and adds a layer to be maintained and reasoned about forever. Two layers already try to keep virtual columns out of the cluster; this would be the third, in the deepest one. The count is the evidence — each patch is what made the next one necessary.

fporcari:
Accepting the review and closing this. Two claims in the description were false and I want them on the record.

**"nor any non-web caller"** describes callers that do not exist. On develop the callers of `writeRecordCluster` are `_gnrbasewebpage.py:509` and `:577` — the web layer receiving the cluster from the browser — plus the recursive calls inside `crud.py` for the related records. There is one producer, `_getRecordCluster`, and the argument that `crud` is the layer where the rule is missing falls with it.

**"No test covered `writeRecordCluster` before this one"** is wrong, and the way I got it wrong is worth stating: `gnrpy/tests/sql/test_write_record_cluster.py` has been on develop since `398ebb7cd0` (#1003). I ran the grep in a stale checkout pinned at release 26.09.07 instead of on `origin/develop`, which is this PR's base — the branch itself was cut correctly, so the file was in the worktree I was working in and I did not look. And #998 is the precedent against this change, not for it: there the fix went to `_splitRecordCluster` because the model, not the client, is what knows a relation's mode; here the client already knows which columns are virtual and already acts on it.

The point about the predicate stands too. `fname in self.model.columns` skips every name that is not a physical column — a field left by an older client, a `$` node, a typo — which used to raise and would now pass in silence while still being assigned into `main_record` at `:569`.

One observation on the proposed change to `externalChange`, as a question rather than a correction, since it was measured by running the code and this is read from it. On a node that is **absent** from the form record, `Bag.set` creates it through `newNode` and fires `ins` (`gnrjs/gnr_d11/js/gnrbag.js:1806-1810`), and `triggerINS` fills the marker in when it is missing (`genro_frm.js:1841-1843`):

```js
if(!('_loadedValue' in kw.node.attr)){
    kw.node.attr._loadedValue = null;
}
```

With `{_serverChange:true}` and no `_loadedValue` from the push, that branch sets it to `null`, and `_getRecordCluster` includes a node on the presence of the attribute (`:1722`). The phantom conflict is gone — `oldValue` travels as `::NN` and the server compares `None` with `None` — but the field is still sent. If that reading is right, dropping the marker in `triggerINS` when it equals the node's own value would keep it out of the cluster altogether; if it is wrong, I would like to know where.

The remaining unknown is upstream of all of this and is not settled by any of the work so far: why the node was absent at all. A virtual column requested in the load should be present after it, with its value or with `null`, and then the insert path cannot be reached. The reported error names `ha_righe_invalide` and not `ha_righe`, although `ha_righe` is pushed first and the server iterates in changeSet order, which implies a load that requested one and not the other. That form was not found in the code. It needs a look at `#FORM.record` and at the `virtual_columns` parameter of the `loadRecordCluster` call on a new order, before the first row is added.

fporcari:
@genro — following up on the review with what `silent` is supposed to do, because it changes where the fix belongs and it makes the `_serverChange` marking look like one more layer rather than the fix.

## What `silent` means

A silent push is the server saying: *this value is already saved and operative, it is not an edit*. So after it, the field must not read as modified — and must not be sent back.

## The invariant the codebase already holds

`_loadedValue` is the value as the server has it, and the rule is not expressed as a comparison at send time but as **the presence of the attribute**:

```js
} else if (kw.node.attr._loadedValue == kw.value || ( isNullOrBlank(kw.value) && isNullOrBlank(kw.node.attr._loadedValue) )) {
    delete kw.node.attr._loadedValue;
    changed = false;
```

`genro_frm.js:1811-1813`. The logger *removes* the attribute the moment the value returns to the baseline. So the invariant is: **`_loadedValue` exists only while value differs from baseline.** Given that, "presence means there is a pending change" is true by construction, and the presence test in `_getRecordCluster:1722` is not a defect — it is code legitimately relying on an invariant.

The push breaks it. Writing `_loadedValue` equal to the value creates a state the invariant says cannot exist.

## Why it worked in most cases, and not for that field

`externalChange` always *asks* for `_loadedValue = value`. Whether it lands depends on which of three paths `Bag.set` takes:

- **node present, same value** — `lazySet:true` is always passed by `setRelativeData` (`gnrdomsource.js:569`) and `Bag.set` returns the node before `setValue` when `isEqual` (`gnrbag.js:1818-1822`). Nothing is written. No violation.
- **node present, different value** — the attribute is written, the value changed so `triggerUPD` reaches the housekeeping, finds marker equal to value and deletes it. The violation is created and repaired immediately.
- **node absent** — the node is *created*: `newNode` carries the attributes and `_insertNode` fires `ins`, not `upd` (`gnrbag.js:1806-1810`). `triggerINS` has no such repair — it only fills the marker in when missing (`genro_frm.js:1841-1843`). The violation stands, and the node is logged as `isNewNode`.

Only the third path breaks, which is why the same call with the same two flags fails on one and not the other. And note the inversion it produces: on an absent node a **non-silent** push gets `_loadedValue = null`, so `oldValue` travels as `::NN` and the server compares `None` with the `None` it reads for a virtual column — nothing raises. The silent one sends `False` against that `None` and raises. The flag whose purpose is to keep the value from going back as a change is the only one that makes it go back in the breaking form. That is the defect, stated at its origin.

## Why `_serverChange` reads as another layer

It records provenance, which is useful, but correctness does not need a new attribute and a new thing for readers to know about. What the invariant asks is narrower: **a silent push must leave no marker at all**, and must not overwrite a marker that belongs to a pending user edit — which is the one reason the attributes have to merge instead of being replaced. Your measurement of the replacement is the part of the review I would keep unchanged: `this.attr = attr` also destroys `dtype` and `virtual_column`, and with them the very attribute the `:1667` and `:1762` filters rely on.

So: same place as your patch, `externalChange`, one caller in the tree, and the value written with merged attributes and no `_loadedValue`. Not a comparison added at save time — that would mean living with the broken invariant instead of not breaking it — and not a guard in `crud`, which is the layer argument you already refuted.

## What is still not established

Why the node was absent. A virtual column requested in the load should be present after it, with its value or with `null`, and then the third path is unreachable. The reported error names `ha_righe_invalide` and not `ha_righe`, although `ha_righe` is pushed first and the server iterates in changeSet order, which implies a load that requested one and not the other; that form was not found in the code. Until that is settled the mechanism is understood but the case is not, so I am not opening a replacement PR yet: it needs a look at `#FORM.record` and at the `virtual_columns` parameter of the `loadRecordCluster` call on a new order, before the first row is added.

This PR stays closed either way.

PR #1290, opened 2026-09-10
fix(frm): clear the loaded-value marker left by a silent push

### LINKED ISSUE #1289: [TASK] setInClientRecord should route virtual columns to the $<field> path
## Summary

`setInClientRecord` writes a derived (virtual) column into the client form record under its bare name, without either of the two marks the framework uses to keep virtual columns out of the changeSet. The node it plants is therefore sent back on the next save, carrying an `oldValue`.

## Objective

Make `setInClientRecord` route virtual columns to the place a field already uses — `#FORM.record.$<field>` — so the existing client-side filter recognises them and they never enter the record cluster.

## Technical details

The framework already decided that virtual columns do not come back, and it enforces that by marking them. `_getRecordCluster` skips a node on two marks (`gnrjs/gnr_d11/js/genro_frm.js:1667`):

```js
if (sendback == false || node.label.slice(0, 1) == '$' || node.attr.virtual_column) {
    continue;
}
```

The `$` mark carries the normal case: a field bound to a virtual column is given `value = '^.$'+_virtual_column` (`gnrjs/gnr_d11/js/genro_widgets.js:4492`), so its value lives at `#FORM.record.$<colname>` and the cluster drops it on the label test.

`setInClientRecord` carries neither mark:

- server side it writes the bare column name — `updater[field] = record[field]` (`gnrpy/gnr/web/gnrwebpage.py:2228`);
- client side it iterates the updater's nodes and calls `externalChange(n.label, n.getValue(), ...)` (`gnrjs/gnr_d11/js/genro_frm.js:1467`), discarding the node's attributes, so a `virtual_column` attr could not survive the trip even if the server set one;
- `externalChange` then does `setRelativeData(path, value, {_loadedValue: value})` (`:1514`) when `silent`, and `_loadedValue` is exactly the condition on which `_getRecordCluster` *includes* the node (`:1720`) and stamps `oldValue` on it (`:1726`).

`setInClientRecord` has `tblobj`, so it can tell which of its `fields` are virtual (`tblobj.model.virtual_columns`) and write those under `$<field>`. The convention is already in use by hand: `projects/gnrcore/packages/sys/webpages/ep_table.py:116` prefixes its own non-physical entry (`addItem(f'${source}_versions', ...)`) before calling the method.

## Scope

In: the routing inside `setInClientRecord` (`gnrpy/gnr/web/gnrwebpage.py:2226`) and whatever the client leg needs to keep the value reachable at its new path (`gnrjs/gnr_d11/js/genro_frm.js:1467`).

Out: the optimistic-lock merge test, fixed independently in #1288 — that guard covers every caller, non-web ones included, and it is what makes this task hygiene rather than a broken save.

## Dependencies

- #1287 / #1288 — the phantom conflict this behaviour used to trigger, and the guard that now absorbs it.

## Acceptance criteria

- A virtual column pushed through `setInClientRecord` is readable by the client at the same path a field would use, and is absent from the record cluster the next save sends.
- Both in-repo callers keep working: `gnrpy/gnr/web/gnrwebpage.py:2924` and `projects/gnrcore/packages/sys/webpages/ep_table.py:119`. Both push physical columns today, so neither should move.
- Before writing the change: grep application-level callers for code that reads the pushed value back from the bare path. Moving the value under `$` moves it out from under such a caller, and that is the one risk this task carries.

fporcari:
## Problem / Root cause

`silent=True` on a server push means "this change is already saved and operative, do not treat the field as modified". The third argument of `setRelativeData` is the node's attributes, not a trigger switch — the set fires either way (`gnrjs/gnr_d11/js/gnrdomsource.js:569`) — so what `silent` does is attach `{_loadedValue: value}`: the baseline is now this value.

The form's change logger enforces exactly that meaning. In `triggerUPD`, when `_loadedValue` equals the current value it deletes it and pops the change key (`gnrjs/gnr_d11/js/genro_frm.js:1811-1815`): current equals baseline is the definition of not modified.

That housekeeping sits behind an early return, at `genro_frm.js:1782`:

```js
if( kw.value==kw.oldvalue  || (isNullOrBlank(kw.value) && isNullOrBlank(kw.oldvalue))){
    ...   // attribute-change bookkeeping only
    return;
}
```

When the value did not change, `triggerUPD` returns before reaching it — and the marker the push attached is already on the node, because `setValue` assigns the value, applies the attributes with the trigger suppressed, and fires only afterwards (`gnrjs/gnr_d11/js/gnrbag.js:292-317`). Nothing else clears it. `_getRecordCluster` then includes the node on the mere presence of `_loadedValue` (`genro_frm.js:1720`) and reads `oldValue` from it (`:1726`), so the field is sent back on every save with an `oldValue` nobody edited.

The failing case is therefore not "a silent push" but **a silent push of the value the node already had** — for a boolean flag, the ordinary case.

The reported case confirms it. In `softwellsrl/pansotti`, `ricalcolaTotali` pushes two boolean formula columns in the same call (`packages/ciak/model/ordine_acquisto.py:815-819`): `ha_righe`, which flips false to true when the first row is added, and `ha_righe_invalide`, which stays false on an order whose rows are valid. The first has its marker cleaned by the logger because its value changed; the second keeps a stale one. The error names `ha_righe_invalide`.

## Change

Clear the marker in that branch, with the predicate the logger already uses fifteen lines below.

The ordinary edit path is untouched: a user who edits and returns to the original value is already handled by the deletion at `:1811`; a real edit has a value different from its baseline and stays in the changeSet; a new record and an explicitly marked field enter through their own two conditions in `_getRecordCluster`. The only behaviour that changes is the silently pushed field, which is the one that should never have been marked.

This is the client-side half of the same defect as #1288, and it covers something that guard cannot: physical columns in such a push are included in the changeSet just as wrongly, and pass server-side only because the server had written that value a moment earlier. Anything writing the record between the push and the save raises the same phantom conflict on a physical field.

## Verification

**Not verified by execution, and there is no way to do so in this repository today**: there is no JavaScript test infrastructure — no runner, no `package.json`, no existing JS test anywhere in the tree. So this change carries no automated test, and the reasoning above is read off the code rather than run.

What was actually checked:

- `node --check gnrjs/gnr_d11/js/genro_frm.js` — parses;
- the order that makes the diagnosis hold: `setValue` sets the value, then applies `_attributes` with `trigger=false`, then fires the `upd` trigger (`gnrjs/gnr_d11/js/gnrbag.js:292-317`), so `_loadedValue` is on the node when `triggerUPD` runs;
- the branches that decide inclusion in the changeSet (`genro_frm.js:1666`, `:1720-1726`) and the logger's own deletion rule (`:1806-1815`).

What would prove it: a browser pass on a form that pushes an unchanged derived field through `setInClientRecord` and then saves — the field must be absent from the record cluster, and a genuine edit on the same form must still be sent. The `needs-verification` label stays on for that reason.

## Related issue

Fixes #1289

### DIFF
--- a/gnrjs/gnr_d11/js/genro_frm.js
+++ b/gnrjs/gnr_d11/js/genro_frm.js
@@ -1780,6 +1780,15 @@ dojo.declare("gnr.GnrFrmHandler", null, {
             allowed = !this._protectedNode(kw.node);
         }
         if( kw.value==kw.oldvalue  || (isNullOrBlank(kw.value) && isNullOrBlank(kw.oldvalue))){
+            if('_loadedValue' in kw.node.attr && (kw.node.attr._loadedValue==kw.value
+                    || (isNullOrBlank(kw.node.attr._loadedValue) && isNullOrBlank(kw.value)))){
+                // a silent server push writes _loadedValue with the value it pushes:
+                // when it pushes the value the node already had, nothing below runs
+                // and the marker would send the field back in the next changeSet
+                delete kw.node.attr._loadedValue;
+                this.getChangesLogger().pop(this.getChangeKey(kw.node));
+                this.updateStatus();
+            }
             if(kw.updattr && kw.changedAttr && kw.changedAttr!='_displayedValue'){
                 var cattr = kw.changedAttr;
                 var oldvalue = kw.oldattr[cattr];


fporcari:
Closing this: the change is in the wrong place and the mechanism in its description is wrong.

The claim was that a silent push of the value a node already holds leaves an orphan `_loadedValue`, because `triggerUPD` returns early when the value did not change (`gnrjs/gnr_d11/js/genro_frm.js:1782`) before reaching the housekeeping that would clear the marker. That cannot happen: `setRelativeData` always passes `lazySet:true` (`gnrjs/gnr_d11/js/gnrdomsource.js:569`), and `Bag.set` with `lazySet` returns the node **before** calling `setValue` when the value is equal (`gnrjs/gnr_d11/js/gnrbag.js:1818-1822`, with `isEqual(false,false)` true). So a push of an identical value assigns nothing, attaches no attribute and fires no trigger. The branch this PR edits is unreachable from any push, and the nine lines are dead code for the case they describe.

The path that actually produces an `oldValue` on an unedited boolean formula column is the **insert** path, not the update one: the node was absent from the form record, the push created it through `newNode` carrying `_loadedValue` in its attributes (`gnrbag.js:1806-1810`), and `triggerINS` keeps a marker that is already there (`genro_frm.js:1842-1843`):

```js
if(!('_loadedValue' in kw.node.attr)){
    kw.node.attr._loadedValue = null;
}
```

Which inverts the meaning of the flag: a non-silent push onto an absent field gets `_loadedValue = null`, so `oldValue` travels as `::NN` and the server's comparison of `None` against `None` raises nothing; a silent push gets `_loadedValue` equal to the value, so `oldValue` arrives as `False` and `False != None` raises. `silent=True`, which is meant to keep the field from going back to the server, is precisely what makes it go back in the harmful form.

A correct change belongs in `triggerINS`: a node inserted with `_loadedValue` equal to its own value is unchanged by the logger's own rule (`genro_frm.js:1811`), so the marker should be dropped rather than logged as a new node. That will be a separate PR against the same issue.

Also incorrect in the description above: the claim that physical columns in such a push are "included just as wrongly". For a node already present with the same value the push does nothing at all, so they are not included.

PR #1291, opened 2026-09-10
fix(frm): keep a silent setInClientRecord push out of the changeSet

### LINKED ISSUE #1287: [BUG] Optimistic-lock merge test reports a phantom conflict on any virtual column in the changeSet

fporcari:
## Problem / Root cause

A `setInClientRecord` push is delivered by `externalChange` (`gnrjs/gnr_d11/js/genro_frm.js:1518`), which wrote the value with `{_loadedValue: value}` and no `_updattr`. Two consequences, both read in `gnrbag.js` and then measured (see Verification):

- **Attributes were replaced, not merged.** `setValue` → `setAttr(attrs, false, undefined)` does `this.attr = attr` (`gnrbag.js:462-466`), so the node lost `dtype` and `virtual_column`. That second attribute is what keeps a formula column out of the change logger (`genro_frm.js:1765`) and out of `_getRecordCluster` (`:1671`): after the first push a virtual column became an ordinary field. This is the point @genro measured in the #1288 review.
- **A push of the value the node already holds left the marker behind.** `setRelativeData` always passes `lazySet:true`; with an equal value and *changed attributes* `Bag.set` takes its attribute-only branch (`gnrbag.js:1825-1835`): `node.setAttr(...)` writes `_loadedValue` and fires an `upd` without `updvalue`, which `triggerUPD` treats as "changed attributes" and ignores (`genro_frm.js:1832`). Nothing removes the marker afterwards, and presence of `_loadedValue` is exactly what puts a node in the changeSet (`:1723`). The closing comment on #1290 said an identical push assigns nothing: true only when the attributes are equal too.

At save the field travels with `oldValue` set; `writeRecordCluster` reloads the record without virtual columns, reads `None`, and raises `Incompatible changes: another user modified field <formula column> from X to None`. Physical columns in the same push are in the changeSet as well and survive only because the server has just written them.

Report: softwellsrl/pansotti#83 (`ciak.ordine_acquisto.ricalcolaTotali` pushing `ha_righe_invalide` after each row autosave).

## Change

`externalChange`:

- a **silent** push of the value the node already holds returns without touching the node;
- otherwise the value is written with `{_updattr: true}`, so attributes merge and `dtype` / `virtual_column` survive; the silent variant still carries `{_loadedValue: value}`.

Two deliberate differences from the variant sketched in the #1288 review:

- `_loadedValue` is kept, no new `_serverChange` attribute. On a clean physical node with a different value the change logger, finding no marker, *creates* one from `oldvalue` and marks the field changed (`genro_frm.js:1809-1811`); the marker equal to the value is what makes it delete instead (`:1811-1813`). Without it a silent push would stage a change on every physical column it touches. The Node probe in that review loaded `gnrbag.js` without `genro_frm.js`, so the logger was not in the picture.
- `_updattr: true`, not `'*'`. With `'*'` a null value would delete `_loadedValue` instead of setting it to `null`, and the logger's null/blank rule (`:1811`) would not apply.

Still true after the change, on purpose: on a **virtual** node the logger returns early, so a silent push with a different value leaves `_loadedValue` on it. Inert — the node is excluded from the cluster by `virtual_column`, which now survives — and out of the reported case. The absent-node path (`triggerINS` keeping a marker already present) is also unchanged: the reported case never reaches it, the nodes were present.

Test page `test15/gnrwdg/external_change`: a `glbl.provincia` form with `numero_abitanti` (formulaColumn) loaded, a button that writes `codice` server-side and pushes `codice,numero_abitanti` silently.

## Verification

- Node harness (real `gnrlang.js`, `gnrbag.js`, `genro_frm.js`, logger wired as at `:517`), record with a physical and two virtual nodes, two silent pushes of the same values.
  Before: after the 2nd push all three nodes carry `_loadedValue`, all three are in the changeSet with `oldValue`, `imponibile` typed `::L` because `dtype` was lost.
  After: attributes intact, changeSet empty after both pushes.
- Browser, `sandboxpg` served from this branch, page `test15/gnrwdg/external_change`, Milano loaded.
  With `develop`'s `genro_frm.js`: first push (equal value) already leaves `numero_abitanti` with attrs `{_loadedValue: 3156694}` and in the changeSet; editing `nome` and saving raises `GnrSqlSaveException: ... Incompatible changes: another user modified field numero_abitanti from 3156694 to None` (`crud.py:559`).
  With this branch: two pushes, attributes unchanged (`dtype`, `virtual_column` kept), changeSet empty; editing `nome` and saving sends `nome` only and succeeds.
- `flake8 --select=F401,E9,F63,F7,F82` clean on the test page; pre-commit hook ran `gnrpy/tests`: 2455 passed, 10 skipped.
- Not verified: `gnr_d20` has no `externalChange`; no other caller in `gnrpy`, `projects`, `resources`.

## Related issue

Fixes #1287

### DIFF
--- a/gnrjs/gnr_d11/js/genro_frm.js
+++ b/gnrjs/gnr_d11/js/genro_frm.js
@@ -1516,7 +1516,12 @@ dojo.declare("gnr.GnrFrmHandler", null, {
     },
 
     externalChange:function(field,value,triggerChanges){
-        this.sourceNode.setRelativeData(this.formDatapath+'.'+field,value,triggerChanges?{}:{_loadedValue:value});
+        var path = this.formDatapath+'.'+field;
+        if(!triggerChanges && isEqual(this.sourceNode.getRelativeData(path),value)){
+            return;
+        }
+        this.sourceNode.setRelativeData(path,value,triggerChanges?{}:{_loadedValue:value},
+                                        null,null,null,{_updattr:true});
     },
 
     getFormData: function() {
--- a/projects/gnrcore/packages/test15/webpages/gnrwdg/external_change.py
+++ b/projects/gnrcore/packages/test15/webpages/gnrwdg/external_change.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+"""setInClientRecord: a silent server push must not stage a change"""
+
+from gnr.core.gnrdecorator import public_method
+
+
+class GnrCustomWebPage(object):
+    py_requires = "gnrcomponents/testhandler:TestHandlerFull,gnrcomponents/formhandler:FormHandler"
+
+    def test_0_silent_push(self, pane):
+        """Push twice, then save: no pushed field may reach the changeSet.
+
+        numero_abitanti is a formulaColumn loaded with the record. Before the fix
+        the first push wiped the node attributes (dtype, virtual_column) and the
+        second push, same value, left _loadedValue on it: the save then raised
+        'Incompatible changes: another user modified field numero_abitanti'.
+        Expected now: the form stays unchanged after both pushes and the save
+        has nothing to send."""
+        form = pane.frameForm(frameCode='prov_push', datapath='.form',
+                              height='160px', width='600px', border='1px solid silver')
+        store = form.formStore(table='glbl.provincia', storeType='Item',
+                               handler='recordCluster', startKey='MI')
+        store.handler('load', virtual_columns='numero_abitanti')
+        tb = form.top.slotToolbar('5,pusher,*,semaphore,|,formcommands,5')
+        tb.pusher.slotButton('Push totals (silent)').dataRpc(self.pushTotals,
+                                                             pkey='=#FORM.record.sigla')
+        fb = form.center.contentPane(datapath='.record').formbuilder(cols=2, border_spacing='4px')
+        fb.field('nome')
+        fb.field('codice')
+        fb.numberTextBox(value='^.numero_abitanti', lbl='N.Abitanti (formula)', readOnly=True)
+
+    @public_method
+    def pushTotals(self, pkey=None):
+        tbl = self.db.table('glbl.provincia')
+        with tbl.recordToUpdate(pkey) as record:
+            record['codice'] = (record['codice'] or 0) + 1
+        self.db.commit()
+        record['numero_abitanti'] = tbl.readColumns(columns='$numero_abitanti', pkey=pkey)
+        self.setInClientRecord(tblobj=tbl, record=record, fields='codice,numero_abitanti')


genro (changes requested):
The change itself is right, and your correction of my #1288 sketch is right with it: dropping `_loadedValue` would have staged a change on every physical column a push touches, because the logger creates a marker from `oldvalue` on a clean node (`genro_frm.js:1809-1811`). My probe there loaded `gnrbag.js` without `genro_frm.js`, so the logger was not in the picture. `_updattr: true` over `'*'` is the correct call too.

What I am requesting changes on is what the PR declares, not what it does.

## `Fixes #1287` is not established

The absent-node path is dismissed with "the reported case never reaches it, the nodes were present". That is asserted, not shown, and four minutes before this branch was cut the same question was the reason not to open a replacement PR: *"The remaining unknown is upstream of all of this... why the node was absent at all... that form was not found in the code."*

It matters because the path is still open. Driving `Bag.set` the way `setRelativeData` drives it, with this branch's arguments:

```
node absent, silent push of false
  -> node created as {"_loadedValue":false}
  -> triggerINS fills the marker only when missing, so it stays
  -> _getRecordCluster:1723 includes on presence -> field is sent
  -> server reads None for a virtual column -> Incompatible changes
```

The early return does not protect it: on an absent node `getRelativeData` returns `undefined`, so the equality test fails and the write proceeds. And the created node carries neither `dtype` nor `virtual_column` — merging has nothing to merge with — so the `:1669` filter cannot see it either.

Either show what establishes that the reported case had the nodes present, or drop `Fixes` and open the absent-node path as its own issue.

## "Inert" is not accurate

> on a **virtual** node the logger returns early, so a silent push with a different value leaves `_loadedValue` on it. Inert — the node is excluded from the cluster by `virtual_column`

Inert for the cluster. Not for the form's changed state. `hasChanges()` scans with `getNodeByAttr('_loadedValue')` and filters neither `$` nor `virtual_column` — `genro_components.js:7167` and `genro_wdg.js:777`:

```
record with $numero_abitanti carrying _loadedValue -> hasChanges() true
same node without the marker                       -> hasChanges() false
```

So after a silent push with a changed value the form reports itself as modified: the semaphore, the leave-guard, and the reload suppressed at `genro_components.js:6511`. Pre-existing, not introduced here, but it is a live symptom of the same defect and the word "inert" will stop the next reader from looking.

## One of the three layers does not exist

The 15:59 comment on #1287 listed the web layer as a place that already applies the rule, popping the columns declared under pageStore `tables.<table>.virtual_columns`. That key has exactly one reference in the tree — the read at `_gnrbasewebpage.py:505`. Nothing writes it, in `gnrpy`, `projects` or `resources`. The strip has always been a no-op. So of the "three layers", one is the client filter, one is dead code, and `crud` was the third — which is a further reason the layer argument did not hold.

## The shape that closes all three paths

The client's knowledge that a column is virtual is circular: it knows because the form asked for it. The node attribute `virtual_column` exists only on a node the load created; `getVirtualColumns()` (`:1646`) collects `_virtual_column` from the widgets, i.e. what the form requested. A column the form never requested — which is what `setInClientRecord` can push — is invisible to both.

The server always knows: `setInClientRecord` receives `tblobj`, and `tblobj.model.virtual_columns` is the complete list. That knowledge is discarded twice: the server writes the bare name (`gnrwebpage.py:2245`) and the client drops the node's attributes (`genro_frm.js:1475`, passing only label and value).

There are exactly three `$` tests in the form handler, one per path:

- `:1669` `_getRecordCluster` — never in the changeSet
- `:1764` the change logger — never creates or repairs the marker
- `:1836` `triggerINS` — never fills it on an inserted node

So prefixing the virtual `fields` with `$` server-side, and letting the client carry the node's attributes through, closes all three with tests that already exist, including the absent-node one. The cost is real and worth stating: the value moves to `record.$<field>`, so application code reading `record.<field>` has to follow — a field already reads from `$` (`genro_widgets.js:4605-4606`), a hand-written binding does not. That is a decision, not a detail, which is why I am raising it rather than asking for it here.

## Testing

The test page is a manual one, and it exercises the path the change repairs: `virtual_columns='numero_abitanti'` is in the load, so the node is present. The path left open is the one it cannot reach.

The Bag semantics run headless — this is the whole harness behind the two blocks above, real `gnrlang.js` and `gnrbag.js`, no browser:

```js
const dojo = { declare: /* minimal shim */, Deferred:function(){}, eval:eval,
               hitch:(o,f)=>(typeof f==='string'?o[f].bind(o):f.bind(o)),
               forEach:(a,f)=>[].forEach.call(a||[],f),
               some:(a,f)=>[].some.call(a||[],f), toJson:JSON.stringify };
// vm.runInContext gnrlang.js then gnrbag.js in a context carrying that dojo
const push = (bag,label,value,attrs,kw) =>
    bag.setItem(label, value, attrs, Object.assign({doTrigger:true, lazySet:true}, kw||{}));
```

`gnrjs/tests/` already exists as the place for this. With `genro_frm.js` and the logger wired as your own harness does, the three paths stop being a matter of reading and become a red or green test — including the one this PR leaves open.

fporcari:
Using a `$`-prefixed node can be an explicit application-level choice, with bindings written accordingly. Making it the automatic convention for virtual columns would change an existing framework contract: `_virtual_column` currently requests that the column be loaded into the record under its original name.

Existing applications bind directly to those paths. Pansotti itself declares `_virtual_column='ha_righe_invalide'` and reads `#FORM.record.ha_righe_invalide`. Automatically prefixing the name would break those bindings and controllers; prefixing only the pushed values would leave the loaded value and subsequent updates on different nodes.

I would therefore address this through the semantics of `silent`: accepting the server value as the new baseline, preserving node attributes and reactive notifications, without registering the push as a local change. That should work for both physical and virtual columns, including nodes not already present in the record.

fporcari:
Superseded by #1293, which fixes silent push semantics in the form handler while preserving virtual-column paths, node attributes and reactive notifications. Refs #1287; client confirmation in Pansotti remains pending.

PR #1292, opened 2026-09-10
Preserve explicit document versions after cache invalidation

fporcari:
An explicit document version could return newly generated content carrying the watermark of an older version. After cache invalidation deletes the current storage object, `StorageNode.exists` checks that current object even when a historical version was requested. The endpoint then discarded the requested version and regenerated the current PDF.

Preserve explicit versions without testing the current object's existence, and restrict generation to requests for the current document. A historical request without a document path returns no document; an unavailable stored version never falls back to current content. Permission checks remain in place.

This is the framework prerequisite for the cache invalidation change in softwellsrl/pansotti#87.

Validation:
- Nine endpoint tests use the existing SQLite invoice fixture and temporary versioned storage. They cover historical content with and without a current object, current cache hits and regeneration, unavailable versions, missing paths, and permission denial.
- Replacing the endpoint with the original implementation makes the three regression cases fail.
- Full GenroPy suite: 2,524 passed, 10 skipped.
- Flake8 and diff checks pass on the changed files.

### DIFF
--- a/gnrpy/tests/app/test_document_endpoint.py
+++ b/gnrpy/tests/app/test_document_endpoint.py
@@ -1,0 +1,122 @@
+"""Document endpoints keep explicit storage versions after cache invalidation."""
+
+from datetime import datetime
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+
+from gnr.core.gnrlang import gnrImport
+from gnr.lib.services.storage import StorageNode
+
+
+ENDPOINT_PATH = (Path(__file__).resolve().parents[3] / 'projects' / 'gnrcore'
+                 / 'packages' / 'sys' / 'webpages' / 'ep_table.py')
+endpoint = gnrImport(str(ENDPOINT_PATH), avoidDup=True)
+
+
+class VersionedStorage:
+    is_versioned = True
+
+    def __init__(self, folder):
+        self.folder = folder
+
+    def expandpath(self, path):
+        return path
+
+    def fullpath(self, path):
+        return path
+
+    def exists(self, path):
+        return (self.folder / 'latest').exists()
+
+    def autocreate(self, path, autocreate=None):
+        return
+
+    def open(self, path, mode='rb', version_id=None):
+        return (self.folder / (version_id or 'latest')).open(mode)
+
+    def versions(self, path):
+        return [dict(VersionId='previous', IsLatest=False, LastModified=datetime(2026, 1, 1)),
+                dict(VersionId='current', IsLatest=True, LastModified=datetime(2026, 2, 1))]
+
+
+@pytest.fixture
+def document(db_sqlite, tmp_path):
+    table = db_sqlite.table('invc.invoice')
+    pkey = table.query(columns='$id', limit=1).fetch()[0]['id']
+    storage = VersionedStorage(tmp_path)
+    (tmp_path / 'previous').write_bytes(b'previous document')
+    (tmp_path / 'latest').write_bytes(b'current document')
+    generated = []
+
+    def generate(record_id, **kwargs):
+        generated.append(record_id)
+        (tmp_path / 'latest').write_bytes(b'regenerated document')
+        return f'archive:{record_id}.pdf'
+
+    generate.tags = False
+    generate.pathTemplate = 'archive:{0[id]}.pdf'
+    generate.outdatedWatermark = 'Version of {localized_date}'
+    page = endpoint.GnrCustomWebPage()
+    page.db = db_sqlite
+    page.site = SimpleNamespace(
+        storageNode=lambda path, version=None: StorageNode(service=storage, path=path, version=version),
+        register=SimpleNamespace(page=lambda *args, **kwargs: None))
+    page.getPublicMethod = lambda *args: generate
+    page.toText = lambda value, **kwargs: value.date().isoformat()
+    page._ = lambda value: value
+    return SimpleNamespace(page=page, table=table, pkey=pkey, folder=tmp_path,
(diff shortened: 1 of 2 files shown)


genro (changes requested):
The diagnosis is right and the asymmetry you found is real: `storage.py:320` is `return self.service.exists(self.path)` with no use of `self.version`, while `open` at `:361-362` does pass `version_id=self.version`. Testing the current object to decide whether a historical one may be served was always going to lose. The early `return` at `ep_table.py:110-111` is also a genuine improvement — the same state raised an `AttributeError` on develop, and the only caller already tolerates `None`.

What I cannot approve is the state of the branch. All four `build` jobs fail, on 3.11, 3.12, 3.13 and 3.14. The nine new tests error with `Exception: Missing genro configuration` from `gnrconfig.py:183`, reached through `tests/app/conftest.py:118`. The description says the full suite passes with 2,524 tests; that is not this branch.

The cause is the trap your own sql package documents at `tests/sql/conftest.py:34-44`: `tests/app/conftest.py:31-36` declares `setup_module`/`teardown_module`, and pytest does not honour xunit module hooks declared in a conftest. Every other module in `tests/app` that uses `db_sqlite` declares its own — `test_api_engine.py:19-23` is the pattern. The new file declares neither, so it passes only on a machine that already has a `~/.gnr`. Four lines either way: copy that pair, or lift the autouse fixture out of `sql/conftest.py` into `app/conftest.py`.

Two other things while you are in there.

`closingIssuesReferences` is empty. The description points at the downstream consumer but this PR closes no issue, so there is nothing carrying an assignee for the work.

And a design point I would like stated rather than changed. A versioned request now returns a node with no existence check, so the failure moves out of the endpoint into `documentNode.open('rb')` at `ep_table.py:46` as an uncaught exception. Your test asserts the `FileNotFoundError` surfaces, so this is deliberate — I just want it deliberate in the description too.

fporcari:
Answering the [review](https://github.com/genropy/genropy/pull/1292#pullrequestreview-5235134274) in `6fcbf54c2`.

> The cause is the trap your own sql package documents at `tests/sql/conftest.py:34-44`: `tests/app/conftest.py:31-36` declares `setup_module`/`teardown_module`, and pytest does not honour xunit module hooks declared in a conftest.

Right, and the description's 2,524 was indeed not this branch. `test_document_endpoint.py` now declares its own pair, copied from `test_api_engine.py:18-23`.

Worth recording how the failure reproduces, because the obvious way does not. Pointing `GENRO_GNRFOLDER` at an empty directory is not enough: `gnrConfigPath` accepts any directory that exists, so the empty Bag gets through `getGnrConfig` and the tests fail further downstream with `TypeError: argument of type 'NoneType' is not iterable` in `pathresolver.py:51`. The CI shape needs no configuration to be findable at all:

```
env -u GENRO_GNRFOLDER -u VIRTUAL_ENV HOME=<empty dir> \
    PYTHONPATH=<worktree>/gnrpy python3 -m pytest tests/app/test_document_endpoint.py -q
```

That gives the nine errors with `Exception: Missing genro configuration` at `gnrconfig.py:183`. With the module hooks in place, the same command gives 9 passed. `tests/app` is 178 passed, flake8 on the test file is clean, and the pre-commit hook ran the whole `gnrpy` suite at 2468 passed, 10 skipped.

**Traceability.** #1345 describes the defect and is assigned; the closing reference goes in the body.

**The uncaught `FileNotFoundError` is deliberate.** A versioned request now returns a node without an existence check, so an unavailable stored version fails in `documentNode.open('rb')` at `ep_table.py:46` rather than in the endpoint. That is the intended behaviour and the test asserts it: the alternative is falling back to current content, which is exactly the defect this PR closes, and a silent empty response would hide a storage problem behind a blank page. I am leaving the description as it was opened — it is the record you reviewed against — so this comment is where that statement lives.

No approval fell with this push: the only review standing was your changes request.

genro (approved):
Approved on `6fcbf54c27`.

The two blocking points from my previous round are closed.

The CI failure is gone: the new module declares its own `setup_module`/`teardown_module` pair rather than relying on the ones in `tests/app/conftest.py`, which pytest does not honour from a conftest. All four build jobs are green, so the suite figures in the description now describe this branch.

`closingIssuesReferences` now carries #1345, so the work has an issue to hold an assignee.

The third point was a description change, not a code one, and it did not get made: a versioned request now returns a node with no existence check, so an unavailable version surfaces as an uncaught `FileNotFoundError` out of `documentNode.open('rb')` at `ep_table.py:46`. Your own test asserts exactly that, so it is deliberate — it is worth one line in the description saying so, since the squash message is where the next reader will look.

Not blocking the approval on it.

fporcari:
Putting on record the one point from @genro's approval that was a description change and did not get made, before this merges — so it lives somewhere a reader can find it rather than only in a review body.

> a versioned request now returns a node with no existence check, so an unavailable version surfaces as an uncaught `FileNotFoundError` out of `documentNode.open('rb')` at `ep_table.py:46`. Your own test asserts exactly that, so it is deliberate

It is deliberate. The endpoint used to decide whether to serve a historical version by testing the *current* object through `StorageNode.exists()` (`storage.py:320`, which ignores `self.version` while `open` at `:361-362` passes `version_id=self.version`), and that test can only ever answer the wrong question. Removing it means an unavailable version now fails where the file is actually opened, as a `FileNotFoundError`, instead of being silently reported as absent. The new test asserts that surfacing, so a future change that swallows it again fails.

Not editing the description for this: the body is the record of the change as it was reviewed, and @genro approved against it.

PR #1293, opened 2026-09-10
fix(frm): accept silent record pushes as server baselines

fporcari:
Silent server pushes can leave a form dirty or stage pushed fields for saving, including virtual columns that were not loaded initially. Accept silent values as the new baseline so a later user edit compares against the server value.

Use a dedicated change reason to skip form tracking while preserving reactive notifications and node attributes. Clear prior changes only for the pushed field or Bag subtree; retain unrelated edits and track writes made by reactive controllers normally. Store the original data path in change-log entries so subtree cleanup does not confuse dotted paths with underscore-containing field names. Explicit non-silent pushes retain normal change tracking. Virtual-column names and the default `silent=True` remain unchanged.

Supersedes #1291. Refs #1287. Client confirmation in Pansotti is still pending; this PR does not close that issue automatically.

Validation:
- 31 JavaScript regression tests pass using the real Bag, source-node and form-handler code. The same tests expose 28 failures on develop and 20 on #1291. A dedicated GitHub Actions job runs them.
- Full Python suite: 2515 passed, 10 skipped. The added test page passes flake8; JavaScript syntax checks and `git diff --check` pass.
- Real-browser check at `test15/gnrwdg/external_change`, backed by PostgreSQL: loaded and initially absent virtual pushes stay clean; an unrelated Name edit survives the pushes and saves to the database; `silent=False` is tracked; an equal silent push acknowledges a local edit. No console errors or failed HTTP requests observed. Test-row values restored afterward.

### DIFF
--- a/.github/workflows/javascript-tests.yml
+++ b/.github/workflows/javascript-tests.yml
@@ -1,0 +1,16 @@
+name: JavaScript tests
+
+on: [push, pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+      - run: node --test gnrjs/tests/*.test.js
--- a/gnrjs/gnr_d11/js/genro_frm.js
+++ b/gnrjs/gnr_d11/js/genro_frm.js
@@ -1516,7 +1516,14 @@ dojo.declare("gnr.GnrFrmHandler", null, {
     },
 
     externalChange:function(field,value,triggerChanges){
-        this.sourceNode.setRelativeData(this.formDatapath+'.'+field,value,triggerChanges?{}:{_loadedValue:value});
+        if(!triggerChanges){
+            this.resetChangesAtPath(field);
+        }
+        this.sourceNode.setRelativeData(this.formDatapath+'.'+field,value,{},null,
+                                       triggerChanges?null:'externalChange',null,{_updattr:true});
+        if(!triggerChanges){
+            this.updateStatus();
+        }
     },
 
     getFormData: function() {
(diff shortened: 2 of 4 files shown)


genro (approved):
Approving. This answers the review on #1291 point by point, and the shape is better than the one I proposed there.

Verified on 17838e57:

- **The absent-node path is closed.** `triggerINS` returning on `reason == 'externalChange'` is what #1291 was missing, and `silent push creates a clean node` covers it.
- **The `$` route is not needed after all.** A dedicated reason closes all three paths without moving the value to `record.$<field>`, so application code reading `record.<field>` keeps working. That was the cost I flagged as a decision rather than a detail; this form does not pay it.
- **`hasChanges` is covered** by `assertClean()`, which asserts `record.getNodeByAttr('_loadedValue') == null` and not just an empty cluster. That was the "inert" point.
- **`Fixes` is gone.** `Refs #1287` with the Pansotti confirmation stated as pending is the right claim for what is established.
- **The tests are where they can be run.** 31 in `gnrjs/tests/`, real Bag and form handler, 28 failing on develop and 20 on #1291, with a CI job. That turns the three paths from a matter of reading into a red or green result.

One thing I checked and got wrong, recorded so nobody re-checks it: `resetChangesAtPath` pops from the changes logger while iterating `getNodes()`, which returns the live `_nodes`. That looked like a skipped-element bug and is not one — `_pop` rebuilds the array instead of splicing it, with a comment saying callers rely on that. Driving it with four entries, three matching, removes all three.

**One decision, not a defect, and it should be visible rather than only in a test.** A silent push at a field the user is editing discards the pending edit — `resetChangesAtPath` deletes the marker and the logger entry, then the server value is written and becomes the baseline. `silent push acknowledges an edited field` pins it: edit to 11, push, form clean, and the next edit compares against the server value. Edits on other fields survive, and there is a test for that.

That is defensible — the server has just written the record, so a pending edit on that field is stale. The opposite reading is also defensible: a user edit is not discarded silently. The behaviour is deliberate and tested, so it is a decision; it deserves a line in the docstring of `resetChangesAtPath` saying that a silent push at a path drops a pending edit there, so the next reader meets it in the code and not only in the suite.

PR #1294, opened 2026-09-11
fix: portable AFM metrics loading for pip-installed deployments

### LINKED ISSUE #978: [BUG] calcRowsNumber crashes with FileNotFoundError on pip-installed deployments (afm_widths.json not found)

dgpaci:
## Motivation

`_load_afm_widths()` in `gnr/utils/fonts.py` resolved `afm_widths.json` by climbing three directory levels from the module to the repository root. That layout only exists in a development checkout: on a pip-installed genropy the shared `resources/` directory is packaged as the `gnr.resources` package inside site-packages, so the lookup pointed at a non-existent path and any print measuring text crashed with `FileNotFoundError`. This broke invoice printing in a production deployment and forced a revert of the row-height migration on genromed.

Since #981 landed on develop, `auto_row_height` is on by default in `GnrBagHtmlBuilder`: `calcRowHeight` now reaches the metrics on **every grid print**, not only on the reports that explicitly overrode `getRowWrapField()`. The bug went from latent to systematic on every non-editable install, which is why the fix belongs here rather than on the hotfix branch this replaces.

Fixes #978. Replaces #979, which targeted `master`, where the AFM path is still strictly opt-in and no stock print reaches it.

## Solution

The metrics file is resolved through an ordered list of candidates, first readable one wins:

1. every path declared under `resources` in `environment.xml` — the same `resources:#a.path` lookup used by `resource_name_to_path` in `gnrwsgisite.py` and `gnrresourceloader.py`, so a deployment can override the metrics like any other shared resource
2. `resources` inside the `gnr` package — on an installed layout this is exactly where `MANIFEST.in`/`pyproject.toml` put the file (`site-packages/gnr/resources/common/fonts/afm_widths.json`), so it resolves with no configuration at all
3. the former checkout-relative path, for a repository used without installing (CI)

Candidate 2 is what makes the fix independent of configuration: `build_environment_xml` in `gnrdeploy.py` does write `resources.genropy` pointing at `<genropy_home>/resources` in both layouts, but a deployment with a hand-written or stale `environment.xml` would otherwise silently lose the real metrics.

If no candidate is readable, string widths degrade to approximate values (average char width) with a logged warning instead of raising — a metrics problem can never take a print down again.

The `getGnrConfig` import is deferred inside the resolver to break the `gnrstring → fonts → gnrconfig → gnrstring` import cycle.

## Tests

Seven tests added to `gnrpy/tests/utils/fonts_test.py`, all end-to-end with real files and real configuration (no mocks), driving resolution through the public `GENRO_GNRFOLDER` entry point:

- `test_no_declaration_finds_shipped_metrics`: no `resources` declaration → the metrics shipped with the code are still found
- `test_config_resources_priority`: metrics declared in `environment.xml` win over the shipped ones
- `test_config_resources_without_file_skipped`: a declared resources dir without the file is skipped
- `test_invalid_metrics_file_skipped`: a corrupt metrics file is skipped without raising
- `test_installed_layout_unconfigured`: regression test for #978 — module copy loaded from a simulated `site-packages/gnr/utils`, metrics in `site-packages/gnr/resources`, **nothing declared in `environment.xml`**
- `test_installed_layout_declared_resources_win`: same layout, declaration overrides the packaged metrics
- `test_metrics_file_missing_never_raises`: worst case, no usable metrics anywhere → approximate widths, no `FileNotFoundError`

Both new resolution candidates were mutation-checked: removing the package candidate fails `test_installed_layout_unconfigured`, removing the config candidates fails `test_config_resources_priority` and `test_installed_layout_declared_resources_win`.

Suite on this branch: 1302 passed excluding `gnrpy/tests/sql`, which errors locally only for a missing postgres binary (`initdb: program "postgres" is needed`), unrelated to this change. flake8 clean on both modified files.

### DIFF
--- a/gnrpy/gnr/utils/fonts.py
+++ b/gnrpy/gnr/utils/fonts.py
@@ -6,13 +6,18 @@
 # Units are 1/1000 of the font size in points.
 # Example: at 10pt, 'A' in Helvetica = 667/1000 * 10 = 6.67 pt wide.
 #
-# Width data is stored in resources/common/fonts/afm_widths.json and loaded
-# at first use.
+# Width data is stored in the shared resources directory
+# (common/fonts/afm_widths.json) and loaded at first use.
 
 import json
 import os
 
+from gnr.core.gnrsys import expandpath
+from gnr.utils import logger
+
 _AFM_WIDTHS = None
+
+_AFM_RELATIVE_PATH = os.path.join('common', 'fonts', 'afm_widths.json')
 
 _DEFAULT_CHAR_WIDTH = 556  # fallback for unknown chars (avg lowercase Helvetica)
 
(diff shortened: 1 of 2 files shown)


fporcari (approved):
Approved.

Verified on the head commit, not just on the description:

- the packaged candidate is real: `gnr.resources` is declared both in `packages` and in `package-dir` (`"gnr.resources" = "../resources"`), with `include-package-data = true` and `recursive-include ../resources *` in MANIFEST.in, so `site-packages/gnr/resources/common/fonts/afm_widths.json` is where the file lands on an installed layout — which is exactly what candidate 2 computes from `__file__`;
- the config candidate matches the existing lookup: `gnrwsgisite.resource_name_to_path` uses the same `'resources' in gnr_config['gnr.environment_xml']` guard and the same `digest('resources:#a.path')`;
- the checkout candidate resolves to the same repo-root path as the old code;
- the deferred `getGnrConfig` import is necessary and not cosmetic: `gnrconfig` imports `gnrstring`, which imports this module at load time;
- the degraded map never raises downstream: `_get_widths` returns `{}` (not `None`) for Helvetica, and `string_width` falls back to `_DEFAULT_CHAR_WIDTH` per char for every font;
- tests: 20 passed in `tests/utils/fonts_test.py` and 20 in `tests/core/gnrbaghtml_test.py`, flake8 clean on both modified files. Mutation-checked the package candidate myself: removing it fails `test_installed_layout_unconfigured` and nothing else, so the regression test really pins the mechanism.

One non-blocking note: the degraded result is cached in `_AFM_WIDTHS` for the lifetime of the process, so a transient read failure on the first call leaves that worker on approximate widths until restart, silently after the single warning. Given that candidate 2 ships with the code, that only happens on a genuinely broken install, so it is not worth re-scanning on every `string_width` call — just worth knowing when a report comes in saying the heights are slightly off on one worker only.

The merge is yours. Consider assigning the PR to yourself: an unassigned PR reads as unclaimed to everybody else.

